PR #350, opened 2020-10-22
Add ability to create and assign host to a host set in one step

gsusmi:
host-create-and-assign-workflow-revised

Introduces ability to create a host and assign it to a host set automagically from within a host set. This removed the current need to navigate to a host catalog to create a host and then navigate into a host set to assign it. The current assignment of hosts in a host set workflow hasn't been changed.

### DIFF
--- a/addons/api/addon/models/host-set.js
+++ b/addons/api/addon/models/host-set.js
@@ -40,6 +40,16 @@ export default class HostSetModel extends GeneratedHostSetModel {
   }
 
   /**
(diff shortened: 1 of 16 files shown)


randallmorey:
@gsusmi this looks great!  Thanks for pulling this together.  I like how you created a new route for this, makes sense to me going forward.  One opportunity to refactor is to reuse the existing host form.  If we avoid duplicating resource forms and their tests, we save some extra work when their fields change.

PR #357, opened 2020-11-03
Feature role grant scope

randallmorey:
This PR introduces grant scope field support on roles via a radio select in the role form.

Screenshot 2020-11-04 at 15 02 37


gsusmi (approved):
Looks good. Ty!
Minor note: In one of my tests, the grants dropdown added icons to all of the nested tree and flicked to only apply it to scopes correctly. Very hard to reproduce. Something for us to watch out for.

PR #377, opened 2020-11-11
chore: Configure ipc host as env var in desktop

gsusmi:
Making ipc host configurable as I tend to use non standard (4200) port to launch desktop app in dev.


randallmorey (changes requested):
We should probably remove the config option and just make the origin `window.location.origin` at runtime.  The origin will always be the same as the origin within which the code is running.

Detect via the regular the window property obtained through ember (it's part of the service already).  This would also require updating the test suite to ignore origin (it no longer needs to be passed in explicitly).

randallmorey:
Closing due to #379

PR #380, opened 2020-11-12
Feature auth

randallmorey:
Adds password auth to desktop app.


gsusmi (approved):
Missing translation strings to be addressed in a future PR. Approving. Ty!

PR #384, opened 2020-11-18
Refactor translations extraction

randallmorey:
This PR introduces the `core` addon and ports i18n-related concerns into it.  The core addon is where common code, such as translations, mocks, components, and services will live for easy reuse across UIs.


gsusmi (approved):
LGTM. Ty!

PR #390, opened 2020-11-19
refactor: 💡 port error/message component to core addon

randallmorey:


addons/core/tests/dummy/app/templates/application.hbs
```diff
@@ -1,3 +1,3 @@
 <h2 id="title">Welcome to Ember</h2>
 
-{{outlet}}
\ No newline at end of file
+{{t "errors.401.title"}}
```
gsusmi, line 3: Is this for an easy intl check?

randallmorey: Yes.  The dummy app app or index tends to be whatever we were last working on and doesn't affect anything.

gsusmi: Awesome. Just checking. Will merge.

PR #403, opened 2020-12-02
feat: 🎸 implement configurable IPC timeout, with a 2s default

randallmorey:
Adds timeouts to IPC requests.  If the request does not complete within the timeout period, the request rejects with an error.  The duration is configurable per request and defaults to 2 seconds.


randallmorey:
This may be unnecessary.  The CLI is expected to exit with an error on timeout, which we can

PR #404, opened 2020-12-02
Feature breezy onboarding

randallmorey:
This PR improves both UX and DX.  Firstly, origin and auth fields get autofocus.  And in development, the origin field is auto populated with the current origin.  This allows developers to simply press enter twice to onboard.

### DIFF
--- a/ui/desktop/package.json
+++ b/ui/desktop/package.json
@@ -39,6 +39,7 @@
     "core": "*",
     "doctoc": "^1.4.0",
     "ember-auto-import": "^1.6.0",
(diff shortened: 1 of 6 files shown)


ui/desktop/package.json
```diff
@@ -39,6 +39,7 @@
     "core": "*",
     "doctoc": "^1.4.0",
     "ember-auto-import": "^1.6.0",
+    "ember-autofocus-modifier": "^1.1.0",
```
gsusmi, line 42: Nice!

PR #407, opened 2020-12-08
feat: Add node spawn promise wrapper

gsusmi:


ui/desktop/app/routes/scopes/scope/projects/targets.js
```diff
@@ -47,10 +47,17 @@ export default class ScopesScopeProjectsTargetsRoute extends Route {
   @action
   async connect(model) {
     try {
-      await this.ipc.invoke('connect', {
+      await this.ipc.invoke('cli');
```
randallmorey, line 50: Can the method name be more descriptive?  It's not immediately obvious that this is a check.  Something like `cli-exists'

ui/desktop/app/routes/scopes/scope/projects/targets.js
```diff
@@ -47,10 +47,17 @@ export default class ScopesScopeProjectsTargetsRoute extends Route {
...
+      const connect = this.ipc.invoke('connect', {
         target_id: model.target.id,
         token: this.session.data.authenticated.token,
       });
+
+      connect.then((data) => {
+        this.notify.success(JSON.stringify(data));
+      }, (e) => {
```
randallmorey, line 58: The error case is the same as the next one.  You don't need to handle it here too.

ui/desktop/electron-app/src/spawn-promise.js
```diff
@@ -0,0 +1,65 @@
...
+
+module.exports = function spawnPromise (command) {
+  return new Promise((resolve, reject) => {
+    const childProcess = spawn('boundary', command);
+    let outputStream = '';
+    let errorStream = '';
+
+    childProcess.stdout.on('data', (data) => {
+      outputStream += data.toString();
+      resolveOutput(resolve, outputStream);
```
randallmorey, line 47: I'd avoid passing a function if you don't need to:
```js
if (isJSON(output)) resolve(JSON.parse(output));
```

ui/desktop/electron-app/src/spawn-promise.js
```diff
@@ -0,0 +1,65 @@
...
+    childProcess.on('close', (code) => {
+      if(code === 0) {
+        resolveOutput(resolve, outputStream);
+      }else {
+        rejectError(reject, errorStream);
+      }
+    });
+
+    // Handle error
+    childProcess.on('error', (err) => rejectError(reject, err));
```
randallmorey, line 60: I'd reverse this to something like reject(normalizeError(err)) so that your own function just does the transformation.

randallmorey: Avoid passing functions around if possible.

randallmorey (changes requested):

ui/desktop/app/routes/scopes/scope/projects/targets.js
gsusmi: Updated this to 'cliPath' as I'm returning the actual path of cli.

ui/desktop/electron-app/src/spawn-promise.js
gsusmi: Updated. I like this approach of not passing a function.

ui/desktop/electron-app/src/spawn-promise.js
gsusmi: Corrected.

gsusmi:
@randallmorey , this is ready for another round of review. It contains confirm dialog integration so we'll have a decent connection workflow in place with this PR.

ui/desktop/app/routes/scopes/scope/projects/targets.js
```diff
@@ -47,12 +48,24 @@ export default class ScopesScopeProjectsTargetsRoute extends Route {
   @action
   async connect(model) {
     try {
-      await this.ipc.invoke('connect', {
+      // Check for cli
+      const cliPath = await this.ipc.invoke('cliPath');
+      if(!cliPath) throw new Error('Cannot find boundary cli.')
+
+      // Create target session
+      const connection = this.ipc.invoke('connect', {
```
randallmorey, line 56: This section could be rewritten more fluently as:

```js
const connectionDetails = await this.ipc.invoke('connect', {
   target_id: model.target.id,
   token: this.session.data.authenticated.token,
 });

this.confirm.confirm(connectionDetails, { isConnectSuccess: true });
```

ui/desktop/electron-app/src/handlers.js
```diff
@@ -20,51 +20,25 @@ handle('getOrigin', () => origin);
  */
 handle('setOrigin', (requestOrigin) => origin = requestOrigin);
 
+/**
+ * Detect boundary cli path
+ */
+handle('cliPath', async () => {
+  const cliPath = await lookpath('boundary');
+  if(!cliPath) throw new Error({ message: 'Cannot find boundary cli.'})
```
randallmorey, line 28: Two things here:

- The targets route throws its own, duplicate error for this.  Ideally it would only need to be thrown in one place.
- Should this method return a boolean?  It makes a bit more sense to me that it would just be a boolean `true` / `false`.  Either the CLI is found or it's not.

ui/desktop/electron-app/src/spawn-promise.js
```diff
@@ -0,0 +1,46 @@
+const { spawn } = require('child_process');
+
+// Verify json
+const isJSON = (data) => {
```
randallmorey, line 4: It seems like this method is not exactly "is JSON".  It's more "is an object in object or JSON string form".

ui/desktop/electron-app/src/spawn-promise.js
```diff
@@ -0,0 +1,46 @@
...
+// POJO representation and throw it or promise->reject it.
+module.exports = function spawnPromise (command) {
+  return new Promise((resolve, reject) => {
+    const childProcess = spawn('boundary', command);
+    let outputStream = '';
+    let errorStream = '';
+
+    childProcess.stdout.on('data', (data) => {
+      outputStream += data.toString();
+      if(isJSON(outputStream)) resolve(JSON.parse(outputStream));
```
randallmorey, line 38: There's a lot of JSON stringification and parsing happening in order to resolve.  Can this be simplified?

ui/desktop/electron-app/src/spawn-promise.js
```diff
@@ -0,0 +1,46 @@
...
+    });
+
+    childProcess.stderr.on('data', (data) => {
+      errorStream += data.toString();
+      if(isJSON(errorStream)) reject(new Error(JSON.parse(errorStream)?.error));
```
randallmorey, line 43: Same here.

randallmorey (changes requested):

ui/desktop/electron-app/src/spawn-promise.js
gsusmi: Updated - by introducing a `jsonify` function that returns stream as a json, is possible.

ui/desktop/electron-app/src/spawn-promise.js
gsusmi: Updated - by introducing a `jsonify` function that returns stream as a json, is possible.

ui/desktop/electron-app/src/spawn-promise.js
gsusmi: Updated to `jsonify` function that returns data as json, is possible.

ui/desktop/electron-app/src/handlers.js
gsusmi: Removed cli path errors from handler. It makes sense for path checks to happen on the side of things that invoked it.
In addition, separated out 'exists' and 'path' functions for convenience.

ui/desktop/electron-app/src/spawn-promise.js
```diff
@@ -0,0 +1,45 @@
...
+module.exports = function spawnPromise (command) {
+  return new Promise((resolve, reject) => {
+    const childProcess = spawn('boundary', command);
+    let outputStream = '';
+    let errorStream = '';
+
+    childProcess.stdout.on('data', (data) => {
+      outputStream += data.toString();
+      const jsonData = jsonify(outputStream);
+      if(jsonData) resolve(jsonData);
```
randallmorey, line 36: 👍 This looks great!  Very readable.

randallmorey (approved):

PR #411, opened 2020-12-10
fix: 🐛 fix ember-notify style dupe/override problem

randallmorey:
This PR improves on the ember-notify style fix.  Previously we had duplicated styles and used `!important` to override.  The better approach was to bring the addon local to each app rather than providing it via core.


gsusmi:
Much simpler fix that what I thought it'd end up being. Awesome!.

PR #421, opened 2021-01-06
Refactor typed auth methods

randallmorey:
Previously, the "password" auth method type was a hard-coded assumption in both the auth method CRUD as well as the authentication route.  This PR introduces _typed auth methods_:

- Auth method CRUD form is a thin wrapper around type-specific auth method forms, allowing the admin routes to be type-agnostic.
- Authenticate form is also a thin wrapper, with additional logic to handle clearing of sensitive credentials independently of authentication success or failure
- Authenticate route is now entirely type-agnostic


ui/admin/app/components/form/authenticate/password/index.hbs
```diff
@@ -0,0 +1,28 @@
+<Rose::Form
+  class="full-width"
+  @onSubmit={{fn this.submit @submit (hash
+    identification=this.identification
+    password=this.password
+  )}}
+  @cancel={{false}}
```
gsusmi, line 7: Shouldn't this be `@cancel={{@cancel}}`?

randallmorey: There is no cancel action for the auth form.

gsusmi (approved):

PR #423, opened 2021-01-07
Add sessions sidebar in desktop

gsusmi:
desktop-sessions

Surfaces filtered sessions view in desktop. Filtering is based on current user id.

Todo for next PR:
- extract copyable (and maybe breadcrumbs?) component from admin into core addon
- figure out if it's worth it to extract session styles into core addon.


gsusmi:
Thing to watch out for in this PR is the assumption mirage makes with authenticated user id: 'user123'. Since sessions in desktop client need to be filtered to authenticated user, mirage has been updated to ensure 'user123' has sessions. Happy to rework this if the current mirage user creation is hard to follow.

addons/api/mirage/factories/auth-method.js
```diff
@@ -10,7 +10,10 @@ export default factory.extend({
   withAccountsAndUsers: trait({
     afterCreate(authMethod, server) {
       const { scope } = authMethod;
-      server.createList('user', 5, { scope }).map(user => {
+      const users = server.createList('user', 5, { scope });
+      // Create authorized user: user123
+      users.push(server.create('user', { scope, id: 'user123' }));
```
randallmorey, line 15: We can't hardcode the user ID in this way.  This factory could potentially be run on multiple auth methods.  But a given user can only be associated with a single auth method.  What is it you're trying to achieve with this hardcoded user ID?

addons/api/mirage/factories/session.js
```diff
@@ -41,6 +41,29 @@ export default factory.extend({
...
       }
     }
+  }),
+
+  /**
+   * Sessions generated with this trait will automatically form
+   * associations to target, host set, and host (if available) within
+   * the current scope and for specified user.
+   */
+  withUserAssociations: trait({
```
randallmorey, line 51: This is essentially a duplicate of `withAssociations` on the sessions factory.  Why is this necessary?

addons/api/mirage/scenarios/default.js
```diff
@@ -29,6 +29,9 @@ export default function(server) {
     server.createList('host-catalog', 2, { scope }, 'withChildren');
     server.createList('target', 2, { scope }, 'withRandomHostSets');
     server.createList('session', 4, { scope }, 'withAssociations');
+    // Create sessions for all users
+    server.schema.users.where({ scopeId: scope.scope.id }).models.forEach(user =>
```
randallmorey, line 33: Why?

ui/desktop/app/styles/app.scss
```diff
@@ -100,15 +100,8 @@
 }
 
 .rose-header {
-  -webkit-app-region: drag;
```
randallmorey, line 103: Why were these styles removed?

randallmorey (changes requested):

addons/api/mirage/factories/auth-method.js
gsusmi: In desktop, list of all sessions need to be filtered down to ones related to logged in user. Seeing as we don't actually create a user for authentication in mirage ('user123' is returned as part user_id of auth payload without any associated user creation in mirage schema), I'm creating the authenticated user and populating sessions data for this user. I was hoping I could inject authenticated user sessions as part of /sessions response in mirage but this request doesn't have access to authenticated user id. There might be some mirage magic to support this scenario. Do you have any pointers on how to do this more cleanly?

ui/desktop/app/styles/app.scss
gsusmi: Might be a leftover from an experiment. Oops. Will recheck this.

addons/api/mirage/factories/auth-method.js
randallmorey: We don't have to model every detail of the API (and we shouldn't, because we'd recreate the API).  In this case it's probably enough to return any sessions without filtering them by user?

gsusmi: Ah. The user filtering happens in ember route action so if we leave things as they are on main, sessions view won't display any data.

ui/desktop/app/styles/app.scss
gsusmi: Oh, I remember why padding was needed. Desktop electron shell needs the padding to accommodate min/max icons. Updated by reverting deletion.

addons/api/mirage/factories/session.js
gsusmi: Removed.

addons/api/mirage/scenarios/default.js
gsusmi: Cleaned up by creating authenticated user sessions list in `withAssociations`

addons/api/mirage/factories/auth-method.js
gsusmi: Updated mock auth payload to use first available user and mock session records to randomly pick between auth user and any user available.

randallmorey: Can this be made deterministic somehow?  For example, could it alternate between the two options?  Otherwise it's possible to have no sessions assigned to the logged in user.

addons/api/mirage/route-handlers/auth.js
```diff
@@ -33,12 +33,13 @@ export function authHandler({ scopes, authMethods }, request) {
     const scope = scopes.find(authMethod.scopeId);
     const scopeAttrs =
       this.serialize(scopes.find(scope.id));
+    const user = users.first();
```
randallmorey, line 36: Would it make more sense to pass in a specific user rather than a list from which we assume there is a first user?

addons/api/mirage/factories/session.js
```diff
@@ -26,8 +26,10 @@ export default factory.extend({
       const scopeId = record.scopeId;
       if (record.scope.type === 'project') {
         const orgScopeId = record.scope?.scope.id;
-        const user =
+        const orgUser =
           server.schema.users.where({ scopeId: orgScopeId }).models[0];
+        const authenticatedUser = server.schema.users.first();
+        const user = random.arrayElement([orgUser, authenticatedUser]);
```
randallmorey, line 32: Can this be made deterministic?

randallmorey (changes requested):

gsusmi:
Updates: Moved session population for logged in user into mirage route handler instead of factories.

addons/api/mirage/scenarios/default.js
```diff
@@ -13,6 +13,8 @@ export default function(server) {
     scope: { id: globalScope.id, type: globalScope.type }
   }, 'withChildren')[0];
 
+  server.create('user', { id: 'authenticateduser' });
```
randallmorey, line 16: Is this necessary?

addons/core/tests/unit/authenticators/password-test.js
```diff
@@ -44,6 +44,7 @@ module('Unit | Authenticator | password', function (hooks) {
     const authMethod = { id: 'paum_1234' };
     this.server.create('scope', { id: 'global', type: 'global' });
     this.server.create('auth-method', { id: authMethod.id, scopeId: 'global' });
+    this.server.create('user', { scopeId: 'global' });
```
randallmorey, line 47: Is this necessary?

randallmorey (changes requested):

addons/api/mirage/scenarios/default.js
gsusmi: When responding to sessions query, I'm updating userId of alternate session records to `authenticateduser`. Due to this association, the actual user object needs to exist otherwise mirage throws a ` Mirage: Couldn't find user with id = authenticateduser Error: Mirage: Couldn't find user with id = authenticateduser`

addons/core/tests/unit/authenticators/password-test.js
gsusmi: Good catch! Forgot about this from my previous mocking of authenticated user. Removed.

PR #425, opened 2021-01-08
Feature csp

randallmorey:
This PR introduces a strict content security policy to the renderer process.  It also handles updating the CSP to dynamically allow the origin in `content-src`.


gsusmi (approved):
LGTM. Til about `axios`

PR #426, opened 2021-01-11
refactor: 💡 move authenticate forms to core and use in desktop

randallmorey:
Simple PR to move the authenticate forms into addons/core for reusability.  The password form is now shared across admin and desktop, as will the OIDC form once built.


mgaffney (approved):
Great job! Very readable.

PR #429, opened 2021-01-22
Feature oidc authenticate

randallmorey:
OIDC authentication for admin UI.


addons/auth/addon/authenticators/oidc.js
```diff
@@ -0,0 +1,99 @@
...
+  async authenticate(json) {
+    console.info('Authenticated with OIDC provider');
+    return this.normalizeData(json);
+  }
+
+  /**
+   *
+   */
+  async attemptOIDCAuthentication(options) {
+    const url = this.buildAuthEndpointURL(options);
```
gsusmi, line 82: Think of renaming `buildAuthEndpointURL` to `buildOIDCAuthEndpointURL`. This would match the pattern set in `startAuthentication` function.

addons/core/addon/authenticators/oidc.js
```diff
@@ -0,0 +1,65 @@
...
+   * @return {string}
+   */
+  buildAuthEndpointURL({
+    scope: { id: scopeID },
+    authMethod: { id: authMethodID },
+  }) {
+    const adapter = this.store.adapterFor('application');
+    const options = { adapterOptions: { scopeID, method: 'authenticate' } };
+    return adapter.buildURL('auth-method', authMethodID, options, 'findRecord');
+  }
```
gsusmi, line 52: Potential for creating a common function that takes `scope`, `authMethod`, and current `adapterOptions.method` string as a param to combine `buildAuthEndpointURL` and `buildStartAuthEndpointURL`. Unless this separation is future proofing for config variations.

ui/admin/app/router.js
```diff
@@ -10,7 +10,9 @@ Router.map(function () {
   this.route('scopes', function () {
     this.route('scope', { path: ':scope_id' }, function () {
       this.route('authenticate', function () {
-        this.route('method', { path: ':auth_method_id' });
+        this.route('method', { path: ':auth_method_id' }, function() {
+          this.route('oidc');
```
gsusmi, line 14: Can/should authenticate/method route evolve to handle 'password' route? If we are going to call out oidc maybe we can separate out password route too (although it don't do as much as oidc).

gsusmi (commented):
Minor queries. Great first steps to OIDC @randallmorey!

ui/admin/app/router.js
randallmorey: The full auth flow for password can be completed on the method route (it's a one-step process).  In the case of OIDC, there's an indefinite polling period while the user authenticates.  This route serves as a kind of waiting room that is only needed for OIDC at this time.

addons/core/addon/authenticators/oidc.js
randallmorey: OIDC has different needs and there is no single "auth" endpoint, so this is now split into `buildStartAuthEndpointURL` and `buildTokenAuthEndpointURL`.

addons/auth/addon/authenticators/oidc.js
randallmorey: This is now `buildTokenAuthEndpointURL`

gsusmi (approved):
Awesome! Ty!

PR #436, opened 2021-02-03
View hosts in a target

gsusmi:
Navigate into a target to view hosts. Hosts are loaded through host sets associated to the target.

hosts-in-a-target


ui/desktop/app/routes/scopes/scope/projects/targets/target.js
```diff
@@ -0,0 +1,43 @@
...
+    return this.store.findRecord('target', target_id, { reload: true });
+  }
+
+  /**
+   * Navigate to hostsets within a target.
+   * @param {TargetModel} model 
+   */
+  afterModel(model) {
+    this.transitionTo('scopes.scope.projects.targets.target.hosts', model);
+  }
```
gsusmi, line 22: Is there a better way to load into hosts route? I could have used 'index' route and places hosts templates in it but we called this out as a separate route on admin.

ui/desktop/app/routes/scopes/scope/projects/targets/target/hosts.js
```diff
@@ -0,0 +1,35 @@
...
+    );
+    
+    // Fetch all hosts
+    const hostPromises = await all(
+      hostIds.uniq().map((hostId) => 
+        this.store.findRecord('host', hostId)
+      )
+    )
+
+    return all(hostPromises);
```
gsusmi, line 33: This was hard to work through - to try to get to hosts without going through host catalogs. Looking for feedback.

ui/desktop/app/routes/scopes/scope/projects/targets/target.js
randallmorey: The redirect should happen in the target/index route, using the `redirect()` route method.  You also do not need to (and shouldn't) pass the model.  The target/hosts route does not take a target instance.

ui/desktop/app/routes/scopes/scope/projects/targets/target/hosts.js
randallmorey, line 33: Hosts were already wrapped in `all()` and awaited, meaning they're fully resolved by now.  No need to `all()` again.

ui/desktop/app/routes/scopes/scope/projects/targets/target/hosts.js
```diff
@@ -0,0 +1,35 @@
...
+    const hostSetsPromises = await all(
+      host_sets.map(({ host_set_id }) => 
+        this.store.findRecord('host-set', host_set_id)
+      )
+    );
+
+    // Extract host ids from all host set
+    const hostIds = A();
+    hostSetsPromises.map(({ host_ids }) => 
+      host_ids.forEach(({ value }) => hostIds.push(value) )
```
randallmorey, line 23: Anytime I see `push()` I try to find a better way.  Can this be mapped to an array of arrays instead, and then flattened?

ui/desktop/app/routes/scopes/scope/projects/targets/target/hosts.js
```diff
@@ -0,0 +1,35 @@
...
+export default class ScopesScopeProjectsTargetsTargetHostsRoute extends Route {
+  // =methods
+
+  /**
+   * Loads all host-sets and associated hosts under the current target.
+   * @return {Promise{[HostModel]}}
+   */
+  async model() {
+    const { host_sets } = this.modelFor('scopes.scope.projects.targets.target');
+    const hostSetsPromises = await all(
```
randallmorey, line 14: I think you can call this `hostSets` if you're awaiting.  The resolved value will be host sets, not promises.

ui/desktop/app/routes/scopes/scope/projects/targets/target/hosts.js
```diff
@@ -0,0 +1,35 @@
...
+    );
+
+    // Extract host ids from all host set
+    const hostIds = A();
+    hostSetsPromises.map(({ host_ids }) => 
+      host_ids.forEach(({ value }) => hostIds.push(value) )
+    );
+    
+    // Fetch all hosts
+    const hostPromises = await all(
```
randallmorey, line 27: If you await the value, you get hosts, not promises.

randallmorey (changes requested):

gsusmi:
As recommended offline, looked through migrating hosts onto target model based on existing hostSets property. Alas, the current hostSets property doesn't seems to be populating and not being used outside of api addon. Going to do a separate PR after additional debugging.

ui/desktop/app/routes/scopes/scope/projects/targets/target/hosts.js
```diff
@@ -0,0 +1,35 @@
...
+    // Extract host ids from all host sets
+    const hostIds = hostSets.map(({host_ids}) =>
+      host_ids.content.map((hostId) =>
+        hostId.value
+      ) 
+    )
+
+    // Load unique hosts
+    const uniqueHostIds = new Set(hostIds.flat());
+    return all(
```
randallmorey, line 29: Just a note that this works for now, but it will break down once we have very large host sets with many hosts.  In that case we'll need to paginate.  Same problem will occur in admin.

randallmorey (approved):

PR #437, opened 2021-02-04
Persist proxy info

randallmorey:
After a successful connection, the proxy info returned by the CLI is persisted (in memory) into the associated session instance for display in the UI.  This allows users to refer back to the proxy address and port as needed.  The info is transient since it is not returned by the API.

Screenshot 2021-02-04 at 10 46 06


gsusmi (approved):
Looks good. Ty for setting up copyable. Had it on my todo.

PR #438, opened 2021-02-04
Show active sessions in a target

gsusmi:
active-sessions-in-target

TODO: Add polling to refresh active sessions. Update mirage to populate active sessions for targets.


ui/desktop/app/routes/scopes/scope/projects.js
```diff
@@ -25,4 +27,17 @@ export default class ScopesScopeProjectsRoute extends Route {
...
+  /**
+   * Cancels the specified session and notifies user of success or error.
+   * @param {SessionModel}
+   */
+  @action
+  @notifyError(({ message }) => message, { catch: true })
+  @notifySuccess('notifications.canceled-success')
+  async cancelSession(session) {
+    await session.cancelSession();
+  }
```
gsusmi, line 42: Since we have both _/projects/sessions_ and _/projects/targets/:target_id/sessions_ have actions to cancel sessions, I surfaced the action to parent route _/projects_. Thoughts?

randallmorey: It makes sense as a way to remove duplication.  At the same time it moves the action to an unexpected place.  Generally the actions relating to a route resource should live with that route.  Because you could argue that all actions simply live on application.  So developers looking through our code will never be certain where to find an action if the principal is "on the related route, except if it's convenient to move it up to a common ancestor so it can be reused".  Because then you need to know where else the action is being used.  Thoughts?

randallmorey: I guess my position is "put the action on the related route and no higher".  Duplication is ok for now.  And it's not that bad because most of the logic is encapsulated in `cancelSession`.

gsusmi: Agreed. Since we had a pattern of placing relevant actions in their routes, this was a push to find a boundary. Glad we agree that it's ok to have duplication when it helps with context. Updated PR. Ty!

PR #442, opened 2021-02-08
fix: 🐛 partition (and thus browser storage) should be persisted

randallmorey:


gsusmi (approved):
Awesome! Can't believe how simple this fix is.

PR #444, opened 2021-02-09
feat: Connect to a host within target

gsusmi:
Introduces button to connect to a specific host within a target.

Screen Shot 2021-02-08 at 20 23 40

TODO: There's bit of duplication between connecting to a host and connecting to a target. Good point of refactor for either a new decorator or a service.


randallmorey (changes requested):
Can it just say "connect"?  The wrapped text in the button doesn't look nice.

gsusmi:
Updated button text:
connect-to-a-host

randallmorey (approved):

PR #452, opened 2021-02-10
refactor: 💡 use recursive listing for desktop queries

randallmorey:
This PR updates desktop to use recursive queries for scopes, targets, and sessions.  It adds recursion support to mocks for the resources as well (and only these for now).  Also, session filtering and sorting is moved to the controller to simplify the route model.

### DIFF
--- a/addons/api/mirage/config.js
+++ b/addons/api/mirage/config.js
@@ -23,13 +23,16 @@ export default function() {
 
   // Scope resources
 
-  this.get('/scopes', ({ scopes }, { queryParams: { scope_id } }) => {
+  this.get('/scopes', ({ scopes }, { queryParams: { scope_id, recursive } }) => {
     // Default parent scope is global
     if (!scope_id) scope_id = 'global';
-    const results = scopes.where(scope => {
-      return scope.scope ? scope.scope.id === scope_id : false
-    });
-    return results;
+    if (recursive && scope_id === 'global') {
+      return scopes.all();
+    } else {
+      return scopes.where(scope => {
+        return scope.scope ? scope.scope.id === scope_id : false
+      });
+    }
   });
   this.post('/scopes', function ({ scopes }) {
     // Parent scope comes through the payload via `scope_id`, but this needs
@@ -287,7 +290,16 @@ export default function() {
 
   // target
 
-  this.get('/targets', function ({ targets }, { queryParams: { scope_id } }) {
+  this.get('/targets', function ({ targets }, { queryParams: { scope_id, recursive } }) {
+    if (recursive && scope_id === 'global') {
+      return targets.all();
+    } else if (recursive) {
+      return targets.where(target => {
+        const targetModel = targets.find(target.id);
+        return (target.scopeId === scope_id)
+          || (targetModel?.scope?.scope?.id === scope_id)
+      });
+    }
     return targets.where(target => target.scopeId === scope_id);
   });
   this.post('/targets', function ({ targets }) {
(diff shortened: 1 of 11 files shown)


randallmorey (commented):
I need to add tests for the changes in API addon.

addons/api/mirage/config.js
```diff
@@ -23,13 +23,16 @@ export default function() {
...
-  this.get('/scopes', ({ scopes }, { queryParams: { scope_id } }) => {
+  this.get('/scopes', ({ scopes }, { queryParams: { scope_id, recursive } }) => {
     // Default parent scope is global
     if (!scope_id) scope_id = 'global';
-    const results = scopes.where(scope => {
-      return scope.scope ? scope.scope.id === scope_id : false
-    });
-    return results;
+    if (recursive && scope_id === 'global') {
+      return scopes.all();
```
randallmorey, line 30: This may not be exactly right.  @jefferai, if recursive scopes are queried with ID `global`, it should return all scopes _except_ global, correct?  And if scopes are queried _with only_ recursive (no ID), would all scopes including global be returned?

ui/desktop/app/controllers/scopes/scope/projects/sessions/index.js
```diff
@@ -0,0 +1,27 @@
...
+
+export default class ScopesScopeProjectsSessionsIndexController extends Controller {
+
+  // =sessions
+
+  @service session;
+
+  // =attributes
+
+  @computed('model.@each.{created_time,active}', 'session.data.authenticated.user_id')
```
randallmorey, line 13: This needs a comment.

ui/desktop/app/routes/scopes/scope.js
```diff
@@ -89,14 +89,14 @@ export default class ScopesScopeRoute extends Route {
...
-   * In global scope, redirect to first org on authentication.
-   */
-  redirect(model) {
-    const { isAuthenticated } = this.session;
-    const { isGlobal } = model;
-    if (isAuthenticated && isGlobal && this.scopes.orgs?.length) {
-      this.replaceWith('scopes.scope', this.scopes.orgs.firstObject.id);
-    }
-  }
+  // /**
```
randallmorey, line 92: This can be removed.

addons/api/mirage/config.js
jefferai: Nope, recursion _includes_ the scope being queried. If you aren't recursing, we'd show scopes where the parent is the one given, but when recursing scopes from global we include global as well, because otherwise there'd be no way to include it (since it doesn't have a parent).

ui/desktop/app/routes/scopes/scope/projects.js
```diff
@@ -20,10 +20,9 @@ export default class ScopesScopeProjectsRoute extends Route {
    * @return {Promise
    */
   model() {
-    const currentScope = this.modelFor('scopes.scope');
-    return this.store.query('scope', {
-      scope_id: currentScope.id,
-    });
+    return this.store
```
randallmorey, line 23: This could use an explanation because it's unclear why we would return _all_ projects when we're under scopes/scope route.  In desktop, we simply want to prime the store with all projects.  It may not even make sense to have this route any more now that we have recursive listings (this query can be made from scopes or scopes/scope).

ui/desktop/app/routes/scopes/scope/projects/targets.js
```diff
@@ -33,7 +32,7 @@ export default class ScopesScopeProjectsTargetsRoute extends Route {
   @task(function * () {
     while(true) {
       yield timeout(POLL_TIMEOUT_SECONDS * 1000);
-      yield this.refreshSessions();
+      //yield this.refreshSessions();
```
randallmorey, line 35: OOPS!

gsusmi (approved):
LGTM!

PR #455, opened 2021-02-10
fix: Remove global scope from header org dropdown

gsusmi:
As selecting 'global' scope in header after authentication loads first org by default, remove option from dropdown to avoid confusion.


randallmorey (commented):
I fix this in the refactor branch.  The user is no longer redirected to the first org.  If they auth with global, they should stay in global until they navigate.  Recursive listing means we can still show their resources from global.

gsusmi:
Awesome! I'll close this PR without merging in favor of using recurse API for listing.

PR #461, opened 2021-02-11
fix: 🐛 target sessions should update on poll

randallmorey:
This fix ensures that target/sessions as well as overall sessions are properly updated when polling.

### DIFF
--- a/ui/desktop/app/templates/scopes/scope/projects/targets/target/sessions.hbs
+++ b/ui/desktop/app/templates/scopes/scope/projects/targets/target/sessions.hbs
@@ -1,6 +1,6 @@
 <Rose::Table as |table|>
   <table.body as |body|>
-    {{this.sessions.length}} / {{this.cancelableUserSessions.length}}
+    
     {{#each this.cancelableUserSessions as |session|}}
       <body.row as |row|>
         <row.cell class="nowrap">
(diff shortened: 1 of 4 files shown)


ui/desktop/app/templates/scopes/scope/projects/targets/target/sessions.hbs
```diff
@@ -1,6 +1,7 @@
 <Rose::Table as |table|>
   <table.body as |body|>
-    {{#each this.selfActiveSessions as |session|}}
+    {{this.sessions.length}} / {{this.cancelableUserSessions.length}}
```
gsusmi, line 3: Screen Shot 2021-02-11 at 17 09 39
 It'd be great to have a label here to explain these numbers. Something like: '2 out of 4 active sessions' or 'Active Sessions: 2. Total Sessions: 4'.

gsusmi (changes requested):
When a connection is added using 'Connect' button in header of 'Active Sessions' tab, the copyable component isn't triggered correctly so session details cannot be copied from the updated list. Can you take a look into this?

ui/desktop/app/templates/scopes/scope/projects/targets/target/sessions.hbs
randallmorey: Oops.  This was debug output.  Removed.

gsusmi (approved):

PR #463, opened 2021-02-11
Add desktop release trigger

99:
Adding github workflow to trigger boundary-ui releases in circleci.


.github/workflows/ui-release-trigger.yml
```diff
@@ -0,0 +1,20 @@
+name: "Trigger Boundary UI release"
+
+on:
+  push:
+    tags:
+      - '*'
```
shore, line 6: Do we really want to release for any tag?  This would preclude the use of tags for any other purpose.

.github/workflows/ui-release-trigger.yml
```diff
@@ -0,0 +1,20 @@
...
+jobs:
+  trigger-circleci-release:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@master
+      - name: Trigger Release job
+        env:
+          TAG: ${{ github.event.release.tag_name }}
+        run: |
```
shore, line 17: Should we do any validation on the tag name before attempting a release?  Should we require it look like a semver version?

randallmorey: Tagging @99 for visibility.  I agree with @shore.  For the time being let's use what's here.  We'd like to push out a release today.

.github/workflows/ui-release-trigger.yml
randallmorey: Tagging @99 for visibility.  I agree with @shore.  For the time being let's use what's here.  We'd like to push out a release today.

99: @randallmorey I'm happy to modify this,  please suggest what tag syntax the team is going to use for prod releases?
For now I'm seeing different tags(v*, 1*).
Also please note that our publishing is following semver syntax (tag like this v1.2.3 will be modified to 1.2.3).
We can also add `tag-ignore` field if needed to make sure release is not triggered for some tags .

99: we can maybe do something like this releases/1.2.0 not sure if there is any hashicorp standarts for doing this

randallmorey: The main boundary repo uses a `v0.0.0` convention.  We should probably be consistent across the product so that gets my vote.  The point about not being able to do any other release type besides desktop is a good one.  We could do `desktop/v0.0.0*` or similar.

PR #476, opened 2021-02-16
docs: add changelog

randallmorey:
Proposed changelog takes the same approach as Boundary.

### DIFF
--- a/ui/desktop/CHANGELOG.md
+++ b/ui/desktop/CHANGELOG.md
@@ -1,10 +1,8 @@
-# Boundary UI CHANGELOG
+# Boundary Desktop CHANGELOG
 
-Canonical reference for changes, improvements, and bugfixes for Boundary UI.
+Canonical reference for changes, improvements, and bugfixes for Boundary Desktop.
 
 
 ## 0.1.0 (2021.02.16)
-
-### Desktop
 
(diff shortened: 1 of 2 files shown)


randallmorey:
One question @gsusmi @malnick is heading structure for UI types.  Should it be:

- version
  - UI name (admin or desktop)
    - changes/additions/fixes

OR

- version
  - changes/additions/fixes
    - UI name (admin or desktop)

gsusmi:
I prefer version/ui name as it's much easier to identify client changes that one is interested in. How are we doing to identify features that get released to both? Something like : admin/desktop/all for UI name?

randallmorey:
Actually, I wasn't thinking clearly.  We're not versioning for admin so it's not clear how we'd changelog it.  Let's just worry about desktop for now?

ui/desktop/CHANGELOG.md
```diff
@@ -0,0 +1,8 @@
+# Boundary Desktop CHANGELOG
+
+Canonical reference for changes, improvements, and bugfixes for Boundary Desktop.
+
+
+## 0.1.0 (2021.02.16)
```
gsusmi, line 6: Should this heading be `v1.0.0-alpha`?

gsusmi (approved):
LGTM. Ty!

PR #483, opened 2021-02-25
chore: 🤖 support api_errors and drop errors format flag

randallmorey:
This PR adds support for hashicorp/boundary#962.


ui/desktop/electron-app/src/boundary-cli.js
```diff
@@ -33,13 +35,13 @@ module.exports = {
       `-target-id=${sanitized.target_id}`,
       `-token=${sanitized.token}`,
       `-addr=${addr}`,
-      '-format=json',
-      '--output-json-errors'
+      '-format=json'
```
gsusmi, line 38: Are json errors captured as part of this change in 0.1.6 or is it supported in next cli release?

gsusmi: desktop client release will pickup latest cli release. Resolving.

PR #490, opened 2021-02-27
Fix a tag parameter for the release pipeline

99:
Fixing tag issue to tigger release job

### DIFF
--- a/.github/workflows/ui-release-trigger.yml
+++ b/.github/workflows/ui-release-trigger.yml
@@ -15,5 +15,5 @@ jobs:
           echo "Triggering release with a TAG ${GITHUB_REF#refs/tags/}"
 
           curl -u ${{ secrets.CIRCLE_TOKEN }}: -X POST --header "Content-Type: application/json" \
-          -d '{ "branch": "main", "parameters": {"TAG": "${GITHUB_REF#refs/tags/}" }}' \
+          -d '{ "branch": "main", "parameters": {"TAG": ${GITHUB_REF#refs/tags/} }}' \
           https://circleci.com/api/v2/project/github/hashicorp/boundary-ui-releases/pipeline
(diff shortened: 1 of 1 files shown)


.github/workflows/ui-release-trigger.yml
```diff
@@ -15,5 +15,5 @@ jobs:
           echo "Triggering release with a TAG ${GITHUB_REF#refs/tags/}"
 
           curl -u ${{ secrets.CIRCLE_TOKEN }}: -X POST --header "Content-Type: application/json" \
-          -d '{ "branch": "main", "parameters": {"TAG": "${GITHUB_REF#refs/tags/}" }}' \
+          -d '{ "branch": "main", "parameters": {"TAG": ${GITHUB_REF#refs/tags/} }}' \
```
samsalisbury, line 18: I haven't tested this code to check, but this looks as though it will produce invalid JSON. The `parameters.TAG` field will have the literal value `${GITHUB_REF#refs/tags/}` (without the needed double quotes for JSON). You would need to add double quotes for JSON's sake, and then also use single quotes to break out of the literal JSON string, and allow bash to interpret the GITHUB_REF substitution and return its value. See suggestion below.

```suggestion
          -d '{ "branch": "main", "parameters": {"TAG": "'${GITHUB_REF#refs/tags/}'" }}' \
```

samsalisbury: What would be useful here is a test run that exercises this code path. I know this is hard to do, but it would give us much more certainty about correctness.

samsalisbury (approved):

.github/workflows/ui-release-trigger.yml
99: thank you for suggestion. Let me come up and add the test.

PR #506, opened 2021-03-26
Feature primary auth method

randallmorey:
### Background

Support for a **primary** auth method is coming soon as part of OIDC.  With primary support, only one auth method per scope will auto-vivify users.  While today all auth methods auto-vivify, in the future it's possible for _none_ to, which could lead to login failures.  It's important for users to understand the impact of the primary auth method and to elect one if needed.

### Implementation

Users may specify a primary auth method via the auth method form under "Manage".  A primary auth method may be removed as primary in the same dropdown.  When no primary is detected for the scope, a warning is issued to users of admin UI.  Lastly, the primary auth method is presented first at authentication time.

This feature is disabled by a feature flag until API support is released.  To test locally, enable the feature flag in `config/environment.js` and start the admin UI with mocks enabled.

Screenshot 2021-03-30 at 10 41 41
Screenshot 2021-03-30 at 10 42 09
Screenshot 2021-03-30 at 10 41 57

Screenshot 2021-03-29 at 18 48 34
Screenshot 2021-03-29 at 18 48 30
Screenshot 2021-03-29 at 16 54 27
Screenshot 2021-03-29 at 16 54 41
Screenshot 2021-03-29 at 16 54 17

### DIFF
--- a/addons/api/tests/unit/serializers/auth-method-test.js
+++ b/addons/api/tests/unit/serializers/auth-method-test.js
@@ -1,0 +1,66 @@
+import { module, test } from 'qunit';
+import { setupTest } from 'ember-qunit';
+
+module('Unit | Serializer | auth method', function (hooks) {
+  setupTest(hooks);
+
+  // Replace this with your real tests.
+  test('it exists', function (assert) {
+    let store = this.owner.lookup('service:store');
+    let serializer = store.serializerFor('auth-method');
+
+    assert.ok(serializer);
+  });
+
+  test('it serializes records', function (assert) {
+    let store = this.owner.lookup('service:store');
+    let record = store.createRecord('auth-method', {});
+
+    let serializedRecord = record.serialize();
+
+    assert.ok(serializedRecord);
+  });
+
+  test('it sorts primary first in normalizeArrayResponse', function (assert) {
+    assert.expect(2);
+    const store = this.owner.lookup('service:store');
+    const serializer = store.serializerFor('auth-method');
+    const modelClass = store.createRecord('auth-method').constructor;
+    const payload = {
+      items: [{ id: 1 }, { id: 2, primary: true }, { id: 3 }],
+    };
+    const normalizedArray = serializer.normalizeArrayResponse(
+      store,
+      modelClass,
+      payload
+    );
+    assert.ok(payload.items[1].primary, 'Second payload item is primary');
+    assert.deepEqual(
+      normalizedArray,
+      {
+        included: [],
+        data: [
+          {
+            id: '2',
+            type: 'auth-method',
+            attributes: { primary: true },
+            relationships: {},
+          },
+          {
+            id: '1',
+            type: 'auth-method',
+            attributes: {},
+            relationships: {},
+          },
+          {
+            id: '3',
+            type: 'auth-method',
+            attributes: {},
+            relationships: {},
+          },
+        ],
+      },
+      'First normalized itme is primary'
+    );
+  });
+});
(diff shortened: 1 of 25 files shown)


ui/admin/app/controllers/scopes/scope/auth-methods/index.js
```diff
@@ -0,0 +1,13 @@
...
+
+export default class ScopesScopeAuthMethodsIndexController extends Controller {
+  // =attributes
+
+  /**
+   * True if the current scope has a primary auth method set.
+   * @type {boolean}
+   */
+  get hasPrimaryAuthMethod() {
+    return this.scopeModel.primary_auth_method_id;
```
randallmorey, line 11: Hmm this actually needs to be wrapped in `Boolean()` to become a boolean.

mochanghc:
Some thoughts on the text for the confirmation dialog. It isn't clear as to why a user shouldn't remove the primary auth method or add one in the messaging. We could rephrase to


1)
! Remove primary auth method?
------------------------------------------
Scopes without a primary auth method can lead to login failures. Learn more
[Remove primary auth method] [Cancel]

2)
Do we need to confirm choosing a different primary auth method? I do not know how problematic this could be if someone chose the wrong auth method and needed to revert, etc. Minor problem? Major problem? 

3)
I don't know what auto-vivify means in this context. Could we change to something more understandable, more around the effect of not choosing a primary scope?

! Recommendation: Select a primary auth method
------------------------------------
Scopes without a primary auth method can lead to login failures. Learn more
[Select primary auth method]

ui/admin/app/templates/scopes/scope/auth-methods/auth-method/-header.hbs
```diff
@@ -1,5 +1,14 @@
 <h2>
   {{t "resources.auth-method.title"}}
+  {{#if (feature-flag 'primary-auth-method')}}
+    {{#if @model.isPrimary}}
+      &nbsp;
+      <Rose::Badge @style="success-outline" @size="">
```
gsusmi, line 6: Can default size be used instead of nullifying it?

ui/admin/config/environment.js
```diff
@@ -83,6 +83,10 @@ module.exports = function (environment) {
     'ember-cli-mirage': {
       directory: '../../addons/api/mirage',
     },
+
+    featureFlags: {
+      'primary-auth-method': false,
+    },
```
gsusmi, line 89: Wondering if we should introduce feature flags in dev mode first and then move them upto production to avoid enabling them in prod by accident. Thoughts?

ui/admin/tests/acceptance/auth-methods-test.js
```diff
@@ -10,6 +10,7 @@ import {
   //   //currentSession,
   //   //invalidateSession,
 } from 'ember-simple-auth/test-support';
+import { enableFeature } from 'ember-feature-flags/test-support';
```
gsusmi, line 13: Love the test support for feature flags!

addons/api/tests/unit/serializers/auth-method-test.js
```diff
@@ -0,0 +1,66 @@
...
+          },
+          {
+            id: '3',
+            type: 'auth-method',
+            attributes: {},
+            relationships: {},
+          },
+        ],
+      },
+      'First normalized itme is primary'
```
gsusmi, line 63: `item` typo.

gsusmi (approved):
Looks great! Couple of minor notes, nothing to block PR merge.

PR #516, opened 2021-04-02
Add static auth callback pages to admin UI for notifying success or error.

randallmorey:
`/authentication-complete`
`/authentication-error`

Screenshot 2021-04-02 at 16 01 15
Screenshot 2021-04-02 at 16 01 03

### DIFF
--- a/addons/core/translations/en-us.yaml
+++ b/addons/core/translations/en-us.yaml
@@ -2,13 +2,13 @@ auth:
   titles:
     no-scopes: No Scopes Available
     flow-success: Successful authentication
-    flow-error: Could not authenticate to provider
+    flow-error: Could not authenticate
   descriptions:
     no-scopes: No scopes were found.  To authenticate, ask your administrator to create an scope and grant you access.
     choose-org: Choose a scope to authenticate.
     authenticating-with-org: "You are authenticating with {name}."
     flow-success: You've successfully authenticated.  You may now close this window and return to the client.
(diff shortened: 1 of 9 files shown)


addons/core/translations/en-us.yaml
```diff
@@ -1,10 +1,14 @@
 auth:
   titles:
     no-scopes: No Scopes Available
+    flow-success: Successful authentication
+    flow-error: Could not authenticate to provider
```
gsusmi, line 5: Could we eliminate `provider`? As we are using `authentication` instead of identifying oidc or provider in this route, we could simplify this:
> Could not authenticate ~~to provider~~
> Something went wrong while authenticating ~~with the provider~~.  Please close this window and try again.  If the problem persists, notify your administrator.

gsusmi (approved):
LGTM! Ty! Minor note about using `provider` in message.

addons/core/translations/en-us.yaml
jimlambrt: +1 vote on @gsusmi suggestions.  Most users will have no context to understand what a provider is.

jimlambrt (approved):
ty!

addons/core/translations/en-us.yaml
randallmorey: Good point!  I even intentionally made the translation keys generic in anticipation of (get excited) SAML.  And yet the content is not quite generic.  Will update.

PR #518, opened 2021-04-08
Style auth form

randallmorey:
This PR introduces a new component `Rose::Frame`, which is applied to the admin auth form per latest designs.

Screenshot 2021-04-07 at 11 55 11


gsusmi (approved):
LGTM. Ty!

PR #519, opened 2021-04-08
fix: 🐛 silently drop trailing slashes from origin input - ICU-1293

randallmorey:
This resolves an issue where Boundary Desktop may generate API endpoint URLs with double slashes if the origin includes a trailing slash.  The solution is to silently drop trailing slashes  should the user include them.  This ensures users may enter origins with or without a trailing slashes and Boundary Desktop will still work as expected.


gsusmi (approved):
Ty!

PR #525, opened 2021-04-13
fix: Trim origin url in desktop

gsusmi:


ui/desktop/app/services/origin.js
```diff
@@ -66,7 +66,8 @@ export default class OriginService extends Service {
       // Silently drop trailing slashes if present.
       // This is important since API paths that will be appended to the origin
       // will include preceeding slashes already.
-      origin = origin.replace(/\/*$/, '');
+      // Trim whitespaces around origin.
+      origin = origin.replace(/\/*$/, '').trim();
```
randallmorey, line 70: We may want to trim first, before removing slashes.  It's common for whitespace to appear at the beginning or end of the string.  The regex is looking for slashes at the immediate end, meaning that if there is whitespace at the end `foobar//    `, the slashes would remain and then the whitespace removed, resulting in `foobar//`.  The result we want is `foobar`.

randallmorey (changes requested):

ui/desktop/app/services/origin.js
gsusmi: 👍 Good catch. Updated

PR #532, opened 2021-04-14
fix: Check for update archive before prompting download

gsusmi:


gsusmi:
Update: Refactor out suppressing dialog prompt from function that shows it.

PR #539, opened 2021-04-28
chore: 🤖 Update node version engines

calcaide:
Update the node versions our UI's can use. These versions are specified
on the engines key in the package.json of each UI (admin and desktop).
Our UI's can run from node v10 to v14, we can not run on v15 because
ember-cli does not support it yet (info [here](https://github.com/ember-cli/ember-cli/blob/master/docs/node-support.md)).

**How to test?**
 I use [nvm](https://github.com/nvm-sh/nvm) to switch node version in my dev environment.
- Run `yarn run start` using `node` version `10`, `12` and `14`, should run no problem.
- Run `yarn run start` using `node` version `15`, should not run. You should see in the terminal a message like this `The engine "node" is incompatible with this module. Expected version ">=10 <=14". Got "15.14.0"`

### DIFF
--- a/addons/rose/package.json
+++ b/addons/rose/package.json
@@ -87,7 +87,7 @@
     "sass-lint": "^1.13.1"
   },
   "engines": {
-    "node": ">=10 <=14"
+    "node": ">=10 <=15"
   },
   "ember": {
     "edition": "octane"
--- a/addons/rose/package.json
+++ b/addons/rose/package.json
@@ -36,8 +36,7 @@
     "ember-named-blocks-polyfill": "^0.2.4",
     "ember-radio-button": "^2.0.1",
     "ember-uuid": "^2.1.0",
-    "emberx-select": "github:adopted-ember-addons/emberx-select#bc5f774",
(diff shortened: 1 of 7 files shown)


randallmorey (changes requested):
I suggest adding this to the other packages as well:  addons and the top-level workspaces package.

addons/rose/package.json
```diff
@@ -36,8 +36,7 @@
     "ember-named-blocks-polyfill": "^0.2.4",
     "ember-radio-button": "^2.0.1",
     "ember-uuid": "^2.1.0",
-    "emberx-select": "github:adopted-ember-addons/emberx-select#bc5f774",
-    "ember-named-blocks-polyfill": "^0.2.4"
```
calcaide, line 40: @randallmorey the deletion of `ember-named-blocks-polyfill` is unrelated to topic of this PR, but as you can see in the line 36, it is already included, so is a duplicate package.

randallmorey: Good catch.  Thank you!

randallmorey (approved):

PR #545, opened 2021-05-05
feat: 🎸 Add Form fieldset component

calcaide:
Add a simple form fieldset component, basic storybook page and tests.

✅ Closes: ICU-1428

Screen Shot 2021-05-05 at 2 05 42 PM

### DIFF
--- a/addons/rose/addon/components/rose/form/fieldset/index.hbs
+++ b/addons/rose/addon/components/rose/form/fieldset/index.hbs
@@ -1,0 +1,20 @@
+<fieldset ...attributes class='rose-form-fieldset'>
+    <legend 
+        id={{concat 'title-' this.id}}
+        aria-describedby={{concat 'description-' this.id}}
+        class='rose-form-fieldset-title'
+    >
+        {{yield to='title'}}
+    </legend>
+    <div
+        id={{concat 'description-' this.id}}
+        aria-labelledby={{concat 'title-' this.id}}
+        class='rose-form-fieldset-description'
+    >
+        {{yield to='description'}}
+    </div>
+
+    <div class='rose-form-fieldset-body'>
+        {{yield to='body'}}
+    </div>
+</fieldset>
--- a/addons/rose/addon/components/rose/form/fieldset/index.hbs
+++ b/addons/rose/addon/components/rose/form/fieldset/index.hbs
@@ -1,19 +1,20 @@
-<fieldset ...attributes class='rose-form-fieldset'>
+<fieldset ...attributes class='rose-form-fieldset' aria-describedby={{concat 'title-' this.id}}>
     <legend 
         id={{concat 'title-' this.id}}
-        aria-describedby={{concat 'description-' this.id}}
         class='rose-form-fieldset-title'
     >
         {{yield to='title'}}
     </legend>
-    <div
-        id={{concat 'description-' this.id}}
-        aria-labelledby={{concat 'title-' this.id}}
-        class='rose-form-fieldset-description'
-    >
-        {{yield to='description'}}
-    </div>
-
+    
+    {{#if (has-block "description")}}
+        <div
+            id={{concat 'description-' this.id}}
+            class='rose-form-fieldset-description'
+        >
+            {{yield to='description'}}
+        </div>
+    {{/if}}
+    
     <div class='rose-form-fieldset-body'>
         {{yield to='body'}}
     </div>
--- a/addons/rose/addon/components/rose/form/fieldset/index.hbs
+++ b/addons/rose/addon/components/rose/form/fieldset/index.hbs
@@ -1,4 +1,4 @@
-<fieldset ...attributes class='rose-form-fieldset' aria-describedby={{concat 'title-' this.id}}>
+<fieldset ...attributes class='rose-form-fieldset' aria-describedby={{if (has-block "description") (concat 'description-' this.id)}}>
     <legend 
         id={{concat 'title-' this.id}}
         class='rose-form-fieldset-title'
--- a/addons/rose/tests/integration/components/rose/form/fieldset-test.js
+++ b/addons/rose/tests/integration/components/rose/form/fieldset-test.js
@@ -11,19 +11,29 @@ module('Integration | Component | rose/form/fieldset', function (hooks) {
   this.mockDescription = 'Description';
   this.mockBody = 'Rest of the form';
 
-  hooks.beforeEach(async function () {
-    // Render the component before each test
-    await render(hbs`
+  // Wrap the rendering in a function to avoid repeat code
+  async function renderComponent(noDescription) {
+    if (noDescription === true) {
+      return render(hbs`
+        <Rose::Form::Fieldset>
+          <:title>{{this.mockTitle}}</:title>
+          <:body>{{this.mockBody}}</:body>
+        </Rose::Form::Fieldset>
+    `);
+    }
+
+    return render(hbs`
       <Rose::Form::Fieldset>
         <:title>{{this.mockTitle}}</:title>
         <:description>{{this.mockDescription}}</:description>
         <:body>{{this.mockBody}}</:body>
       </Rose::Form::Fieldset>
     `);
-  });
+  }
 
-  test('it renders', async function (assert) {
+  test('it renders the full component', async function (assert) {
     assert.expect(4);
+    await renderComponent(false);
     assert.ok(find('.rose-form-fieldset'));
     assert.equal(
       find('.rose-form-fieldset-title').textContent.trim(),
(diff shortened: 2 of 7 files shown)


addons/rose/addon/components/rose/form/fieldset/index.hbs
```diff
@@ -0,0 +1,20 @@
+<fieldset ...attributes class='rose-form-fieldset'>
+    <legend 
+        id={{concat 'title-' this.id}}
+        aria-describedby={{concat 'description-' this.id}}
+        class='rose-form-fieldset-title'
+    >
+        {{yield to='title'}}
+    </legend>
+    <div
```
randallmorey, line 9: Will everything render as expected if a description is not passed?

addons/rose/addon/components/rose/form/fieldset/index.hbs
```diff
@@ -0,0 +1,20 @@
+<fieldset ...attributes class='rose-form-fieldset'>
+    <legend 
+        id={{concat 'title-' this.id}}
+        aria-describedby={{concat 'description-' this.id}}
+        class='rose-form-fieldset-title'
+    >
+        {{yield to='title'}}
```
randallmorey, line 7: I like the use of named blocks.

addons/rose/tests/integration/components/rose/form/fieldset-test.js
```diff
@@ -0,0 +1,59 @@
...
+    );
+  });
+
+  test('it supports aria description for title', async function (assert) {
+    assert.expect(2);
+    const titleElement = find('.rose-form-fieldset-title');
+    const descriptionElement = find('.rose-form-fieldset-description');
+
+    assert.equal(
+      titleElement.getAttribute('aria-describedBy'),
```
randallmorey, line 48: It appears this is passing due to case-insensitivity.  But ideally the attributes here would be all lowercase to match the actual template above.

addons/rose/tests/integration/components/rose/form/fieldset-test.js
```diff
@@ -0,0 +1,59 @@
...
+      descriptionElement.id,
+      'Title is described by description'
+    );
+
+    assert.equal(
+      descriptionElement.getAttribute('aria-labelledBy'),
```
randallmorey, line 54: It appears this is passing due to case-insensitivity.  But ideally the attributes here would be all lowercase to match the actual template above.

addons/rose/addon/components/rose/form/fieldset/index.hbs
```diff
@@ -0,0 +1,20 @@
...
+    </legend>
+    <div
+        id={{concat 'description-' this.id}}
+        aria-labelledby={{concat 'title-' this.id}}
```
randallmorey, line 11: I think in this case labeledby isn't unnecessary, since `legend` serves this semantic purpose and should be understood by assistive technology.

addons/rose/addon/components/rose/form/fieldset/index.hbs
```diff
@@ -0,0 +1,20 @@
+<fieldset ...attributes class='rose-form-fieldset'>
+    <legend 
+        id={{concat 'title-' this.id}}
+        aria-describedby={{concat 'description-' this.id}}
```
randallmorey, line 4: `describedby` would go on the `fieldset`.

randallmorey (changes requested):

addons/rose/addon/components/rose/form/fieldset/index.hbs
calcaide: @randallmorey Do you mean the `aria-describedby` pertains to the fieldset and then the `aria-labelledby` should be move to legend?

addons/rose/addon/components/rose/form/fieldset/index.hbs
calcaide: @gsusmi show me about them so I kindda cheat here :stuck_out_tongue_winking_eye:

addons/rose/tests/integration/components/rose/form/fieldset-test.js
calcaide: Ups! really miss that one. Already fixed in both assertions (will push it with the rest of changes). Nice catch!

addons/rose/addon/components/rose/form/fieldset/index.hbs
calcaide: Yes. I just tested with this code as render no problem:
```javascript
<Rose::Form::Fieldset>
  <:title>Title</:title>
  <:body>Rest of the form</:body>
</Rose::Form::Fieldset>
```

addons/rose/addon/components/rose/form/fieldset/index.hbs
randallmorey: Fieldset should have the `aria-describedby` attribute.  And we can probably eliminate the `aria-labelledby`?  `legend` semantically serves the purpose so there should be no ambiguity to assistive tech.

addons/rose/addon/components/rose/form/fieldset/index.hbs
randallmorey: Yes I think this will be important to handle because of the `aria-describedby`.  It may be a poor experience for some users if they rely on that attribute and there is no associated text.

addons/rose/addon/components/rose/form/fieldset/index.hbs
calcaide: Ahh gotcha that makes sense

addons/rose/addon/components/rose/form/fieldset/index.hbs
calcaide: That is right. Even I am deleting the `aria-describedby` as per your suggestion, I agree if does not receive description to not render it.

gsusmi (approved):
Loving the discussions. Nothing to add from my end.

addons/rose/addon/components/rose/form/fieldset/index.hbs
```diff
@@ -0,0 +1,21 @@
+<fieldset ...attributes class='rose-form-fieldset' aria-describedby={{concat 'title-' this.id}}>
+    <legend 
+        id={{concat 'title-' this.id}}
+        class='rose-form-fieldset-title'
+    >
+        {{yield to='title'}}
+    </legend>
+    
+    {{#if (has-block "description")}}
```
calcaide, line 9: @randallmorey this fix the issue of rendering an empty `<div>` is description is not provided.

addons/rose/tests/integration/components/rose/form/fieldset-test.js
```diff
@@ -0,0 +1,79 @@
...
+module('Integration | Component | rose/form/fieldset', function (hooks) {
+  setupRenderingTest(hooks);
+
+  // Global values for mocking
+  this.mockTitle = 'Title';
+  this.mockDescription = 'Description';
+  this.mockBody = 'Rest of the form';
+
+  // Wrap the rendering in a function to avoid repeat code
+  async function renderComponent(noDescription) {
```
calcaide, line 15: @randallmorey Here I changed the test. Previously I wrapped the rendering function within `beforeEach()` but since we need to render without description too this is the solution I came with.

randallmorey (changes requested):

addons/rose/addon/components/rose/form/fieldset/index.hbs
```diff
@@ -0,0 +1,21 @@
+<fieldset ...attributes class='rose-form-fieldset' aria-describedby={{concat 'title-' this.id}}>
```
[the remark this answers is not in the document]
randallmorey: Try `aria-describedby={{if (has-block "description") (concat 'title-' this.id)}}`.  It _might_ work but needs to be validated.

randallmorey (changes requested):

addons/rose/tests/integration/components/rose/form/fieldset-test.js
```diff
@@ -0,0 +1,79 @@
...
+    if (noDescription === true) {
+      return render(hbs`
+        <Rose::Form::Fieldset>
+          <:title>{{this.mockTitle}}</:title>
+          <:body>{{this.mockBody}}</:body>
+        </Rose::Form::Fieldset>
+    `);
+    }
```
gsusmi, line 23: Wondering if `renderComponent` should always render the full component (without any conditionals) and the test for missing description should define it's own template without using `renderComponent`. This approach would clarify why missing description block is not a normal state. Thoughts?

addons/rose/addon/components/rose/form/fieldset/index.hbs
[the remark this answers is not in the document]
calcaide: That works like a charm, thanks Randy!!

addons/rose/tests/integration/components/rose/form/fieldset-test.js
calcaide: I see it as description is an optional element of the fieldset. If is provided awesome, but if is not still a normal state (IMHO).
I create the function because I was repeating code in each test case, and if you check the last changes on the tests we will be repeating code even more since I just add a new test for aria when description is not provided.
But happy to discuss and change the approach.

randallmorey: If only we had a standardized way to document component inputs.  :sigh:  This part of Ember still feels a little rough.

randallmorey (approved):

PR #548, opened 2021-05-08
fix: 🐛 migrate IPC to service-based window for reliable testing

randallmorey:
This resolves a longstanding issue where IPC tests rely on the real
window.postMessage and MessageChannel APIs.  These are evented and
asynchronous, resulting in nondeterministic tests.  This commit updates
tests with synchronous mocked APIs.


gsusmi (approved):
I was wondering last week why we haven't seen any origin tests fail recently :). Ty for updating!

PR #549, opened 2021-05-11
feat: Close local proxy on session cancellation in Desktop

gsusmi:
On a session cancellation through in desktop ui, close spawned node process that handles boundary connect command.


ui/desktop/electron-app/src/spawn-session.js
```diff
@@ -0,0 +1,44 @@
+const process = require('process');
```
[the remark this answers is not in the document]
calcaide: I am not the best when it comes to English and I do not have a strong opinion, but seeing the code comments I would say `spawned` feels more appropriate?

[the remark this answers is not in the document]
randallmorey: I'm increasingly concerned about the growing number of unorganised source files in electron-app.  Something to think about.

Should we have a proper data model/class for sessions?  Maybe refactor the current sessions code to take an OOP approach.  That said, this feels like it should be combined with the code that actual spawns the sessions.  It is more than just a data structure, since it is empowered to kill sessions.

ui/desktop/electron-app/src/spawn-promise.js
```diff
@@ -45,22 +46,37 @@ module.exports = {
...
         const jsonData = jsonify(outputStream);
-        if (jsonData) resolve(jsonData);
+        if (jsonData) {
+          if (jsonData.termination_reason) {
+            spawnSession.cancelProcess(childProcess.pid);
+          } else {
+            // Track only successfully launched child processes
+            spawnSession.add({ childProcess, data: jsonData });
+          }
+          outputStream = '';
```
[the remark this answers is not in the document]
calcaide: Indeed. Because `outputStream` is scoped to the function, I think could be declared as `const` on line 47. So we do not need to concat values and clean it up at the end?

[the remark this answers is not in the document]
randallmorey: Ah.  A promise may only be resolved or rejected once.  So it doesn't make sense to continue receiving input after the initial resolution.  Can the data event handler be removed after resolution?

ui/desktop/electron-app/src/spawn-promise.js
```diff
@@ -45,22 +46,37 @@ module.exports = {
...
       });
 
+      // @todo Cancel spawned session childprocess on error? Does it get automatically cancelled?
+      // Test on connection limit errors especially
       childProcess.stderr.on('data', (data) => {
         errorStream += data.toString();
         const jsonData = jsonify(errorStream);
         if (jsonData) {
           const error = jsonData.api_error || jsonData.error;
+          errorStream = '';
```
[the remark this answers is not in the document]
calcaide: Ahhh gotcha, miss that one! thanks for the explanation!

ui/desktop/electron-app/src/session/index.js
```diff
@@ -0,0 +1,118 @@
...
+    return this.#proxyDetails;
+  }
+
+  /**
+   * Using cli, initialize a session to a target.
+   * Tracks local proxy details if successful.
+   */
+  start() {
+    const command = this.cliCommand();
+    return spawnAsyncJSONPromise(command).then((spawnedSession) => {
```
randallmorey, line 73: You might return proxy details from the start promise to make it easier to get these.

ui/desktop/electron-app/src/session/manager.js
```diff
@@ -0,0 +1,51 @@
...
+   * Returns session proxy details if successfully started.
+   * Only successfully started sessions will be tracked.
+   * @param {string} addr
+   * @param {string} target_id
+   * @param {string} token
+   * @param {string} host_id
+   */
+  async start(addr, target_id, token, host_id) {
+    const session = new Session(addr, target_id, token, host_id);
+    return session.start().then(() => {
```
randallmorey, line 21: Suggestion:
```
this.#sessions.push(session);
return session.start(); // the start promise could return proxy details.
```

ui/desktop/electron-app/src/session/manager.js
```diff
@@ -0,0 +1,51 @@
...
+   * @param {string} session_id
+   */
+  async stopById(session_id) {
+    const index = this.#sessions.findIndex(
+      (session) => session.id === session_id
+    );
+    if (index === -1) return;
+    const session = this.#sessions[index];
+    return session.stop().then(() => {
+      this.#sessions.splice(index, 1);
```
randallmorey, line 38: I'd actually leave sessions in the array in case we want to do anything with them, even after they're killed.

ui/desktop/electron-app/src/session/manager.js
```diff
@@ -0,0 +1,51 @@
...
+      return session.proxyDetails;
+    });
+  }
+
+  /**
+   * Stop a session and remove it from tracked sessions.
+   * @param {string} session_id
+   */
+  async stopById(session_id) {
+    const index = this.#sessions.findIndex(
```
randallmorey, line 32: would `const session = this.#sessions.find((session) => session.id === sessionID)` work?

ui/desktop/electron-app/src/session/manager.js
```diff
@@ -0,0 +1,51 @@
...
+    return session.stop().then(() => {
+      this.#sessions.splice(index, 1);
+    });
+  }
+
+  stopAll() {
+  }
+
+  get sessions() {
+    return this.#sessions;
```
randallmorey, line 46: Why expose this?

ui/desktop/electron-app/src/session/manager.js
gsusmi: My approach was to only track successful session processes so that I could cancel all of them on quitting the app with a simple loop. I will reconfigure this to check for an active session. It'll simplify code as you suggested.

ui/desktop/electron-app/src/session/manager.js
gsusmi: I was playing around with interface. Will delete.

gsusmi:
👋 @randallmorey , @calcaide : Ready for another review. Ty for your patience!

ui/desktop/app/routes/scopes/scope/projects/sessions.js
```diff
@@ -83,5 +84,6 @@ export default class ScopesScopeProjectsSessionsRoute extends Route {
   @notifySuccess('notifications.canceled-success')
   async cancelSession(session) {
     await session.cancelSession();
+    await this.ipc.invoke('cancel', { session_id: session.id });
```
randallmorey, line 87: On the IPC side, it's more of a "stop" or "kill" concept if the session has already been canceled via the API.

ui/desktop/electron-app/src/helpers/spawn-promise.js
```diff
@@ -42,25 +33,37 @@ module.exports = {
...
       let errorStream = '';
 
-      childProcess.stdout.on('data', (data) => {
+      const processData = (data) => {
         outputStream += data.toString();
         const jsonData = jsonify(outputStream);
-        if (jsonData) resolve(jsonData);
-      });
+        if (jsonData) {
+          childProcess.removeListener('data', processData);
```
randallmorey, line 40: Probably want to remove both event listeners (including error)?  And remove both in the error handler below too.

ui/desktop/electron-app/src/session/manager.js
```diff
@@ -0,0 +1,38 @@
...
+  async stopById(session_id) {
+    const session = this.#sessions.find((session) => session.id === session_id);
+    return session.stop();
+  }
+
+  stopAll() {
+  }
+}
+
+const sessionManager = new SessionManager();
```
randallmorey, line 37: I suggest instantiating this outside the class file so that the class itself can be reused if we wanted to.

gsusmi:
To test:

1. Launch desktop client and connect to a boundary controller
2. Connect to a target
3. Verify creation of new boundary process using 'Activity Monitor' or 'ps'. 
4. Cancel session to target in desktop.
5. Verify that previously created process no longer exists.

ui/desktop/electron-app/src/session/index.js
```diff
@@ -0,0 +1,111 @@
...
+    const sanitized = {
+      target_id: escapeAndValidateBase62(this.#targetId),
+      token: escapeAndValidateBase62(this.#token),
+    };
+
+    const command = [
+      'connect',
+      `-target-id=${sanitized.target_id}`,
+      `-token=${sanitized.token}`,
+      `-addr=${this.#addr}`,
```
randallmorey, line 99: This unsanitized input is bugging me a little.  We know it's highly unlikely to be unsafe since we validate the origin.  But considering the code in isolation, this feels a bit unsafe.  It can't use the same base62 scheme but is there a quick and dirty way to improve this situation?

ui/desktop/electron-app/src/session/manager.js
```diff
@@ -0,0 +1,38 @@
+const Session = require('./index.js');
+
+class SessionManager {
+  #sessions;
+
+  constructor() {
+    this.#sessions = [];
```
randallmorey, line 7: Isn't this equivalent to initializing it in the declaration above?

randallmorey (commented):
Can you do a quick audit to remove `async` from functions that don't use `await`?

ui/desktop/electron-app/src/helpers/spawn-promise.js
gsusmi: Good point. Will update.

ui/desktop/electron-app/src/session/manager.js
gsusmi: Enforced creation on initial load to ensure there will only ever be one session manager for the app. Would you recommend updating current folder structure to include `models` to contain class files?

ui/desktop/electron-app/src/session/manager.js
gsusmi: Focused on maintaining a class pattern by defining private vars first, followed by constructors, getters, setters, and methods. I see your point.

ui/desktop/electron-app/src/session/manager.js
randallmorey: It may be a singleton, but what if we wanted to unit test the class?  Or what if we wanted to segregate sessions by some arbitrary characteristic using multiple managers?  It is bit more flexible to separate the class from the instantiation even for singletons.

calcaide: I see the point of this. But for learning purposes, where will be a good place to instantiate this? within the `electron-app` or within `desktop`? Just trying to figure out good practices here.

randallmorey: I'm noticing that `RuntimeSettings` uses this pattern so let's set this discussion aside for another time and address both cases at once.

ui/desktop/app/routes/scopes/scope/projects/sessions.js
gsusmi: Updated to 'stop'. Ty!

ui/desktop/electron-app/src/helpers/spawn-promise.js
gsusmi: Updated to remove all listeners before resolving/rejecting promise.

ui/desktop/electron-app/src/session/manager.js
gsusmi: I took a stab at this by moving `Session` and `SessionManager` class definitions to a new `models` folder. Singleton for session manager now exists under `src/session/manager.js`. Happy to refactor `RuntimeSettings` if this approach makes sense.

ui/desktop/electron-app/src/session/index.js
gsusmi: Attempted to use `encodeURI` to mimic sanitization. Any better alternatives?

randallmorey: Not sure.  We need to prevent injection attacks and `encodeURI` allows most punctuation characters through, including semicolons.  This could still be used for script injection.  https://auth0.com/blog/preventing-command-injection-attacks-in-node-js-apps/

ui/desktop/electron-app/src/session/manager.js
randallmorey: Looks good.  I wonder if "services" or "utils" is more applicable since something like the manager isn't exactly a model.  It's debatable whether a session is in this case.

calcaide: Ok now I understand this better. I do not have a strong opinion, but I see your point here. That said, I see it more as a service than as util.

calcaide (approved):
LGTM! Performed testing and works as expected. When the session is terminated on the desktop client, the process is kill on the activity monitor. Awesome!

ui/desktop/electron-app/src/session/index.js
calcaide: Very interesting point here. I will do some research about this too. In case we can't find something "default" create a helper function to sanitize URI's within `core` addon (so we can share it between admin and desktop) is a possibility?

ui/desktop/electron-app/src/session/manager.js
gsusmi: I'm leaning towards 'services' too. Will update.

ui/desktop/electron-app/src/session/index.js
randallmorey: I'm not sure there's much ability to share code between electron (Node) and non-electron (Ember) code.  The article above cites this tool:  https://www.npmjs.com/package/shell-quote for escaping shell inputs.

ui/desktop/electron-app/src/session/manager.js
gsusmi: Moved 'session/manager' to 'services/session-manager' plus additional refactoring of origin runtimesettings into new 'services/origin'.

gsusmi:
Pending change - figure out what to do with origin address sanitization.

gsusmi:
Update: Add origin url validation using nodejs [url api](https://nodejs.org/api/url.html).

ui/desktop/electron-app/src/utils/sanitizer.js
```diff
@@ -0,0 +1,33 @@
...
+      Could not invoke command:
+      input contained unsafe characters.
+    `);
+  },
+  /**
+   * Return a quoted string suitable for using in shell commands.
+   * @param {string} str
+   * @returns {string}
+   */
+  urlValidate: (str) => {
```
calcaide, line 21: Nice!

ui/desktop/electron-app/src/services/origin.js
```diff
@@ -0,0 +1,8 @@
+const RuntimeSettings = require('../models/runtime-settings.js');
+
+// Provides a singleton class instance to enable a consistent view of
```
randallmorey, line 3: Runtime settings is more general than the "origin" filename would suggest.

randallmorey: Runtime settings is more of a service.

ui/desktop/electron-app/src/utils/sanitizer.js
```diff
@@ -0,0 +1,33 @@
...
+    try {
+      const url = new URL(str);
+      return url.href;
```
randallmorey, line 24: Probably want to be a bit more intentional here.  `href` will return most of the URL:

```
> new URL('http://localhost:9200?foo=bars;test').href
'http://localhost:9200/?foo=bars;test'
```

I suggest reconstructing the origin: `${url.protocol}://${url.hostname}:${url.port}`

randallmorey: Which raises the question, should this parsing and reconstitution happen within the runtime settings when origin is set in the first place?

randallmorey (approved):

ui/desktop/electron-app/src/services/origin.js
gsusmi: This was a bit confusing - as to why did we name it `origin` in the first place. I'll update it to `runtimeSettings` service for future sanity.

gsusmi: Updated.

ui/desktop/electron-app/src/utils/sanitizer.js
gsusmi: When origin is being updated, it will now be validated that it is a valid url in addition to verifying scopes api reachability. I left boundary command url validation as is for now. We can remove it in the future if it is found to be frivolous.

PR #550, opened 2021-05-11
fix: 🐛 Update trim package version resolution

calcaide:
Update the version we resolve of dependency package trim since older version has a security vulnerability.

Tests performed:
- Run all the tests suite.
- Run `yarn run build:ui:admin` and `yarn run build:ui:desktop` to make sure this does not brake anything.


gsusmi (approved):
LGTM. Ty!

PR #554, opened 2021-05-11
Refactor rose table component

calcaide:
We refactor the component adding a condensed variant of the table, refactor the hidden feature. We also provide different alignments for a cell. Updated tests and storybook page.


addons/rose/tests/dummy/app/templates/index.hbs
```diff
@@ -1,93 +1,208 @@
...
+      <row.headerCell >Address</row.headerCell>
+      <row.headerCell>Host Set</row.headerCell>
+      <row.headerCell>Name</row.headerCell>
+      <row.headerCell>Description</row.headerCell>
+    </header.row>
+  </table.header>
+  <table.body as |body|>
+    <body.row as |row|>
+      <row.cell>1.1.1.1</row.cell>
+      <row.cell @alignCenter={{true}}>host-set-1, host-set-2</row.cell>
```
calcaide, line 66: Because the linter, adding mustaches on boolean values. I do follow this approach because I saw it in our code base, but if there is other preferred way to go, please let me know.

calcaide: Also, decided to go for a boolean approach rather than string such as `@align="center"` because we do not expect more than two options, align center and right. Left alignment is by default.

addons/rose/tests/integration/components/rose/table/row/cell-test.js
```diff
@@ -0,0 +1,70 @@
...
+    assert.notOk(find('.rose-table-cell'));
+  });
+
+  test('it adds a style class', async function (assert) {
+    assert.expect(1);
+    await render(hbs`<Rose::Table::Row::Cell @style='expand' />`);
+    assert.ok(find('.rose-table-cell-expand'));
+  });
+
+  test('it renders with align center in a header cell', async function (assert) {
```
calcaide, line 39: Since we allow alignment in header cells and body cells, I found best to split tests for it. But let me know if you have a better way of doing so.

calcaide:
@gsusmi OMG!! my bad!! While working in local I was adding different commits with not much context. When I finished I wanted to re-do my none sense commits to one explaining properly what has been done and I think I mess it up... I am so sorry!

gsusmi:
No worries at all! Just curious. :)

addons/rose/addon/styles/rose/components/table/_table.scss
```diff
@@ -35,7 +39,33 @@
     }
   }
 
+  // Grow to full available width
+  .rose-table-cell-expand {
+    width: 100%;
+
+    & > * {
```
randallmorey, line 46: For performance reasons, we should avoid `*` selectors.  What is the purpose of this block and can it be rewritten?

addons/rose/addon/styles/rose/components/table/_table.scss
```diff
@@ -13,6 +13,10 @@
   .rose-table-cell,
   .rose-table-header-cell {
     vertical-align: middle;
+    white-space: nowrap;
```
randallmorey, line 16: This seems too prescriptive doesn't it?  This isn't always desirable.

addons/rose/addon/styles/rose/components/table/_table.scss
```diff
@@ -35,7 +39,33 @@
     }
   }
 
+  // Grow to full available width
+  .rose-table-cell-expand {
```
randallmorey, line 43: I was thinking the main use case is actually for a "shrink" concept rather than expand.

addons/rose/addon/styles/rose/components/table/_table.scss
```diff
@@ -35,7 +39,33 @@
...
+    }
+  }
+
   [class*='rose-button'] {
     margin-bottom: 0;
   }
+
+  &.rose-table-condensed {
+    .rose-table-cell {
+      @include type.type(s);
```
randallmorey, line 58: Is the type actually small in the designs?

addons/rose/addon/styles/rose/components/table/_table.scss
```diff
@@ -35,7 +39,33 @@
...
+      $lineHeight: 36 / 14;
```
randallmorey, line 59: The type mixing can handle line height.  Is it nonstandard?

addons/rose/addon/components/rose/table/index.stories.mdx
```diff
@@ -60,6 +87,76 @@ A table component to display data in structure of rows and columns. It supports
...
+</Preview>
+
+<Preview>
+  <Story name="Basic Key/Value Table">{{
+    template: hbs`
+      <Rose::Table @style="condensed" as |table|>
+        <table.header as |header|>
+          <header.row @hidden=true as |row|>
+            <row.headerCell>Key</row.headerCell>
+            <row.headerCell @style="expand">Value</row.headerCell>
```
randallmorey, line 119: This doesn't really feel like a "style".  It could be a boolean.  In either case, I think the need is more for a "shrink" concept.  `@shrink={{true}}`

randallmorey (changes requested):

addons/rose/addon/styles/rose/components/table/_table.scss
gsusmi: `condensed` styles have been extracted from existing key/value component styles as one of the goals on this PR was to make key/value component obsolete.

addons/rose/addon/styles/rose/components/table/_table.scss
gsusmi: Same note as above regarding extracting styles from key/value component for condensed table style.

addons/rose/addon/styles/rose/components/table/_table.scss
randallmorey: I doubled checked and the designs use 14px or "small" so this should be fine.

addons/rose/addon/components/rose/table/index.stories.mdx
gsusmi: Could we use a `@expand={{true}}` and reuse existing styles? Mainly coz it's easier to apply a width of 100% cell rather than figuring out how small a cell (including it's header) can possibly be.

addons/rose/addon/components/rose/table/index.stories.mdx
```diff
@@ -1,27 +1,44 @@
 import { Meta, Story, Preview } from '@storybook/addon-docs/blocks';
 import { hbs } from 'ember-cli-htmlbars';
+import { boolean, select } from '@storybook/addon-knobs';
 
 <Meta title="Rose|Table" component="Table" />
 
 # Table
 
-A table component to display data in structure of rows and columns. It supports `header`, `body` and `footer` optional definitions. Each definition allows for a table `row` to be defined with either a `cell` or `headerCell` table element.
+A table component to display data in structure of rows and columns. It supports `header`, `body` and `footer` optional definitions. Each definition allows for a table `row` to be defined with either a `cell` or `headerCell` table element. Rows can be visually hidden, styled for a key/value visualization, cells can be shrinked to its content with, and center/right cell alignment is supported.
```
calcaide, line 9: @gsusmi Not sure if my wording is correct here :sweat_smile:

addons/rose/addon/components/rose/table/index.stories.mdx
```diff
@@ -36,22 +53,34 @@ A table component to display data in structure of rows and columns. It supports
...
-    <Rose::Table as |table|>
+    <Rose::Table @style={{style}} as |table|>
       <table.header as |header|>
-        <header.row as |row|>
-          <row.headerCell>Address</row.headerCell>
-          <row.headerCell>Host Set</row.headerCell>
-          <row.headerCell>Name</row.headerCell>
-          <row.headerCell>Description</row.headerCell>
+        <header.row @hidden={{hideHeaderRow}} as |row|>
+          <row.headerCell @shrink={{cellShrink}} @alignCenter={{cellAlignmentCenter}} @alignRight={{cellAlignmentRight}}>Address</row.headerCell>
```
calcaide, line 59: I add all of this JUST for storybook demonstration purposes. We will never have both alignment, they are deterministic.

addons/rose/addon/components/rose/table/index.stories.mdx
```diff
@@ -36,22 +53,34 @@ A table component to display data in structure of rows and columns. It supports
...
+          <row.headerCell @alignCenter={{cellAlignmentCenter}} @alignRight={{cellAlignmentRight}}>Host Set</row.headerCell>
+          <row.headerCell @shrink={{cellShrink}} @alignCenter={{cellAlignmentCenter}} @alignRight={{cellAlignmentRight}}>Name</row.headerCell>
+          <row.headerCell @alignCenter={{cellAlignmentCenter}} @alignRight={{cellAlignmentRight}}>Description</row.headerCell>
         </header.row>
       </table.header>
       <table.body as |body|>
         <body.row as |row|>
-          <row.cell>1.1.1.1</row.cell>
+          <row.cell @shrink={{cellShrink}}>1.1.1.1</row.cell>
```
calcaide, line 67: Define `shrink` in every cell it is not mandatory but recommended. Otherwise the wider cell with shrink applied in the row, will determinate the width of the whole row.

addons/rose/addon/styles/rose/components/table/_table.scss
```diff
@@ -10,9 +10,23 @@
     text-align: left;
   }
 
+  .rose-table-row-key-value {
+    td:nth-child(2) {
```
calcaide, line 14: `Key/Value` styles defined with a nth-child. 
Assumption:
- `key/value` rows will have JUST two cells. One for the key and another for the value.
- The cell that contains the value can have 1 or more elements. In example, an input + a button. That's why I defined with `display: flex`. So if we add more elements they will work just fine.
- If new use cases of `key/value` are shown in designs, we might need to tweak it, but for the use cases we have now its fine.

addons/rose/addon/styles/rose/components/table/_table.scss
calcaide: I assume, regarding the previous comment this should be fine. Otherwise let me know and I will check.

addons/rose/tests/integration/components/rose/table/row-test.js
```diff
@@ -7,17 +7,20 @@ module('Integration | Component | rose/table/row', function (hooks) {
   setupRenderingTest(hooks);
 
   test('it renders', async function (assert) {
+    assert.expect(2);
```
calcaide, line 10: In this test file I added assert expectations for each test. I do not know how strict we want to be with the assert expectation. If we want be more relaxed I can delete it.

randallmorey: I think these are very useful especially for async tests.

randallmorey (changes requested):

addons/rose/addon/styles/rose/components/table/_table.scss
calcaide: Actually the style applied it is just for the `value` cell. I do not encounter any issue with the `key` cell. That said, if in any use case the `key` cell gets too bigger I agree we can rely on `shrink` for it.

About the `value` cell, I used the special-casing mostly to apply `display: flex` so the items we add on that cell flow. But happy to consider other options.

calcaide: Update: after a zoom call we decide to forget about `display:flex` so we delete the `rose-table-row-key-value` style and approach. The path to follow is use 3 rows in these cases (key/value lookalike editing mode).

randallmorey (approved):

PR #556, opened 2021-05-13
feat: 🎸 implement basic OIDC auth method attributes modeling

randallmorey:
This PR introduces basic OIDC attributes, excluding array fields, and does not include custom methods.


addons/api/addon/models/auth-method.js
```diff
@@ -1,9 +1,31 @@
 import GeneratedAuthMethodModel from '../generated/models/auth-method';
+import { fragment } from 'ember-data-model-fragments/attributes';
 import { equal } from '@ember/object/computed';
 
+export const signingAlgorithms = [
```
gsusmi, line 5: This isn't being used anywhere. Is this for reference in future PRs?

addons/api/mirage/generated/factories/auth-method.js
```diff
@@ -11,4 +11,22 @@ export default Factory.extend({
   created_time: () => date.recent(),
   updated_time: () => date.recent(),
   disabled: () => datatype.boolean(),
+
+  attributes: function () {
+    let attrs;
+    switch (this.type) {
```
calcaide, line 17: I do asume we will have more auth methods in the future?

addons/api/addon/models/auth-method.js
randallmorey: Yes.  We don't necessarily have a great way to model `enum`-type fields.  But UI code will definitely need to know about allowable values to display them in, say, a `select`.

addons/api/mirage/generated/factories/auth-method.js
randallmorey: We already have two types:  `password` and `oidc`.  However, only `oidc` uses the `attributes` field right now.

gsusmi (approved):
Ty!

addons/api/mirage/generated/factories/auth-method.js
calcaide: Yep. I was asking about it for the usage of `switch`. I understand we are using a `switch` statement because we expect 3 or more cases in the near future.

calcaide (approved):
LGTM. Thank you!

PR #557, opened 2021-05-13
feat: 🎸 add OIDC change state custom method

randallmorey:
This PR adds support for the `change-state` custom method of OIDC auth methods which allows state to be changed.  Under normal updates, `state` is read only.


gsusmi (approved):
LGTM. Ty!

PR #558, opened 2021-05-14
Feature oidc account claim maps

randallmorey:
This PR introduces modelling for the OIDC `account_claim_maps` field.  This field is represented in the API as an array of strings:

```
account_claim_maps: ["oid=sub", "full_name=name"]
```

And is normalized into the auth method model as a fragment array of objects:

```
authMethod.attributes.account_claim_maps = [{from: 'old', to: 'sub'}, {from: 'full_name', to: 'name'}];
```


addons/api/addon/serializers/auth-method.js
```diff
@@ -80,10 +81,19 @@ export default class AuthMethodSerializer extends ApplicationSerializer {
...
+   * @see FragmentScope
+   * @param {Model} typeClass
+   * @param {Object} hash
+   * @return {Object}
    */
-  normalize() {
-    const normalized = super.normalize(...arguments);
+  normalize(typeClass, hash, ...rest) {
+    let normalizedHash = copy(hash, true);
+    // switch (normalizedHash.type) {
```
randallmorey, line 91: I left this commented out because we may eventually have OIDC-specific normalization.

gsusmi (approved):
LGTM. Ty!

addons/api/addon/serializers/auth-method.js
```diff
@@ -80,10 +81,19 @@ export default class AuthMethodSerializer extends ApplicationSerializer {
...
    * @override
-   * @return {object}
+   * @see FragmentScope
+   * @param {Model} typeClass
+   * @param {Object} hash
+   * @return {Object}
    */
-  normalize() {
-    const normalized = super.normalize(...arguments);
+  normalize(typeClass, hash, ...rest) {
```
gsusmi, line 89: +1 the use of `...rest` here.

PR #559, opened 2021-05-17
Fix improper use of url.protocol in desktop

gsusmi:
`protocol` in node's url implementation includes `:`. This PR removes appending it to url.
Ref: https://nodejs.org/api/url.html#url_url_strings_and_url_objects


calcaide (approved):
LGTM! Just checked the official doc for `url.protocol` and this makes all the sense, nice!

PR #563, opened 2021-05-20
Feature more oidc modeling

randallmorey:
This PR completes OIDC auth method modelling with the missing string array fields.  Unlike other string array fields so far, these are updated via conventional PUT/PATCH requests, rather than via custom methods.  Thus no field-specific serialization methods are required as are found in previous serializers

### DIFF
--- a/addons/api/addon/initializers/fragment-string.js
+++ b/addons/api/addon/initializers/fragment-string.js
@@ -1,0 +1,12 @@
+// app/initializers/fragment-serializer.js
+import FragmentStringSerializer from '../serializers/fragment-string';
+
+export function initialize(application) {
+  application.register('serializer:fragment-string', FragmentStringSerializer);
+}
+
+// Fragment serializers must be manually registered
+export default {
+  name: 'fragment-string',
+  initialize: initialize,
+};
--- a/addons/api/tests/unit/initializers/fragment-string-test.js
+++ b/addons/api/tests/unit/initializers/fragment-string-test.js
@@ -1,0 +1,32 @@
+import Application from '@ember/application';
+
+import { initialize } from 'dummy/initializers/fragment-string';
+import { module, test } from 'qunit';
+import Resolver from 'ember-resolver';
+import { run } from '@ember/runloop';
+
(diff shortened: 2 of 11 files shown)


addons/api/tests/unit/initializers/fragment-string-test.js
```diff
@@ -0,0 +1,32 @@
...
+
+  hooks.afterEach(function () {
+    run(this.application, 'destroy');
+  });
+
+  // TODO: Replace this with your real tests.
+  test('it works', async function (assert) {
+    await this.application.boot();
+
+    assert.ok(true);
```
calcaide, line 30: I guess this is the default test ember cli created?

addons/api/tests/unit/serializers/fragment-string-test.js
```diff
@@ -0,0 +1,49 @@
+import { module, test } from 'qunit';
+import { setupTest } from 'ember-qunit';
+
+module('Unit | Serializer | fragment string', function (hooks) {
+  setupTest(hooks);
+
+  // Replace this with your real tests.
```
calcaide, line 7: I think this comment is not valid anymore?

addons/api/addon/initializers/fragment-string.js
```diff
@@ -0,0 +1,12 @@
+// app/initializers/fragment-serializer.js
```
calcaide, line 1: This comment shows where the serializer lives?

randallmorey: This is default generated.

addons/api/tests/unit/initializers/fragment-string-test.js
randallmorey: Yes.  We sometimes don't override these.

addons/api/tests/unit/serializers/fragment-string-test.js
randallmorey: This is generated along with the "it exists" test by default.  There are a lot of these comments hanging around.  PR to clean this up is welcome.

gsusmi (approved):
LGTM. Ty!

PR #567, opened 2021-05-21
OIDC standard crud - ICU-1365

randallmorey:
This PR enables standard CRUD configuration for OIDC auth methods.

Screenshot 2021-05-21 at 16 51 14


ui/admin/app/components/form/auth-method/oidc/index.hbs
```diff
@@ -0,0 +1,168 @@
...
+      </field.errors>
+    {{/if}}
+  </form.input>
+
+  <form.checkbox
+    @label={{t 'form.disable_discovered_config_validation.label'}}
+    @helperText={{t 'form.disable_discovered_config_validation.help'}}
+    @checked={{@model.attributes.disable_discovered_config_validation}}
+    onChange={{fn
+      (mut @model.attributes.disable_discovered_config_validation)
```
gsusmi, line 79: Today I learnt about [mut](https://api.emberjs.com/ember/3.26/classes/Ember.Templates.helpers/methods/mut?anchor=mut). Awesome!

gsusmi (approved):
Ty!

ui/admin/app/components/form/auth-method/oidc/index.hbs
randallmorey: It's Ember magic that allows a value to be mutated somewhat declaratively, so that it we don't need to write a trivial action to do it.

PR #568, opened 2021-05-21
feat: Generate config file for desktop build

gsusmi:
Due to changes to forge config, about menu option has to be configured.
Previous dialog:
original

Updated dialog:
future-1


gsusmi:
Updates:
`config/config.js` file is generated at build time. Electron forge config uses this configuration file to read in desktop build details.
Example `config/config.js`

```
module.exports = {
  "name": "boundary",
  "productName": "Boundary",
  "bundleId": "com.electron.boundary",
  "copyright": "Copyright © 2021 HashiCorp, Inc.",
  "releaseVersion": "1.0.3",
  "icon": "/Users/susmitha/code/boundary-ui/ui/desktop/electron-app/config/app-icons/icon",
  "macos": {
    "hardenedRuntime": true,
    "entitlements": "/Users/susmitha/code/boundary-ui/ui/desktop/electron-app/config/macos/entitlements.plist",
    "signatureFlags": "library",
    "dmg": {
      "diskIcon": "/Users/susmitha/code/boundary-ui/ui/desktop/electron-app/config/macos/disk.icns",
      "background": "/Users/susmitha/code/boundary-ui/ui/desktop/electron-app/config/macos/background.png"
    }
  }
}
```

ui/desktop/electron-app/config/desktop.js
```diff
@@ -0,0 +1,63 @@
...
+    );
+  }
+
+  console.log(`[desktop-config] Release commit: ${config.releaseCommit}`);
+  console.log(`[desktop-config] Release version: ${config.releaseVersion}`);
+
+  // MacOS signing identity
+  if (!config.macos.signingIdentity)
+    console.warn(
+      '[desktop-config] WARNING: Could not find signing identity. Proceeding without signing.'
```
randallmorey, line 43: This is just the config file generator.  It should contain almost no logic.  This feels like a forge config concern.

randallmorey: This file is generated even during development, which would trigger this warning, correct?

ui/desktop/electron-app/config/desktop.js
```diff
@@ -0,0 +1,63 @@
...
+        diskIcon: path.join(macosConfigPath, 'disk.icns'),
+        background: path.join(macosConfigPath, 'background.png'),
+      },
+    },
+  };
+
+  if (!config.releaseVersion) {
+    config.releaseVersion = '0.0.0';
+    console.warn(
+      `[desktop-config] Release version not specified. Using: ${config.releaseVersion} instead`
```
randallmorey, line 33: This is a tough read since the variable being used "instead" is the one that was missing in the first place.

randallmorey (changes requested):
I think it's super close.  It still feels too tightly couple to the build configuration though.  There's still some static stuff and purely build-time concerns that don't need to be output into a generated file.  Think of the config as divorced from any particular use.  It should contain essential meta about the application like name, versions, etc.  Logic such as warnings are build-time concerns, not config generation concerns.  Let's review during our standup.

There's also a question as to whether the generator script would be bundled into the application?  It probably shouldn't be.

gsusmi:
Updates:
- Refactored assets from `electron-app/config` into new `assets` folder.
- Refactored electron forge options out from build time generated desktop config file. Example shown below:
```
module.exports = {
  "name": "boundary",
  "productName": "Boundary",
  "copyright": "Copyright © 2021 HashiCorp, Inc.",
  "releaseVersion": "0.0.0",
  "releaseCommit": "47c453e979462c5fc8ee1d8e769b992c529e1091"
}
```

ui/desktop/electron-app/config/desktop.js
gsusmi: Instead of raising a warning, updated desktop config to set a release version if missing.

ui/desktop/electron-app/config/desktop.js
gsusmi: Removed signing warning as using developer signed builds is discouraged.

calcaide:
I like it the SOC in `assets` also new config file is pretty clean :grinning:

calcaide (approved):

randallmorey (approved):

PR #574, opened 2021-05-25
chore: 🤖 fix imports that don't make sense

randallmorey:
Core helpers were exported from a source that doesn't exist.  And some Rose components were exported that don't exist at all.  This PR resolves these issues as a small step toward Embroider support.


gsusmi (approved):
I'd like to learn how you found these inconsistencies. Are there linting logs I'm not reading correctly?
Changes LGTM. Ty!

randallmorey:
These are reported by Webpack when using Embroider.  Unfortunately a working build is still a ways off.  But these were some easy wins.

calcaide (approved):
Following to the previous comments. When I worked with webpack + react, webpack has plenty of tools to point packages that are processed/build but not use. But also did eslint. I need to get better knowledge on this with ember.
That said, awesome changes, loving our code is caring! Thank you

PR #575, opened 2021-05-25
feat: 🎸 textarea and select should support contexual usage

randallmorey:
This PR adds support for contextual usage of select and textarea components, bringing them in line with the input component.  Contextual usage allows labels and fields to be rendered in arbitrary locations by the developer, which is useful when rendering form fields within a table.


calcaide (approved):
LGTM

gsusmi (approved):

PR #577, opened 2021-05-31
First iteration of connect dialog component

calcaide:
Assumptions:
- I Am just using a s simple dialog that eventually will be swap for the component. Even I created the component (at least the view) I am not using it yet, the dialog is the base of the component tho.
- The component does not have a route or model and I think it needs one.
- I discover and issue with the z-index within the dialog component. I will work on that later it is not blocking for now.

Screen Shot 2021-05-31 at 9 29 06 AM

### DIFF
--- a/addons/api/addon/models/session.js
+++ b/addons/api/addon/models/session.js
@@ -2,12 +2,13 @@ import GeneratedSessionModel from '../generated/models/session';
 import { computed } from '@ember/object';
 import { equal } from '@ember/object/computed';
 import { attr } from '@ember-data/model';
+import { tracked } from '@glimmer/tracking';
 
 export default class SessionModel extends GeneratedSessionModel {
   // =attributes
 
-  @attr('boolean', { readOnly: true }) acknowledged;
-  @attr('boolean', { readOnly: true }) started_desktop_client;
+  @tracked acknowledged;
+  @tracked started_desktop_client;
   @attr('string', { readOnly: true }) proxy_address;
   @attr('number', { readOnly: true }) proxy_port;
 
--- a/addons/api/addon/models/session.js
+++ b/addons/api/addon/models/session.js
@@ -6,6 +6,8 @@ import { attr } from '@ember-data/model';
 export default class SessionModel extends GeneratedSessionModel {
   // =attributes
 
+  @attr('boolean', { readOnly: true }) acknowledged;
+  @attr('boolean', { readOnly: true }) started_desktop_client;
   @attr('string', { readOnly: true }) proxy_address;
   @attr('number', { readOnly: true }) proxy_port;
 
--- a/addons/api/addon/models/session.js
+++ b/addons/api/addon/models/session.js
@@ -1,7 +1,6 @@
 import GeneratedSessionModel from '../generated/models/session';
 import { computed } from '@ember/object';
 import { equal } from '@ember/object/computed';
-import { attr } from '@ember-data/model';
 import { tracked } from '@glimmer/tracking';
 
 export default class SessionModel extends GeneratedSessionModel {
@@ -9,8 +8,8 @@ export default class SessionModel extends GeneratedSessionModel {
 
   @tracked acknowledged;
   @tracked started_desktop_client;
-  @attr('string', { readOnly: true }) proxy_address;
-  @attr('number', { readOnly: true }) proxy_port;
+  @tracked proxy_address;
+  @tracked proxy_port;
 
   @equal('status', 'active') isActive;
   @equal('status', 'pending') isPending;
--- a/ui/desktop/app/components/dialog-detail/index.hbs
+++ b/ui/desktop/app/components/dialog-detail/index.hbs
@@ -8,14 +8,16 @@
   @modal={{true}}
   as |dialog|
 >
-  {{yield (hash
-    notification=(component 'rose/notification' class='connection-dialog-notification')
-  )}}
+  {{yield
+    (hash
+      notification=(component
+        'rose/notification' class='connection-dialog-notification'
+      )
+    )
+  }}
 
   <dialog.body>
-    {{yield (hash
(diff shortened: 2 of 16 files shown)


calcaide:
Updated the PR with this look:
Screen Shot 2021-06-02 at 9 26 24 AM

ui/desktop/app/components/dialog-detail/index.hbs
```diff
@@ -0,0 +1,117 @@
...
+>
+  {{yield (hash
+    notification=(component 'rose/notification' class='connection-dialog-notification')
+  )}}
+
+  <dialog.body>
+    {{yield (hash
+      body=(component 'rose/anonymous')
+    )}}
+    {{!--
```
calcaide, line 19: Please, ignore this commented code, I am just leaving it temporary for reference.

ui/desktop/app/styles/app.scss
```diff
@@ -173,7 +173,6 @@
   .copyable-button {
     margin: 0;
     padding: 0 sizing.rems(xxs);
-    position: relative;
```
calcaide, line 176: @randallmorey I deleted this `position: relative` because was the cause of the copy icon messing up with the `z-index`. I tested the uses cases I am aware and did change nothing, but let me know your thoughts.

ui/desktop/app/styles/app.scss
```diff
@@ -291,3 +290,64 @@
...
+    }
+      
+    div:nth-child(2) {
+      display: flex;
+      justify-content: flex-end;
+      width: 50%;
+    }
+  }
+
+  // .credentials-card {
```
calcaide, line 323: Please, ignore this commented code. I am using it for reference and I will get rid of it pretty soon.

ui/desktop/package.json
```diff
@@ -28,7 +28,7 @@
     "format:js": "prettier --write '{app,config,stories,tests,electron-app/src}/**/*.js' *.js",
     "format:sass": "prettier --write app/**/*.scss",
     "lint:js:fix": "eslint . --fix",
-    "start": "ember serve",
+    "start": "ember serve --port 3200",
```
calcaide, line 31: @randallmorey @gsusmi be aware of this change. I add this to could work in the desktop with the browser without any port issues with the admin.

addons/api/addon/models/session.js
```diff
@@ -6,6 +6,8 @@ import { attr } from '@ember-data/model';
 export default class SessionModel extends GeneratedSessionModel {
   // =attributes
 
+  @attr('boolean', { readOnly: true }) acknowledged;
+  @attr('boolean', { readOnly: true }) started_desktop_client;
```
gsusmi, line 10: Is this an indication that user clicked on 'connect' button and got a successful response back?

ui/desktop/app/templates/scopes/scope/projects/targets.hbs
```diff
@@ -1,3 +1,63 @@
...
+      {{#unless session.acknowledged}}
+        <DialogDetail
+          @heading={{concat 'Target connection details: ' session.id}}
+          @dismiss={{route-action 'acknowledge' session}}
+          @dismissButtonText={{t 'actions.ok'}}
+          as |dialogDetail|
+        >
+          <dialogDetail.notification
+            @style='success'
+            @heading='Successfully connected'
```
gsusmi, line 15: Use translation helper for heading string to future proof multi-language support, something like `{{t resources.target.messages.connection-success}}`

gsusmi: Actually, can you double check this file for naked string? There are a few within heading tags. Ty!

ui/desktop/package.json
gsusmi: I'd vote to leave port configurations out of our start scripts, mainly because we standardize on start scripts in every available `package.json` and all of us have our port range preferences. I personally use `yarn start -p <port>` when launching multiple clients/addons.

addons/api/addon/models/session.js
randallmorey: Essentially yes.  This is a client-side-only property indicating that the session was started _via this client_ as opposed to some other client.  However upon further reflection, these could be `@tracked` properties instead of model attributes.  When changing the values of tracked properties, you'll need to do so directly on the instances @calcaide, rather than via the store push method we currently use.

ui/desktop/package.json
calcaide: Fair enough. I do agree that should be more developer preference, thanks for the tip on the `--port` flag for the yarn script :wink:

addons/api/addon/models/session.js
calcaide: @randallmorey cool! that makes sense. I will check some info about `@tracked` and do such change.

ui/desktop/app/templates/scopes/scope/projects/targets.hbs
calcaide: Absolutely! I missed the translation helper, but now you mention I will check the file 😉 .

Thanks for the heads up with the translation helper, I will put it in the top of my mind while working in other files.

addons/api/addon/models/session.js
calcaide: These two variables has been converted to `@tracked`.

ui/desktop/app/templates/scopes/scope/projects/targets.hbs
calcaide: I checked this file, now it uses the translation helper.

calcaide:
@randallmorey Updated the dialog to could be usable in the app. Here the latest screenshot, I tried to mimic the actual confirmation dialog with this one, I temporary removed the credentials or any new fields.

Screen Shot 2021-06-08 at 8 25 45 AM

PR #578, opened 2021-05-31
chore: 🤖 Fix WS security vulnerability

calcaide:
Update the resolutions to resolve a version of WS that has no security
vulnerabilities.


calcaide:
Adding @gsusmi and @randallmorey for visibility. Once approved I will merge right away to get rid of the bot alert.

PR #581, opened 2021-06-01
[desktop] Enable windows platform support

gsusmi:
PR results in generating an EXE application on windows platform.

99adda3e0aa4880cff25384e7411b9fc

Merging this PR will allow release pipeline to run parallel macOS and windows builds.
Known issue: Cannot get past authentication screen. To be addressed in next PR.


ui/desktop/electron-app/config/cli.js
```diff
@@ -1,25 +1,39 @@
 const fs = require('fs');
+const os = require('os');
 const path = require('path');
 const https = require('https');
 const decompress = require('decompress');
-const os = require('os');
+const { isMac, isWindows } = require('../src/helpers/platform.js');
```
calcaide, line 6: Love the naming, can't be more clear :grin:

ui/desktop/electron-app/src/helpers/platform.js
```diff
@@ -0,0 +1,6 @@
+const os = require('os');
+
+module.exports = {
+  isMac: () => Boolean(os.platform().match(/(darwin)/i)),
+  isWindows: () => Boolean(os.platform().match(/(win32)/i)),
```
calcaide, line 5: Silly question. Does all supported versions of windows including 64bits return `win32`?

ui/desktop/electron-app/config/cli.js
```diff
@@ -1,25 +1,39 @@
...
+  const archivePlatform = {};
+  // TODO: Support all available build machines
+  if (isMac()) {
+    archivePlatform.name = 'darwin';
+    archivePlatform.arch = 'amd64';
+  }
+
+  if (isWindows()) {
+    archivePlatform.name = 'windows';
+    archivePlatform.arch = 'amd64';
```
calcaide, line 21: Does `arch` always return `amd64`? do we have no cases where is `x86`?

calcaide (approved):
LGTM! good work!

ui/desktop/electron-app/src/helpers/platform.js
gsusmi: It's a limitation of [os.platform()](https://nodejs.org/api/os.html#os_os_platform). We could use [os.arch()](https://nodejs.org/api/os.html#os_os_arch) in the future, although we don't have a need right now.

ui/desktop/electron-app/config/cli.js
gsusmi: For windows, cli for is supported for amd64 and 386. However, desktop is aimed for Windows 10 and above, thus hardcoded to 'amd64'.

ui/desktop/electron-app/src/helpers/platform.js
calcaide: For sure! not need for it now ;).

randallmorey (approved):

ui/desktop/electron-app/config/cli.js
calcaide: Ahh gotcha! for now more than enough, I was using these questions as a shortcuts to learn too.

PR #583, opened 2021-06-02
[draft] feat: Add oidc status change dropdown to auth method actions

gsusmi:
https://user-images.githubusercontent.com/111036/120415535-b1438600-c329-11eb-8c6e-14ac14573aa1.mov


addons/core/translations/resources/en-us.yaml
```diff
@@ -45,6 +45,7 @@ auth-method:
     delete: Delete Auth Method
     make-primary: Make Primary
     remove-as-primary: Remove as Primary
+    change-state: State
```
randallmorey, line 48: Not sure this key->value mapping makes sense.

randallmorey (changes requested):

gsusmi:
Todo: 

- Disable state change dropdown when model is dirty
- Introduce static class (maybe emberobject) to list all possible states
- Add acceptance test for state change API errors

addons/core/translations/resources/en-us.yaml
gsusmi: Refactored this to `status: Status` to better represent auth-method status.

PR #585, opened 2021-06-03
feat: 🎸 New Card component with header, body and footer

calcaide:
Add the new Card component with header, body and footer for the desktop client. This card will be used within the connection dialog to show credentials.

✅ Closes: ICU-1432

Ignore the dialog where the Card component is used.

Screen Shot 2021-06-03 at 8 44 14 AM


ui/desktop/app/styles/app.scss
```diff
@@ -292,3 +292,38 @@
...
+
+  .credentials-card-header {
+    border-bottom: sizing.rems(xxxxs) solid var(--ui-gray-subtler-3);
+    display: flex;
+    flex-direction: row;
+    padding: sizing.rems(s);
+
+    div {
+      display: flex;
+      flex-grow: 1;
```
calcaide, line 314: This makes the header ready to receive more than 1 `<div>` (as item) and rearrange itself. `div` it is very specific, if you like, I can make it more generic creating a class for it, something like `credentials-card-header-item`

ui/desktop/app/styles/app.scss
```diff
@@ -292,3 +292,38 @@
...
+  border: sizing.rems(xxxxs) solid var(--ui-gray-subtler-3);
+  display: flex;
+  flex-direction: column;
+  margin-bottom: sizing.rems(m);
+
+  &:last-child {
+    margin-bottom: 0;
+  }
+
+  .credentials-card-header {
```
calcaide, line 306: The naming is not definitive. I call it `credentials-card` because we already have a `card` in rose. Should I call it just card? should I call it something else?

calcaide, line 306: The heather align its items to the right by default, but we can create a class to align them to the left?

randallmorey (changes requested):
Great start!  I think for convenience and standardisation, this component should accept a `heading` attribute (string) and an option `icon` attribute (also string).  Those should be rendered by the component itself so they always turn out the same.

The `header` named block could live _next to_ the heading and icon.  We might want the header block to default to a right-alignment (with flex box).  n=1 but I'd say this is good place to start.

ui/desktop/app/components/card/index.hbs
```diff
@@ -0,0 +1,15 @@
+<section class='credentials-card'>
```
randallmorey, line 1: I'd use different terminology than `credentials`, since the card is more general.  Credentials just happens to be one (and currently only) use case.

randallmorey (changes requested):

calcaide:
Refactored the component with the suggestions:
- Change the name to be more generic (card).
- Add attributes for heading. For now `@heading` and `@icon` we can discuss later on if we want to add more (like `@iconSize`) but I think as a start should be enough.

ui/desktop/app/components/card/index.hbs
```diff
@@ -0,0 +1,17 @@
+<section class='card'>
+  <header class='card-header'>
+    <div>
+      {{#if @icon}}
+        <Rose::Icon @name={{@icon}} @size='medium' />
+      {{/if}}
+      {{@heading}}
+    </div>
+    {{yield (hash header=(component 'rose/anonymous'))}}
```
randallmorey, line 9: When we're just defining an empty region, let's use named blocks, e.g.
```
{{yield to="header"}}
```

Where developers will use it thusly:
```
<Card>
  <:header>This is my header!</:header>
</Card>
```

randallmorey: Also I think by default, the header region should be right-aligned to reflect the design.

ui/desktop/app/components/card/index.hbs
```diff
@@ -0,0 +1,17 @@
...
+  </header>
+  {{yield
```
randallmorey, line 11: Same feedback here.  Let's use named blocks.

randallmorey (changes requested):

calcaide:
**Updated the PR with the suggestions**

Screenshot of the card using `@heading` and providing code within the header (1 div):
Screen Shot 2021-06-16 at 1 09 46 PM

Code used in the screenshot:
```html
<Card @heading='This is @heading'>
    <:header>
        <div>This is :header content</div>
    </:header>
    <:body>This is the body</:body>
    <:footer>This is the footer</:footer>
</Card>
```

Screenshot of the card using `@heading` and providing code withint the header (2 div):
Screen Shot 2021-06-16 at 1 12 29 PM

Code used in the screenshot:
```html
<Card @heading='This is @heading'>
    <:header>
        <div>This is :header content</div>
        <div>This is :header content 2</div>
    </:header>
    <:body>This is the body</:body>
    <:footer>This is the footer</:footer>
</Card>
```

Changes applied:
- Apply forgoten padding and alignment styles for header.
- Apply conditional rendering for when @heading is not provide. Before, if was not provided we render an empty <div>.
- Refactor to use named block with in the card component.

ui/desktop/tests/integration/components/card-test.js
```diff
@@ -0,0 +1,46 @@
...
+
+  test('it renders', async function (assert) {
+    await render(hbs`<Card />`);
+    assert.ok(find('.card'));
+  });
+
+  test('it renders with content', async function (assert) {
+    await render(hbs`
+      <Card as |card|>
+        <card.header></card.header>
```
calcaide, line 17: @randallmorey Didn't realize till now I was double checking the PR... Should I change in each test the render function and use the named blocks?

ui/desktop/app/components/card/index.hbs
```diff
@@ -0,0 +1,15 @@
+<section class='card'>
+  <header class='card-header'>
+    {{#if @heading}}
+      <div>
```
randallmorey, line 4: Can this be an `h2` instead?

randallmorey (changes requested):

ui/desktop/app/styles/app.scss
```diff
@@ -360,35 +359,44 @@
...
+    align-items: center;
+    border-bottom: sizing.rems(xxxxs) solid var(--ui-gray-subtler-3);
+    display: flex;
+    justify-content: flex-start;
+    padding: sizing.rems(s);
 
-  //     div {
-  //       display: flex;
-  //       width: 50%;
+    h2 {
```
calcaide, line 378: I changed this `h2` to look like in the figma design.

PR #586, opened 2021-06-03
fix: 🐛 Deprecate Slash as division operator

calcaide:
Deprecation of slash as division operator as we were getting warnings in the terminal about it. More info [here](https://sass-lang.com/documentation/breaking-changes/slash-div).


randallmorey (approved):

gsusmi (approved):
LGTM. Ty!

PR #588, opened 2021-06-04
Use frameless desktop shell on Windows OS

gsusmi:
Screenshot 2021-06-06 030039

Todo:
- [Windows] ~~Run tests for killing spawned process on app quit~~
- [MacOS] Test _everything_

### DIFF
--- a/ui/desktop/app/styles/app.scss
+++ b/ui/desktop/app/styles/app.scss
@@ -124,6 +124,10 @@
   padding-left: sizing.rems(l) * 3;
   position: relative;
 
+  &.header-shrink {
+    padding-left: 0;
+  }
+
   .logotype {
     display: none;
   }
--- a/ui/desktop/app/styles/app.scss
+++ b/ui/desktop/app/styles/app.scss
@@ -141,11 +141,12 @@
 
 .rose-header {
   -webkit-app-region: drag;
-  -webkit-user-select: none;
   min-height: sizing.rems(l) * 2;
   position: relative;
+  -webkit-user-select: none;
 
-  .rose-header-brand, .rose-header-nav,
+  .rose-header-brand,
+  .rose-header-nav,
   .rose-header-utilities .rose-dropdown,
   .rose-header-utilities button {
     -webkit-app-region: no-drag;
--- a/ui/desktop/app/styles/app.scss
+++ b/ui/desktop/app/styles/app.scss
@@ -42,7 +42,6 @@
     overflow-y: scroll;
   }
 
-  // Windows OS specific overflow.
   &.layout-scroll {
     .rose-layout-global-body {
       overflow-y: auto;
@@ -156,16 +155,13 @@
     padding-left: sizing.rems(l) * 3;
   }
 
-  .utilities-actions {
-    display: flex;
-    justify-content: center;
-
-    button {
-      background: transparent;
-      border: transparent;
-      color: var(--white);
-      margin: sizing.rems(xs);
-    }
+  .button-window-minimize,
+  .button-window-fullscreen,
+  .button-window-close {
+    background: transparent;
+    border: transparent;
+    color: var(--white);
+    margin: sizing.rems(xs);
   }
 
   .logotype {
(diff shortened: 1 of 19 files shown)


ui/desktop/app/templates/application.hbs
```diff
@@ -1,17 +1,45 @@
...
       {{#if this.session.isAuthenticated}}
         <header.nav>
           {{outlet 'header-nav'}}
         </header.nav>
+      {{/if}}
 
-        <header.utilities as |utilities|>
+      <header.utilities as |utilities|>
+        {{#if this.isWindowsOS}}
+          <span class='utilities-actions'>
```
randallmorey, line 23: This name feels too generic.  These are specifically window controls.  Is there a more specific name that makes sense?

ui/desktop/app/templates/application.hbs
```diff
@@ -1,17 +1,45 @@
...
+            <Rose::Button
```
randallmorey, line 24: Maybe instead of styling buttons within a container, we give these windowing buttons their own class name and style on that?  `button-desktop-windowing`?

ui/desktop/electron-app/src/models/session.js
```diff
@@ -60,10 +62,18 @@ class Session {
    */
   stop() {
     return new Promise((resolve, reject) => {
+      debugger;
```
randallmorey, line 65: This should be removed right?

randallmorey (changes requested):

ui/desktop/app/templates/application.hbs
gsusmi: I was trying to keep in pattern with rose-like generic naming. Will reconsider.

ui/desktop/app/styles/app.scss
```diff
@@ -29,13 +29,34 @@
...
+      overflow: auto;
+    }
   }
 
   .rose-layout-sidebar-sidebar,
   .rose-layout-sidebar-body {
     overflow-y: scroll;
   }
+
+  // Windows OS specific overflow.
```
calcaide, line 45: How do we apply these CSS JUST to windows OS? or we applied for both clients but just affects windows?

ui/desktop/electron-app/src/cli/path.js
```diff
@@ -0,0 +1,11 @@
+const path = require('path');
+const { isWindows } = require('../helpers/platform.js');
+
+module.exports = {
+  // Returns boundary cli path
+  path: () => {
+    let name = 'boundary';
```
calcaide, line 7: How about `const name = (isWindows)? 'boundary.exe' : 'boundary';` ? not a big fan of one line conditionals, but in this case is pretty short.

ui/desktop/app/styles/app.scss
gsusmi: The style is applied selectively based on platform as shown [here](https://github.com/hashicorp/boundary-ui/pull/588/files#diff-de7e5aba3ecc486ff8ba63bcd57bf8e6660eb3d3dd45530d712a2417548bf65cR4). I cleaned up.

ui/desktop/electron-app/src/cli/path.js
gsusmi: Good idea. Updated.

ui/desktop/electron-app/src/models/session.js
gsusmi: Oops. Removed.

ui/desktop/app/templates/application.hbs
gsusmi: Added `button-window-minimize`, `button-window-fullscreen`, and `button-window-close` classes to Windows window actions to simplify and removed wrapper span.

ui/desktop/app/styles/app.scss
calcaide: Ahh makes all the sense, I miss the condition on the view. :thumbsup:

calcaide (approved):
LGTM! :ok_hand:

randallmorey (approved):

PR #590, opened 2021-06-05
fix: 🐛 add read-only hmac field and enable oidc auth creation

randallmorey:


ui/admin/app/components/form/auth-method/oidc/index.hbs
```diff
@@ -6,6 +6,14 @@
   @showEditToggle={{if @model.isNew false true}}
   as |form|
 >
+
+  <form.actions
```
randallmorey, line 10: This was added because the form is long and the actions fall below the fold where they may not be seen.

ui/admin/app/templates/scopes/scope/auth-methods/index.hbs
```diff
@@ -24,6 +24,12 @@
         >
           {{t 'resources.auth-method.types.password'}}
         </dropdown.link>
+        <dropdown.link
+          @route='scopes.scope.auth-methods.new'
+          @query={{hash type='oidc'}}
```
gsusmi, line 29: Nice!

gsusmi (approved):
LGTM. Ty!

PR #596, opened 2021-06-08
chore: tweak dialog styles and language

randallmorey:


addons/core/translations/resources/en-us.yaml
```diff
@@ -233,7 +233,12 @@ target:
     none:
       title: No Targets Available
       description: No Targets to display yet. Contact your Boundary admin if you don't see a Target you expect access to.
-    connection-success: Successfully connected
+    connection-success:
```
calcaide, line 236: Sorry I miss the translations! Thanks!

ui/desktop/app/templates/scopes/scope/projects/targets.hbs
```diff
@@ -6,40 +6,36 @@
...
-            session.id
+            target.id
           }}
           @dismiss={{route-action 'acknowledge' session}}
           @dismissButtonText={{t 'actions.ok'}}
-          as |dialogDetail|
+          as |dialog|
         >
-          <dialogDetail.notification
+          <dialog.notification
```
calcaide, line 17: I am not sure (yet) if the notification should live out of the dialog body. Initially I went this way to not have to override paddings from the dialog body, but I think we need to start working with this component to determinate, for now here is goo enough.

calcaide (approved):
LGTM! thanks for the changes

PR #597, opened 2021-06-08
fix: about dialog should display properly

randallmorey:
This resolves a bug where the about dialog was not displaying in macOS desktop.


calcaide (approved):
LGTM!

PR #605, opened 2021-06-11
feat: 🎸 Create reveal component for desktop client

calcaide:
Create reveal component for desktop client.

✅ Closes: ICU-1432

Assumptions made, for visibility and context. But happy to discuss other approaches or changes:
- I did not wrap all the component (all meaning `<details>` + `<content>`) in a parent container like a `div`.
- I choose `reveal` as a name for the component.
- The icon that toggles as caret, it is not an attribute but hardcoded.
- The component will expect two attributes for toggle the summary text between open and close state.
- The component does not use an attribute to keep state (open/close) but a predefined `@tracked`.
- I did not spend any time on the `<pre> ` and `<code>` styling since it is out of topic for this PR (as per my understanding).

_Do not pay attention to the dialog I add the component for testing please._

Screenshot component close:
Screen Shot 2021-06-11 at 1 56 44 PM

Screenshot component open:
Screen Shot 2021-06-11 at 1 57 16 PM

Screenshot code to test the component:
Screen Shot 2021-06-11 at 1 58 20 PM


randallmorey (changes requested):
@calcaide Let's model this on the existing dropdown as much as possible.  In fact, you can inherit the component class from rose dropdown.

One main difference is that the `open` attribute isn't handled by the component JS, it's on the element itself.  And the icon is set via CSS.

calcaide:
@randallmorey Updated PR with suggested changes.

**Changes:**

- Refactor RevealComponent class to inherit from RoseDropdown, no logic need it. Now there is no need to keep any logic or state.
- Refactor component, to toggle Summary text and set chevron icon with CSS.
- Removed old commented code. Please feel free to suggest a change to bring it back if can be useful.

Screenshot component close:
Screen Shot 2021-06-18 at 10 47 17 AM

Screenshot component open:
Screen Shot 2021-06-18 at 10 47 40 AM

**How to use the component?**

```html
<Reveal
	@textWhenOpen='Hide raw API output'
	@textWhenClose='View raw API output'
>
  <pre>
	<code>
	  {
		"auth": null,
		"data": {
		  "foo": "bar",
		  "ttl": "1h"
		},
		"lease_duration": 3600,
		"lease_id": "",
		"renewable": false
	  }
	</code>
  </pre>
</Reveal>
```

ui/desktop/app/components/reveal/index.hbs
```diff
@@ -0,0 +1,14 @@
...
+  <summary class='reveal-summary'>
+    <span class='reveal-summary-text-close'>
+      {{@textWhenClose}}
+    </span>
+    <span class='reveal-summary-text-open'>
+      {{@textWhenOpen}}
+    </span>
+
+  </summary>
+  <content class='reveal-content' {{on-click-inside this.close}}>
```
randallmorey, line 11: This is one thing that differs from dropdown.  The reveal component should probably remain open until the user explicitly closes it.  There could be important information revealed they want to copy/paste or even interact with.  So we don't need either `on-click-outside` or `-inside` modifiers.

ui/desktop/app/components/reveal/index.js
```diff
@@ -0,0 +1,7 @@
+import RoseDropdownComponent from 'rose/components/rose/dropdown';
+
+export default class RevealComponent extends RoseDropdownComponent {
+  constructor() {
```
randallmorey, line 4: Is this necessary?

randallmorey (changes requested):

ui/desktop/app/components/reveal/index.js
calcaide: I don't think so. I forgot to add a comment to ask you exactly this. Now I know lol thanks!!

ui/desktop/app/components/reveal/index.hbs
calcaide: Awesome! yeah Iw as wondering about this behaviour will change it.

gsusmi (approved):

randallmorey (approved):

PR #607, opened 2021-06-14
chore: 🤖 Upgrade ember-jquery

calcaide:
Uprgrade ember-jquery package

### DIFF
--- a/ui/desktop/package.json
+++ b/ui/desktop/package.json
@@ -49,7 +49,7 @@
   },
   "devDependencies": {
     "@doyensec/electronegativity": "^1.8.1",
-    "@ember/jquery": "^1.1.0",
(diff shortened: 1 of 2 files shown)


randallmorey (changes requested):
@calcaide Let's eliminate our need for jQuery entirely.  If [this delegated event handler](https://github.com/hashicorp/boundary-ui/blob/40ee317f47100e0be459f9c6aee12e966efd7961/ui/desktop/app/routes/application.js#L41) can be rewritten in vanilla JS, we can remove the jquery dep.

randallmorey:
@calcaide closing for now.  See my comment regarding jQuery.  Let's see if we can eliminate it as a dep.

PR #611, opened 2021-06-18
Feature oidc accounts

randallmorey:
This PR enables CRUD UI for OIDC accounts, as well as more helpful account details where accounts are displayed (e.g. user "add accounts" UX).

Screenshot 2021-06-18 at 11 58 54
Screenshot 2021-06-18 at 11 59 12
Screenshot 2021-06-18 at 12 00 24
Screenshot 2021-06-18 at 11 59 30


addons/api/mirage/generated/factories/account.js
```diff
@@ -1,15 +1,14 @@
...
+import { date, datatype } from 'faker';
 
 /**
  * GeneratedAccountFactory
  */
 export default Factory.extend({
   type: 'password',
-  name: () => random.words(),
-  description: () => random.words(),
+  // name: () => random.words(),
```
randallmorey, line 9: These fields are no longer given mock values, since in reality they are rarely set for this resource type.

calcaide (approved):
LGTM!

PR #613, opened 2021-06-21
chore: 🤖 Update resolutions for glob-parent

calcaide:
Update resolutions field in package json to avoid security vulnerability with glob-parent older versions.


calcaide:
**How was this tested?**
I run the main yarn scripts to validate we did not brake any process:
- `build` and all its variants `ui:admin`, `ui:desktop` and such.
- `lint` and `test`.

PR #617, opened 2021-06-24
feat: Update connect dialog with credentials api response

gsusmi:
https://user-images.githubusercontent.com/111036/123192040-23624300-d470-11eb-828e-925d26d3a1a9.mov

What's next?
- ~~Show raw json output instead of '[object object]'. Probably needs a handlerbar helper to generate json string~~ Fixed 
- Should credential type be surfaced in header instead of displaying it in credential details dropdown?
- ~~Test connect error detail dialog~~ Done


randallmorey (changes requested):
Fantastic!  Only thing I noticed is that the raw API output is just showing "object" and is very indented.

calcaide (approved):
AWESOME!! looks grate

gsusmi:
Fixed raw api output display:
raw-api-output

gsusmi:
Update: Integrated onto `feature-more-session-creds`:

https://user-images.githubusercontent.com/111036/123338847-7e02aa00-d517-11eb-916d-5a67b4f8485f.mov

PR #618, opened 2021-06-24
feat: 🎸 Add hidden secret component

calcaide:
Add hidden secret component adhoc to desktop client. This is the first iteration of the component, expect changes in the near future.

This first iteration does not include to toggle (vissible/hidden) secret. Just the ability to copy it.

Screen Shot 2021-06-24 at 12 07 56 PM

How to use the component:
`<HiddenSecret @secret='AAaaBBccDDeeOTXzSMT1234BB_Z8JzG7JkSVxI' />`

### DIFF
--- a/addons/core/translations/resources/en-us.yaml
+++ b/addons/core/translations/resources/en-us.yaml
@@ -256,4 +256,3 @@ target:
       add:
         title: Add Host Sets
         description: Select host sets to assign to this target.
-    secret_label: ■ ■ ■ ■ ■ ■ ■ ■ ■ ■ ■ ■
--- a/addons/core/translations/resources/en-us.yaml
+++ b/addons/core/translations/resources/en-us.yaml
@@ -256,3 +256,4 @@ target:
       add:
         title: Add Host Sets
         description: Select host sets to assign to this target.
+    secret_label: ■ ■ ■ ■ ■ ■ ■ ■ ■ ■ ■ ■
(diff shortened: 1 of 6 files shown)


ui/desktop/app/components/hidden-secret/index.hbs
```diff
@@ -0,0 +1,11 @@
+<div class='hidden-secret'>
+  {{! TODO: The span wrap within a div because later we will add the toggle secret icon/button}}
+  <div>
+    <span>{{t 'resources.target.host-set.secret_label'}}</span>
```
calcaide, line 4: To skip the `no-bare-strings` rule I decided to add the secret characters as a translation string.

addons/core/translations/resources/en-us.yaml
```diff
@@ -256,3 +256,4 @@ target:
       add:
         title: Add Host Sets
         description: Select host sets to assign to this target.
+    secret_label: ■ ■ ■ ■ ■ ■ ■ ■ ■ ■ ■ ■
```
calcaide, line 259: Not sure if this is the bet translation file to keep this string. Thoughts?

gsusmi: What do you think about putting it in https://github.com/hashicorp/boundary-ui/blob/main/addons/core/translations/form/en-us.yaml?

ui/desktop/app/components/hidden-secret/index.hbs
```diff
@@ -0,0 +1,11 @@
...
+  </div>
```
gsusmi, line 5: This content can be wrapped with <Copyable>.

calcaide: https://gph.is/g/Zx1WqN7

PR #624, opened 2021-06-28
fix: Use name in connect header

gsusmi:
This PR removes unused credential purpose property from UI and displays name in credential header (instead of purpose).

### DIFF
--- a/ui/desktop/app/templates/scopes/scope/projects/targets.hbs
+++ b/ui/desktop/app/templates/scopes/scope/projects/targets.hbs
@@ -39,13 +39,11 @@
 
               {{#each session.credentials as |credential|}}
                 <Card
-                  @heading={{concat
-                    (t 'resources.session.credential.title')
-                    (if
-                      credential.library.name
-                      (concat ' (' credential.library.name ')')
-                    )
-                  }}
+                  @heading='{{t 'resources.session.credential.title'}}
+                    {{if
+                    credential.library.name
+                    (concat ' (' credential.library.name ')')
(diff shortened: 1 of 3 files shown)


ui/desktop/app/templates/scopes/scope/projects/targets.hbs
```diff
@@ -41,7 +41,10 @@
                 <Card
                   @heading={{concat
                     (t 'resources.session.credential.title')
-                    (if credential.purpose (concat ' (' credential.purpose ')'))
+                    (if
+                      credential.library.name
+                      (concat ' (' credential.library.name ')')
```
randallmorey, line 46: Name is optional, so the parens should be conditional.  You could try an inline if:

```
(if credential.library.name (concat ' (' credential.library.name ')'))
```

ui/desktop/app/templates/scopes/scope/projects/targets.hbs
```diff
@@ -71,12 +74,6 @@
                         <:key>{{t 'form.type.label'}}</:key>
                         <:value>{{credential.library.type}}</:value>
                       </dropdown.key-value>
-                      {{#if credential.library.name}}
```
randallmorey, line 74: I think it's worth keeping this in the dropdown for visibility.  It's another way to get the same information, particularly because in the card header the key isn't labeled.

randallmorey (changes requested):

randallmorey (approved):

PR #634, opened 2021-06-30
chore:  update to reflect secret { raw, decoded } schema, where decod…

randallmorey:
…ed is optional


gsusmi (approved):
Huge thanks for taking this on.

PR #638, opened 2021-07-08
feat: Enable credential feature flag in dev mode for admin

gsusmi:
This would enable conditional check in templates for credential feature: `{{#if (feature-flag 'credential-store')}}`


calcaide (approved):
LGTM! Thank you

PR #640, opened 2021-07-09
fix: Update text font color and weight 

DhariniJeeva:
This pr fixes the font color and weight to be in sync with the design  

Before:
Screen Shot 2021-07-09 at 10 56 59 AM

After: 
Screen Shot 2021-07-09 at 10 55 48 AM

### DIFF
--- a/ui/admin/app/styles/app.scss
+++ b/ui/admin/app/styles/app.scss
@@ -94,6 +94,8 @@
 .rose-table {
   p {
     margin-bottom: 0;
+    color: var(--ui-gray-subtler-1);
+    font-weight: normal;
   }
 
   .rose-form-checkbox {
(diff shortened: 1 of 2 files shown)


ui/admin/app/styles/app.scss
```diff
@@ -94,6 +94,8 @@
 .rose-table {
   p {
```
gsusmi, line 95: To reduce accidentally applying this style to all table columns, could this be restricted to just the header cells in a table?

gsusmi (commented):
Minor note about nesting style to a header cell. Looks great otherwise! Ty!

ui/admin/app/styles/app.scss
DhariniJeeva: good catch. updated!

calcaide (approved):
LGTM!

gsusmi (approved):

PR #646, opened 2021-07-18
feat: List credential libraries in a credential store

gsusmi:
list-credential-libraries

Empty list:
nothing-available


calcaide (approved):
LGTM!!

PR #650, opened 2021-07-20
Test acceptance credential stores

calcaide:
Add acceptance tests for list, create and edit credential stores.

Missing Delete tests because the delete feature is not available yet.

### DIFF
--- a/ui/admin/tests/acceptance/credential-stores-test.js
+++ b/ui/admin/tests/acceptance/credential-stores-test.js
@@ -4,6 +4,7 @@ import { setupApplicationTest } from 'ember-qunit';
 import setupMirage from 'ember-cli-mirage/test-support/setup-mirage';
 import a11yAudit from 'ember-a11y-testing/test-support/audit';
 import { authenticateSession } from 'ember-simple-auth/test-support';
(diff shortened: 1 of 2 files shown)


ui/admin/tests/acceptance/credential-stores-test.js
```diff
@@ -32,17 +41,198 @@ module('Acceptance | credential-stores', function (hooks) {
...
     await a11yAudit();
     assert.equal(currentURL(), urls.credentialStores);
+    await visit(urls.credentialStore);
+    await a11yAudit();
+    assert.equal(currentURL(), urls.credentialStore);
+  });
+
+  // This test is returning an issue regarding not finding resource.
+  // It is a false positive.
+  test.skip('visiting an unknown credential store display 404 message', async function (assert) {
```
calcaide, line 74: @randallmorey I am skipping this test because I do believe there is an issue causing a false positive. Host catalog has the same issue. Perhaps we should re-think this test?

randallmorey: Let me jump in and take a look!  Thanks for catching this.

randallmorey (approved):

PR #654, opened 2021-07-21
API: Target Credential Library api modeling

gsusmi:
Updates api addon to read new `application_credential_library_ids` in target response.
`add-credential-libraries` and `remove-credential-libraries` methods are added to target API.


calcaide (approved):
LGTM!!

Only question I had was resolved o slack DM due to my lack of knowledge of ember modeling.
Question was regarding to re-use `model/target` methods `addCredentialLibraries` and `removeCredentialLibraries` since the only code change is the `method` within the `defaultAdapterOptions`. In the future we will look in to this, but since it is a new feature makes more sense to not over do it now.

Thanks for the work and lesson @gsusmi

randallmorey (approved):

PR #657, opened 2021-07-22
chore: 🤖 Update eslint dependency

calcaide:
Update Eslint dependency version.

**Risk:** After checking the changelog I see no risk of doing so.
**How I test it?** I run eslint tasks in all the projects within the monorepo and reported no problems.


gsusmi (approved):
LGTM. Ty!

PR #656, opened 2021-07-21
use Fieldset in credential library create page

DhariniJeeva:

PR #660, opened 2021-07-22
fix: 🐛 update fieldset font weight

DhariniJeeva:
Updating `fieldset` weight to be in sync with the label style. conversation [here](https://github.com/hashicorp/boundary-ui/pull/656)


gsusmi (approved):
LGTM. Ty!

PR #661, opened 2021-07-22
Credential Library - Read

gsusmi:
credential-library-read

Todo: Acceptance tests


randallmorey (commented):
It feels a bit cluttered to have two labels and two descriptions for the body field.  Can these be consolidated somehow?

gsusmi:
Good point. Consolidated options labeling into:
http-post-options-form

calcaide (approved):
LGTM!! :hugs:

PR #675, opened 2021-07-27
Add Credential Store Attributes

gsusmi:
screencapture-localhost-4200-scopes-p-1234567890-credential-stores-csvlt-hEoaNqCiBx-2021-07-27-13_59_34

Errors:
screencapture-localhost-4032-scopes-s-xk1i4dk50v2-credential-stores-new-2021-07-27-14_29_15


calcaide (approved):
LGTM

randallmorey (changes requested):
We should probably display the HMAC values (token, client cert key) in read-only fields to be consistent with the OIDC form.

Also, the fields which have associated HMAC values (token, client cert) may need to be deleted from the serialisation when they are falsy.  This is always true for token (it may not be unset).  And sometimes true for client cert key (it may not be unset as long as client cert has a value).

gsusmi:
Updates:
#### Edit mode for credential store displays HMAC values for token and client certificate key:
screencapture-localhost-4032-scopes-s-xk1i4dk50v2-credential-stores-credential-store-id-0-2021-07-28-13_09_16


#### Fix: TLS skip verify checkbox error handler
Screen Shot 2021-07-28 at 13 12 22

randallmorey (approved):

PR #676, opened 2021-07-27
Test credential library create errors

calcaide:
Add test credential library create errors and fix saving credential library behaviour to not redirect after a save.

The behaviour change was discussed on slack thread [here](https://hashicorp.slack.com/archives/C012GTH1C4E/p1627409871077400)


DhariniJeeva (approved):
ty! LGTM

ui/admin/tests/acceptance/credential-library/credential-library-test.js
```diff
@@ -137,6 +137,112 @@ module('Acceptance | credential-library', function (hooks) {
...
+    );
+  });
+
+  test('can cancel changes to existing credential library', async function (assert) {
+    assert.expect(2);
+    await visit(urls.credentialLibrary);
+    await click('form [type="button"]', 'Activate edit mode');
+    await fillIn('[name="name"]', 'random string');
+    await click('.rose-form-actions [type="button"]');
+    assert.notEqual(instances.credentialStore.name, 'random string');
```
randallmorey, line 160: To be confident that this is true, you may want to look up the instance at assertion time (like in the save case above).  This is because these instances may not be updated on change (they're more like "snapshots").

ui/admin/tests/acceptance/credential-library/credential-library-test.js
```diff
@@ -137,6 +137,112 @@ module('Acceptance | credential-library', function (hooks) {
...
+    assert.equal(currentURL(), urls.credentialLibrary);
+
+    try {
+      await visit(urls.credentialLibraries);
+    } catch (e) {
+      assert.ok(find('.rose-dialog'));
+      await click('.rose-dialog-footer button:first-child', 'Click Discard');
+      assert.equal(currentURL(), urls.credentialLibraries);
+      assert.notEqual(
+        this.server.schema.credentialLibraries.all().models[0].name,
```
randallmorey, line 217: Nice.

randallmorey (changes requested):

ui/admin/tests/acceptance/credential-library/credential-library-test.js
calcaide: Makes sense, I will do the change.

Since I copy this pattern from `host-catalogs` and also reproduce it on `credential-stores` test. I will provide on a separate PR these test following this pattern.

randallmorey (approved):

PR #680, opened 2021-07-29
fix: credential library naming does not change

randallmorey:


gsusmi (approved):
Ty for handling this.

PR #688, opened 2021-08-05
API support for UI capabilities

randallmorey:
This PR introduces capabilities support including `read`, `update`, and `delete` authorisations for singular resource instances.  Based on Ember Can, the base abilities may be called from templates to selectively hide UI when a user does not have proper authorisation.

This PR is foundational and does not apply capabilities to the UI.  Applied capabilities will follow in additional PRs.

Usage example:
```hbs
{{#if (can 'delete model' @model)}}
  <button>Delete</button>
{{/if}}
```


gsusmi (approved):
LGTM. Ty!

DhariniJeeva (approved):

PR #689, opened 2021-08-05
feat: 🎸 apply base singular capabilities to account

randallmorey:
This PR applies base capabilities to singular accounts (`read`, `update`, `delete`, and `save`).

### DIFF
--- a/addons/api/mirage/factories/account.js
+++ b/addons/api/mirage/factories/account.js
@@ -1,8 +1,15 @@
 import factory from '../generated/factories/account';
 import { random, internet } from 'faker';
+import permissions from '../helpers/permissions';
 
 export default factory.extend({
-  authorized_actions: () => ['no-op', 'read', 'update', 'delete'],
+  authorized_actions: () =>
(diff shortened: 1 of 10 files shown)


randallmorey (commented):
Note that acceptance tests have be reorganised to include separate test files for `accounts/delete`, `accounts/update`, and `accounts/read`, all of which now have new test cases for capabilities.  This cleans up the main acceptance test file which was growing too large.

gsusmi:
Looks great! Especially like breaking up tests into actions.
Query: Where are the ember-can abilities defined? Is that not necessary??

gsusmi:
@randallmorey , I can't see `can 'update model'` being used in this example - is `can 'save model'` the preferred method?

randallmorey:
> @randallmorey , I can't see `can 'update model'` being used in this example - is `can 'save model'` the preferred method?

Yes.  The `save` permission is a composite that includes an `update` check.  There's a comment in the `canSave` getter of the ability explaining why it is necessary.

DhariniJeeva (approved):

PR #691, opened 2021-08-06
Default platform/arch support

jboero:
Explicit platform checks are prohibitive.  *NIX builds by default but not with explicit checks.


gsusmi:
Appreciate the PR. As we don't officially support *nix yet for the Desktop Client, we'll circle back to this once we add support.

PR #692, opened 2021-08-06
chore: 🤖 Update tar dependency resolutions

calcaide:
Update tar transitive dependency resolutions to avoid security vulnerability.

### DIFF
--- a/package.json
+++ b/package.json
@@ -75,7 +75,8 @@
     "mem": "^4.0.0",
     "minimist": "^1.2.3",
     "normalize-url": "^4.5.1",
-    "glob-parent": "^5.1.2"
+    "glob-parent": "^5.1.2",
+    "tar": "^3.2.3 - ^4.4.15 - ^5.0.7 - ^6.1.2"
   },
   "config": {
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -75,7 +75,8 @@
     "mem": "^4.0.0",
     "minimist": "^1.2.3",
     "normalize-url": "^4.5.1",
-    "glob-parent": "^5.1.2"
+    "glob-parent": "^5.1.2",
+    "tar": "^3.2.3 - ^4.4.15 - ^5.0.7 - ^6.1.2"
```
calcaide, line 79: **Why this syntax?** This is a transitive dependency and after exploring the `yarn.lock` file realize we are resolving different versions of it. Forcing all these transitive dependencies to use just one version (`6.1.2`) could be risky. Following the security vulnerability report, they provided fixes for their stable versions `3.x`, `4.x`, `5.x` and `6.x` so that's why I decided to go this way.

randallmorey (approved):

randallmorey (approved):

PR #697, opened 2021-08-10
chore: 🤖 adds support for development capabilities toggle

randallmorey:
This PR adds development time support for capabilities toggling on mock resources.  To use the toggle, add a query string to the URL, with as many resources and permissions as desired.  For convenience, these should persist even as you navigate through the application.  Example:

`?authorized_actions={"account": ["read"]}`

Note that this feature is provided as a convenience and does not come with rigorous tests.  As it is only intended for use by our Mirage mocks, it only needs to be just _good enough_ to solve the development use case (and not cause tests to fail).

### DIFF
--- a/addons/api/mirage/helpers/permissions.js
+++ b/addons/api/mirage/helpers/permissions.js
@@ -1,0 +1,58 @@
+import { copy } from 'ember-copy';
+
+/**
+ * Allows development-time toggling of resource permissions via the window
+ * querystring.  Pass a JSON object into the query to toggle permissions.
+ * Note that the value must be a valid JSON object in order to be parsed.
+ *
+ * @example
+ *   ?authorized_actions={"account": ["read"]}
+ *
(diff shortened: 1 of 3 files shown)


DhariniJeeva:
looks good! thank you. 
quick question- do we need to commit the yarn.lock file?  are there any dependency updates in this pr?

addons/api/mirage/helpers/permissions.js
```diff
@@ -0,0 +1,58 @@
...
+ *   import { Factory } from 'ember-cli-mirage';
+ *   import permissions from '../helpers/permissions';
+ *
+ *   export default Factory.extend({
+ *     authorized_actions: () =>
+ *       permissions.authorizedActionsFor('account') ||
+ *       ['no-op', 'read', 'update', 'delete']
+ *   });
+ */
+class ToggledCapabilities {
```
gsusmi, line 23: Excellent idea.

PR #700, opened 2021-08-12
Feat user capabilities

DhariniJeeva:


randallmorey (approved):
Approved.  But it looks like jsconfig.json files were committed?  What are these for?  Are they necessary?

DhariniJeeva:
@randallmorey  regarding the jsconfig.json files, I'm not sure, I think vscode generated those?  Don't think they are necessary. I'll remove those in a separate pr.

PR #701, opened 2021-08-12
Groups capabilities

DhariniJeeva:


ui/admin/tests/acceptance/groups/update-test.js
```diff
@@ -0,0 +1,100 @@
...
+import setupMirage from 'ember-cli-mirage/test-support/setup-mirage';
+import { Response } from 'miragejs';
+import {
+  authenticateSession,
+  // These are left here intentionally for future reference.
+  //currentSession,
+  //invalidateSession,
+} from 'ember-simple-auth/test-support';
+
+module('Acceptance | groups', function (hooks) {
```
randallmorey, line 13: Do we need a test case here for when the user does not have the update permission?

DhariniJeeva: good point. I will add a test case for that.

randallmorey (approved):

PR #702, opened 2021-08-12
feat: 🎸 role capabilities

DhariniJeeva:
this pr adds read/delete/update capabilities to roles


randallmorey (changes requested):
This needs an update test for when authorisation is not present.

DhariniJeeva:
> This needs an update test for when authorisation is not present.

added the missing test and removed the jsconfig.files that were added in this [pr](https://github.com/hashicorp/boundary-ui/pull/700)

randallmorey (approved):

PR #705, opened 2021-08-16
chore: 🤖 Update tar resolutions

calcaide:
I applied a change for this same library days ago. Apparently the pattern I provided to resolve versions is not correct. So there is no other solution but resolve just one version.


gsusmi (approved):
Good to know. Ty!

PR #713, opened 2021-08-19
Feature collection abilities

randallmorey:
This PR introduces ability models for collection capabilities `list`, `create`, and a composite capability `navigate`.

### DIFF
--- a/ui/admin/app/abilities/collection.js
+++ b/ui/admin/app/abilities/collection.js
@@ -1,0 +1,30 @@
+import CollectionAbility from 'api/abilities/collection';
+import { inject as service } from '@ember/service';
+
+export default class OverrideModelAbility extends CollectionAbility {
+  // =services
+
+  @service features;
+
+  // =permissions
+
+  /**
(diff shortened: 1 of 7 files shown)


DhariniJeeva (approved):

ui/admin/app/abilities/collection.js
```diff
@@ -0,0 +1,30 @@
+import CollectionAbility from 'api/abilities/collection';
+import { inject as service } from '@ember/service';
+
+export default class OverrideCollectionAbility extends CollectionAbility {
```
calcaide, line 4: Curiosity question: this class is declared within admin and desktop with the same code. Also, it inherits from a class declared within addons. I was wondering if we could share also this class between clients? why or why not?

randallmorey: That's a good point.  The reason I did this is because at the application level, capabilities depend on `ember-feature-flags` configuration which is per-app.  However the dependency itself comes from `core`, so this _might_ still work if it were pulled into `core`.

PR #714, opened 2021-08-19
Host catalog capabilities

DhariniJeeva:


randallmorey (approved):

ui/admin/app/templates/scopes/scope/host-catalogs/index.hbs
```diff
@@ -41,12 +41,14 @@
...
-                  {{hostCatalog.displayName}}
-                </LinkTo>
+                {{#if (can 'read model' hostCatalog)}}
+                  <LinkTo
+                    @route='scopes.scope.host-catalogs.host-catalog'
+                    @model={{hostCatalog.id}}
+                  >
+                    {{hostCatalog.displayName}}
+                  </LinkTo>
+                {{/if}}
```
randallmorey, line 51: Else clause?

randallmorey (changes requested):

ui/admin/app/components/form/host-catalog/index.hbs
```diff
@@ -45,12 +45,12 @@
...
   </form.textarea>
-
-  <form.actions
-    @disabled={{if @model.cannotSave @model.cannotSave}}
-    @enableEditText={{t 'actions.edit-form'}}
-    @submitText={{t 'actions.save'}}
-    @cancelText={{t 'actions.cancel'}}
-  />
-
+  {{#if (can 'save model' @model)}}
```
calcaide, line 48: I am missing an else case and wondering if the user has no save capabilities, it just sees the form with no actions? Perhaps the else is not need here?

DhariniJeeva: oops, seems like my changes didn't get pushed. It needs an else case.

DhariniJeeva: yes, we don't need the else case here. The user can't edit the form if they don't have `update` capability.  `save model` is a composite that includes `update`.  see discussion [here](https://github.com/hashicorp/boundary-ui/pull/689)

calcaide: Nice! I miss that discussion

calcaide (approved):

randallmorey (approved):

PR #716, opened 2021-08-19
feat: 🎸 host read/del/update capability

DhariniJeeva:


ui/admin/app/templates/scopes/scope/host-catalogs/host-catalog/hosts/index.hbs
```diff
@@ -13,12 +13,14 @@
...
-              {{host.displayName}}
-            </LinkTo>
+            {{#if (can 'read model' host)}}
+              <LinkTo
+                @route='scopes.scope.host-catalogs.host-catalog.hosts.host'
+                @model={{host.id}}
+              >
+                {{host.displayName}}
+              </LinkTo>
+            {{/if}}
```
randallmorey, line 23: Does this need an else clause?

randallmorey (changes requested):

ui/admin/app/templates/scopes/scope/host-catalogs/host-catalog/hosts/index.hbs
DhariniJeeva: oops, it does! will update this and the other two PRs

ui/admin/tests/acceptance/host-catalogs/hosts/read-test.js
```diff
@@ -0,0 +1,94 @@
...
+    projectScope: null,
+    hostCatalogs: null,
+    hostCatalog: null,
+    hosts: null,
+    host: null,
+    unknownHost: null,
+    newHost: null,
+  };
+
+  hooks.beforeEach(function () {
```
calcaide, line 38: @DhariniJeeva @randallmorey @gsusmi 
I am writing this as a question, not than as a statement: I see a recurrent pattern on all the capabilities tests. We are creating a test file per capability, which is a nice separation of concerns. But we are also repeating code through these files, all the instances, urls and setting those within the beforeEach statements are the same across capability.
I wonder if we could create a test helper file per capability where we define and export the repeating patterns, so then the test file just import those and we avoid repeating code.

For this use case, could be something like read-capability-test-helper.js? exports instances and urls initialization (basically generating the object equal to null) and generate all the resources.

Let me know your thoughts, perhaps this is not a good idea.

calcaide (approved):

ui/admin/tests/acceptance/host-catalogs/hosts/read-test.js
randallmorey: This is something I've thought about as well.  Essentially we need a standard way to model a test scenario and all the resources that go into it.  And from there, we might be able to move the standard test cases themselves into helpers.  This is housekeeping and hopefully we can find time in the future.

randallmorey (approved):

ui/admin/tests/acceptance/host-catalogs/hosts/read-test.js
calcaide: Sure! yeah I should mentioned before this message was OUT of topic of this PR and more a housekeeping thing.
Thanks!

DhariniJeeva: Thanks for bringing this up @calcaide.

PR #717, opened 2021-08-20
User create/list collection capabilities UI

randallmorey:
This PR implements create and list capabilities UI for the users resource and establishes a pattern for similar resources.


addons/api/mirage/factories/scope.js
```diff
@@ -4,6 +4,19 @@ import { trait } from 'ember-cli-mirage';
 export default factory.extend({
   type: 'global',
 
+  authorized_collection_actions: () => {
+    return {
+      users: ['create', 'list'],
+      // groups: ['create', 'list'],
```
DhariniJeeva, line 10: do we need the commented-out block here?

DhariniJeeva:
LGTM!

PR #723, opened 2021-08-26
user capabilities list test case

DhariniJeeva:


ui/admin/tests/acceptance/users/list-test.js
```diff
@@ -27,11 +26,12 @@ module('Acceptance | users | list', function (hooks) {
       org: null,
     },
     users: null,
+    orgScope: null,
   };
 
   hooks.beforeEach(function () {
     instances.scopes.global = this.server.create('scope', { id: 'global' });
-    orgScope = this.server.create(
+    urls.orgScope = this.server.create(
```
randallmorey, line 34: `urls` is meant to contain the URLs (strings) of the routes used in the test.  Model instances such as the one created here may be contained by `instances`.

randallmorey (changes requested):

ui/admin/tests/acceptance/users/list-test.js
DhariniJeeva: updated to instances!

ui/admin/tests/acceptance/users/list-test.js
```diff
@@ -41,17 +40,36 @@ module('Acceptance | users | list', function (hooks) {
...
+    assert.expect(2);
+    instances.orgScope.authorized_collection_actions.users = [];
+    await visit(orgURL);
+    assert.notOk(
+      instances.orgScope.authorized_collection_actions.users.includes('list')
+    );
+    assert.notOk(find(`[href="${urls.users}"]`));
   });
 
-  test('cannot navigate to users with proper authorization', async function (assert) {
```
calcaide, line 54: Just wondering and asking out of the context, pardon me if I ask something too obvious :sweat_smile: Are we planning to add test where the user can not navigate because has not permissions?

DhariniJeeva: ahh. it should be `without`. good catch!

DhariniJeeva: we renamed it to `User cannot navigate to index without either list or create actions` on L59

calcaide: ahhh that makes sense!! see, it was an obvious one lol. Terrific job with all the capabilities testing @DhariniJeeva :clap: :clap:

calcaide (approved):
LGTM!! Awesome work!

randallmorey (approved):

PR #732, opened 2021-09-01
chore: 🤖 Update tar version resolutions

calcaide:
Update tar version resolutions to avoid security vulnerability


gsusmi (approved):
LGTM. Ty!

PR #733, opened 2021-09-01
Sessions

DhariniJeeva:


addons/api/mirage/factories/scope.js
```diff
@@ -4,6 +4,19 @@ import { trait } from 'ember-cli-mirage';
 export default factory.extend({
   type: 'global',
 
+  authorized_collection_actions: () => {
+    return {
+      users: ['create', 'list'],
+      // groups: ['create', 'list'],
```
calcaide, line 10: This is minor thing, I understand that leaving all these comments means we are going to use them in the near future? if so, can we add a note on top of it saying so? so we know we should not clean up them.

DhariniJeeva: these will be uncommented when I merge in the PRs today/tomorrow

randallmorey (approved):

PR #753, opened 2021-09-08
desktop: Update README with instructions for using non-release versions

vancluever:
This documents the use of `BYPASS_CLI_SETUP` to allow desktop development
to use the version of Boundary manually placed in
`ui/desktop/electron-app/cli` rather than downloading a release version.


ui/desktop/README.md
```diff
@@ -67,7 +67,17 @@ To run as a desktop app:
 The Boundary CLI is downloaded and extracted to `electron-app/cli/` folder as part of
 build. CLI version is defined in `electron-app/config/cli.js`.
 
-#### Environment Variables
+### Developing Using Non-Release Versions of Boundary
```
gsusmi, line 70: Thoughts on renaming to `Building using local Boundary instance` or something that does not indicate `non-release`. This scenario generates unsigned builds for local testing only so we need to be careful about encouraging it's use.

randallmorey: Agree in principal though Boundary CLI is always local for Desktop builds, regardless of origin.

vancluever: Yeah, up to yall. Only thing that's important to me is that we document how to actually get the binary there. 🙂

randallmorey (approved):

PR #754, opened 2021-09-13
fix: Enable credential store feature in prod

gsusmi:


gsusmi:
Ignore. Feature flag isn't being used in template.

PR #757, opened 2021-09-16
Add chromeless window option for desktop

gsusmi:
Introduces new `DISABLE_WINDOW_CHROME` env var to hide window chrome entirely regardless of OS.


randallmorey (approved):

randallmorey (changes requested):
Maybe I'm misreading this, but the conditional `{{#if this.isWindowChromeless}}` seems like when true, it displays windows action buttons?

gsusmi:
`this.isWindowChromeless` refers to state of electron's OS frame. When window doesn't have a frame - i.e returning `true` - we need to add window buttons to add chrome functionality into ember template. Does `isBrowserWindowChromeless` read better?

randallmorey (changes requested):
Sorry I may not have been clear on the overall request.  The states are:

- All platforms:  "frameless".
- macOS:  show default window actions
- Windows & Linux:  show custom window actions
- All platforms:  If "DISABLE_WINDOW_CHROME" is `true`, completely hide all window actions, default and custom.  There should be no visible window action buttons at all.

Additionally, the boolean that controls the window action buttons could be more direct, such as `showCustomWindowActions` or something.

ui/desktop/electron-app/src/index.js
```diff
@@ -113,11 +113,18 @@ app.on('ready', async () => {
   }
   Menu.setApplicationMenu(menuTemplate);
 
+  /**
+   * Enable electron OS window frame/chrome for MacOS only.
+   * Disable frame/chrome regardless of OS when `DISABLE_WINDOW_CHROME` is set.
+   */
+  let showWindowChrome = isMac();
+  if (process.env.DISABLE_WINDOW_CHROME) showWindowChrome = false;
```
calcaide, line 121: Curiosity question, when we use `DISABLE_WINDOW_CHROME` should we just define it? or also give it a value? I mean, on the terminal, when executing desktop app, should we do `$ DISABLE_WINDOW_CHROME=true yarn start:desktop` or `$ DISABLE_WINDOW_CHROME yarn start:desktop` ?

DhariniJeeva (approved):
looks good to me! 👍

ui/desktop/electron-app/src/index.js
gsusmi: Good point. Checking for `DISABLE_WINDOW_CHROME` truthy value is clearer. Will update.

gsusmi: Made minor modification to comment text to specify that `DISABLE_WINDOW_CHROME=true` is used.

ui/desktop/app/routes/application.js
```diff
@@ -31,10 +31,14 @@ export default class ApplicationRoute extends Route {
   }
 
   /**
-   * Add OS config on controller.
+   * Add window frame config on controller.
    */
   async setupController(controller) {
-    controller.set('isMacOS', await this.ipc.invoke('isMacOS'));
+    controller.set('hasMacOSChrome', await this.ipc.invoke('hasMacOSChrome'));
```
randallmorey, line 37: I like that these are performed once and statically set on the controller.  Initially I questioned this but since these values don't change at runtime, this is less costly.

randallmorey (approved):

ui/desktop/electron-app/src/index.js
```diff
@@ -113,11 +113,18 @@ app.on('ready', async () => {
   }
   Menu.setApplicationMenu(menuTemplate);
 
+  /**
+   * Enable electron OS window frame/chrome for MacOS only.
+   * Disable frame/chrome regardless of OS when `DISABLE_WINDOW_CHROME` is set.
+   */
+  let showWindowChrome = isMac();
```
randallmorey, line 121: We should plan a way to move platform-dependent configuration into a separate file and set relevant values there, rather than inline with our normal code.  But that can happen in a separate PR.

PR #760, opened 2021-09-16
feat: 🎸 add connect target capability UI

randallmorey:


addons/api/addon/abilities/target.js
```diff
@@ -0,0 +1,15 @@
+import ModelAbility from './model';
+
+/**
+ * Generic ability provides abilities common to all models in the API.
```
randallmorey, line 4: Oops

gsusmi (approved):
LGTM. Ty!

PR #761, opened 2021-09-16
Bump flight icons

amyrlam:
- Test of new version bump (already published to npm) https://github.com/hashicorp/flight/pull/227 that I made with @didoo 
- Addon has new namespacing, does this work for you @randallmorey?
- I think the find and replace second commit is a-ok, but please lmk if I missed something
- Note: For Boundary's end, AFAICT doesn't seem necessary to make this change now, I wanted to test the bump anyways so using Boundary as a great real-world example!

### DIFF
--- a/addons/core/addon/components/doc-link/index.hbs
+++ b/addons/core/addon/components/doc-link/index.hbs
@@ -4,5 +4,5 @@
   target="_blank"
   rel="noreferrer noopener"
   title={{t "actions.get-topic-help"}}>
-  <Rose::Icon @name="flight-icons/help-24" @size={{@iconSize}} />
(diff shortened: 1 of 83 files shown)


amyrlam:
Whoops, let me look at my Prettier settings in VSCode, probably don't have it on right. Can squash in and rerun CI

randallmorey:
> Whoops, let me look at my Prettier settings in VSCode, probably don't have it on right. Can squash in and rerun CI

I suspect that the change caused some lines to exceed the max line length.  If you run `yarn format` in the affected addon or UI, it should clean it up for you.

amyrlam:
FYI @MelSumner @Dhaulagiri @didoo, tagging to keep you informed!

randallmorey (approved):

PR #767, opened 2021-09-22
chore: 🤖 Fix ember-fetch unmet dependency

calcaide:
Fix ember-fetch unmet dependency warning from Yarn.

Screen Shot 2021-09-22 at 3 45 33 PM

### DIFF
--- a/addons/auth/package.json
+++ b/addons/auth/package.json
@@ -23,6 +23,7 @@
   "dependencies": {
     "ember-cli-babel": "^7.26.6",
     "ember-cli-htmlbars": "^5.7.1",
+    "ember-fetch": "^8.1.1",
     "ember-simple-auth": "^4.0.0"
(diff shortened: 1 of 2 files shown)


randallmorey (approved):

calcaide:
After some reaserch, we do not need to take care of peerDependencies because we are developing and app and not a library, so we are not exposing our code to be used by others.
I will open a thread on #tech-frontend slack channel to validate that.

I will close this PR for the moment, I will reopen the PR in case we have to deal with peerDependencies.

calcaide:
As per [this thread](https://hashicorp.slack.com/archives/C012GTH1C4E/p1633075817031000) we find out all the dependencies shown on warnings are not errors because are hoisted. Conclusion: we should install the peer dependencies on the packages that consumes them.

PR #773, opened 2021-09-24
Bump @hashicorp/flight-icons to 1.0.0 with ^

amyrlam:
Bump package to 1.0.0 🚀 

I think dependabot will pick up future bumps with the `^` and we are following SemVer.

IS:
image

cc design systems folks @didoo @Dhaulagiri @MelSumner

### DIFF
--- a/addons/rose/package.json
+++ b/addons/rose/package.json
@@ -43,7 +43,7 @@
     "@ember/test-helpers": "^2.4.2",
     "@glimmer/component": "^1.0.4",
     "@glimmer/tracking": "^1.0.4",
-    "@hashicorp/flight-icons": "0.0.9-beta",
+    "@hashicorp/flight-icons": "^1.0.0",
     "@storybook/addon-actions": "^6.3.8",
     "@storybook/addon-docs": "^6.3.8",
     "@storybook/addon-knobs": "^6.3.1",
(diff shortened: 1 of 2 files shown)


randallmorey (approved):

calcaide (approved):
LGTM!

didoo:
Please wait before committing this. We never actually discussed the "bump" strategy for the consumers, and there are a few downsides in doing it automatically (using "^") vs manually (using specific versions). I would prefer to discuss with the Design System team and then make the changes for the consumers.

amyrlam:
@randallmorey synced with @didoo and team, this is good to go as `^1.0.0` 

there was some discussion that we may want to lock people to a version so that when new icons are added, the version bump is more clear

thanks for dealing with the churn!

randallmorey:
> @randallmorey synced with @didoo and team, this is good to go as `^1.0.0`
> 
> there was some discussion that we may want to lock people to a version so that when new icons are added, the version bump is more clear
> 
> thanks for dealing with the churn!

Thanks @amyrlam!  How will versions be bumped going forward when icons are added, removed, or changed?

amyrlam:
@randallmorey good question, there was a versioning RFC drafted...but it got put on the back burner. we plan to follow semver

i'll take a stab at it, but this is my personal opinion:
- i don't think any icons will be removed, but if they were, that would be a breaking major verion bump
- if icons are added, a minor bump
- if icons are changed...it prob depends on what the change is. maybe a patch if something with the svg was slightly off? 

there will likely be more versioning happening on the https://www.npmjs.com/package/@hashicorp/ember-flight-icons side, like if the wrapping component gets edits 

does that make sense? happy to get more info also if preferred!

PR #777, opened 2021-09-27
chore: 🤖 Add nth-check desktop resolutions

calcaide:
Add nth-check, transitive dependency, to desktop resolutions to avoid security vulnerability


gsusmi (approved):
LGTM. Ty!

PR #779, opened 2021-09-28
Feature remove sources capability

randallmorey:
Adds an ability check for target remove *-sources.

### DIFF
--- a/addons/api/addon/abilities/target.js
+++ b/addons/api/addon/abilities/target.js
@@ -12,4 +12,18 @@ export default class TargetAbility extends ModelAbility {
   get canConnect() {
     return this.hasAuthorizedAction('authorize-session');
   }
+
+  /**
+   * @type {boolean}
+   */
+  get canRemoveHostSources() {
+    return this.hasAuthorizedAction('remove-host-sources');
+  }
+
+  /**
+   * @type {boolean}
+   */
+  get canRemoveCredentialSources() {
+    return this.hasAuthorizedAction('remove-credential-sources');
+  }
 }
--- a/addons/api/mirage/factories/target.js
+++ b/addons/api/mirage/factories/target.js
@@ -14,7 +14,14 @@ export default factory.extend({
       'update',
       'delete',
       'authorize-session',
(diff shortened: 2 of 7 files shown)


addons/api/addon/abilities/target.js
```diff
@@ -12,4 +12,18 @@ export default class TargetAbility extends ModelAbility {
...
+  get canRemoveHostSources() {
+    return this.hasAuthorizedAction('remove-host-sources');
+  }
+
+  /**
+   * @type {boolean}
+   */
+  get canRemoveCredentialSources() {
+    return this.hasAuthorizedAction('remove-credential-sources');
+  }
```
gsusmi, line 28: Will `add-*-sources` permissions be added in the future? Apologies if they already exists and I missed them.

ui/admin/tests/acceptance/targets/credential-sources-test.js
```diff
@@ -160,6 +160,16 @@ module('Acceptance | targets | credential-libraries', function (hooks) {
     assert.equal(findAll('tbody tr').length, count - 1);
   });
 
+  test('cannot remove credential libraries without proper authorization', async function (assert) {
```
DhariniJeeva, line 163: should this be for `credential stores`?

randallmorey: No.  A target's "credential sources" refers to credential libraries.  In the future it might also allow one-off credentials (one level below library).

addons/api/mirage/factories/target.js
```diff
@@ -14,7 +14,14 @@ export default factory.extend({
       'update',
       'delete',
       'authorize-session',
+      'add-host-sources',
+      'set-host-sources',
```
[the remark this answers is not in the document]
randallmorey: Yeah.  I just copied the real grants over to mocks.

addons/api/addon/abilities/target.js
randallmorey: Yes.  This PR is just for remove.

PR #780, opened 2021-09-28
Feature add sources capability

randallmorey:


addons/api/tests/unit/abilities/target-test.js
```diff
@@ -26,6 +37,17 @@ module('Unit | Abilities | Target', function (hooks) {
     assert.notOk(service.can('removeHostSources target', model));
   });
 
+  test('it reflects when a given target may add crednetial sources', function (assert) {
```
DhariniJeeva, line 40:
```suggestion
  test('it reflects when a given target may add credential sources', function (assert) {
```

gsusmi (approved):
LGTM. Ty!

PR #784, opened 2021-10-01
fix: 🐛 Fix babel-core unmet peer dependency in Rose

calcaide:
Fix @babel-core unmet peer dependency in Rose.

Warning before fix:
Screen Shot 2021-10-01 at 11 39 56 AM

No warning in Rose regarding `@babel-core`:
Screen Shot 2021-10-01 at 11 58 40 AM

### DIFF
--- a/addons/rose/package.json
+++ b/addons/rose/package.json
@@ -29,6 +29,7 @@
     "doc:toc": "doctoc README.md"
   },
   "dependencies": {
+    "@babel/core": "^7.15.5",
     "babel-loader": "^8.2.2",
     "ember-cli-babel": "^7.26.6",
     "ember-cli-htmlbars": "^5.7.1",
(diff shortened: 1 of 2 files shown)


randallmorey (changes requested):
Does this work as part of dev deps?  I prefer to be conservative here and add things to dev deps as much as possible to avoid adding to the final bundle sizes.

calcaide:
@randallmorey You are right.

Initially, I thought on put `@babel-core` on `devDependencies` but I saw `babel-loader` [in dependencies](https://github.com/hashicorp/boundary-ui/blob/main/addons/rose/package.json#L32) and did not think twice.

But now that you mention both these packages should live in `devDependencies`, will add this change to the PR.

Good catch!!

DhariniJeeva (approved):

PR #789, opened 2021-10-04
feature: apply theme-color meta for more immersive UX

randallmorey:
[Applies the `theme-color` meta](https://css-tricks.com/safari-15-new-ui-theme-colors-and-a-css-tricks-cameo/) to Boundary UI, which adds brand color to the latest versions of macOS and iOS Safari.

Screenshot 2021-10-04 at 09 35 01

### DIFF
--- a/ui/admin/app/index.html
+++ b/ui/admin/app/index.html
@@ -6,6 +6,7 @@
     <title>Admin</title>
     <meta name="description" content="Simple and secure remote access with HashiCorp Boundary.">
     <meta name="viewport" content="width=device-width, initial-scale=1">
+    <meta name="theme-color" content="#f24c53">
 
     {{content-for "head"}}
 


ui/admin/app/index.html
```diff
@@ -6,6 +6,7 @@
     <title>Admin</title>
     <meta name="description" content="Simple and secure remote access with HashiCorp Boundary.">
     <meta name="viewport" content="width=device-width, initial-scale=1">
+    <meta name="theme-color" content="#f24c53">
```
calcaide, line 9: I do not have a strong opinion, but thinking loud to get your thoughts. 

Can we use a env var to define this color? I think we are just using it here, so then makes not much sense to create a env var. But if this color is something we will end up defining in other places, would be easier for maintenance to define it in one place.

calcaide (approved):
Looks awesome!! Thank you

ui/admin/app/index.html
randallmorey: @calcaide long-term, yes we should factor out hardcoded content.  This file is one exception and we can revisit it in the future.

PR #791, opened 2021-10-06
Show user "account name" wherever users are listed (if possible)

ghost:
This PR addresses feedback that, when an OIDC user has no associated `name`, it is displayed as a bare ID in user listings.  Users do come with information about their primary account:  email, full name, and login_name.  We now display this account name when possible.  Closes hashicorp/boundary#1467.

users-index
group-members
group-add-members
role-principals
role-add-principals


DhariniJeeva:
looks good to me! 
minor q - any particular reason for using break tags instead of CSS?

lucardcoder:
my users that are created with okta oidc does not populate the name in the account section. how can i ensure it does get automatically populated in the account? When they sign in in the UI the name section is empty.

Screenshot 2023-11-30 210606

nameokta

PR #800, opened 2021-10-13
[WIP] Add sb metadata

amyrlam:
design-system-website is reading /metadata.json as HTML

Ref: https://github.com/hashicorp/structure/pull/81 and https://github.com/hashicorp/design-system-website/pull/20

//

Part of: https://github.com/hashicorp/design-system-website/issues/17

PR #798, opened 2021-10-12
Add sb extract to enable Storybook composition

amyrlam:
This PR generates a https://boundary-ui-storybook-n10zglibh-hashicorp.vercel.app/stories.json so that it can be consumed by `design-system-website` Storybook.

Ref: https://github.com/hashicorp/cloud-ui/pull/1457 and https://github.com/hashicorp/design-system-website/blob/main/taskbox/.storybook/main.js

PR #803, opened 2021-10-13
Add Storybook metadata.json

amyrlam:
WAS: 

The Boundary /metadata.json was being read as HTML like the Vault metadata.json in this pic. That caused the Boundary Storybook to not be read in Storybook composition (like the Vault one has the error rn on http://design-system-website.vercel.app/)

image

***

IS:

WIth the Boundary /metadata.json, Boundary is visible in external Storybook.

image

***

Follow-on to https://github.com/hashicorp/boundary-ui/pull/798 and replaces closed https://github.com/hashicorp/boundary-ui/pull/800

### DIFF
--- a/ui/desktop/tests/end2end/authentication.spec.js
+++ b/ui/desktop/tests/end2end/authentication.spec.js
@@ -1,14 +1,14 @@
 /* eslint-disable no-undef */
 const { test, expect } = require('@playwright/test');
 const { _electron: electron } = require('playwright');
-const {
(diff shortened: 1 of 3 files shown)


didoo:
@randallmorey for context: this is a temporary website that should have a short life, around 3-6 months (don't worry, is not that kind of thing that starts as "temp" and then remains for years, is really temporary! ) so the risk of changing the url is acceptable for us.

/cc @amy

PR #828, opened 2021-10-29
Make UI reverse proxy path compatible

rjshrjndrn:
Hi,
I'm running the boundary in my infrastructure, where I have a reverse proxy and an FQDN(myorg.com) with a valid SSL certificate. So theoretically I should be able to access boundary UI and the API server behind the proxy with a path like `myorg.com/boundary`, which will offload tasks like maintaining another domain/SSL etc. 
API will work with `myorg.com/bounday` but UI won't; because when the initial request for `/index.html` is passed, the response has all resources pointing to the root path `/style.css` rather than `/boundary/style.css` which will pass the request to another root ( because proxy will forward all request to `/boundary` to my boundary instance, and `/` to default app)

So if there is any way to prepend the proxy_path then the reverse proxy can pass the requests to `/boundary/<resource>`

image

### DIFF
--- a/ui/admin/config/environment.js
+++ b/ui/admin/config/environment.js
@@ -2,12 +2,13 @@
 
 const APP_NAME = process.env.APP_NAME || 'Boundary';
 const API_HOST = process.env.API_HOST || '';
+const ROOT_URL = process.env.ROOT_URL || '/';
 
 module.exports = function (environment) {
   let ENV = {
     modulePrefix: 'admin',
     environment,
-    rootURL: '/',
+    rootURL: ROOT_URL,
     locationType: 'auto',
     EmberENV: {
       FEATURES: {
--- a/ui/desktop/config/environment.js
+++ b/ui/desktop/config/environment.js
@@ -1,12 +1,13 @@
 'use strict';
 
 const APP_NAME = process.env.APP_NAME || 'Boundary';
+const ROOT_URL = process.env.ROOT_URL || '/';
 
 module.exports = function (environment) {
   let ENV = {
     modulePrefix: 'desktop',
     environment,
-    rootURL: process.env.EMBER_CLI_ELECTRON ? '' : '/',
+    rootURL: process.env.EMBER_CLI_ELECTRON ? '' : ROOT_URL,
     locationType: process.env.EMBER_CLI_ELECTRON ? 'hash' : 'auto',
     EmberENV: {
       FEATURES: {


rjshrjndrn:
@calcaide Hi, This is my first contribution to this project. This feature will be useful for lots of consumers, as we can easily deploy Boundary under the same domain name/path. If you can review and provide suggestions, that'll be greatly helpful. Cheers :)

calcaide:
Hi @rjshrjndrn

Thanks for the contribution!
Can you provide a description of the use case you want to apply this solution to? It will help to understand the change.

rjshrjndrn:
@calcaide I've updated the description. Please have a look.

randallmorey:
@rjshrjndrn this is an interesting use case.  What is the benefit of hosting Boundary on an arbitrary path, as opposed to a subdomain?

rjshrjndrn:
Same thing as not to manage separate domain name and ssl.

For example, for Dev environment you can have dev.domain.org/boundary, rather than boundary.dev.domain.org(which need a separate ssl) 
Same goes for staging.domain.org preprod.domain.org... So on and so forth. 

On November 2, 2021 7:30:17 PM GMT+05:30, Randall Morey ***@***.***> wrote:
***@***.*** this is an interesting use case.  What is the benefit of hosting Boundary on an arbitrary path, as opposed to a subdomain?

calcaide:
Hi @rjshrjndrn

After further discussion with the team seems this use case is common and we will investigate a more robust solution, but we can not promise a release date at this time.

In the meantime, we assume your actual workaround will work for your use case.

rjshrjndrn:
@calcaide Glad to hear this in your Roadmap. But may I know what you meant by robust solution? I think I depict the use case in description diagram. And the only issue here is assets path, AFAIK.

randallmorey:
@rjshrjndrn Thanks again for raising this issue.  This is a great approach for static scenarios, where the path is known at UI build time.  However in production scenarios (where the UI is prebuilt into the Boundary binary) we need to support a dynamically-configured path (known only at runtime).  And we may discover that some BE features require knowledge of the path as well, so more investigation is required on our end.

rjshrjndrn:
I got your point of investigating other features knowledge about boundary path( for cli, if the address is not 127.0.0.1 you've to pass the - addr flag always.). But that applies for aubdoians also, right. Rather than giving subdoin, we'll give the path with FQDN. So I don't think that'll be a blocker.

But the change which I made is to read the value from env variable when the application starts. And that's not static, but the system's env variable. 

On November 5, 2021 9:19:38 PM GMT+05:30, Randall Morey ***@***.***> wrote:
***@***.*** Thanks again for raising this issue.  This is a great approach for static scenarios, where the path is known at UI build time.  However in production scenarios (where the UI is prebuilt into the Boundary binary) we need to support a dynamically-configured path (known only at runtime).  And we may discover that some BE features require knowledge of the path as well, so more investigation is required on our end.

randallmorey:
The UI is built statically and baked into the binary.  The environment variables found in the config are read-once at build time, then output to the static build.  In order to achieve dynamic runtime paths, we'll need a holistic solution where the UI can receive configuration from the server.

The UI works anywhere it runs on the root path, which includes subdomains, with no additional configuration.  The problem only occurs when hosting Boundary on a non-root path.

rjshrjndrn:
Oh okay got it. I thought the config was read on run time. Makes sense.

PR #832, opened 2021-11-01
Checkbox admin sessions filter

DhariniJeeva:
This pr adds `checkbox group` to the toolbar component on sessions admin.  onChange functionality will be addressed in the upcoming PR


calcaide (changes requested):
Notice you are submitting `ui/admin/package-lock.json`. This file does not existed before, reason why is because we run `yarn install` within the root of the project. So `boundary-ui/yarn.lock` is the one we consider the source of truth.

I also think, you ran `npm install` rather than `yarn install` that's why it created a `package-lock.json` rather than `yarn.lock`.

DhariniJeeva:
> yarn install



> Notice you are submitting `ui/admin/package-lock.json`. This file does not existed before, reason why is because we run `yarn install` within the root of the project. So `boundary-ui/yarn.lock` is the one we consider the source of truth.
> 
> I also think, you ran `npm install` rather than `yarn install` that's why it created a `package-lock.json` rather than `yarn.lock`.

you are right. my bad. fixing this

ui/admin/package.json
```diff
@@ -63,6 +63,7 @@
     "ember-cli-sri": "^2.1.1",
     "ember-cli-template-lint": "^2.0.2",
     "ember-cli-terser": "^4.0.2",
+    "ember-composable-helpers": "^4.5.0",
```
calcaide, line 66: Nice! We are not seeing a change in `yarn.lock` because we already add `ember-composable-helpers: ^4.5.0` So all good!

calcaide (approved):
LGTM

ui/admin/app/controllers/scopes/scope/sessions/index.js
```diff
@@ -0,0 +1,9 @@
+import Controller from '@ember/controller';
+import { inject as service } from '@ember/service';
+import { tracked } from '@glimmer/tracking';
+export default class ScopesScopeSessionsIndexController extends Controller {
+  // =services
+
+  @service intl;
+  @tracked selectedItems;
```
gsusmi, line 8: `items` doesn't fit quite right. Thoughts on `selectedFilters` or even just `filters`?

ui/admin/app/templates/scopes/scope/sessions/index.hbs
```diff
@@ -22,7 +22,23 @@
               <Rose::Dropdown
                 @text={{t 'form.status.label'}}
                 @ignoreClickInside={{true}}
-              />
+                as |dropdown|
+              >
+                <Rose::Form::Checkbox::Group
+                  @items={{this.availableSessions}}
+                  @selectedItems={{this.selectedItems}}
+                  @onChange={{this.checkboxGroupChanged}}
```
gsusmi, line 30: Is `checkboxGroupChanged` already defined? If this is a route action, we should call out that it's related to filters somehow. Thoughts on changing fn name to `onFilterSelection` or `filterUpdate`?

ui/admin/app/controllers/scopes/scope/sessions/index.js
DhariniJeeva: filters sound good! I'll update

ui/admin/app/templates/scopes/scope/sessions/index.hbs
DhariniJeeva: this is a route action and can change to a better sounding name. We can go with `filterUpdate` to be in parity with other route actions

ui/admin/app/routes/utils/setupFilters.js
```diff
@@ -0,0 +1,4 @@
+export function setupFilters(controller, route, queryParam) {
+  const item = controller.controllerFor(route)[`filter-${queryParam}`];
+  return item ? JSON.parse(item) : null;
```
calcaide, line 3: Is this return working?

Assuming `item` is a `string` or `object`. I recreate this return with an `if` statement and did not work for me, maybe I am missing something.

randallmorey:
Closing in favour of newer PRs.

PR #836, opened 2021-11-05
fix: 🐛 fix broken LinkButton component and associated new tests

randallmorey:
The LinkButton was silently broken due to a problem with a recent refactor.  This PR fixes that problem and updates the create acceptance tests in admin to test for the expected URLs in link buttons.


addons/rose/addon/components/rose/link-button/index.hbs
```diff
@@ -1,8 +1,6 @@
 <LinkTo
   class="rose-link-button {{if @style (concat "rose-button-" @style)}} {{if @iconOnly 'has-icon-only'}} {{if @iconLeft 'has-icon-left'}} {{if @iconRight 'has-icon-right'}}"
-  @route={{if @route @route}}
-  @query={{if @query @query}}
```
DhariniJeeva, line 4: we don't need these two arguments anymore?

DhariniJeeva (approved):
LGTM! thank you!

PR #837, opened 2021-11-05
Remove ember-get-config

randallmorey:
Boundary UI used `ember-get-config` in multiple places, but this was ultimately unnecessary.  Removing our dependence on this addon brings us a step closer to adopting [Embroider](https://github.com/embroider-build/embroider).

### DIFF
--- a/addons/api/tests/unit/adapters/application-test.js
+++ b/addons/api/tests/unit/adapters/application-test.js
@@ -1,6 +1,5 @@
 import { module, test } from 'qunit';
 import { setupTest } from 'ember-qunit';
-import config from 'ember-get-config';
 import RESTAdapter from '@ember-data/adapter/rest';
 import { InvalidError } from '@ember-data/adapter/error';
 import setupMirage from 'ember-cli-mirage/test-support/setup-mirage';
@@ -9,6 +8,12 @@ import { Response } from 'miragejs';
 module('Unit | Adapter | application', function (hooks) {
   setupTest(hooks);
   setupMirage(hooks);
+
+  let config;
+
+  hooks.beforeEach(function () {
+    config = this.owner.resolveRegistration('config:environment');
+  });
 
   test('its namespace is equal to the configured namespace', function (assert) {
     assert.expect(2);
--- a/addons/core/addon/helpers/doc-url.js
+++ b/addons/core/addon/helpers/doc-url.js
@@ -1,20 +1,23 @@
-import { helper } from '@ember/component/helper';
+import Helper from '@ember/component/helper';
+import { getOwner } from '@ember/application';
 import { assert } from '@ember/debug';
-import config from 'ember-get-config';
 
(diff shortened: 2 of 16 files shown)


addons/api/tests/unit/adapters/application-test.js
```diff
@@ -10,6 +9,12 @@ module('Unit | Adapter | application', function (hooks) {
   setupTest(hooks);
   setupMirage(hooks);
 
+  let config;
+
+  hooks.beforeEach(function () {
```
calcaide, line 14: Do we want to retrieve the value of `config` before each test? or once before all tests?

addons/core/addon/helpers/doc-url.js
```diff
@@ -1,20 +1,23 @@
...
+export default class extends Helper {
+  compute(params) {
+    const config = getOwner(this).resolveRegistration('config:environment');
+    const baseURL = config.documentation.baseURL;
+    const docKey = params[0] || '';
+    const configuredPath = config.documentation.topics[docKey];
+    if (docKey) {
+      assert(
+        `
+        Documentation for "${docKey}" could not be found. Please ensure that
```
calcaide, line 14: Asking with future improvements in mind: there is a possibility to include this text to the translations files? or does not make sense to add it to translations?

addons/core/addon/services/confirm.js
```diff
@@ -17,13 +17,21 @@ import config from 'ember-get-config';
...
   /**
    * When the confirm service is disabled, it always returns resolving promises.
    * By default the service is enabled but it is useful to disable it for test
    * purposes.
    * @type {boolean}
    */
-  enabled = config.enableConfirmService ?? true;
+  get enabled() {
+    const config = getOwner(this).resolveRegistration('config:environment');
+    return this.#enabled ?? config.enableConfirmService ?? true;
```
calcaide, line 30: This return is a bit weird to me :sweat_smile: is something ember related or vanila js?

calcaide (approved):
LGTM. Thanks to prepare the ground for the coming changes :muscle:

addons/core/addon/services/confirm.js
randallmorey: We sometimes set these values in tests, which we need to account for here.  Essentially it's saying:  return the first value you find in the list:  private enabled, config enabled, or literally true.

addons/api/tests/unit/adapters/application-test.js
randallmorey: It can only be retrieved in the context of a test case, so it is looked up every time.

addons/core/addon/helpers/doc-url.js
randallmorey: This is an error message for developers, not end users.  Currently, only strings presented to the user are intended for translation.

PR #838, opened 2021-11-05
Add managed-groups basic list route to admin

calcaide:
Add basic managed-groups list within auth methods.

Screen Shot 2021-11-08 at 11 16 44 AM


gsusmi (approved):
LGTM. Ty!

PR #841, opened 2021-11-09
Feature managed groups read CRUD

calcaide:
Add read CRUD operation to managed groups.
- The breadcrumbs are not working properly, we will take care of those later.
- The form has no action buttons yet, will be added on a future PR.
- Actions dropdown is working but not (yet) the delete function. That will be implemented on the delete PR.

[**Preview Link**](https://boundary-ui-git-feature-managed-group-read-hashicorp.vercel.app/scopes/global/auth-methods/auth-method-id-0/managed-groups/managed-groups-id-0)

Screen Shot 2021-11-09 at 10 30 00 AM


randallmorey (changes requested):
The managed group tab should only appear when the auth-method has a type of `oidc`.  Password auth methods don't have managed groups.

As an aside, you might also tweak the mocks such that managed groups are _only generated for OIDC auth methods_.  This isn't critical to the mocks, but makes it a bit more realistic.

DhariniJeeva:
more of a design question - thoughts on having a description for `filter` same as the above fields?

calcaide:
> The managed group tab should only appear when the auth-method has a type of `oidc`. Password auth methods don't have managed groups.
> 
> As an aside, you might also tweak the mocks such that managed groups are _only generated for OIDC auth methods_. This isn't critical to the mocks, but makes it a bit more realistic.

Absolutely! nice point. I already applied both changes.
- Just show the tab 'managed groups' when the type of the auth method is `oidc`. 
- Create managed-groups just on the auth method with type `oidc`.
- Bonus track. I added more than just 1 auth-method.

calcaide:
> more of a design question - thoughts on having a description for `filter` same as the above fields?

@DhariniJeeva that is an excellent point. I did add description to `filter` [based on our documentation](https://www.boundaryproject.io/docs/concepts/domain-model/managed-groups#filter).

@gsusmi @randallmorey What do you guys think? should we keep that description on `filter`? Also, should I add description to type?

Screen Shot 2021-11-09 at 11 28 56 AM

ui/admin/app/templates/scopes/scope/auth-methods/auth-method/index.hbs
```diff
@@ -1,3 +1,4 @@
+<h2>{{@model.type}}</h2>
```
randallmorey, line 1: Is this for development purposes?
Screenshot 2021-11-10 at 08 36 02

randallmorey (changes requested):

ui/admin/app/templates/scopes/scope/auth-methods/auth-method/index.hbs
calcaide: That was definitely unintentional :sweat_smile:

gsusmi (approved):
LGTM. Ty!

addons/core/translations/resources/en-us.yaml
```diff
@@ -354,3 +354,6 @@ scope:
 managed-group:
   title: Managed Group
   title_plural: Managed Groups
+  description: A managed group is a resource that represents a collection of accounts.
```
randallmorey, line 357: "A managed group represents a collection of accounts."  I'd also mention something about the automated/managed nature via a filter.  BE might be able to help here.

randallmorey (approved):
Provisional approval.  Had one comment in the code.  It would also be nice to get acceptance tests included with this, but they can come later if that's easier.

addons/core/translations/resources/en-us.yaml
calcaide: - Change done.
- Also, asked BE some help with enhancing this description.

calcaide:
> Provisional approval. Had one comment in the code. It would also be nice to get acceptance tests included with this, but they can come later if that's easier.

Definitely all the tests are coming on a separate PR just for testing.

PR #853, opened 2021-11-11
Checkbox style updates

DhariniJeeva:
dropdown with checkboxes-https://boundary-ui-storybook-53via7z7a-hashicorp.vercel.app/?path=/story/rose-dropdown--dropdown-with-checkboxes


addons/rose/addon/components/rose/form/checkbox/index.stories.mdx
```diff
@@ -78,6 +78,7 @@ Renders a form input checkbox.
   @error={{false}}
   @disabled={{false}}
   @inline={{false}}
+  @outline={{false}}
```
randallmorey, line 81: Outline shouldn't be a component attribute.  Just make it easy to toggle with CSS in case we decide not to use it for dropdown items.

This also does not apply to the base checkbox.  The base checkbox must always have an outline.

addons/rose/addon/styles/rose/components/form/_checkbox.scss
```diff
@@ -44,8 +44,6 @@
       height: sizing.rems(m);
       pointer-events: none;
       content: ' ';
-      border: sizing.rems(xxxxs) solid var(--ui-gray-subtler-2);
```
randallmorey, line 47: Make no changes to the base checkbox.

randallmorey (changes requested):
It looks like the checkbox items are very large compared to the radio items.  They should be the same size.  We also need a hover state similar to radio.  And possibly a highlighted state for when the item is checked.  That would really make it stand out.  Basically, just copy the radio items, but use a checkbox.

The outline itself feels large compared to original checkbox and the radio.  Can it be the same size as a standard checkbox?

Screenshot 2021-11-11 at 14 42 11

Screenshot 2021-11-11 at 14 42 08

addons/rose/addon/styles/rose/components/form/_checkbox.scss
DhariniJeeva: removed this. will update `outline` logic

randallmorey: The diff is showing that these lines are still removed.  There should be no changes to the base checkbox styles.

DhariniJeeva:
@randallmorey updated the styles!

randallmorey (approved):

PR #855, opened 2021-11-12
feat: List accounts within a managed group

gsusmi:
managed-group-accounts


addons/core/translations/resources/en-us.yaml
```diff
@@ -362,3 +362,6 @@ managed-group:
     welcome:
       title: Welcome to Managed Groups
       description: Create and manage managed groups for an auth method.
+    accounts:
+      title: Welcome to Accounts
+      description: No accounts in this managed group.
```
calcaide, line 367: This shouldn't be part of empty state?

calcaide (approved):
LGTM!! Thank you for the work

calcaide:
Updated PR to refactor `accounts` for `members` within `managed-groups`. Added breadcrumbs

Screen Shot 2021-11-12 at 10 54 15 AM

addons/core/translations/resources/en-us.yaml
gsusmi: This is used in a new `members` tab, thus isn't an empty state for managed group.

addons/core/translations/resources/en-us.yaml
```diff
@@ -362,3 +362,9 @@ managed-group:
     welcome:
       title: Welcome to Managed Groups
       description: Create and manage managed groups for an auth method.
+    member:
+      title: Welcome to Accounts
```
gsusmi, line 366:
```suggestion
      title: Welcome to Members
```

addons/core/translations/resources/en-us.yaml
```diff
@@ -362,3 +362,9 @@ managed-group:
...
+      description: No accounts in this managed group.
```
gsusmi, line 367:
```suggestion
      description: No members in this managed group.
```

addons/core/translations/resources/en-us.yaml
```diff
@@ -362,3 +362,9 @@ managed-group:
...
+member:
+  title: Member
+  title_plural: Members
```
gsusmi, line 370: Thoughts on nesting this under managed-group? `target` resource is a good example of this.

ui/admin/app/controllers/scopes/scope/auth-methods/auth-method/managed-groups/members.js
```diff
@@ -1,7 +1,7 @@
 import Controller from '@ember/controller';
 import { inject as service } from '@ember/service';
 
-export default class ScopesScopeAuthMethodsAuthMethodManagedGroupsNewController extends Controller {
+export default class ScopesScopeAuthMethodsAuthMethodManagedGroupsMembersController extends Controller {
```
gsusmi, line 4: This file needs to be nested under `../auth-method/managed-groups/managed-group/member.js`

addons/core/translations/resources/en-us.yaml
calcaide: Definitely including this, makes all the sense :)

ui/admin/app/controllers/scopes/scope/auth-methods/auth-method/managed-groups/members.js
calcaide: Updated. Good catch!

PR #856, opened 2021-11-12
Fix: Add empty string and array filter checks to resource filters

gsusmi:


randallmorey:
This is fixed in `feature-auth-methods-filters`

PR #861, opened 2021-11-16
feat: 🎸 Add managed groups featuer flag for production

calcaide:


DhariniJeeva:
could we remove the dev and test flags for `managed-groups`?

DhariniJeeva (commented):
and I think we need to enable the flag on L100

randallmorey (changes requested):
Let's flip the feature on line 100 so that it's on by default for all environments, and then remove environment-specific managed group flags.

DhariniJeeva (approved):
awesome! thank you!

gsusmi (approved):
LGTM. Ty!

PR #874, opened 2021-11-30
feat: 🎸 add filter to cred stores

DhariniJeeva:


randallmorey:
Great work on this!  This is perfect _but_ I think we should sit on it until we have multiple cred store types.  Otherwise we're showing a filter UI with only one option.  Thoughts @gsusmi ?

If that's agreeable to folks, I suggest changing the conditional in the template to always hide the filter UI (e.g. `{{#if false}}` or something, if that's even allowed).  That way we're ready to go quickly when we have multiple types to show.

DhariniJeeva:
closing this for now. will bring it back up when we have multiple types. ty!

PR #877, opened 2021-12-06
Refactor resource filters

randallmorey:
This PR introduces dynamic resource filtering, static usage, and updates naming of related concepts.


addons/core/addon/decorators/resource-filter.js
```diff
@@ -0,0 +1,251 @@
...
+    defaultValue,
+    serialize,
+    findBySerialized
+  ) {
+    const owner = getOwner(routeInstance);
+    const containerKey = `resource-filter:${name}@${routeInstance.routeName}`;
+    const factory = owner.factoryFor(containerKey);
+    if (!factory) {
+      // create factory
+      class AppliedRouteResourceFilter extends this {
```
calcaide, line 154: Is this class declaration extending from `RouteResourceFilter` ?

addons/core/addon/decorators/resource-filter.js
```diff
@@ -0,0 +1,251 @@
...
+   */
+  get filterKey() {
+    return `filter-${this.name}`;
+  }
+
+  /**
+   * The allowed values for this filter.  See `allowed`.
+   * @type {?array}
+   */
+  get allowedValues() {
```
calcaide, line 40: For all the computed functions declared here, do we plan to unit test them?

calcaide (approved):
LGTM! AWESOME job on the dynamic filtering system Randy!! :muscle: :muscle:

addons/core/addon/decorators/resource-filter.js
randallmorey: Yes.  `this` in the context refers to the class itself, because the method is `static`.

addons/core/addon/decorators/resource-filter.js
randallmorey: I've been lax on this because filtering is in flux.  Testing is something I'd like to come back to in another pass.

PR #880, opened 2021-12-10
Desktop sessions filtering

randallmorey:
This PR introduces dynamic filtering of sessions in Desktop by project.  This is useful UX enhancement because Desktop aggregates sessions across scopes.

[Demo link](https://boundary-ui-desktop-2136btvqm-hashicorp.vercel.app/scopes/global/projects/sessions?filter-project=%5B%22s_s9n819zgko7%22%2C%22s_xk1i4dk50v8%22%5D)

Screenshot 2021-12-13 at 13 59 05

### DIFF
--- a/addons/api/mirage/config.js
+++ b/addons/api/mirage/config.js
@@ -26,16 +26,18 @@ export default function () {
 
   this.get(
     '/scopes',
-    ({ scopes }, { queryParams: { scope_id, recursive } }) => {
+    ({ scopes }, { queryParams: { scope_id, recursive, filter } }) => {
+      let resultSet;
       // Default parent scope is global
       if (!scope_id) scope_id = 'global';
       if (recursive && scope_id === 'global') {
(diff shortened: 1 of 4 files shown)


randallmorey:
@joshuaogle one question from a design perspective is how to display projects. Since the filter dropdown could show projects from multiple orgs, should we qualify them?  e.g. `org-name/project-name`

calcaide (approved):
LGTM!

joshuaogle:
@randallmorey Yeah, we definitely should. We could do it that way, but I'm a little afraid that the org names would make us run out of room (are they truncated or do they wrap?).

What do you think about adding a little height to add the org name above the project name as helper text?

image

calcaide:
@joshuaogle @randallmorey Perhaps we can add some ellipsis to the org name. So if it's too long we cut it and add the `...` at the end?

PR #881, opened 2021-12-14
feat: 🎸 Add table header for hosts list within targets

calcaide:
Add table header for hosts list within targets.

Also change the columns order to reflect latest design, [link here](https://www.figma.com/file/a8vdiYPLJqAw5RxGwpTMvM/Dynamic-Hosts-Catalog?node-id=305%3A9700)

[Direct link to desktop](https://boundary-ui-desktop-git-feature-add-targets-ho-66b587-hashicorp.vercel.app/scopes/global/projects/targets/1/hosts)

Before: No header
Screen Shot 2021-12-14 at 10 23 52 AM

After:
Screen Shot 2021-12-14 at 10 19 54 AM


randallmorey (changes requested):
We should drop the type column.

randallmorey (approved):

PR #886, opened 2021-12-30
add static port and address mappings

maksim-paskal:
This PR add simple static port mappings logic that will create new proxy on fixed local port and IP on selected host in Boundary Desktop

This change can help users that need host proxy on specific local port, in my case I have Mysql Workbench that connect via Boundary Desktop to remote cloud databases, I have 20+ database templates that contains database credentials - fixed port to databases is really helps my to work with multiples databases in various clouds

If target name has name that starts with 3 digits and dash (regexp `^(\d{3})-.+$`), and host name starts with number between 0-99 and dash (regexp `^(\d{1}|\d{2})-.+$`) - only in this case Boundary Desktop will create proxy to selected host on port that will contains digits from target and host name name

If host name has name that ends with dash and IP address (regexp `.+-(\d+.\d+.\d+.\d+)$`) - in this case Boundary Desktop will create proxy that will be listened on selected IP

| Target name  | Host name | Proxy address that will be created |
| ------------- | ------------- | ------------- |
| 181-Database service AZURE PROD  | 5-az-dbs-payment-master  | 127.0.0.1:18105  |
| 161-Database service AWS PROD  | 17-promo-master  | 127.0.0.1:16117  |
| 171-Database service AWS DEV  | 45-main-master-0.0.0.0  | 0.0.0.0:17145  |

Signed-off-by: Maksim Paskal <paskal.maksim@gmail.com>

### DIFF
--- a/ui/desktop/app/routes/scopes/scope/projects/targets.js
+++ b/ui/desktop/app/routes/scopes/scope/projects/targets.js
@@ -95,7 +95,27 @@ export default class ScopesScopeProjectsTargetsRoute extends Route {
         token: this.session.data.authenticated.token,
       };
 
-      if (host) options.host_id = host.id;
+      if (host) {
+        options.host_id = host.id;
+
+        // If target name starts with 3-numbers and dash example 123-test-target
+        // And host name starts with 2-numbers and dash example 1-test-host
+        // Boundary Desktop will create local proxy to selected target on local port based on numbers in name of target and host example 127.0.0.1:12301
+        const targetPortMatch = model.name.match(/^(\d{3})-.+$/)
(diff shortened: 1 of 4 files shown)


calcaide:
@maksim-paskal 
Support for configuration of a local proxy port on connect in Boundary Desktop would improve the UX for use cases like yours. We really appreciate your raising this. We will account for this new UX requirement internally, but we don't have a timeline for it just yet. 

This PR is a great proof-of-concept and gives us something to think about. We will need to integrate information from other users and use cases as well and the final result may differ from your proposal. 

Stay tuned!
Thank you for your interest in Boundary; we appreciate your contribution.

PR #887, opened 2022-01-02
refactor: 💡 copyable component no longer uses ember-concurrency

randallmorey:
In an effort to remove uses of ember-concurrency in preparation for Ember 4.0, this PR refactors the copyable components to use pure Ember code instead.  Using `later()`, the same effect is achieved with no third-party dependency.

This PR has no user-facing effect.


DhariniJeeva (approved):

gsusmi (approved):
LGTM. Ty!

PR #898, opened 2022-01-10
chore: 🤖 specify only LTS node versions in packages

randallmorey:
This PR pins the allowable versions of Node to LTS 12, 14, and 16 only.  A question came up recently as to which version to use and we had accidentally specified Node 17 as allowable, but it appears to be incompatible.


addons/auth/package.json
```diff
@@ -66,7 +66,7 @@
     "qunit-dom": "^2.0.0"
   },
   "engines": {
-    "node": ">=12 <=17"
+    "node": "12.* || 14.* || 16.*"
```
ChaiWithJai, line 69: `thought (non-blocking)`:

Typically, I use [ember-new-output](https://github.com/ember-cli/ember-new-output/blob/v3.28.3/package.json) to model dependency versions. From there, I notice the following setting:

`"12.* || 14.* || >= 16"`

From here, I reviewed the change log for `16.*`. I noticed there was an upgrade to `npm 8` which does have breaking changes for dropping `node: 12` support.

`question:`

Would the above statement impact `ember-addons` or other dependencies? I'm noticing that the latest version of `ember-cli` isn't shipping with `node:  16.*`. Wondering why that would be the case.

randallmorey: I want to get away from catchall ranges like `>=16`, which means _everything in the future forever_.  It's just not an assertion we can realistically support.  Regarding NPM, we assume that the environment will have a compatible node/npm combination.

As for `ember-cli` the [package.json](https://github.com/ember-cli/ember-cli/blob/master/package.json) specifies Node `>= 12`, which is also a catchall assertion we don't necessarily agree with.

I see these fields mostly as guidance to developers though.  We're essentially making a statement that we believe this project will build in Node 12, 14, and 16.

calcaide (approved):
LGTM thank you for this Randy!!

PR #900, opened 2022-01-11
feat: 🎸 Add host catalog type basic support

calcaide:
Add host catalog type basic support. Delete previous conditional rendering to just render static types.

This PR already adds this model changes within Mirage.

[Direct link](https://boundary-ui-23zp7iqbu-hashicorp.vercel.app/scopes/s_xk1i4dk50v2/host-catalogs)

Screenshot:
Screen Shot 2022-01-12 at 12 23 31 PM

### DIFF
--- a/addons/api/addon/models/host-catalog.js
+++ b/addons/api/addon/models/host-catalog.js
@@ -40,4 +40,16 @@ export default class HostCatalogModel extends GeneratedHostCatalogModel {
   get compositeType() {
     return this.isPlugin ? this.plugin.name : this.type;
   }
+
+  /**
+   * Sets type, if type is different than static, sets plugin name to type
+   */
+  set compositeType(type) {
+    if (type === 'static') {
+      this.type = 'static';
+    } else {
+      this.type = 'plugin';
+      this.plugin.name = type;
+    }
+  }
 }
--- a/addons/api/addon/models/host-catalog.js
+++ b/addons/api/addon/models/host-catalog.js
@@ -2,6 +2,14 @@ import GeneratedHostCatalogModel from '../generated/models/host-catalog';
 
 export default class HostCatalogModel extends GeneratedHostCatalogModel {
   /**
+   * Returns if the host-catalog is static type or not.
+   * @type {boolean}
+   */
+  get isStatic() {
+    return this.type === 'static';
+  }
+
+  /**
    * Returns if the host-catalog is a plugin or not.
    * @type {boolean}
    */
@@ -23,5 +31,13 @@ export default class HostCatalogModel extends GeneratedHostCatalogModel {
    */
   get isAzure() {
     return this.plugin.name === 'azure';
+  }
+
+  /** If host-catalog is a plugins returns its name,
+   * otherwise returns the host-catalog type
+   * @type {string}
+   */
+  get compositeType() {
+    return this.isPlugin ? this.plugin.name : this.type;
   }
 }
--- a/addons/api/addon/models/host-catalog.js
+++ b/addons/api/addon/models/host-catalog.js
@@ -49,7 +49,7 @@ export default class HostCatalogModel extends GeneratedHostCatalogModel {
       this.type = 'static';
     } else {
       this.type = 'plugin';
-      this.plugin = { name: type }
+      this.plugin = { name: type };
     }
   }
 }
--- a/addons/api/addon/models/host-catalog.js
+++ b/addons/api/addon/models/host-catalog.js
@@ -2,9 +2,26 @@ import GeneratedHostCatalogModel from '../generated/models/host-catalog';
 
 export default class HostCatalogModel extends GeneratedHostCatalogModel {
   /**
+   * Returns if the host-catalog is a plugin or not.
    * @type {boolean}
    */
-  get isStatic() {
-    return this.type === 'static';
+  get isPlugin() {
+    return this.type === 'plugin';
+  }
+
+  /**
+   * Returns if host-catalog plugin is AWS or not
+   * @type {boolean}
+   */
+  get isAWS() {
+    return this.plugin.name === 'aws';
+  }
+
+  /**
+   * Return if host-catalog plugin is Azure or not.
+   * @type {boolean}
+   */
+  get isAzure() {
+    return this.plugin.name === 'azure';
   }
 }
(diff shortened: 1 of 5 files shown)


addons/api/addon/models/host-catalog.js
```diff
@@ -2,9 +2,42 @@ import GeneratedHostCatalogModel from '../generated/models/host-catalog';
...
+    return this.plugin.name === 'azure';
+  }
+
+  /** If host-catalog is a plugins returns its name,
+   * otherwise returns the host-catalog type
+   * @type {string}
+   */
+  get compositeType() {
+    return this.isPlugin ? this.plugin.name : this.type;
+  }
```
randallmorey, line 42: Let's add a `set compositeType(type)` field that handles setting values of `static` or `*`.

- If setting this field to `static`, the underlying field `type` should be set to `static`.
- For any other value, `type` should be set to `plugin` and `plugin.name` should be set to the value.

addons/api/mirage/generated/factories/host-catalog.js
```diff
@@ -11,4 +11,15 @@ export default Factory.extend({
...
   updated_time: () => date.recent(),
   disabled: () => datatype.boolean(),
+  plugin: function () {
+    if (this.type == 'plugin') {
+      return {
+        id: `pl_${datatype.hexaDecimal(5)}`,
+        name: random.arrayElement(['azure', 'aws']),
+        description: random.words(),
+      };
+    } else {
```
randallmorey, line 21: The else is unnecessary since `undefined` is the default return value.

randallmorey (changes requested):
See review comments.  We need to support _setting_ of the `compositeType` field as well as getting.  Ideally application code doesn't interact with `plugin` field directly.

addons/api/tests/unit/models/host-catalog-test.js
```diff
@@ -19,4 +19,75 @@ module('Unit | Model | host catalog', function (hooks) {
...
+    assert.expect(3);
+    const store = this.owner.lookup('service:store');
+    const modelPlugin = store.createRecord('host-catalog', {
+      type: 'plugin',
+      plugin: { name: 'Test name' },
+    });
+    const modelStatic = store.createRecord('host-catalog', {
+      type: 'static',
+    });
+    modelPlugin.set('compositeType', 'aws');
```
calcaide, line 86: @randallmorey I could not find a pattern to test set in our codebase. I found this in a old ember documentation. Thoughts?

addons/api/addon/models/host-catalog.js
```diff
@@ -2,9 +2,54 @@ import GeneratedHostCatalogModel from '../generated/models/host-catalog';
...
+  get isPlugin() {
+    return this.type === 'plugin';
+  }
+
+  /**
+   * Returns if host-catalog plugin is AWS or not
+   * @type {boolean}
+   */
+  get isAWS() {
+    return this.plugin.name === 'aws';
```
randallmorey, line 25: Should this use `this.compositeType`?

addons/api/addon/models/host-catalog.js
```diff
@@ -2,9 +2,54 @@ import GeneratedHostCatalogModel from '../generated/models/host-catalog';
...
+
+  /**
+   * Sets type, if type is different than static, sets plugin name to type
+   */
+  set compositeType(type) {
+    if (type === 'static') {
+      this.type = 'static';
+    } else {
+      this.type = 'plugin';
+      this.plugin.name = type;
```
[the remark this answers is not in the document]
randallmorey: The only time it is valid to _set_ the `plugin` field is for a new model instance.  In this case, the field is expected to be initially undefined, so it should set the whole object.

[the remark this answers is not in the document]
randallmorey: If we were to ever set `plugin.id` we'd need to revisit this.  But for now I think it's fine.

addons/api/mirage/generated/factories/host-catalog.js
```diff
@@ -11,4 +11,13 @@ export default Factory.extend({
   created_time: () => date.recent(),
   updated_time: () => date.recent(),
   disabled: () => datatype.boolean(),
+  plugin: function () {
+    if (this.type == 'plugin') {
+      return {
+        id: `pl_${datatype.hexaDecimal(5)}`,
+        name: random.arrayElement(['azure', 'aws']),
```
randallmorey, line 18: The current index of the instance is passed into the function, so be sure to add the parameter on line 14, then:
```suggestion
        name: i % 2 ? 'aws' : 'azure',
```

addons/api/tests/unit/models/host-catalog-test.js
```diff
@@ -19,4 +19,75 @@ module('Unit | Model | host catalog', function (hooks) {
...
+    assert.equal(modelPlugin.compositeType, 'Test name');
+    assert.equal(modelStatic.compositeType, 'static');
+  });
+
+  test('set compositeType sets expected values', async function (assert) {
+    assert.expect(3);
+    const store = this.owner.lookup('service:store');
+    const modelPlugin = store.createRecord('host-catalog', {
+      type: 'plugin',
+      plugin: { name: 'Test name' },
```
randallmorey, line 81:
```suggestion
      compositeType: 'aws',
```

addons/api/tests/unit/models/host-catalog-test.js
randallmorey, line 86:
```suggestion

```

addons/api/tests/unit/models/host-catalog-test.js
```diff
@@ -19,4 +19,75 @@ module('Unit | Model | host catalog', function (hooks) {
...
+    });
+    const modelStatic = store.createRecord('host-catalog', {
+      type: 'static',
```
randallmorey, line 84:
```suggestion
      compositeType: 'static',
```

randallmorey (changes requested):

addons/api/mirage/generated/factories/host-catalog.js
```diff
@@ -11,4 +11,13 @@ export default Factory.extend({
   created_time: () => date.recent(),
   updated_time: () => date.recent(),
   disabled: () => datatype.boolean(),
+  plugin: function () {
```
randallmorey, line 14:
```suggestion
  plugin: function (i) {
```

randallmorey (changes requested):

randallmorey (approved):

PR #901, opened 2022-01-12
Feat radio helper text

DhariniJeeva:


DhariniJeeva:
> I'm fixing up the style a bit. will update here once I'm done.

done!

calcaide (approved):
LGTM

addons/rose/addon/components/rose/form/radio/radio/index.hbs
```diff
@@ -12,7 +12,7 @@
     @disabled={{@disabled}}
     @changed={{@changed}} />
 
-  <label class="rose-form-radio-label" for={{this.id}}>
+  <label class="rose-form-radio-label{{if @helperText " helper-text"}}" for={{this.id}}>
```
randallmorey, line 15: Is there a way to avoid adding another class here?

randallmorey (changes requested):

addons/rose/addon/components/rose/form/radio/radio/index.hbs
DhariniJeeva: yes! works w/o class.

randallmorey (changes requested):
I noticed minor vertical layout changes when completely removing helper text, vs when it is present.  It seems like the layout is correct when helper text is absent and then it switches up when it's added.

DhariniJeeva:
> I noticed minor vertical layout changes when completely removing helper text, vs when it is present. It seems like the layout is correct when helper text is absent and then it switches up when it's added.

good catch! I added a fix for that.

randallmorey (approved):
Perfect!

PR #902, opened 2022-01-12
updates to rose radio component

DhariniJeeva:
This pr migrates the main component class `rose-form-radio` from a static hard-coded value in the template to a component class attribute. this update allows passing in different class names to support radioCard field (upcoming pr)


calcaide (approved):
LGTM

PR #907, opened 2022-01-18
Feat: create radio card component

DhariniJeeva:
This pr introduces `radioCard` field with a card-like style in two formats- tile and wide 
[radio card](https://boundary-ui-storybook-7qlr8ewrf-hashicorp.vercel.app/?path=/story/rose-form-radio-radio-card--basic-tile&knob-disabled=true&knob-error=true&knob-helperText=Record%20financial%20transactions%20using&knob-selectedValue=value%20and%20selectedValue%20must%20match%20to%20select%20radiofield&knob-text=Azure&knob-value=value%20and%20selectedValue%20must%20match%20to%20select%20radiofield)


randallmorey (changes requested):
Can this live under `Rose::Form::Radio::Card`?

To be consistent with the rest of the CSS:
- Use spacing helpers instead of pixel units directly.
- Use type sizing helpers for font size and line height.
- It looks like failure subtler 1 is duplicated with a slightly different naming scheme.  If possible, reuse the colors that already exist.

addons/rose/addon/components/rose/form/radio/card/tile/index.stories.mdx
```diff
@@ -0,0 +1,56 @@
...
+import { text, boolean, select } from '@storybook/addon-knobs';
+
+<Meta title="Rose|Form/Radio/Radio/Card/Tile" component="Form/Radio/Radio/Card/Tile" />
+
+# Form RadioCard field in tile format
+
+Renders a form radioCard field in a tile format.
+
+```htmlbars
+<Rose::Form::Radio::Card::Tile
```
randallmorey, line 12: Can this be moved to `Rose::Form::Radio::Card`, including the associated file paths?

addons/rose/addon/components/rose/form/radio/card/wide/index.js
```diff
@@ -0,0 +1,7 @@
+import RoseFormRadioRadioComponent from '../../radio';
+
+export default class RoseFormRadioRadioCardComponent extends RoseFormRadioRadioComponent {
```
randallmorey, line 3: I see this more as a different style of the card component, rather than an entirely different component.  Open to discussing.

randallmorey (changes requested):
I see this as a single radio card component with a different styles/layouts rather than two separate components.  How many different types do we have and what component attributes would be needed to specify them?

Let's use `_radio-card.scss` instead of `_radioCard.scss`.  We avoid uppercase in filenames because macOS and Linux treat these differently.

addons/rose/addon/components/rose/form/radio/card/wide/index.js
DhariniJeeva: good point! I think that helps contain the layouts of this component.  we can pass in a component attribute such as `layout` and this can be easily extended to other layouts if we decide to add more card styles.

randallmorey (changes requested):
I think we can consolidate the CSS for the two layouts quite a bit.  They differ only in width.  Given the minor difference in widths, I'm not sold on two "styles" or layouts.  Perhaps the width changes only when a description is included?

A couple things I noticed in the storybook:
- The position and sizing of the radio circle/dot is inconsistent between selected and error.  The standard/selected positioning and sizing looks good.  Can this be replicated to the error state?
- Ensure the icon is closer to the designed size and test with logos (these appear in flight icons).

addons/rose/addon/styles/rose/components/form/_radio-card.scss
```diff
@@ -0,0 +1,219 @@
+@use '../../variables/sizing';
+@use '../../utilities/type';
+@use '../../variables/typography';
+
+$l: sizing.rems(l); // 16
```
calcaide, line 5: @randallmorey @DhariniJeeva 

I am not quite understading what is the purpose of defining variables that are sizing. 
Also, I find a lack of consistency on the naming, i.e. `$l` for the size l, but `$size-xxs` for the size xxs.

Why not using directly? Unless I am missing something and taking as example line 21.

Why: `height: $l + $size-xxs`; rather than `height: sizing.rems(l) + sizing.rems(xxs);`?

Mentioning both of you because I think for the sake of consistency and mantainability we should choose a pattern and stick to it (unless specific cases).

randallmorey: The helper is verbose so that's probably why the values get aliased.  As for naming, I agree this should be consistent.

addons/rose/addon/styles/rose/components/form/_radio-card.scss
```diff
@@ -0,0 +1,219 @@
...
+$size-xs: sizing.rems(xs); // 8
+$size-xxs: sizing.rems(xxs); // 4
+$size-xxxs: sizing.rems(xxxs); // 2
+$size-xxxxs: sizing.rems(xxxxs); // 1
+
+.rose-form-radio-card-wide {
+  position: relative;
+  display: flex;
+  align-items: center;
+  width: ($m) * 23;
```
calcaide, line 16: Unless there is a justification for this I will recommend to never use fixed numbers i.e `23` but doing math between the already defined sizes.

I will change `width: ($m) * 23;` to `width: sizing.rems(m) * (sizing.rems(l) - sizing.rems(xxxxs))` or to be honest I will directly `width: sizing.rems(m) * sizing.rems(l)`

addons/rose/addon/styles/rose/components/form/_radio-card.scss
```diff
@@ -0,0 +1,219 @@
...
+    width: 100%;
+    bottom: 0;
+    right: 0;
+  }
+  .rose-form-radio-label {
+    font-style: normal;
+    font-weight: bold;
+    font-size: typography.font-size(s);
+    color: var(--black);
+    width: ($m) * 23;
```
calcaide, line 32: - Avoid hardcoding a measurement number (use sizing).
- Avoid using odd numbers for sizing.

addons/rose/addon/styles/rose/components/form/_radio-card.scss
```diff
@@ -0,0 +1,219 @@
...
+    height: ($m) * 7.5;
```
calcaide, line 33: - Avoid hardcoding a measurement number (use sizing).
- Avoid using odd numbers for sizing.

addons/rose/addon/styles/rose/components/form/_radio-card.scss
```diff
@@ -0,0 +1,219 @@
...
+    display: block;
+    position: relative;
+    text-align: left;
+    word-break: break-all;
+    border: $size-xxxxs solid #bac1cc;
```
calcaide, line 38: Unless there is a specific reason, we should use the colors in the styleguide with its variable.

addons/rose/addon/styles/rose/components/form/_radio-card.scss
```diff
@@ -0,0 +1,219 @@
...
+        ~ .rose-form-radio-label {
+          border: $size-xxxxs solid var(--ui-gray);
+        }
+      }
+      &:checked {
+        ~ .rose-form-radio-label {
+          .radio-card-overlay {
+            background-color: var(--action-subtler-1h);
+          }
+          border: $size-xxxxs solid #1563ff;
```
calcaide, line 68: Unless there is a specific reason, we should use the colors in the styleguide with its variable.

addons/rose/addon/styles/rose/components/form/_radio-card.scss
```diff
@@ -0,0 +1,219 @@
...
+          }
+          border-color: var(--failure);
+        }
+        &:checked {
+          opacity: 0;
+          ~ .rose-form-radio-label {
+            &::after {
+              position: absolute;
+              bottom: $size-xxs + $size-xxxs;
+              right: sizing.rems(m) * 10.9;
```
calcaide, line 97: Avoid hardcoding a measurement number (use sizing).
Avoid using odd numbers for sizing.

addons/rose/addon/styles/rose/components/form/_radio-card.scss
randallmorey: This is ok to do for multiplication, IMHO.  We do it elsewhere.  We're taking a size with units and multiplying by a unitless coefficient to get some arbitrary width.

The reason to do this is because the design system doesn't come with tokens for larger values.  In many cases, the design itself is the source of an arbitrary width.  This means the code is an accurate reflection of the design intent.

addons/rose/addon/styles/rose/components/form/_radio-card.scss
```diff
@@ -0,0 +1,219 @@
...
+      }
+    }
+  }
+}
+
+.rose-form-radio-card-tile {
+  position: relative;
+  display: flex;
+  align-items: center;
+  width: ($m) * 15;
```
calcaide, line 122: Avoid hardcoding a measurement number (use sizing).
Avoid using odd numbers for sizing.

addons/rose/addon/styles/rose/components/form/_radio-card.scss
```diff
@@ -0,0 +1,219 @@
...
+          }
+          border-color: var(--failure);
+        }
+        &:checked {
+          opacity: 0;
+          ~ .rose-form-radio-label {
+            &::after {
+              position: absolute;
+              bottom: $size-xxs;
+              right: sizing.rems(m) * 6.8;
```
calcaide, line 200: Avoid hardcoding a measurement number (use sizing).
Avoid using odd numbers for sizing.

addons/rose/addon/styles/rose/components/form/_radio-card.scss
randallmorey: Before I make any suggestions I'd like to better understand why this coefficient was used.  @DhariniJeeva what is the reason for this?  Is it intending to "subtract" the borders from the value?

addons/rose/addon/styles/rose/components/form/_radio-card.scss
randallmorey: See above comments.  This is OK to do and in many cases necessary.

DhariniJeeva:
I will update the sizing variables to be consistent.  
As far as the color variable, I tried to follow the codes on figma @calcaide  are you suggesting we pick whichever is closer to that color code in our predefined color guide?

calcaide (changes requested):
Left some comments on the PR

randallmorey:
> I will update the sizing variables to be consistent.
> As far as the color variable, I tried to follow the codes on figma @calcaide are you suggesting we pick whichever is closer to that color code in our predefined color guide?

@DhariniJeeva generally yes we should apply a "snap to design system" approach.  Unless a value is way off, defer to the system.

addons/rose/addon/styles/rose/components/form/_radio-card.scss
DhariniJeeva: I don't think this works.  
`rem*rem isn't a valid CSS value` this is what I get when I try to multiply rems based on your suggestion.

addons/rose/addon/styles/rose/components/form/_radio-card.scss
calcaide: I see the aliased advantatge of cutting verbose. To avoid re-defining these in each file we need it, what do you guys think on defining these variables  within `_sizing.scss` file? so they can be use as `sizing.l`, `sizing.xss` or something like that?

addons/rose/addon/styles/rose/components/form/_radio-card.scss
DhariniJeeva: @randallmorey  yes, that is right

calcaide:
> I will update the sizing variables to be consistent. As far as the color variable, I tried to follow the codes on figma @calcaide are you suggesting we pick whichever is closer to that color code in our predefined color guide?

Yeah, if is very close I will pick what we have. If is completly off I will suggest letting Joshua know about it. In any case we can just ask in the UI channel what is the preference.

addons/rose/addon/styles/rose/components/form/_radio-card.scss
calcaide: You are right @DhariniJeeva , I got wrong a previously implementation I did operating with sizing.

As per Randy comment, then ignore mine. All good

addons/rose/addon/styles/rose/components/form/_radio-card.scss
calcaide: Deleting this and other comments where this is ok to perform.

addons/rose/addon/styles/rose/components/form/_radio-card.scss
DhariniJeeva: happy to add a ticket for this work and address it later. we use aliased format In another place as well.

addons/rose/addon/styles/rose/components/form/_radio-card.scss
```diff
@@ -0,0 +1,199 @@
...
+    .rose-icon {
+      --size: #{$size-m};
+    }
+  }
+
+  &.tile {
+    width: ($size-m) * 15;
+   
+    .rose-icon {
+      --size: #{$size-m + $size-xxs + $size-xxxs};
```
calcaide, line 47: Perhaps we can simplify it to: `size-l` (24) - `size-xxxs` (2) = 22?

addons/rose/addon/styles/rose/components/form/_radio-card.scss
```diff
@@ -0,0 +1,199 @@
...
+$size-l: sizing.rems(l); // 24
+$size-m: sizing.rems(m); // 16
+$size-s: sizing.rems(s); //12
+$size-xs: sizing.rems(xs); // 8
+$size-xxs: sizing.rems(xxs); // 4
+$size-xxxs: sizing.rems(xxxs); // 2
+$size-xxxxs: sizing.rems(xxxxs); // 1
+
+
+$radio-svg: (
```
calcaide, line 16: Nice one!! :grinning:

addons/rose/addon/styles/rose/components/form/_radio-card.scss
```diff
@@ -0,0 +1,199 @@
+@use '../../variables/sizing';
+@use '../../utilities/type';
+@use '../../variables/typography';
+
+$size-xxl: sizing.rems(xxl); //48
+$size-xl: sizing.rems(xl); //48
```
calcaide, line 6: I think xl = 36

addons/rose/addon/styles/rose/components/form/_radio-card.scss
```diff
@@ -0,0 +1,199 @@
...
+    position: absolute;
+    height: $size-l + $size-xxs;
+    background-color: var(--ui-gray-subtler-5);
+    width: 100%;
+    bottom: 0;
+    right: 0;
+  }
+  .rose-form-radio-label {
+    font-style: normal;
+    font-weight: bold;
```
calcaide, line 74: I am not sure about this, perhaps @randallmorey can bring more knowledge.

But I am sure I saw in the past in our codebase use `type` to define font size, style and weight at once. I.E, [this code](https://github.com/hashicorp/boundary-ui/blob/main/addons/rose/addon/styles/rose/components/card/_link.scss#L86) `@include type.type(m, semibold);`

DhariniJeeva: updated

addons/rose/addon/styles/rose/components/form/_radio-card.scss
DhariniJeeva: it should be 30 actually. fixed!

addons/rose/addon/styles/rose/components/form/_radio-card.scss
randallmorey: Just thinking out loud, might be worth a try:

```
(sizing.rems(m) * 11) - ($size-xxxs * 2);  // or whatever the correct token value is
```

I'd suggest adding comments to make the intent clear.

randallmorey: Hmm you might even define a variable within the selector to define width for the card:  `$cardWidth: sizing.rems(m) * 11` and then a border width:  `$borderWidth: $size-xxxs`.  Then the thing becomes very readable:

```
$cardWidth - ($borderWidth * 2);
```

DhariniJeeva: thank you! updated!

randallmorey (changes requested):
Some minor things and then I think we can ship it:

- Is the radio circle icon centered vertically within the card footer?
- Are the colors for the error state fully applied?
- The logo/icon needs to be slightly larger and have more whitespace like the design (in the design, the logo bounding box is 40x40).
- The card should have slightly rounded corners.

DhariniJeeva:
> Some minor things and then I think we can ship it:
> 
> * Is the radio circle icon centered vertically within the card footer?
> * Are the colors for the error state fully applied?
> * The logo/icon needs to be slightly larger and have more whitespace like the design (in the design, the logo bounding box is 40x40).
> * The card should have slightly rounded corners.

addressed!

randallmorey (approved):

calcaide (approved):

PR #910, opened 2022-01-18
Support nested attributes with no dependencies

randallmorey:
Boundary UI has generally relied on ember model fragments to support nested fields (e.g. under `attributes`).  However this approach is complex (requiring additional models and serialisers) and couples Boundary to another third-party dependency.  This PR introduces first-class support for nested attributes by hoisting them up into the payload body at normalisation time.  It then serialises them back out to the nested attributes field.  To start, the **Target** resource is refactored to use the new approach.

For example in a model:

```js
@attr('string', { isNestedAttribute: true }) fieldName;
```


calcaide (approved):
LGTM! Much appreciate it this work and showing the pattern to follow.

PR #913, opened 2022-01-20
feat: 🎸 Add fields to host-catalog model

calcaide:
Add new fields to host-catalog model for plugin type host-catalog.
Fields are corresponding to Azure and Aws plugins.


addons/api/tests/unit/serializers/host-catalog-test.js
```diff
@@ -0,0 +1,108 @@
...
+    const serializedRecord = serializer.serialize(snapshot);
+    const expectedResult = {
+      name: 'static',
+      description: 'this is a static host-catalog',
+      type: 'static',
+    };
+    assert.deepEqual(serializedRecord, expectedResult);
+  });
+
+  test('it serializes an AWS plugin as expected, ignoring read-only and azure fields', async function (assert) {
```
randallmorey, line 47: We don't need to explicitly exclude or test for the exclusion of type-specific fields.  We only need to test the _inclusion_ of expected fields.  In practice, a model instance should never come to us with fields set for another type.  And even if it did, we want to be somewhat flexible to support unknown types in the future.

I'd modify this test case to test only AWS fields.

randallmorey (changes requested):

addons/api/addon/serializers/host-catalog.js
```diff
@@ -0,0 +1,41 @@
+import ApplicationSerializer from './application';
+
+export default class HostCatalogSerializer extends ApplicationSerializer {
```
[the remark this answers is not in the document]
calcaide: I think we do need it.

I tested following next steps:
1. Serializer returns `serialized` when the record is `static` type (rather than calling `serializeStatic`).
2. Static unit test creates a record as follows, and expect this as result. 
```javascript
{
  name: 'static',
  description: 'this is a static host-catalog',
  type: 'static',
}
```
3. Test fails because it received:
```javascript
{
  attributes: {
    client_id: null,
    disable_credential_rotation: false,
    region: null,
    subscription_id: null,
    tenant_id: null
  },
  description: "this is a static host-catalog",
  name: "static",
  type: "static"
}
```
4. As per my understanding, we should receive `attributes` but we are, that's why I add the `delete.attributes` within the `serializeStatic`

If you think I am doing something wrong or missing something, happy to jump on a zoom call.

addons/api/tests/unit/serializers/host-catalog-test.js
```diff
@@ -0,0 +1,108 @@
...
+
+  test('it serializes a static host-catalog as expected', async function (assert) {
+    assert.expect(1);
+    const store = this.owner.lookup('service:store');
+    const serializer = store.serializerFor('host-catalog');
+    const record = store.createRecord('host-catalog', {
+      name: 'static',
+      description: 'this is a static host-catalog',
+      type: 'static',
+      attributes: {
```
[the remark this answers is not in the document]
calcaide: You are right, this should be at top-level in the model. Thanks for the clarification and sorry for the mess about this specific topic 😅

addons/api/tests/unit/serializers/host-catalog-test.js
calcaide: Gotcha, you are right.
Here I made next (wrong) assumption we want to exclude those fields does not belong to the model or are unexpected as `fake_field`.

addons/api/addon/serializers/host-catalog.js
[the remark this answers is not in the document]
calcaide: Same behaviour with `serializeDynamic`. If is an aws host-catalog brings (set to null) fields from azure, that's why I delete them.

addons/api/addon/generated/models/host-catalog.js
```diff
@@ -44,6 +44,11 @@ export default class GeneratedHostCatalogModel extends BaseModel {
   })
   attributes;
 
+  @attr({
+    description: 'Secrets specific to the plugin type',
+  })
+  secrets;
```
randallmorey, line 50: This is like `attributes` in that it's handled transparently by the serializer.  We don't need to declare it in the model.

addons/api/addon/generated/models/host-catalog.js
```diff
@@ -54,4 +59,68 @@ export default class GeneratedHostCatalogModel extends BaseModel {
...
+
+  @attr('string', {
+    isNestedAttribute: true,
+    description: '',
+  })
+  region;
+
+  @attr('string', {
+    isNestedSecret: true,
+    writeOnly: true,
```
randallmorey, line 81: `writeOnly` goes away.  It is no longer supported or necessary.

addons/api/mirage/generated/factories/host-catalog.js
```diff
@@ -1,6 +1,22 @@
...
+  }
+}
+
+function isAwsPlugin(plugin) {
+  if (plugin && plugin.name === 'aws') {
+    return true;
+  }
+}
 /**
  * GeneratedHostCatalogModelFactory
```
randallmorey, line 21: This logic is far too bespoke for the generated factory and should live in the regular factory.

Mirage mocks must return `attributes`, since it mimics the API (not the Ember models).

Refer to the account factory as a reference for type-specific mocking.  It uses a switch statement that is fairly readable.

randallmorey (changes requested):

addons/api/mirage/factories/host-catalog.js
```diff
@@ -16,6 +17,26 @@ export default factory.extend({
       'host-sets': ['create', 'list'],
     };
   },
+  attributes() {
+    if (this.plugin) {
+      switch (this.plugin.name) {
```
randallmorey, line 22: Would this work as a way to eliminate the outer conditional?
```suggestion
      switch (this.plugin?.name) {
```

addons/api/mirage/generated/factories/host-catalog.js
```diff
@@ -20,4 +20,5 @@ export default Factory.extend({
       };
     }
   },
+  disable_credential_rotation: () => datatype.boolean(),
```
randallmorey, line 23: This is an `attributes` field.

addons/api/tests/unit/serializers/host-catalog-test.js
```diff
@@ -0,0 +1,187 @@
...
+
+  test('it serializes an Azure plugin as expected, ignoring AWS fields', async function (assert) {
+    assert.expect(1);
+    const store = this.owner.lookup('service:store');
+    const record = store.createRecord('host-catalog', {
+      compositeType: 'azure',
+      name: 'Azure',
+      description: 'this is a Azure plugin host-catalog',
+      disable_credential_rotation: true,
+      // these are AWS fields and should be included
```
randallmorey, line 81: This comment is wrong.  Copy/paste error

addons/api/tests/unit/serializers/host-catalog-test.js
```diff
@@ -0,0 +1,187 @@
...
+        type: 'host-catalog',
+        attributes: {
+          type: 'plugin',
+          name: 'aws',
+          description: 'test description',
+          plugin: {
+            name: 'aws',
+          },
+          region: 'andorra',
+          access_key_id: 'foobars',
```
randallmorey, line 125: With `store.push` we're simulating a "saved" record.  Saved records would not have values for secrets fields.  Your test case is _technically_ correct but this condition is not expected to occur in real life.

addons/api/tests/unit/serializers/host-catalog-test.js
```diff
@@ -0,0 +1,187 @@
...
+      },
+      secrets: {
+        access_key_id: 'foobars',
+        secret_access_key: 'testing',
+      },
+    };
+    assert.deepEqual(record.serialize(), expectedResult);
+  });
+
+  test('it serializes azure correctly on update', function (assert) {
```
randallmorey, line 147:
```suggestion
  test('it serializes an existing azure record correctly', function (assert) {
```

addons/api/tests/unit/serializers/host-catalog-test.js
```diff
@@ -0,0 +1,187 @@
...
+      },
+      secrets: {
+        secret_id: 'a1b2c3',
+        secret_value: 'a1b2c3',
+      },
+    };
+    assert.deepEqual(record.serialize(), expectedResult);
+  });
+
+  test('it serializes aws correctly on update', function (assert) {
```
randallmorey, line 110:
```suggestion
  test('it serializes an existing aws record correctly', function (assert) {
```

addons/api/tests/unit/serializers/host-catalog-test.js
```diff
@@ -0,0 +1,187 @@
...
+      },
+      secrets: {
+        access_key_id: 'foobars',
+        secret_access_key: 'testing',
+      },
+    };
+    assert.deepEqual(record.serialize(), expectedResult);
+  });
+
+  test('it serializes an Azure plugin as expected, ignoring AWS fields', async function (assert) {
```
randallmorey, line 73:
```suggestion
  test('it serializes a new azure record as expected, ignoring aws fields', async function (assert) {
```

addons/api/tests/unit/serializers/host-catalog-test.js
```diff
@@ -0,0 +1,187 @@
...
+    });
+    const expectedResult = {
+      name: 'static',
+      description: 'this is a static host-catalog',
+      type: 'static',
+    };
+    assert.deepEqual(record.serialize(), expectedResult);
+  });
+
+  test('it serializes an AWS plugin as expected, ignoring Azure fields', async function (assert) {
```
randallmorey, line 38:
```suggestion
  test('it serializes a new aws plugin as expected, ignoring azure fields', async function (assert) {
```

randallmorey (changes requested):

randallmorey (approved):

PR #917, opened 2022-01-25
Remove renderTemplate usage from groups resource

randallmorey:
This PR removes usage of `renderTemplate` for the groups resource [because it is deprecated](https://deprecations.emberjs.com/v3.x/#toc_route-render-template).

Changes at a glance:
- Partial templates previously used by `renderTemplate` are converted to components
- Components are embedded into each template where they are used


calcaide (approved):
LGTM!

DhariniJeeva (approved):

PR #925, opened 2022-01-26
feat: 🎸 Add basic type support for host-set

calcaide:
Add basic type support for host-set.

### DIFF
--- a/addons/api/mirage/factories/host-set.js
+++ b/addons/api/mirage/factories/host-set.js
@@ -12,4 +12,9 @@ export default factory.extend({
       'remove-hosts',
     ],
   id: (i) => `host-set-id-${i}`,
+  plugin() {
+    if (this.type === 'plugin') {
+      return this.hostCatalog.plugin;
+    }
+  },
 });
--- a/addons/api/mirage/factories/host-set.js
+++ b/addons/api/mirage/factories/host-set.js
@@ -12,6 +12,11 @@ export default factory.extend({
       'remove-hosts',
     ],
   id: (i) => `host-set-id-${i}`,
+  type() {
+    if (this.hostCatalog) {
+      return this.hostCatalog.type;
+    }
+  },
   plugin() {
     if (this.type === 'plugin') {
       return this.hostCatalog.plugin;
--- a/addons/api/mirage/factories/host-set.js
+++ b/addons/api/mirage/factories/host-set.js
@@ -13,9 +13,7 @@ export default factory.extend({
     ],
   id: (i) => `host-set-id-${i}`,
   type() {
-    if (this.hostCatalog) {
-      return this.hostCatalog.type;
-    }
+    return this.hostCatalog?.type || factory.attrs.type;
   },
   plugin() {
     if (this.type === 'plugin') {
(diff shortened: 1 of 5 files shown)


addons/api/mirage/generated/factories/host-set.js
```diff
@@ -5,11 +5,20 @@ import { random, date, datatype } from 'faker';
  * GeneratedHostSetModelFactory
  */
 export default Factory.extend({
-  type: 'static',
+  type: () => random.arrayElement(['static', 'plugin']),
```
calcaide, line 8: @randallmorey I think here we want the same `type` and `plugin.name` than the host-catalog it pertains?

randallmorey: Correct.

calcaide: Gotcha, I updated the PR following recommendations.

addons/api/mirage/generated/factories/host-set.js
```diff
@@ -5,7 +5,7 @@ import { random, date, datatype } from 'faker';
  * GeneratedHostSetModelFactory
  */
 export default Factory.extend({
-  type: 'static',
+  type: () => this.hostCatalog.type,
```
randallmorey, line 8: Right logic.  However the inheritance is "undiscoverable" at the schema level, so this can't be generated.  I think the generated factories ought to express a "ground state" default type.  Inheritance can be expressed in the factory subclasses.

randallmorey (changes requested):

addons/api/mirage/generated/factories/host-set.js
calcaide: Exactly!! I just fixed what a good timing, eh? I just ask you about this in another PR and after it, I realize I have that exact problem. SOLVED!

addons/api/mirage/factories/host-set.js
```diff
@@ -12,4 +12,14 @@ export default factory.extend({
       'remove-hosts',
     ],
   id: (i) => `host-set-id-${i}`,
+  type() {
+    if (this.hostCatalog) {
```
randallmorey, line 16: What about `return this.hostCatalog?.type || super.type;`?  I'd just double check that `super` actually works here.  Not sure.

randallmorey: The super syntax might be something else.  This requires a little research.  Because mirage factories aren't native classes, they may have their own super system.  Hopefully it is super duper.

randallmorey (changes requested):

addons/api/mirage/factories/host-set.js
calcaide: I can reasearch the `super.type` no problem. But your comment got me thinking:
- The factory is extended from `generated/factory` which already has a base type.
- Assuming we have `factory.type` with a value set previously, how about `return this.hostCatalog?.type || factory.type;`

randallmorey: Is `factory.type` accessible?  It might be on a different property, not sure.

calcaide: jsut checked, it is accessible: `factory.attrs.type`

randallmorey (approved):

PR #926, opened 2022-01-26
[draft] Create host catalog form should support static catalog type

gsusmi:
This PR adds 'static' type support to host catalog creation form by splitting  `form/host-catalog/index.hbs` to `form/host-catalog/static/index.hbs`, adding `?type` query param support, and new `changeType` action to update route.

TODOs:
Should `type` be set to `static` host catalog by default?


DhariniJeeva:
I think this looks good and covers what's needed on the host-catalog router side! ty!

gsusmi:
Addressed in: https://github.com/hashicorp/boundary-ui/pull/940

PR #927, opened 2022-01-26
feat: 🎸 Add type badges for host-set

calcaide:
Add type badges for host-set.

When `type` === `static`:
Screen Shot 2022-01-26 at 8 16 33 AM

When `type` is plugin (`aws` or `azure`):
Screen Shot 2022-01-26 at 8 19 15 AM


randallmorey (changes requested):
Let's hold off on this because we will have a concept of "unknown" which will need another clause.

calcaide:
> ill

Ok. I will leave this PR as it is until we get the "unknown" sorted out, then I will provide a commit to support it.

randallmorey (approved):
Looks good.  After reconsideration, `unknown` typing is unnecessary for hosts and host sets, since we will disallow navigation into unknown host catalogs.

PR #928, opened 2022-01-26
feat: 🎸 Add host type basic support

calcaide:
Add host type basic support.

### DIFF
--- a/addons/api/mirage/factories/host.js
+++ b/addons/api/mirage/factories/host.js
@@ -10,4 +10,9 @@ export default factory.extend({
       'delete',
     ],
   id: (i) => `host-id-${i}`,
+  plugin() {
+    if (this.type === 'plugin') {
+      return this.hostCatalog.plugin;
+    }
+  },
 });
--- a/addons/api/mirage/factories/host.js
+++ b/addons/api/mirage/factories/host.js
@@ -10,6 +10,9 @@ export default factory.extend({
       'delete',
     ],
   id: (i) => `host-id-${i}`,
+  type() {
+    return this.hostCatalog?.type || factory.attrs.type;
+  },
   plugin() {
     if (this.type === 'plugin') {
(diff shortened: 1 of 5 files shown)


addons/api/mirage/factories/host.js
```diff
@@ -10,4 +10,9 @@ export default factory.extend({
       'delete',
     ],
   id: (i) => `host-id-${i}`,
+  plugin() {
+    if (this.type === 'plugin') {
+      return this.hostCatalog.plugin;
```
DhariniJeeva, line 15: just curious, why are we returning `hostCatalog.plugin` instead of `host.plugin`

calcaide: Because we want the object `plugin` inherit from hostCatalog.

randallmorey: @DhariniJeeva the type is inherited from the host catalog.  However this raises a problem with the implementation.  We need to apply this inheritance logic to `type` as well @calcaide.

calcaide: @randallmorey 
On it's way! I did it in `host-set` [here](https://github.com/hashicorp/boundary-ui/pull/925/files#diff-c4d10025ef7bd2a9d9275ee8670da51e3d034c096ac9fe931dc4233ea88544b0R8) I will reproduce that pattern here.

DhariniJeeva:
Looks good to me! one minor query- does the host set need support for `unknown` type as well?

randallmorey:
> Looks good to me! one minor query- does the host set need support for `unknown` type as well?

I'm beginning to think we do.  Taking another look.

randallmorey:
> Looks good to me! one minor query- does the host set need support for `unknown` type as well?

For now, no.

randallmorey (approved):

PR #929, opened 2022-01-26
Add support for "unknown" dynamic host catalogs

randallmorey:
This PR introduces support for "unknown" typed dynamic host catalogs.  This could occur if a third-party plugin is installed that is unsupported by the UI.  While we can display such a catalog in an index, we cannot link to a details route or allow editing.  As such, unknown catalogs are always "list only" in the UI.

Screenshot 2022-01-26 at 13 20 03

### DIFF
--- a/addons/api/addon/models/host-catalog.js
+++ b/addons/api/addon/models/host-catalog.js
@@ -1,8 +1,12 @@
 import GeneratedHostCatalogModel from '../generated/models/host-catalog';
 
+export const types = ['aws', 'azure'];
+
 export default class HostCatalogModel extends GeneratedHostCatalogModel {
+  // =attributes
+
   /**
-   * Returns if the host-catalog is static type or not.
+   * True if the host catalog is static.
    * @type {boolean}
    */
   get isStatic() {
@@ -10,7 +14,7 @@ export default class HostCatalogModel extends GeneratedHostCatalogModel {
   }
 
   /**
-   * Returns if the host-catalog is a plugin or not.
+   * True if the host catalog is a plugin.
    * @type {boolean}
    */
   get isPlugin() {
@@ -18,7 +22,15 @@ export default class HostCatalogModel extends GeneratedHostCatalogModel {
   }
 
   /**
-   * Returns if host-catalog plugin is AWS or not
+   * True if the host catalog is an unknown type.
+   * @type {boolean}
+   */
+  get isUnknown() {
+    return this.isPlugin && !types.includes(this.plugin.name);
+  }
+
+  /**
+   * True if host catalog plugin type is AWS.
    * @type {boolean}
    */
   get isAWS() {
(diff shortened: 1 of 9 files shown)


addons/api/addon/models/host-catalog.js
```diff
@@ -1,48 +1,64 @@
 import GeneratedHostCatalogModel from '../generated/models/host-catalog';
 
+export const types = ['aws', 'azure'];
+
 export default class HostCatalogModel extends GeneratedHostCatalogModel {
+  // =attributes
+
   /**
-   * Returns if the host-catalog is static type or not.
+   * True if the host catalog is static.
```
calcaide, line 9: I will apply this comments change also in host-set and host.

addons/api/mirage/generated/factories/host-catalog.js
```diff
@@ -5,19 +5,10 @@ import { random, date, datatype } from 'faker';
  * GeneratedHostCatalogModelFactory
  */
 export default Factory.extend({
-  type: () => random.arrayElement(['static', 'plugin']),
+  type: 'static',
```
calcaide, line 8: This `type` will be override by `mirage/factories/host-catalog` right? if so, perhaps we should not declared here and let the factory do it? I would like to know your logic on this since I have same casses in other tasks :smiley:

addons/api/mirage/scenarios/default.js
```diff
@@ -46,7 +46,7 @@ export default function (server) {
 
   // Other resources
   server.schema.scopes.where({ type: 'project' }).models.forEach((scope) => {
-    server.createList('host-catalog', 4, { scope }, 'withChildren');
+    server.createList('host-catalog', 8, { scope }, 'withChildren');
```
calcaide, line 49: :muscle:

ui/admin/app/templates/scopes/scope/host-catalogs/index.hbs
```diff
@@ -59,7 +59,19 @@
...
+                      (concat
+                        'resources.host-catalog.types.'
+                        hostCatalog.compositeType
+                      )
+                    }}
+                  </Rose::Badge>
+                {{else if hostCatalog.isUnknown}}
+                  <Rose::Badge>
+                    {{t
+                      (concat
```
calcaide, line 71: I will copy this approach for host-set and host once this get merged (I need the changes you add on `resources/en-us.yaml`

calcaide (approved):
LGTM! Awesome work

addons/api/mirage/generated/factories/host-catalog.js
randallmorey: We should have some kind of "ground state" default type for typed resources.  The most sensible default for host catalog is `static`.  It's unclear though if this could feasibly be generated.  We'll have to address the problem of _type discovery_ down the road.

calcaide: Ahh ok that makes sense. I will apply this approach on the PR's I have, thanks for the insights.

DhariniJeeva (approved):

PR #937, opened 2022-01-31
Unify model and collection abilities

randallmorey:
Until now, model and collection abilities have been treated as separate, even though they are both defined on model instances.  This PR combines the ember-can abilities for `model` and `collection`, which simplifies upcoming changes to dynamic host catalogs abilities.

An additional small change was made to remove the feature flag for capabilities, since it is now always enabled.

There are no user-facing changes and capabilities should continue to work as usual:

- List/create-only permissions scenario:  `/scopes/s_xk1i4dk50v2/host-catalogs?authorized_actions={"host-catalog":[]}`
- Read-only scenario:  `/scopes/s_xk1i4dk50v2/host-catalogs/host-catalog-id-0?authorized_actions={"host-catalog":["read"]}`
- Full permissions scenario:  `/scopes/s_xk1i4dk50v2/host-catalogs/host-catalog-id-0?authorized_actions={"host-catalog":["read","update","delete"]}`


gsusmi (approved):
LGTM. Ty!

calcaide (approved):

PR #938, opened 2022-02-01
feat: 🎸 Add new fields to host-set model

calcaide:
Add new fields to host-set model.

### DIFF
--- a/addons/api/addon/serializers/host-set.js
+++ b/addons/api/addon/serializers/host-set.js
@@ -27,7 +27,6 @@ export default class HostSetSerializer extends ApplicationSerializer {
     if (hostIDs) {
       return this.serializeWithHostIDs(snapshot, hostIDs);
     } else {
-      console.log('Carlos: ', snapshot.record.compositeType);
       switch (snapshot.record.compositeType) {
         case 'static':
           return this.serializeStatic(...arguments);
--- a/addons/api/addon/serializers/host-set.js
+++ b/addons/api/addon/serializers/host-set.js
@@ -1,8 +1,8 @@
 import ApplicationSerializer from './application';
 
 const fieldByType = {
-  aws: ['filters'],
-  azure: ['filter'],
+  aws: ['preferred_endpoints', 'filters'],
+  azure: ['preferred_endpoints', 'filter'],
 };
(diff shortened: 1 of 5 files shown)


addons/api/addon/serializers/host-set.js
```diff
@@ -19,10 +23,17 @@ export default class HostSetSerializer extends ApplicationSerializer {
...
    */
   serialize(snapshot) {
-    let serialized = super.serialize(...arguments);
     const hostIDs = snapshot?.adapterOptions?.hostIDs;
-    if (hostIDs) serialized = this.serializeWithHostIDs(snapshot, hostIDs);
-    return serialized;
+    if (hostIDs) {
+      return this.serializeWithHostIDs(snapshot, hostIDs);
+    } else {
+      switch (snapshot.record.compositeType) {
```
calcaide, line 30: A switch for just two cases might be killer, but I assume in the future we will have more cases. Happy to refactor for a simpler `if else` if you think we will not have more cases in the short-term.

addons/api/addon/serializers/host-set.js
```diff
@@ -37,4 +48,24 @@ export default class HostSetSerializer extends ApplicationSerializer {
       host_ids: hostIDs,
     };
   }
+
+  serializeStatic() {
```
randallmorey, line 52: I'd double check, but `static` probably shouldn't have `preferred_endpoints` either.

addons/api/mirage/factories/host-set.js
```diff
@@ -18,4 +19,31 @@ export default factory.extend({
...
+      let result = [];
+      for (let i = 0; i < filtersAmount; ++i) {
+        result.push(`${datatype.string(3)}=${datatype.string(8)}`);
+      }
+      return result;
+    }
+  },
+  // Azure specific
+  filter() {
+    if (this.type === 'plugin' && this.plugin?.name === 'azure') {
```
randallmorey, line 45: Would a single clause `this.plugin?.name === 'azure'` be sufficient?

randallmorey (changes requested):

randallmorey (approved):

PR #939, opened 2022-02-02
Disable non-static hosts navigation

randallmorey:
This PR disallows navigation from dynamic host catalogs directly to hosts, since there may be too many hosts (the existence of which is also dynamic) to be of practical value.  This PR also corrects the grant mocks for dynamic host catalogs to be more realistic.

Static host catalog navigation:
Screenshot 2022-02-02 at 17 38 14

Dynamic host catalog navigation:
Screenshot 2022-02-02 at 17 38 28

### DIFF
--- a/addons/api/mirage/factories/host-catalog.js
+++ b/addons/api/mirage/factories/host-catalog.js
@@ -11,6 +11,11 @@ const pluginTypes = ['aws', 'azure', 'foobar'];
 let pluginTypeCounter = 1;
 
 export default factory.extend({
+  id: (i) => `host-catalog-id-${i}`,
+
+  // Cycle through available types
+  type: (i) => types[i % types.length],
+
   authorized_actions: () =>
     permissions.authorizedActionsFor('host-catalog') || [
       'no-op',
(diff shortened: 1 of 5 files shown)


addons/api/mirage/factories/host-catalog.js
```diff
@@ -11,25 +11,27 @@ const pluginTypes = ['aws', 'azure', 'foobar'];
...
   authorized_actions: () =>
     permissions.authorizedActionsFor('host-catalog') || [
       'no-op',
       'read',
       'update',
       'delete',
     ],
-  authorized_collection_actions: () => {
+  authorized_collection_actions: function () {
+    const isStatic = this.type === 'static';
```
cameronperera, line 27: I noticed that this function changed from an anonymous function to a named one. Is this simply so it gains access to `this`? That is the only thing different that I can see as to why it would need to change.

calcaide (approved):
LGTM

DhariniJeeva (approved):

addons/api/mirage/factories/host-catalog.js
randallmorey: Yes exactly, because this field is dependent on `type`, whereas previously it was not.

gsusmi (approved):

PR #941, opened 2022-02-07
feat: 🎸 Add new fields to host

calcaide:
Add new fields to host and refactor field address to use no more `fragment-host-attribute-model`.

**Be aware:** These changes does NOT include these new fields in mirage, that will come in a separate PR.

### DIFF
--- a/addons/api/addon/generated/models/host.js
+++ b/addons/api/addon/generated/models/host.js
@@ -45,8 +45,26 @@ export default class GeneratedHostModel extends BaseModel {
   version;
 
   @attr({
+    description: 'Attributes specific to the host type',
+  })
+  attributes;
+
+  @attr('string', {
+    description: 'The address (DNS or IP name) used to reach the host',
+    isNestedAttribute: true,
+  })
+  address;
+
+  @attr({
     description: 'Plugin information for this resource.',
     readOnly: true,
   })
   plugin;
+
+  @attr('string', {
+    description: '',
+    readOnly: true,
+    isNestedAttribute: true,
+  })
+  external_id;
 }
--- a/addons/api/addon/generated/models/host.js
+++ b/addons/api/addon/generated/models/host.js
@@ -45,26 +45,8 @@ export default class GeneratedHostModel extends BaseModel {
   version;
 
   @attr({
-    description: 'Attributes specific to the host type',
-  })
-  attributes;
-
-  @attr('string', {
-    description: 'The address (DNS or IP name) used to reach the host',
-    isNestedAttribute: true,
-  })
-  address;
-
-  @attr({
     description: 'Plugin information for this resource.',
     readOnly: true,
   })
   plugin;
-
-  @attr('string', {
-    description: '',
-    readOnly: true,
-    isNestedAttribute: true,
-  })
-  external_id;
 }
--- a/addons/api/addon/models/host.js
+++ b/addons/api/addon/models/host.js
@@ -1,13 +1,28 @@
 import GeneratedHostModel from '../generated/models/host';
+import { attr } from '@ember-data/model';
 import { fragmentArray } from 'ember-data-model-fragments/attributes';
 
 export default class HostModel extends GeneratedHostModel {
+  // =attributes (nested under attributes)
   @fragmentArray('fragment-string', {
     readOnly: true,
     emptyArrayIfMissing: true,
     isNestedAttribute: true,
   })
   ip_addresses;
+
+  @attr('string', {
+    description: 'The address (DNS or IP name) used to reach the host',
+    isNestedAttribute: true,
+  })
+  address;
+
+  @attr('string', {
+    description: '',
+    readOnly: true,
+    isNestedAttribute: true,
+  })
+  external_id;
 
   // Aws specific
   @fragmentArray('fragment-string', {
--- a/addons/api/addon/models/host.js
+++ b/addons/api/addon/models/host.js
@@ -1,34 +1,33 @@
 import GeneratedHostModel from '../generated/models/host';
 import { attr } from '@ember-data/model';
-import { fragmentArray } from 'ember-data-model-fragments/attributes';
 
 export default class HostModel extends GeneratedHostModel {
-  // =attributes (nested under attributes)
-  @fragmentArray('fragment-string', {
-    readOnly: true,
-    emptyArrayIfMissing: true,
-    isNestedAttribute: true,
-  })
-  ip_addresses;
+  // =attributes
 
+  // Static specific fields
   @attr('string', {
     description: 'The address (DNS or IP name) used to reach the host',
     isNestedAttribute: true,
   })
   address;
 
+  // Plugin specific fields
+  @attr({
+    readOnly: true,
+    emptyArrayIfMissing: true,
+  })
+  ip_addresses;
+
   @attr('string', {
     description: '',
     readOnly: true,
-    isNestedAttribute: true,
   })
   external_id;
 
   // Aws specific
-  @fragmentArray('fragment-string', {
+  @attr({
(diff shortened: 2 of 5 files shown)


addons/api/addon/generated/models/host.js
```diff
@@ -44,9 +44,27 @@ export default class GeneratedHostModel extends BaseModel {
   })
   version;
 
+  @attr({
```
randallmorey, line 47: It is unnecessary to formally model `attributes`.

addons/api/addon/generated/models/host.js
```diff
@@ -44,9 +44,27 @@ export default class GeneratedHostModel extends BaseModel {
...
+    description: 'Attributes specific to the host type',
+  })
+  attributes;
+
+  @attr('string', {
```
randallmorey, line 52: I am on the fence about whether attributes fields should be included in the generated model.  However, they should probably be grouped together with other attributes fields, for readability.

addons/api/tests/unit/serializers/host-test.js
```diff
@@ -4,19 +4,48 @@ import { setupTest } from 'ember-qunit';
...
+      },
+    });
   });
 
-  test('it serializes records', function (assert) {
-    let store = this.owner.lookup('service:store');
-    let record = store.createRecord('host', {});
-
-    let serializedRecord = record.serialize();
+  test('it serializes host with type plugin as expected', function (assert) {
```
randallmorey, line 30: We don't need to test serialisation of dynamic hosts because they are read-only.  Only normalisation is important for these.

randallmorey (changes requested):

addons/api/tests/unit/serializers/host-test.js
```diff
@@ -4,19 +4,48 @@ import { setupTest } from 'ember-qunit';
...
+      compositeType: 'static',
+      description: 'Description',
+      host_catalog_id: '123',
+      version: 1,
+      address: '188e:68a9:b342:c05c:2595:2f46:499c:759f',
+    });
+    assert.deepEqual(record.serialize(), {
+      name: 'Host 1',
+      description: 'Description',
+      type: 'static',
```
DhariniJeeva, line 21: could you please explain what is happening here with the compostiteType/type change here?

addons/api/addon/generated/models/host.js
calcaide: Understood. I copied this approach from `host-catalog`. Will provide changes to delete it.

addons/api/addon/generated/models/host.js
calcaide: Understood. As per last time we spoke what's the difference between a simple `model` and a `generated/model` I understood `generated/model` are fields with primitive values (as `address` is).

But I see what you mean my grouping `attributes` together.

That said, IMHO we are going to start messing ember (perhaps internal) good practices with context of the app.

Will perform these changes tho.

addons/api/tests/unit/serializers/host-test.js
calcaide: I will add the `normalisation` tests.

addons/api/addon/generated/models/host.js
randallmorey: There are a few heuristics.  Model generation is aspirational.  Ultimately it's a question of _what can be actually be generated_ when we're ready to do it.  Right now, `attributes` can't be, because they are not present in the schema JSON.  That means `attributes` fields should probably not go in the generated model.  Let's table that for after this work though.

addons/api/tests/unit/serializers/host-test.js
calcaide: Absolutley. I am not 100% sure what you mean by I think I got it.

- This is a serialization test, understanding serialization as a translation from `ember model` (UI client) to `API`.
- When I create the record (so when I use `compositeType`) I am createing a record on `ember model` (UI client). We have `compositeType` available here.
- When I am asserting, I am taking the previously created `ember model`, serialize it, and then comparing to the expectation.
- As explained in the first step, I would expect `type` here, because the object I am expecting is the one the API will receive. The API does not know what `compositeType` is, just the UI.

Happy to jump on zoom call if you need more clarification about this :)

I use a rule of thumb that helps: (thanks @randallmorey 
- Serialization: Ember data -> API.
- Normalization: API -> Ember data.

randallmorey: Yup.  The serialised output is what we send to the API.  The API expects `type`.

DhariniJeeva: got it! that makes sense. ty for the explanation!

calcaide: Np @DhariniJeeva took me a while until Randy told me the rule of thumb, my mind was always a mess with creation vs expectation :sweat_smile:

DhariniJeeva (approved):

randallmorey (changes requested):
In the normalisation test(s), I'd add a complete set of expect fields to the input payload to ensure they make it into the normalised data.

calcaide:
> In the normalisation test(s), I'd add a complete set of expect fields to the input payload to ensure they make it into the normalised data.

@randallmorey 
Thank you for the feedback, I updated the PR with your suggestions.

randallmorey (changes requested):
Host fields aren't nested under attributes.  Also, we don't need to model read-only array fields using fragments.  The utility of fragments is primarily change tracking, which doesn't apply here.

addons/api/addon/models/host.js
```diff
@@ -1,12 +1,36 @@
...
+import { fragmentArray } from 'ember-data-model-fragments/attributes';
 
 export default class HostModel extends GeneratedHostModel {
-  /**
-   * Attributes of this resource, if any, represented as a JSON fragment.
-   * @type {FragmentHostAttributesModel}
-   */
-  @fragment('fragment-host-attributes', { defaultValue: {} }) attributes;
+  // =attributes (nested under attributes)
+  @fragmentArray('fragment-string', {
```
randallmorey, line 7: This is for dynamic.

addons/api/tests/unit/serializers/host-test.js
```diff
@@ -4,19 +4,128 @@ import { setupTest } from 'ember-qunit';
...
-    let record = store.createRecord('host', {});
+  test('it serializes host with type plugin as expected', function (assert) {
+    assert.expect(1);
+    const store = this.owner.lookup('service:store');
+    const recordAws = store.createRecord('host', {
+      name: 'Host 2',
+      description: 'Description',
+      host_catalog_id: '123',
+      compositeType: 'aws',
+      address: '188e:68a9:b342:c05c:2595:2f46:499c:759f',
```
randallmorey, line 38: Address does not apply to dynamic hosts.

addons/api/tests/unit/serializers/host-test.js
```diff
@@ -4,19 +4,128 @@ import { setupTest } from 'ember-qunit';
...
+  test('it normalizes type static record', function (assert) {
+    assert.expect(1);
+    const store = this.owner.lookup('service:store');
+    const serializer = store.serializerFor('host');
+    const host = store.createRecord('host').constructor;
+    const payload = {
+      id: '1',
+      name: 'host test',
+      type: 'static',
+      ip_addresses: ['10.0.0.1', '10.0.0.2', '10.0.0.3'],
```
randallmorey, line 61: Technically, `ip_addresses` and `dns_names` do not apply to `static` hosts.  Only `attributes.address` applies to `static`.

randallmorey (changes requested):

randallmorey (approved):

PR #942, opened 2022-02-08
chore: host catalog form- add rest of the AWS and Azure fields

DhariniJeeva:
currently these are not supported: `access_key_id` and `secret_access_key` 
link [here](https://boundary-ui-kkx8779mx-hashicorp.vercel.app/scopes/s_xk1i4dk50v2/host-catalogs/new?type=aws)
Screen Shot 2022-02-08 at 1 12 11 PM

The POST request to the API is serialized properly, but we are missing these two fields (`access_key_id` and `secret_access_key` ) in the normalization.  This will be covered in a separate PR

### DIFF
--- a/addons/core/translations/form/en-us.yaml
+++ b/addons/core/translations/form/en-us.yaml
@@ -136,3 +136,7 @@ secret_access_key:
   link-text: Learn more
 region: 
   label: Region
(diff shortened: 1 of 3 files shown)


calcaide (approved):
LGTM!

PR #948, opened 2022-02-09
Feat typed host

DhariniJeeva:
this pr splits host into static, azure and AWS components similar to host-catalog and host-set.  

Note: only view mode is supported in AWS and Azure host. create, view and edit functions are supported in static host


addons/core/translations/resources/en-us.yaml
```diff
@@ -155,6 +155,9 @@ host:
     unknown: Unknown
     aws: AWS
     azure: Azure
+  form:
+    provider:
```
randallmorey, line 159: Let's consolidate on a single `provider` translation (maybe from host catalog?).  It's the same across all resources in the feature.

randallmorey (changes requested):

addons/core/translations/resources/en-us.yaml
DhariniJeeva: good idea! I will consolidate

DhariniJeeva: translation consolidation is done and merged [here](https://github.com/hashicorp/boundary-ui/pull/955)

calcaide (approved):
LGTM

PR #950, opened 2022-02-09
chore: 🤖 reuse name/description when changing instance type

randallmorey:
This PR modifies _new host catalog_ behavior such that `name` and `description` are retained when switching among instance types.

### DIFF
--- a/ui/admin/app/routes/scopes/scope/host-catalogs/new.js
+++ b/ui/admin/app/routes/scopes/scope/host-catalogs/new.js
@@ -18,20 +18,23 @@ export default class ScopesScopeHostCatalogsNewRoute extends Route {
   // =methods
 
   /**
-   * Rollback/destroy any new, unsaved instances from this route before
-   * creating another.
-   */
-  beforeModel() {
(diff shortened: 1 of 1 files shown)


ui/admin/app/routes/scopes/scope/host-catalogs/new.js
```diff
@@ -17,21 +17,24 @@ export default class ScopesScopeHostCatalogsNewRoute extends Route {
...
    * Creates a new unsaved host catalog belonging to the current scope.
+   * Also rollback/destroy any new, unsaved instances from this route before
+   * creating another, but reuse name/description if available.
    * @return {HostCatalogModel}
    */
   model() {
     const scopeModel = this.modelFor('scopes.scope');
-    return this.store.createRecord('host-catalog', { scopeModel });
+    let name, description;
+    if (this.currentModel?.isNew) {
```
calcaide, line 29: In the case the model is not new, `name` and `description` will keep being undefiend, right?

randallmorey: Correct, in which case there is no effect.

DhariniJeeva (approved):
💯

PR #953, opened 2022-02-10
refactor: 💡 Refactor app normalize and host catalog normalize

calcaide:
Refactor app normalize to deal with nestedAttributes and nestedSecrets as expected, before it was not working as expected. Refactor the host catalog normalize method. Add normalize test on host catalog to validate changes on normalize methods.

### DIFF
--- a/addons/api/addon/serializers/application.js
+++ b/addons/api/addon/serializers/application.js
@@ -236,7 +236,7 @@ export default class ApplicationSerializer extends RESTSerializer {
     typeClass.attributes.forEach((attribute) => {
       const { isNestedAttribute } = attribute.options;
       const attributeValue = hash.attributes?.[attribute.name];
-      if (isNestedAttribute && attributeValue) {
+      if (isNestedAttribute && typeof attributeValue != 'undefined') {
         hash[attribute.name] = attributeValue;
       }
     });
@@ -261,7 +261,15 @@ export default class ApplicationSerializer extends RESTSerializer {
         name,
         options: { isNestedSecret },
       } = attribute;
-      if (isNestedSecret) hash[name] = null;
+      const attributeValue = hash.secrets?.[name];
+
+      // FIXME: This could be improved
+      if (isNestedSecret) {
+        hash[name] = null;
+        if (attributeValue) {
+          hash[name] = attributeValue;
+        }
+      }
     });
     return hash;
   }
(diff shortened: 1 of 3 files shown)


addons/api/addon/serializers/application.js
```diff
@@ -236,7 +236,7 @@ export default class ApplicationSerializer extends RESTSerializer {
     typeClass.attributes.forEach((attribute) => {
       const { isNestedAttribute } = attribute.options;
       const attributeValue = hash.attributes?.[attribute.name];
-      if (isNestedAttribute && attributeValue) {
+      if (isNestedAttribute && typeof attributeValue != 'undefined') {
```
DhariniJeeva, line 239: just curious. why was this changed? do we not care about other falsy values here?

calcaide: I will try my best to explain, but it is complex. If you think it does not make sense, let me know.

Before, (isNestedAttribute && attributeValue):
- `isNestedAttribute`: simple, just checking if the attribute being normalized has that option set to true on the model.
- `attributeValue`: in this case we were checking if `attributeValue`, defined above, had value. If it had no value, `attributeValue` would be undefined. If undefined will never enter on this conditional. That was ok and expected. 
But then an edge case pops up, an attribute that is type boolean. So attributeValue was NOT undefined, but its value (because it is a boolean in this case) was false. Because of these reasons, the attribute was never set and it had no value in the model.

Now, we change the checking on `attributeValue` to be more explicit with being undefined.

calcaide: Why in `normalizeNestedSecrets ` we are setting some values to null?
I have to say I would like to discuss further that with @randallmorey. This approach works, but I am just not 100% sure if the best way to go.

Biggest difference between `normalizeNestedAttributes` and `normalizeNestedSecrets` is:
- `normalizeNestedAttributes`: we just need to check if is true or false. The logic we apply depending on its value excludes the other.
- `normalizeNestedSecrets`: we need to check if it is true or false. BUT also have some fields declared as `normalizeNestedSecrets` that will have no value when a specific secret field pertains to aws but not azure.
- So the ones we assign to NULL, are the secret fields do not pertain to that specific type. Already checked with Randy, in normalization it is ok to get returned fields with `null` value.

DhariniJeeva: ok, that sounds ok to me. I'm approving this pr.  we can address later to refactor this if necessary!

DhariniJeeva (approved):

addons/api/addon/serializers/application.js
calcaide: Yes. I already set a 1 on 1 with @randallmorey to review this next week. If we find out something else will update with another PR.

PR #957, opened 2022-02-11
feat: 🎸 Add host fields to mirage

calcaide:
Add host new fields to mirage.

### DIFF
--- a/addons/api/mirage/factories/host.js
+++ b/addons/api/mirage/factories/host.js
@@ -1,5 +1,32 @@
 import factory from '../generated/factories/host';
 import permissions from '../helpers/permissions';
+import { internet, datatype } from 'faker';
+
+/**
+ * Returns an array of strings with IP addresses.
+ * @returns {array}
+ */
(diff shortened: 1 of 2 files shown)


addons/api/mirage/factories/host.js
```diff
@@ -18,4 +45,16 @@ export default factory.extend({
...
     }
   },
+  attributes() {
+    return {
+      address: internet.ipv6(),
+      ip_addresses: returnIpAddressesArray(),
+      external_id: datatype.uuid(),
+      // Only aws plugins have dns_names
+      dns_names:
+        this.type != 'static' && this.plugin?.name === 'aws'
```
DhariniJeeva, line 55: do we need the `type` check here?

DhariniJeeva:
nice work! Looks good to me. 

I just left a minor comment.

addons/api/mirage/factories/host.js
calcaide: Actually is redundant to checking plugin. Changing it

PR #959, opened 2022-02-14
Refactor nested fields serialisation.

randallmorey:
This PR is a simple refactor to cleanup serialization/normalization of nested `attributes` and `secrets`:

- It is unnecessary to directly model `attributes` or `secrets` objects.  Only their sub fields should be modelled.
- Nested secrets should be explicitly normalized to `null` when missing.  In all known cases so far, nested secrets are write-only, meaning they should not persist after saving.


calcaide (approved):
LGTM! Thanks for this

gsusmi (approved):

PR #960, opened 2022-02-14
chore: 🤖 updates to labels

DhariniJeeva:
This pr updates type `‘Unknown’` to `plugin` per design/product feedback and also hosts within host-set is updated with the host-set type instead of `plugin`

unknown to [plugin](https://boundary-ui-git-chore-update-to-plugin-hashicorp.vercel.app/scopes/s_xk1i4dk50v2/host-catalogs)

[label change for hosts within plugin host-sets ](https://boundary-ui-git-chore-update-to-plugin-hashicorp.vercel.app/scopes/s_xk1i4dk50v2/host-catalogs/host-catalog-id-3/host-sets/host-set-id-9/hosts)

### DIFF
--- a/addons/core/translations/resources/en-us.yaml
+++ b/addons/core/translations/resources/en-us.yaml
@@ -121,16 +121,6 @@ host-set:
       add-none:
         title: No Hosts Available
         description: No hosts available for selection.
-  types:
-    static:
-      label: Static
-      help: Provide a list of hosts and host sets with static addresses
-    unknown:
-      label: Unknown
-    aws:
-      label: AWS
-    azure:
-      label: Azure
 host:
   title: Host
   title_plural: Hosts
@@ -148,16 +138,6 @@ host:
     add: Add Host
     create: New Host
     delete: Delete Host
-  types:
-    static:
-      label: Static
-      help: Provide a list of hosts and host sets with static addresses
-    unknown:
-      label: Unknown
-    aws:
-      label: AWS
-    azure:
-      label: Azure
 session:
   title: Session
   title_plural: Sessions
--- a/addons/core/translations/resources/en-us.yaml
+++ b/addons/core/translations/resources/en-us.yaml
@@ -89,7 +89,7 @@ host-catalog:
       label: Static
       help: Provide a list of hosts and host sets with static addresses
     unknown:
-      label: Unknown
+      label: Plugin
     aws:
       label: AWS
     azure:
--- a/addons/core/translations/resources/en-us.yaml
+++ b/addons/core/translations/resources/en-us.yaml
@@ -97,13 +97,6 @@ host-catalog:
     dynamic: 
       label: Dynamic
       help: Automate discovery of target hosts and services on Azure or AWS
-  form:
-    provider:
-      label: Provider Details
-      description: This is the provider for this host catalog. The format of the host set filters are specific to each provider.
-      choose_a_provider: 
-        label: Choose a provider 
-        description: There are several built-in plugins for the most common providers. You can choose from on of the included plugins or use a custom plugin.
 host-set:
   title: Host Set
   title_plural: Host Sets
@@ -156,10 +149,15 @@ host:
     create: New Host
     delete: Delete Host
   types:
-    static: Static
-    unknown: Unknown
-    aws: AWS
-    azure: Azure
+    static:
+      label: Static
+      help: Provide a list of hosts and host sets with static addresses
+    unknown:
+      label: Unknown
+    aws:
+      label: AWS
+    azure:
+      label: Azure
 session:
   title: Session
   title_plural: Sessions
--- a/addons/core/translations/resources/en-us.yaml
+++ b/addons/core/translations/resources/en-us.yaml
@@ -84,19 +84,15 @@ host-catalog:
       title: Welcome to Host Catalogs
   actions:
     delete: Delete Host Catalog
+  help: 
+    static: Provide a list of hosts and host sets with static addresses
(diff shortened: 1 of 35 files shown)


calcaide:
I realize this PR introduces an error in missing translation in a host that pertains to a host-set, [here link](https://boundary-ui-git-chore-update-to-plugin-hashicorp.vercel.app/scopes/s_xk1i4dk50v2/host-catalogs/host-catalog-id-3/host-sets/host-set-id-9/hosts).

Here screenshot:
Screen Shot 2022-02-14 at 10 29 04 AM

ui/admin/app/templates/scopes/scope/host-catalogs/host-catalog/host-sets/host-set/hosts.hbs
```diff
@@ -26,7 +26,21 @@
             <p>{{host.description}}</p>
           </row.headerCell>
           <row.cell>
-            <Rose::Badge>{{host.type}}</Rose::Badge>
+            {{#if host.isStatic}}
+              <Rose::Badge>
+                {{t (concat 'resources.host.types.' host.compositeType)}}
```
randallmorey, line 31: Host sets and hosts can reuse the same type translations used by host catalogs, to reduce duplication.

randallmorey (changes requested):

ui/admin/app/templates/scopes/scope/host-catalogs/host-catalog/host-sets/host-set/hosts.hbs
calcaide, line 31: missing the `.label`  within the concat after `host.compositeType`.

ui/admin/app/templates/scopes/scope/host-catalogs/host-catalog/host-sets/host-set/hosts.hbs
```diff
@@ -26,7 +26,21 @@
...
+              </Rose::Badge>
+            {{else}}
+              <Rose::Badge
+                @icon={{concat
+                  'flight-icons/svg/'
+                  host.compositeType
+                  '-color-16'
+                }}
+              >
+                {{t (concat 'resources.host.types.' host.compositeType)}}
```
calcaide, line 41: missing the `.label`  within the concat after `host.compositeType`.

calcaide (changes requested):
Left two comments where we are missing a translation.

ui/admin/app/templates/scopes/scope/host-catalogs/host-catalog/host-sets/host-set/hosts.hbs
DhariniJeeva: do you recommend removing the `types` translations from host/host-set?  
perhaps move the types to the form translation file?

randallmorey: Since we already have these `types` specified for host catalogs, I'd remove mention of them from hosts and host sets and change the references to point to host catalogs:  https://github.com/hashicorp/boundary-ui/blob/main/addons/core/translations/resources/en-us.yaml#L87

DhariniJeeva: sounds good!

randallmorey: Host sets and hosts will always have the same set of types as host catalogs.

DhariniJeeva:
fixing the target host-sources translations.

addons/core/translations/resources/en-us.yaml
```diff
@@ -89,7 +89,7 @@ host-catalog:
       label: Static
```
randallmorey, line 89: Types aren't like form fields and should be simple, e.g. `static: Static` as they are for other resources..  If we need type descriptions I think we need a different way to express them.

randallmorey (changes requested):

addons/core/translations/resources/en-us.yaml
DhariniJeeva: what do you think about doing something like this? (similarly for host/host-set)
```
host-catalog: 
  help: 
     Static: This is static helper text
     Dynamic: This is static helper text
     Azure: This is static helper text
     AWS: This is static helper text
```

ui/admin/app/templates/scopes/scope/host-catalogs/host-catalog/hosts/index.hbs
```diff
@@ -30,7 +30,9 @@
             {{#if host.isStatic}}
```
randallmorey, line 30: This should use the new type badge component.

addons/core/translations/resources/en-us.yaml
```diff
@@ -89,7 +89,7 @@ host-catalog:
...
       help: Provide a list of hosts and host sets with static addresses
     unknown:
```
randallmorey, line 91: This mapping no longer makes sense.

randallmorey (changes requested):

addons/core/translations/resources/en-us.yaml
randallmorey: Works for me.

randallmorey (approved):

calcaide (approved):
LGTM

PR #961, opened 2022-02-14
updates to host-set plugin filters

DhariniJeeva:
This pr enables filter fields on both AWS and Azure host


calcaide (approved):
LGTM awesome work

PR #962, opened 2022-02-14
fix: append ?plugin_name to dynamic host catalog create requests

randallmorey:
The API requires us to send ?plugin_name on create. This is how we set the type for dynamic host catalogs. It does not accept this via the payload.


calcaide (approved):
The code looks good to me.

Can you explain (happy to jump on a zoom call) the purpose of this?

randallmorey:
> The code looks good to me.
> 
> Can you explain (happy to jump on a zoom call) the purpose of this?

Sorry for the lack of PR description.  The API requires us to send `?plugin_name` on create.  This is how we set the type for dynamic host catalogs.  It does not accept this via the payload.

PR #963, opened 2022-02-14
fix: 🐛 static host creation address field

DhariniJeeva:


gsusmi (approved):

ui/admin/app/components/form/host/static/index.hbs
```diff
@@ -56,7 +56,7 @@
   <form.input
     @name='address'
     @type='text'
-    @value={{@model.attributes.address}}
+    @value={{@model.address}}
     @label={{t 'form.address.label'}}
     @error={{@model.errors.name}}
```
gsusmi, line 61: On an unrelated note, shouldn't this be `errors.address`?

DhariniJeeva: it should be! good catch. I'm gonna update this pr

DhariniJeeva: @gsusmi  done!

PR #967, opened 2022-02-15
Expose disable credential rotation field

randallmorey:
This PR exposes the `disable_credential_rotation` field for dynamic host catalogs.  For AWS it is editable.  For Azure, it is always `true`.

Demo:
https://boundary-ui-git-feature-disable-credential-rotation-hashicorp.vercel.app/scopes/s_xk1i4dk50v2/host-catalogs/new?type=aws

### DIFF
--- a/addons/core/translations/form/en-us.yaml
+++ b/addons/core/translations/form/en-us.yaml
@@ -118,7 +118,7 @@ callback_url:
 filter:
   label: Filter
   help: A boolean expression defining a filter run against the provided information.
-tenant_id: 
+tenant_id:
   label: Tenant ID
   help: The tenant (directory) ID for your Azure Active Directory application.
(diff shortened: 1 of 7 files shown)


DhariniJeeva:
this is not a blocker just wanted to bring up a behavior I noticed, after making a selection `disable cred rotation` and hitting save, the page doesn't show `disable cred rotation` as checked.  should it be checked and disabled?

randallmorey:
> this is not a blocker just wanted to bring up a behavior I noticed, after making a selection `disable cred rotation` and hitting save, the page doesn't show `disable cred rotation` as checked. should it be checked and disabled?

Should be fixed now.

DhariniJeeva (approved):
LGTM! ty!

PR #973, opened 2022-02-16
fix: 🐛 missed value field

DhariniJeeva:
client_secret_value was empty due to missing a value field on the input element. 

https://boundary-ui-git-fix-val-hashicorp.vercel.app/scopes/s_xk1i4dk50v2/host-catalogs/host-catalog-id-3
Screen Shot 2022-02-16 at 10 35 42 AM


calcaide (approved):
Awesome!!

gsusmi (approved):

PR #979, opened 2022-02-16
chore: support `filters` when it is a string by converting to an array

randallmorey:
This PR adds (temporary) support for host set `filters` when it is returned by the API as a string.  It is converted to an array in normalisation.  Eventually this can be rolled back once the API consistently returns an array.


gsusmi (approved):
Ty!

PR #980, opened 2022-02-17
chore: make radio cards a bit narrower

randallmorey:
Fixes an issue where radio cards stack at reasonable browser widths, when they should line up horizontally.

Before:
Screenshot 2022-02-17 at 08 34 54

After:
Screenshot 2022-02-17 at 09 21 13


gsusmi (approved):
Good catch. Ty!

calcaide (approved):

PR #995, opened 2022-02-23
updates to AWS and Azure doc links

DhariniJeeva:
This pr adds AWS doc links and refactor the existing dynamic host catalog doc links


ui/admin/app/components/form/host-catalog/aws/index.hbs
```diff
@@ -138,6 +138,7 @@
     @label={{t 'form.access_key_id.label'}}
     @helperText={{t 'form.access_key_id.help'}}
     @linkText={{t 'actions.learn-more'}}
+    @link={{doc-url @awsDoc}}
```
randallmorey, line 141: I think we can pass the doc identifier string directly here, rather than handing it down through the route template and component.  These form components live in the context of the application and aren't promotable to addons (yet).  So we can simplify things here.

randallmorey (changes requested):

ui/admin/app/components/form/host-catalog/aws/index.hbs
randallmorey: Which is to say that abstracting these values out is the right approach for Rose.  But we can be much more informal for in-app components.

DhariniJeeva: updated!

randallmorey (approved):
Perfect! Thank you.

cameronperera (approved):

calcaide (approved):

PR #996, opened 2022-02-24
chore: 🤖 Add nanoid to resolutions

calcaide:
Add `nanoid` to resolutions due to [dependabot alert](https://github.com/hashicorp/boundary-ui/security/dependabot/13).

### DIFF
--- a/package.json
+++ b/package.json
@@ -91,7 +91,8 @@
     "ember-template-recast": "^6.0.0",
     "engine.io": "^5.2.1",
     "markdown-it": "^12.3.2",
-    "mout": "^1.2.3"
+    "mout": "^1.2.3",
+    "nanoid": "^3.1.31"
   },
   "config": {
     "commitizen": {
--- a/yarn.lock
+++ b/yarn.lock
@@ -18180,10 +18180,10 @@ nan@^2.12.1:
   resolved "https://registry.yarnpkg.com/nan/-/nan-2.15.0.tgz#3f34a473ff18e15c1b5626b62903b5ad6e665fee"
   integrity sha512-8ZtvEnA2c5aYCZYd1cvgdnU6cqwixRoYg70xPLWUws5ORTa/lnw+u4amixRS/Ac5U5mQVgp9pnlSUnbNWFaWZQ==
 
-nanoid@3.1.20:
-  version "3.1.20"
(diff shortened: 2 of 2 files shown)


package.json
```diff
@@ -91,7 +91,8 @@
     "ember-template-recast": "^6.0.0",
     "engine.io": "^5.2.1",
     "markdown-it": "^12.3.2",
-    "mout": "^1.2.3"
+    "mout": "^1.2.3",
+    "nanoid": "^3.1.31"
```
cameronperera, line 95: Why does this appear to added and not updated?

gsusmi: Trailing comma (`"mout": "^1.2.3"`  vs `"mout": "^1.2.3",`) triggered an addition.

cameronperera: Sorry, I was referring to the fact that nanoid was not in the package.json prior to this PR but it is in the yarn.lock. Just curious.

gsusmi: Oops. My bad. Rather than adding multiple resolutions for nanoid in every package.json that has a dependency, surfacing it at this monorepo package.json would resolve it for entire app.
You'll notice that electronjs based resolutions are restricted to [ui/desktop/electron-app/package.json](https://github.com/hashicorp/boundary-ui/blob/889461770985f316507a65a7bf232d828637adc5/ui/desktop/electron-app/package.json#L44) as it's not shared across this monorepo.

cameronperera: Cool. Thank you for explaining that for me.

cameronperera (approved):

package.json
calcaide: Thank you @gsusmi for the clarification.

@cameronperera I am not sure if you noticed, but we declare these dependencies as `resolutions`, it is a yarn specific feature, [more info about it here](https://classic.yarnpkg.com/lang/en/docs/selective-version-resolutions/).

In a nutshell, let's say we have `dependencyA` that has a sub-dependency in `dependencyB@1.0.0`.

If we add `dependencyB@1.5.0` within resolutions, `dependencyA` will resolve version `1.5.0` of `dependencyB`.

We use this to be compliant with security vulnerabilities.

cameronperera: I did not notice it was in `resolutions`. Thanks for pointing that out, makes perfect sense now.

randallmorey (approved):

PR #999, opened 2022-02-24
chore: 🤖 Update ember-try

calcaide:
Update `ember-try` from `1.4.0` to `2.0.0`. As [per changelog](https://github.com/ember-cli/ember-try/blob/master/CHANGELOG.md) there is no risk on the update

### DIFF
--- a/addons/api/package.json
+++ b/addons/api/package.json
@@ -59,7 +59,7 @@
     "ember-source": "~3.28.4",
     "ember-source-channel-url": "^3.0.0",
     "ember-template-lint": "^3.7.0",
-    "ember-try": "^1.4.0",
+    "ember-try": "^2.0.0",
     "eslint": "^7.32.0",
     "eslint-config-prettier": "^8.3.0",
     "eslint-plugin-ember": "^10.5.4",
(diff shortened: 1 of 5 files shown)


cameronperera (approved):
What caused all the deletions from the yarn.lock file?

randallmorey (changes requested):
Are we even using Ember Try?  If not, we can probably rip out the dep altogether.

calcaide:
> Are we even using Ember Try? If not, we can probably rip out the dep altogether.

@randallmorey That's actually a good point.

I assumed we use to run tests with different ember versions as is specified in the API project, [here](https://github.com/hashicorp/boundary-ui/tree/main/addons/api#yarn-commands) at the `test:all` row it says:
`Runs tests across multiple Ember versions with ember-try.`

I checked `ember-try` documentation on how to use it ([link here](https://github.com/ember-cli/ember-try)) but the thing is I do not find the implementation in our code base. So I think you are right, we aren't using it :)

I will triple check that, close this PR and provide a new one to get rid of `ember-try`.

Thank you for the feedback and nice catch!!

calcaide:
@randallmorey 

Closing this PR in favour of deprecating `ember-try` in [this new PR](https://github.com/hashicorp/boundary-ui/pull/1002)

PR #1001, opened 2022-02-24
ICU 3488: Display DNS Names for AWS Hosts in Admin UI

cameronperera:
The test cases need to be updated to handle the different types of host-catalogs properly. Test will be updated and added in a later PR. @joshuaogle the wireframe didn't have a suggested name for the title so I went with "DNS Names". Let me know if you have another suggestion.
Old:
Screen Shot 2022-02-25 at 10 10 30 AM

New:
Screen Shot 2022-02-25 at 10 12 23 AM

### DIFF
--- a/addons/core/translations/en-us.yaml
+++ b/addons/core/translations/en-us.yaml
@@ -33,6 +33,7 @@ titles:
   choose-a-provider: Choose a provider
   provider: Provider
   ip-addresses: Latest IP Addresses
+  dns-names: DNS Names
 descriptions:
   empty-set: There are no items to display yet.  You may be able to add items or try back later.
   origin-initialization: To get started, please enter your origin URL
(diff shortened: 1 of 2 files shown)


calcaide (approved):
LGTM!! Congrats on your first PR as member of Boundary team :)

addons/core/translations/en-us.yaml
```diff
@@ -33,6 +33,7 @@ titles:
   choose-a-provider: Choose a provider
   provider: Provider
   ip-addresses: Latest IP Addresses
+  dns-names: DNS Names
```
randallmorey, line 36: `dns_names` is a field, so we could move this translation into `core/translations/form/` and treat is a a label.  Similarly, `ip_addresses` is also a form field and could be moved (as a separate PR).

randallmorey (changes requested):

cameronperera:
Closing PR due to CLA issues. Opened this [PR](https://github.com/hashicorp/boundary-ui/pull/1004) that resolves CLA signing issues.

PR #1002, opened 2022-02-25
refactor: 💡 Deprecate ember-try

calcaide:
Deprecate the usage of `ember-try` because we are not using it.

During this task, I find out perhaps we are not using `ember-source-channel-url`. I will provide a separate PR to deprecate `ember-source-channel-url`.


cameronperera (approved):

randallmorey (approved):

randallmorey (commented):
Converted to draft because I need to consider how this might impact our Ember 4 ambitions.

calcaide:
After a conversation with @randallmorey over slack.

We decide to go ahead with the deprecation of `ember-try`. And if in the future we need it for upgrading to ember `4.0` we will bring it back.

PR #1004, opened 2022-02-25
feat: 🎸 Display DNS names for AWS hosts in Admin UI

cameronperera:
The test cases need to be updated to handle the different types of host-catalogs properly. Test will be updated and added in a later PR. @joshuaogle[](https://user-images.githubusercontent.com/29386339/155749672-b574bcf9-b533-4825-bf64-339600c44d71.png)[](https://user-images.githubusercontent.com/29386339/155749700-a952b76f-68ef-46d2-99e7-3bf3d83144e5.png) the wireframe didn't have a suggested name for the title so I went with "DNS Names". Let me know if you have another suggestion.
Old:
Screen Shot 2022-02-25 at 10 10 30 AM

New:
Screen Shot 2022-02-25 at 2 01 16 PM

### DIFF
--- a/addons/core/translations/form/en-us.yaml
+++ b/addons/core/translations/form/en-us.yaml
@@ -138,4 +138,4 @@ secret:
 disable_credential_rotation:
   label: Disable credential rotation
 dns_names:
-  label: DNS Names
+  label: Latest DNS Names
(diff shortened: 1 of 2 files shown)


calcaide (approved):
LGTM!

randallmorey (approved):

joshuaogle (approved):
This looks great to me! It might be more appropriate to use a larger heading size (a section title instead of a label) for these since it's more than one field, but this is consistent with what we have on other pages so I'm okay with not updating that yet.

randallmorey (approved):

PR #1007, opened 2022-02-28
feat: 🎸 ICU-3529 add rotating credentials warning for azure

cameronperera:
Add "Boundary does not yet support rotating credentials" warning message
for Azure provider

✅ Closes: ICU-3529

Current:
Screen Shot 2022-02-28 at 2 37 55 PM

New:
Screen Shot 2022-02-28 at 2 38 33 PM


addons/core/translations/resources/en-us.yaml
```diff
@@ -97,6 +97,10 @@ host-catalog:
     azure_client_id:
       label: Client ID
       help: The client ID for the client you want Boundary to use.
+    azure_choose_a_provider:
+      description: There are several built-in plugins for the most common providers. You can choose from on of the included plugins or use a custom plugin. Boundary does not yet support rotating credentials.
+    aws_choose_a_provider:
+      description: There are several built-in plugins for the most common providers. You can choose from on of the included plugins or use a custom plugin.
```
cameronperera, line 103: I am open to suggestions about whether this should live here or not. I did keep the original `choose-a-provider` option because it is still being used by the static host-catalog form.

cameronperera:
Not sure what the standard testing practices are around static text, but I would typically add a test to verify that the "Choose a Provider" description is updated correctly depending on the provider a user selected. Thoughts on this?

addons/core/translations/resources/en-us.yaml
randallmorey: Hmm.  I think it's a good idea to have a default `choose_a_provider`, otherwise these will grow with each provider addition even if they're mostly the same.

ui/admin/app/components/form/host-catalog/aws/index.hbs
```diff
@@ -77,7 +77,9 @@
 
     <form.fieldset>
       <:title>{{t 'titles.choose-a-provider'}}</:title>
-      <:description>{{t 'descriptions.choose-a-provider'}}</:description>
+      <:description>
+        {{t 'resources.host-catalog.form.aws_choose_a_provider.description'}}
```
randallmorey, line 81: What if this used the default.

ui/admin/app/components/form/host-catalog/azure/index.hbs
```diff
@@ -76,7 +76,9 @@
     </form.fieldset>
     <form.fieldset>
       <:title>{{t 'titles.choose-a-provider'}}</:title>
-      <:description>{{t 'descriptions.choose-a-provider'}}</:description>
+      <:description>
+        {{t 'resources.host-catalog.form.azure_choose_a_provider.description'}}
```
randallmorey, line 80: And this used the Azure-specific.

randallmorey (changes requested):

randallmorey:
> Not sure what the standard testing practices are around static text, but I would typically add a test to verify that the "Choose a Provider" description is updated correctly depending on the provider a user selected. Thoughts on this?

Our tests are mostly functional and we don't typically test static text content.

We aren't there yet, but one reason why this gets hard is that it duplicates strings between the translations and the test cases.  If we add more languages, we'd want to test those as well.  That's a lot of strings in test cases, which must all be updated whenever the translations change.

If we were to test internationalized strings, we might want to think about a language-agnostic approach that uses translation _keys_ in some kind of data attribute?  With a handful of smoke tests to validate that translations are working at all for a given language.

Thoughts?

addons/core/translations/resources/en-us.yaml
calcaide: Agreed with Randy. We could have a default `choose_a_provider` and then if need, extra information coming from adhoc ones.

calcaide:
> 

I think it is too early to add any translation tests. It will add more work and code for (IMHO) no value.

I will reconsider this, once we are supporting several languages, but ATM I think there is no value for the effort it requires.

addons/core/translations/resources/en-us.yaml
DhariniJeeva: I agree with having a default `choose_a_provider` esp since it is used in `static` forms!

cameronperera: I really like the concat approach @calcaide. And yes, I agree that this could grow out of hand as we add more providers. I will remove the AWS specific one and update it to use concatenation when needed.

cameronperera:
> > Not sure what the standard testing practices are around static text, but I would typically add a test to verify that the "Choose a Provider" description is updated correctly depending on the provider a user selected. Thoughts on this?
> 
> Our tests are mostly functional and we don't typically test static text content.
> 
> One reason why this gets cumbersome is that it duplicates strings between the translations and the test cases. We aren't there yet, but if we add more languages we'd need to test those as well. That's a lot of strings in test cases, which must all be updated whenever the translations change.
> 
> If we were to test internationalized strings, we might want to think about a language-agnostic approach that uses translation _keys_ in some kind of data attribute? With a handful of smoke tests to validate that translations are working at all for a given language.
> 
> Thoughts?

I agree with what is said above. Testing static text typically does not add much value and opens us up to a lot of manual changes if that text ever changes. I mainly asked because we are dynamically rendering different text depending on the provider. However, I still agree that it may not add much value right now.

To your last comment about tests for translations, that sounds like a good approach once we start supporting other translations.

calcaide (approved):
LGTM thank you!

randallmorey (approved):

PR #1012, opened 2022-03-02
feat: 🎸 add sync_interval_seconds to plugin host-sets

DhariniJeeva:
this pr adds the missing attribute `sync_interval_seconds` to plugin host-sets. 

displaying this field in the host-set forms will be handled in a separate PR.

### DIFF
--- a/addons/api/addon/generated/models/host-set.js
+++ b/addons/api/addon/generated/models/host-set.js
@@ -57,6 +57,13 @@ export default class GeneratedHostSetModel extends BaseModel {
   })
   plugin;
 
+  @attr('number', {
+    description:
+      "The number of seconds between the time boundary syncs the hosts in this set using this host set's plugin. If not provided a system determined default is used",
+    defaultValue: 0,
+  })
+  sync_interval_seconds;
+
   // Azure specific
   @attr('string', {
     description: '',
--- a/addons/api/addon/generated/models/host-set.js
+++ b/addons/api/addon/generated/models/host-set.js
@@ -60,7 +60,6 @@ export default class GeneratedHostSetModel extends BaseModel {
   @attr('number', {
     description:
       "The number of seconds between the time boundary syncs the hosts in this set using this host set's plugin. If not provided a system determined default is used",
-    defaultValue: 0,
   })
   sync_interval_seconds;
 
--- a/addons/api/tests/unit/serializers/host-set-test.js
+++ b/addons/api/tests/unit/serializers/host-set-test.js
@@ -12,14 +12,12 @@ module('Unit | Serializer | host set', function (hooks) {
       compositeType: 'static',
       description: 'Description',
       host_catalog_id: '123',
-      sync_interval_seconds: 1,
       version: 1,
     });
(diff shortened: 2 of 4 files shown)


addons/api/addon/generated/models/host-set.js
```diff
@@ -57,6 +57,13 @@ export default class GeneratedHostSetModel extends BaseModel {
   })
   plugin;
 
+  @attr('number', {
+    description:
+      "The number of seconds between the time boundary syncs the hosts in this set using this host set's plugin. If not provided a system determined default is used",
+    defaultValue: 0,
```
randallmorey, line 63: Do we want a client-side default?

DhariniJeeva: open to suggestions! I can't remember if we decided to show the default value.

gsusmi: `Setting to -1 will disable syncing for that host set; setting to zero will cause the set to use Boundary's default. The default may change between releases.`, from [host set proto].(https://github.com/hashicorp/boundary/blob/main/internal/proto/local/controller/api/resources/hostsets/v1/host_set.proto#L64).

randallmorey: Hmm I'm not convinced the client side needs to specify a default though.  I'm not sure we do it in any other number fields.  And is it something that _could_ be machine generated?  If no, then we can omit from here.  If the only way to know that `0` is meaningful is the description, it cannot be machine generated.

randallmorey: What is the difference between `0` and other falsy values here?  @talanknight

talanknight: Great question.  It looks like that comment might be out dated actually.  zero is actually not an accepted value.  It can be either -1, or a positive number.  Setting the field to null will use the Boundary default here.

randallmorey (changes requested):
Let's remove the default value.

randallmorey:
> Let's remove the default value.

Also, should this PR add the field to the form(s)?  Or just the model?

cameronperera (approved):

addons/api/tests/unit/serializers/host-set-test.js
```diff
@@ -12,12 +12,14 @@ module('Unit | Serializer | host set', function (hooks) {
       compositeType: 'static',
       description: 'Description',
       host_catalog_id: '123',
+      sync_interval_seconds: 1,
```
calcaide, line 15: Does `sync_interval_seconds` belongs to `static`, `plugin` or borth types? be aware this test is JUST for static.

DhariniJeeva: seems like only for type `plugin`! I'll add a new test case for this

DhariniJeeva: good catch! I updated serializer, mirage, and tests to include `sync_interval_seconds` only in type `plugin`

randallmorey (approved):

PR #1013, opened 2022-03-03
 feat: 🎸 Add default custom Model for mirage

calcaide:
Add default custom Model for Mirage. All the domain models in mirage now inherit from this new default model.

The goal in this PR is:
- Add the new default model.
- All the domains now have same default attributes assigned at creation time.


addons/api/mirage/models/default-model.js
```diff
@@ -0,0 +1,20 @@
+import { Model } from 'ember-cli-mirage';
+
+export default Model.extend({
+  assignDefaults(defaultAttrs, attrs) {
+    for (const [key, value] of Object.entries(defaultAttrs)) {
+      if (!Object.prototype.hasOwnProperty.call(attrs, key)) {
```
randallmorey, line 6: Is `attrs` ever more than a POJO?  If it's just a POJO this conditional may not be necessary.

randallmorey (changes requested):

addons/api/mirage/models/default-model.js
calcaide: @randallmorey 
The reason for this conditional is to avoid overriding an attribute that already exhisted with the default value.

I saw the `update()` method in the mirage Model parent class ends up calling `save()` so I wanted to prevent overriding. But if you think overriding will not be an issue here definetly I will delete the conditional.

Let me know your thoughts.

addons/api/mirage/models/default-model.js
```diff
@@ -0,0 +1,20 @@
+import { Model } from 'ember-cli-mirage';
+
+export default Model.extend({
```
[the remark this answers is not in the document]
calcaide: Makes sense. `base` or `base-model`? I think `base-model` is more descriptive, but since is already within the `/model` folder I am fin either way.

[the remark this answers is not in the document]
cameronperera: I would lean towards `base` since it is in the `/model` folder already.

addons/api/mirage/models/default-model.js
randallmorey: Gotcha.  Can we simplify it?  Maybe `if (!attrs[key])`?  Can we define it as a POJO?  I can't think of a reason to return a class instance or anything more complex from the defaults method.

calcaide: Answering your previous commnet (for line 5) and this one:
Found a better way of merging these two objects while keeping overriding actual attributes with default values by mistake:
`this.attrs = {...this.defaultAttrs(), ...this.attrs}`
Because the first object is `defaultAttrs`, `attrs` will not be override it.

calcaide: After working on the previous suggested `this.attrs = {...this.defaultAttrs(), ...this.attrs}` does not work as expected.

Implemented suggestions by @randallmorey

PR #1014, opened 2022-03-07
chore: display sync interval seconds on host-set plugin forms

DhariniJeeva:
This pr displays sync_interval_seconds on host-set Azure and AWS forms

 Azure:  https://boundary-ui-ej4eghf2c-hashicorp.vercel.app/scopes/s_xk1i4dk50v2/host-catalogs/host-catalog-id-3/host-sets/host-set-id-11

AWS: https://boundary-ui-ej4eghf2c-hashicorp.vercel.app/scopes/s_xk1i4dk50v2/host-catalogs/host-catalog-id-5/host-sets/host-set-id-15
Screen Shot 2022-03-17 at 12 27 15 PM

 
Screen Shot 2022-03-17 at 12 27 11 PM


addons/core/translations/en-us.yaml
```diff
@@ -32,6 +32,7 @@ titles:
   no-results: No Results
   choose-a-provider: Choose a provider
   provider: Provider
+  sync-interval-seconds: Sync Interval Seconds
```
joshuaogle, line 35: Can we change this title to `Sync Interval`?

And can we also add a description that says:
```
Number of seconds between the time boundary syncs hosts in this set. Uses system default if unchecked.
```

joshuaogle (changes requested):

addons/core/translations/en-us.yaml
DhariniJeeva: added!

ui/admin/app/components/form/host-set/aws/index.js
```diff
@@ -0,0 +1,17 @@
...
+
+export default class FormHostSetAwsComponent extends Component {
+  // =actions
+  @tracked showSyncIntervalSeconds = true;
+
+  @action
+  toggleSyncIntervalSeconds(model) {
+    this.showSyncIntervalSeconds = !this.showSyncIntervalSeconds;
+
+    if (!this.showSyncIntervalSeconds) {
```
calcaide, line 13: What is the type of this.showSyncIntervalSeconds? asking because in case is boolean we might need to check if exists differently.

ui/admin/app/components/form/host-set/aws/index.js
```diff
@@ -0,0 +1,17 @@
...
+      model.sync_interval_seconds = null;
```
calcaide, line 14: Why are we setting the in the model `sync_interval_seconds` to `null` if this.showSyncIntervalSeconds does not exist?

ui/admin/app/components/form/host-set/aws/index.js
```diff
@@ -0,0 +1,17 @@
+import Component from '@glimmer/component';
+import { action } from '@ember/object';
+import { tracked } from '@glimmer/tracking';
+
+export default class FormHostSetAwsComponent extends Component {
+  // =actions
+  @tracked showSyncIntervalSeconds = true;
+
+  @action
+  toggleSyncIntervalSeconds(model) {
```
calcaide, line 10: Assume I might be missing context, so if I ask something too obvious let me know.

We are declaring `toggleSyncIntervalSeconds` within `components/form/host-set/aws/` wich is ok.
But we are using it in `form/host-set/aws` and `form/host-set/azure` templates. Shouldn't this be created in `form/host-set/index.js`. 

I see this functions beeing shared between these child templates, but being declared on a child index. If is shared between childs, I would preffer to have it declared at parent level.

ui/admin/app/components/form/host-set/azure/index.hbs
```diff
@@ -168,7 +168,7 @@
     @value={{@model.filter}}
     @label={{t 'resources.host-set.form.filter.label'}}
     @helperText={{t 'resources.host-set.form.filter.help'}}
-    @linkText={{t 'form. tiltes.learn-more'}}
+    @linkText={{t 'tiltes.learn-more'}}
```
calcaide, line 171: I think this is a typo, `tiltes` should be `titles`.

calcaide (changes requested):
I left some comments

ui/admin/app/components/form/host-set/aws/index.js
DhariniJeeva: cause API expects us to send `null` as a default value when the checkbox is unselected.

ui/admin/app/components/form/host-set/aws/index.js
DhariniJeeva: good point! I'll move this up to the parent level

ui/admin/app/components/form/host-set/aws/index.js
calcaide: As per my understanding, if we are doing this to prepare data for the API, is the serializer who takes care of this.

That said, what I would expect is:
- Model defines that field.
- If the checkbox is unselected, then normalizer will take care of setting that value to `null`.
- If the checkbox is selected, then it will set a value.

Perhaps I am missing context, happy to jump on zoom call to discuss this.

DhariniJeeva: discussed offline: 
1. checkbox field is just for the UI toggling. There is no model state for this field
2. `model.sync_interval_seconds` is set to `null` on the component side when the user unselects the checkbox which clears the text box field and sends the expected `null` to the API.

DhariniJeeva:
after our offline discussion - making the `sync_interval_seconds` a simple textbox field

calcaide:
Does `sync_interval_seconds` apply for both `azure` and `aws`? If you could you provide screenshots of both and update vercel links of both that would be awesome, thank you.

ui/admin/app/components/form/host-set/aws/index.hbs
```diff
@@ -236,6 +236,25 @@
     </:body>
   </Rose::Form::Fieldset>
 
+  <form.input
```
calcaide, line 239: This code looks the same used in `azure` template. Is that right?

If so: Should we move this to a component ? So we do not duplicate code.

Thoughts?

addons/core/translations/en-us.yaml
```diff
@@ -32,6 +32,7 @@ titles:
   no-results: No Results
   choose-a-provider: Choose a provider
   provider: Provider
+  sync-interval: Sync Interval
```
calcaide, line 35: I have no context on how we are organizing translations, perhaps @gsusmi can provide some light.

But as per my understanding (I can be wrong as mentioned above) this file is for root translations, and the content of the translations added here looks to me a form field, Shouldn't they live within `/form`?

thoughts?

calcaide (changes requested):

DhariniJeeva:
@calcaide I added the Vercel link for AWS.  The screenshots will be the same for both as it is the same field so didn't want to clutter the PR description box.  Let me know if you need more details.

ui/admin/app/components/form/host-set/aws/index.hbs
DhariniJeeva: I think that would make sense if it were an `app/template/azure`. This is a form component. I'm not quite sure how that would work here.

cameronperera (changes requested):

addons/core/translations/en-us.yaml
cameronperera: I think that is what @randallmorey suggested when I added the `Latest IP Addresses` label and field to the host form. I agree it and the description should be moved into `/form`.

ui/admin/app/components/form/host-set/aws/index.hbs
calcaide: I was meaning make the field as component. It is ok for now, in the future we might check if this makes sense and how might be best way to do it with Randy.

ui/admin/app/components/form/host-set/azure/index.hbs
```diff
@@ -168,7 +168,7 @@
     @value={{@model.filter}}
     @label={{t 'resources.host-set.form.filter.label'}}
     @helperText={{t 'resources.host-set.form.filter.help'}}
-    @linkText={{t 'form. tiltes.learn-more'}}
+    @linkText={{t 'titles.learn-more'}}
```
calcaide, line 171: I do not have much context on translations, perhaps @susmitha can help with this.

But looking for the same pattern in our codebase, in the past you add this same translation within `actions` which makes more sense to me.

Why is this one in the root of translations? And why we can't re-use the previous pattern you set?

DhariniJeeva: ah, my bad! I noticed a typo here and fixed it. it should be actions. 
but yes, more context on translations would be great.

cameronperera (approved):
LGTM

PR #1018, opened 2022-03-08
feat: 🎸 add type check helper methods to target model

DhariniJeeva:
This pr adds TCP and SSH type check methods to the target model. This is required to support both the target types.

### DIFF
--- a/addons/api/addon/models/target.js
+++ b/addons/api/addon/models/target.js
@@ -207,4 +207,20 @@ export default class TargetModel extends GeneratedTargetModel {
   removeCredentialSource(credentialLibraryID, options) {
     return this.removeCredentialSources([credentialLibraryID], options);
   }
+
+  /**
+   * True if the target type is tcp.
+   * @type {boolean}
+   */
+  get isTCP() {
+    return this.type === 'tcp';
+  }
+
+  /**
+   * True if the target type is ssh.
+   * @type {boolean}
+   */
+  get isSSH() {
+    return this.type === 'ssh';
+  }
 }
--- a/addons/api/tests/unit/models/target-test.js
+++ b/addons/api/tests/unit/models/target-test.js
@@ -395,7 +395,7 @@ module('Unit | Model | target', function (hooks) {
     assert.false(modelSSH.isTCP);
   });
 
-  test('it has isTCP property and returns the expected values', async function (assert) {
+  test('it has isTCP property and returns the expected values', function (assert) {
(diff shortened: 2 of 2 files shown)


addons/api/tests/unit/models/target-test.js
```diff
@@ -384,4 +384,24 @@ module('Unit | Model | target', function (hooks) {
     const model = store.peekRecord('target', '123abc');
     await model.removeCredentialSource('2');
   });
+
+  test('it has isSSH property and returns the expected values', async function (assert) {
```
cameronperera, line 388:
```suggestion
  test('it has isSSH property and returns the expected values', function (assert) {
```
It does not look like we need this to be async.

addons/api/tests/unit/models/target-test.js
```diff
@@ -384,4 +384,24 @@ module('Unit | Model | target', function (hooks) {
...
+    assert.expect(2);
+    const store = this.owner.lookup('service:store');
+    const modelSSH = store.createRecord('target', {
+      type: 'ssh',
+    });
+    assert.true(modelSSH.isSSH);
+    assert.false(modelSSH.isTCP);
+  });
+
+  test('it has isTCP property and returns the expected values', async function (assert) {
```
cameronperera, line 398:
```suggestion
  test('it has isTCP property and returns the expected values', function (assert) {
```
Same as other comment.

cameronperera (changes requested):

addons/api/tests/unit/models/target-test.js
DhariniJeeva: good catch! updated!

cameronperera (approved):

addons/api/addon/models/target.js
```diff
@@ -207,4 +207,20 @@ export default class TargetModel extends GeneratedTargetModel {
   removeCredentialSource(credentialLibraryID, options) {
     return this.removeCredentialSources([credentialLibraryID], options);
   }
+
+  /**
+   * True if the target type is tcp.
+   * @type {boolean}
+   */
+  get isTCP() {
```
calcaide, line 215: Curious, where we will use this? within the targets list?

calcaide (approved):
LGTM, left a question

addons/api/addon/models/target.js
DhariniJeeva: yes! that is right

PR #1022, opened 2022-03-10
chore: 🤖 Refactor cache for dependencies within CI

calcaide:
Refactor the cache for dependencies within the CI.

**How the cache used to work:**
- Cache was working JUST at PR/branch level. So a workflow running on branch A, could not access any other cache that the one generated by branch A.
- Per each new PR we have no cache available (reasons mentioned above), so every PR at first run will need to install all dependencies, after that it generated a cache for that PR.
- The cache could not be shared within the same workflow if the runner Operating System changed.
- Every job of the workflow (3 in this case) were installing dependencies 3 times (but simultaneously).

**How the cache works now:**
- Cache works at project level. So any workflow running within the project, doesn't matter what head the PR is pointing to, has access to the cache.
- Cache now is agnostic of OS runner. The only thing it checks is hashing the yarn.lock file. If there are no changes in dependencies, cache does not change.
- Create one job to take care of dependencies. The job checks if cache is available for that run, if not install dependencies and generates a new cache. All the jobs in the workflow now depend on this one and DO NOT try to install dependencies, it assumes the first job takes care of that.

**Screenshots running the same GH workflow (Validate Monorepo):**

Before the change:
Screen Shot 2022-03-10 at 4 32 01 PM

After the change:
Screen Shot 2022-03-10 at 4 32 24 PM

### DIFF
--- a/.github/workflows/monorepo-validate.yaml
+++ b/.github/workflows/monorepo-validate.yaml
@@ -2,8 +2,8 @@ name: Validate Monorepo
 on: pull_request
 jobs:
 
-  compliance:
-    name: License Compliance
+  dependencies:
+    name: Install dependencies
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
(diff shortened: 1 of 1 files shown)


.github/workflows/monorepo-validate.yaml
```diff
@@ -2,55 +2,76 @@ name: Validate Monorepo
 on: pull_request
 jobs:
 
+  dependencies:
```
gsusmi, line 5: This is vague. How about defining this explicitly to what it does? Something like:
```
node-modules-latest:
...
compliance:
  needs: [node-modules-latest]
...
list:
  needs: [node-modules-latest]
```

calcaide: That is right! `node-modules-latest` is more descriptive. Done! and thanks for the catch.

cameronperera (approved):
Looks great!

DhariniJeeva:
awesome work!!

DhariniJeeva (approved):

PR #1027, opened 2022-03-17
feat: create ssh type target in admin-ui

DhariniJeeva:
:tickets: [Jira ticket](https://hashicorp.atlassian.net/browse/ICU-3654)

## Description

:technologist: 

[TCP](https://boundary-mwzokr1yi-hashicorp.vercel.app/scopes/s_xk1i4dk50v2/targets/new?type=tcp)

[SSH](https://boundary-mwzokr1yi-hashicorp.vercel.app/scopes/s_xk1i4dk50v2/targets/new?type=ssh)

This pr updates existing target forms to support both TCP and SSH types.  

Added a feature flag check to display and create ssh types. 

- If the feature flag[ssh-target] is set to false, then the users are allowed to view and create only the TCP target type.


addons/core/translations/resources/en-us.yaml
```diff
@@ -278,6 +278,9 @@ target:
   types:
     tcp: TCP
     ssh: SSH
+  help: 
+    tcp: This target belongs to an SSH host catalog and will use SSH to connect
+    ssh: This target belongs to an TCP host catalog and will use TCP to connect
```
cameronperera, line 283: Should these be flipped?

```suggestion
  help: 
    tcp: This target belongs to an TCP host catalog and will use TCP to connect
    ssh: This target belongs to an SSH host catalog and will use SSH to connect
```

ui/admin/app/templates/scopes/scope/targets/index.hbs
```diff
@@ -17,9 +17,26 @@
   <page.actions>
     {{#if (can 'create model' this.scope collection='targets')}}
       {{#if @model}}
-        <Rose::LinkButton @route='scopes.scope.targets.new' @style='primary'>{{t
-            'titles.new'
-          }}</Rose::LinkButton>
+        <Rose::Dropdown
```
cameronperera, line 20: 👏🏼 👏🏼 
I love this change.

cameronperera (changes requested):
Overall looks great. I believe @calcaide mentioned something similar to this before but I would love to see if we can refactor the `form/target/[tcp|ssh]/index.hbs` form. From a glance, much of it looks like duplication that we could potentially pull out into a `form/target/base.hbs` maybe.

addons/core/translations/resources/en-us.yaml
DhariniJeeva: ah yes!

DhariniJeeva:
> Overall looks great. I believe @calcaide mentioned something similar to this before but I would love to see if we can refactor the `form/target/[tcp|ssh]/index.hbs` form. From a glance, much of it looks like duplication that we could potentially pull out into a `form/target/base.hbs` maybe.

@gsusmi and I discussed this in the past when we were working on `dynamic host-catalogs` and couldn't get the form data to be shared from the root `index.hbs` hence followed this existing pattern of auth-methods but completely open to discussions on how we can make this better.

cameronperera (approved):
LGTM!

ui/admin/app/components/form/target/index.hbs
```diff
@@ -1,122 +1,6 @@
...
-    <form.actions
-      @disabled={{if @model.cannotSave @model.cannotSave}}
-      @enableEditText={{t 'actions.edit-form'}}
-      @submitText={{t 'actions.save'}}
-      @cancelText={{t 'actions.cancel'}}
-    />
-  {{/if}}
-</Rose::Form>
\ No newline at end of file
+{{component
```
calcaide, line 1: Asking without context:

I see `targets/ssh/index.hbs` and `targets/tcp/index.hbs` being the same except for the reference to the translation within the label and helper text in example: `resources.target.types.ssh`.

As per my understanding both `ssh` and `tcp` have the exact same fields?

My question is, due to that small difference mentioned above, do we want to use difference templates or we can use one template file?

ui/admin/app/templates/scopes/scope/targets/index.hbs
```diff
@@ -37,37 +54,74 @@
...
+                    @text={{target.id}}
+                    @buttonText={{t 'actions.copy-to-clipboard'}}
+                    @acknowledgeText={{t 'states.copied'}}
+                  >
+                    <code>{{target.id}}</code>
+                  </Copyable>
+                </row.cell>
+              </body.row>
+            {{else}}
+              {{#if (eq target.type 'tcp')}}
```
calcaide, line 91: If `feature-flag 'ssh-target'` is false, and `target.type` is not equal to `tcp`. What we will show here? and empty cell?

ui/admin/tests/acceptance/targets/create-test.js
```diff
@@ -52,16 +53,26 @@ module('Acceptance | targets | create', function (hooks) {
...
     assert.expect(1);
     const count = getTargetCount();
-    await visit(urls.newTarget);
+    await visit(urls.newTCPTarget);
+    await fillIn('[name="name"]', 'random string');
+    await click('[type="submit"]');
+    assert.equal(getTargetCount(), count + 1);
+  });
+
+  test('can create new targets of type SSH', async function (assert) {
```
calcaide, line 72: SSH Target creation should be available JUST if the feature flag is enabled right? If so, how this test is passing without activating such feature?

calcaide (changes requested):
I left some comments.

calcaide:
> Overall looks great. I believe @calcaide mentioned something similar to this before but I would love to see if we can refactor the `form/target/[tcp|ssh]/index.hbs` form. From a glance, much of it looks like duplication that we could potentially pull out into a `form/target/base.hbs` maybe.

@cameronperera I add this comment regarding the code duplication between these two: https://github.com/hashicorp/boundary-ui/pull/1027/files#r837785763

ui/admin/tests/acceptance/targets/create-test.js
DhariniJeeva: ssh-target feature flag is enabled for [dev env](https://github.com/hashicorp/boundary-ui/blob/main/ui/admin/config/environment.js#L138).

ui/admin/app/templates/scopes/scope/targets/index.hbs
DhariniJeeva: yeah, we don't want to show ssh type if the `feature-flag 'ssh-target` is set to false. So it will be an empty cell if there are no TCP type.

ui/admin/tests/acceptance/targets/create-test.js
DhariniJeeva: do you mean in prod? 🤔

calcaide: Ahh ok that makes sense.

I was thinking on using this within the test: `enableFeature('ssh-target');` So the test does not depend on the feature being enabled.

But for now I think you are right, better keep it on the feature flag :)

ui/admin/app/templates/scopes/scope/targets/index.hbs
randallmorey: Do we have an `isTCP` computed property?  We generally prefer them over equality checks in the template.

calcaide: Gotcha :ok_hand:

addons/core/translations/resources/en-us.yaml
```diff
@@ -278,6 +278,9 @@ target:
   types:
     tcp: TCP
     ssh: SSH
+  help: 
+    tcp: This target belongs to an TCP host catalog and will use TCP to connect
```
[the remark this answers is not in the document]
randallmorey: I'd do a quick review to see how we're handling this (if we're handling this) for other cases.  Do other "types" have descriptions or help text?

ui/admin/app/components/form/target/index.hbs
randallmorey: We have been distinguishing by type.  I'm not opposed to sharing a single component if they are the same, but they should still be separated from the "index" component, since future types may indeed be different.  The problem might be naming.  What would we call it?  It's not a "base" because future types that are in fact different aren't really a subclass or child.  Something to think about.

randallmorey: Also wondering if it should be an additional layer to achieve simplicity in the index component.  e.g. index -> ssh -> "generic" (rather than index -> "generic")

cameronperera: I am a fan of "generic" and I like the index -> ssh -> approach. My assumption is we would use this component in the `targets/ssh/index.hbs` and just pass the more dynamic parts like translations.

ui/admin/app/templates/scopes/scope/targets/index.hbs
DhariniJeeva: @randallmorey good idea! I'll check that.

addons/core/translations/resources/en-us.yaml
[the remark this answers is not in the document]
DhariniJeeva: we do have a `help` in [host-catalog](https://github.com/hashicorp/boundary-ui/blob/main/addons/core/translations/resources/en-us.yaml#L87)

ui/admin/app/templates/scopes/scope/targets/index.hbs
DhariniJeeva: updated!

ui/admin/app/components/form/target/index.hbs
DhariniJeeva: I went with `details.hbs` and followed the suggested index->ssh->details. feedback is much appreciated!

ui/admin/app/components/form/target/ssh/index.hbs
```diff
@@ -0,0 +1,7 @@
+{{component
+  (concat 'form/target/details')
```
cameronperera, line 2: Could you explain what this line is doing exactly please? I am not entirely sure why `concat` is being called here.

DhariniJeeva: this has been updated to invoke the component directly!

addons/core/translations/resources/en-us.yaml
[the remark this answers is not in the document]
DhariniJeeva: For now, this will live here but will be updated when we change `types` to have a nested structure.

calcaide (approved):
LGTM

PR #1038, opened 2022-03-28
feat: 🎸 update target route to support ssh type

DhariniJeeva:
:tickets: [Jira ticket](https://hashicorp.atlassian.net/browse/ICU-3655)

## Description

:technologist: [Admin preview](https://boundary-ui-92vznttxr-hashicorp.vercel.app/scopes/s_xk1i4dk50v2/targets/new)

update existing target route to include type query parameter and make TCP the default type.

### DIFF
--- a/ui/admin/app/routes/scopes/scope/targets/new.js
+++ b/ui/admin/app/routes/scopes/scope/targets/new.js
@@ -3,14 +3,24 @@ import Route from '@ember/routing/route';
 export default class ScopesScopeTargetsNewRoute extends Route {
   // =methods
 
+  // =attributes
+  queryParams = {
+    type: {
+      refreshModel: true,
+    },
+  };
   /**
    * Creates a new unsaved target in current scope.
    * @return {TargetModel}
    */
-  model() {
+
+  model(params) {
     const scopeModel = this.modelFor('scopes.scope');
+    if (!params.type) {
+      params.type = 'tcp'; //tcp is the default type
+    }
     return this.store.createRecord('target', {
-      type: 'tcp',
+      type: params.type,
       scopeModel,
     });
   }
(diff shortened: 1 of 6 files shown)


ui/admin/app/routes/scopes/scope/targets/new.js
```diff
@@ -3,14 +3,24 @@ import Route from '@ember/routing/route';
...
    * Creates a new unsaved target in current scope.
    * @return {TargetModel}
    */
-  model() {
+
+  model(params) {
     const scopeModel = this.modelFor('scopes.scope');
+    if (!params.type) {
+      params.type = 'tcp'; //tcp is the default type
+    }
```
cameronperera, line 21: Are we able to test this functionality?

randallmorey: I might create a new local variable too, rather than assigning into the params object.  We receive this object from Ember internals.  Probably there's no harm but good to be safe.

randallmorey: @cameronperera the acceptance test could attempt to navigate to the new route with no type, then check that the TCP form is displayed.

DhariniJeeva: @cameronperera the display test case is covered in [PR](https://github.com/hashicorp/boundary-ui/pull/1027) and we could discuss refactoring the CRUD test files offline tomorrow.

DhariniJeeva: discussed offline and @cameronperera will be adding a few more test cases to ssh target CRUD work in a separate PR.

cameronperera (approved):

calcaide (approved):

PR #1043, opened 2022-04-05
chore: 🤖 Add build admin UI workflow

calcaide:
Add build admin UI github actions workflow.

**Assumption**
This github action triggers after merging with main. It assumes before merging, the worflow `validate-monorepo` (compliance, lint and tests) has run. So no need to run again such workflow, but open to discussion if you think otherwise.

:tickets: [Jira ticket](https://hashicorp.atlassian.net/browse/ICU-3849)


.github/workflows/build-admin-ui.yaml
```diff
@@ -0,0 +1,29 @@
...
+    steps:
+      - uses: actions/checkout@v3
+      - uses: actions/cache@v3
+        id: node-modules-cache
+        with:
+          path: '**/node_modules'
+          key: node-modules-${{ hashFiles('**/yarn.lock') }}
+          restore-keys: |
+            node-modules-
+      - uses: actions/setup-node@v3
```
DhariniJeeva, line 20: this is not a blocker and can be addressed later. I wonder if we should update our setup-node version [here](https://github.com/hashicorp/boundary-ui/blob/main/.github/workflows/monorepo-validate.yaml#L55) as well?

DhariniJeeva:
LGTM!

DhariniJeeva (approved):

.github/workflows/build-admin-ui.yaml
calcaide: @DhariniJeeva You got it! Nice catch! It was on the toDo list but you are right, there is no reason for it to be included on this PR.
I changed to the V3 in all the actions, they release it when I started testing this 

Nice catch!

cameronperera (approved):
LGTM

PR #1044, opened 2022-04-05
Chore update babel parser

calcaide:
As part of updating `eslint` to `v8` we need first to update deprecated `babel-eslint` in favour of `@babel/eslint-parser`.

More information about the deprecation [here](https://github.com/babel/babel-eslint).

The approach used in this PR is to NOT use a babel config file `.babelrc`, but happy to discuss a change on that.


addons/api/.eslintrc.js
```diff
@@ -2,13 +2,17 @@
 
 module.exports = {
   root: true,
-  parser: 'babel-eslint',
+  parser: '@babel/eslint-parser',
   parserOptions: {
     ecmaVersion: 2018,
     sourceType: 'module',
     ecmaFeatures: {
```
calcaide, line 9: @randallmorey 
I think this is legacy from the parser we are deprecating. After we merge this, I would like to perform some testing to triple check before deleting it.

addons/api/.eslintrc.js
```diff
@@ -2,13 +2,17 @@
...
       legacyDecorators: true,
     },
+    requireConfigFile: false,
```
calcaide, line 12: This is the approach to NOT have to use a specific babel config file. Because we are just adding 1 plugin I think is not enough to justify such file. But happy to discuss if you think it deserves to have a babel config file aka `.babelrc`

addons/api/.eslintrc.js
```diff
@@ -2,13 +2,17 @@
...
+    babelOptions: {
+      plugins: [['@babel/plugin-proposal-decorators', { legacy: true }]],
```
calcaide, line 14: This is need due to the usage of decorators (ember uses them a lot).

addons/api/.eslintrc.js
calcaide: Never mind my previous comment. As per discussion on FE tech time through zoom we will try to deprecate this old code within this PR.

cameronperera (approved):
LGTM. Thanks for the explanation on the reason to not have a babel config.

DhariniJeeva (approved):

calcaide:
> LGTM. Thanks for the explanation on the reason to not have a babel config.

@cameronperera Thank you for asking :raised_hands:

PR #1045, opened 2022-04-07
Chore update eslint

calcaide:
**Beware: this PR depends on:** [Chore update babel parser](https://github.com/hashicorp/boundary-ui/pull/1044) and will not be merged until the dependency is resolved.

Updates `eslint` and `eslint-plugin-ember` to the latest version.

:tickets: [Jira ticket](ICU-3596)


cameronperera (approved):
LGTM

ui/admin/app/controllers/scopes/scope/scopes.js
```diff
@@ -12,7 +12,7 @@ export default class ScopesScopeScopesController extends Controller {
   /**
    * @type {string}
    */
-  @computed('model.currentScope.{isGlobal,isOrg}')
+  @computed('intl', 'model.currentScope.{isGlobal,isOrg}')
```
ZedLi, line 15: Still trying to understand ember but can you explain what this change is for? Is it so the getter can also use `intl`?

randallmorey: Did eslint complain about not declaring `intl` as a dependent key?

addons/api/addon/serializers/account.js
```diff
@@ -33,7 +33,7 @@ export default class AccountSerializer extends ApplicationSerializer {
    * `attributes` entirely on update.
    */
   serializeOIDC(snapshot) {
-    const { isNew } = snapshot?.record;
+    const { isNew } = snapshot?.record || null;
```
DhariniJeeva, line 36: what's the reason behind adding this `null` value here?  Did it error out during this update?

calcaide: @DhariniJeeva
This error come with the update, as a side effect of this update also updating the `eslint-ember-plugin` and the recommended rules for JS.

The exact rules that affects this is `no-unsafe-optional-chaining`, more [info about here](https://eslint.org/docs/rules/no-unsafe-optional-chaining).

In a nutshell, if `snapshot.record` does not exist, `isNew` will not be destructured and it will have a `undefined` value. To avoid that I thought best will be setting the value to null as fallback.
But happy to discuss other options here :smiley:

ui/admin/app/controllers/scopes/scope/scopes.js
calcaide: @ZedLi @randallmorey 

This error comes with the update of `eslint-plugin-ember` and the update of the recommended rules.

As Randy mentioned, this change is due to eslint complaining about dependencies to be declared statically in computed properties. [More info about this rul in here](https://github.com/ember-cli/eslint-plugin-ember/blob/master/docs/rules/require-computed-property-dependencies.md).

Happy to discuss other approaches :smiley:

ZedLi (approved):
LGTM!

addons/api/addon/serializers/account.js
randallmorey:
```suggestion
    const { isNew } = snapshot?.record || {};
```

addons/api/addon/serializers/account.js
```diff
@@ -70,7 +70,7 @@ export default class AccountSerializer extends ApplicationSerializer {
     }
     // Change password custom method
     if (snapshot?.adapterOptions?.method === 'change-password') {
-      const { currentPassword, newPassword } = snapshot?.adapterOptions;
+      const { currentPassword, newPassword } = snapshot?.adapterOptions || null;
```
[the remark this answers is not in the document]
calcaide: That is correct @DhariniJeeva !! Updated the PR with that suggestion. Nice catch! :female_detective:

DhariniJeeva (approved):
Awesome work! LGTM!

PR #1048, opened 2022-04-08
Feat add build admin UI workflow

calcaide:
Add build admin UI github actions workflow.

**Assumption**
This github action triggers after merging with main. It assumes before merging, the worflow `validate-monorepo` (compliance, lint and tests) has run. So no need to run again such workflow, but open to discussion if you think otherwise.

:tickets: [Jira ticket](https://hashicorp.atlassian.net/browse/ICU-3849)


DhariniJeeva (approved):

cameronperera (approved):
LGTM! Thanks for the explanation offline.

PR #1051, opened 2022-04-11
feat: 🎸 Add mirage data for the worker_filter

ZedLi:
✅ Closes: ICU-3888

:tickets: [Jira ticket]()

## Description

This adds the mirage data for the `worker_filter` attribute

### DIFF
--- a/addons/api/mirage/factories/target.js
+++ b/addons/api/mirage/factories/target.js
@@ -25,6 +25,7 @@ export default factory.extend({
    * -1 means "unlimited" and we want to generate these on occasion.
    */
   session_connection_limit: () => random.arrayElement([-1, datatype.number()]),
+  worker_filter: (i) => (i % 2 === 0 ? random.words() : null),
   type: (i) => types[i % types.length],
   /**
    * Generates attributes fields by type.
--- a/addons/api/mirage/generated/factories/target.js
+++ b/addons/api/mirage/generated/factories/target.js
@@ -9,6 +9,7 @@ export default Factory.extend({
   name: () => random.words(),
   session_max_seconds: () => datatype.number(),
   session_connection_limit: () => datatype.number(),
+  worker_filter: () => random.words(),
   description: () => random.words(),
   created_time: () => date.recent(),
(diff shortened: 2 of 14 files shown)


addons/api/mirage/generated/factories/target.js
```diff
@@ -9,6 +9,7 @@ export default Factory.extend({
   name: () => random.words(),
   session_max_seconds: () => datatype.number(),
   session_connection_limit: () => datatype.number(),
+  worker_filter: () => random.words(),
```
ZedLi, line 12: What's the difference between the usage of `addons/api/mirage/generated/factories/target.js` vs `addons/api/mirage/factories/target.js`? It seems like the former overrides/extends the latter?

addons/api/mirage/factories/target.js
```diff
@@ -25,6 +25,7 @@ export default factory.extend({
    * -1 means "unlimited" and we want to generate these on occasion.
    */
   session_connection_limit: () => random.arrayElement([-1, datatype.number()]),
+  worker_filter: (i) => (i % 2 === 0 ? random.words() : null),
```
ZedLi, line 28: Do we generally try and return different scenarios for the data?

addons/api/mirage/generated/factories/target.js
randallmorey: In a perfect world, we would automatically generate models and mocks from the open API schema definition.  The code is structured in a way to allow this, though no such automation yet exists.  This arrangement allows us to override the generated code if we need to _without directly editing the generated files_.

calcaide (approved):
Looks good to me. Awesome first PR @ZedLi

cameronperera (approved):
LGTM!

addons/api/mirage/factories/target.js
cameronperera: As of right now, I do not think it is used anywhere else in the UI. It is an optional field but we weren't sure if it would be worth while to verify that it shows up properly when it is present vs when it is left null.

PR #1053, opened 2022-04-13
feat: 🎸 Add worker filter field in target form

cameronperera:
✅ Closes: ICU-3891

:tickets: [Jira ticket](https://hashicorp.atlassian.net/browse/ICU-3891)

## Description

Add worker filter field in target form

:technologist: [Admin preview](https://boundary-ui-git-icu-3891-view-worker-filter-in-ccfc7b-hashicorp.vercel.app/scopes/s_xk1i4dk50v2/targets/1)

Screen Shot 2022-04-14 at 2 24 13 PM


addons/core/translations/form/en-us.yaml
```diff
@@ -41,6 +41,9 @@ session_max_seconds:
 session_connection_limit:
   label: Maximum Connections
   help: The maximum number of connections allowed per session.  For unlimited, specify "-1".
+worker_filter:
+  label: Worker Filter
+  help: Target filters are regular expressions that reference worker tags.
```
cameronperera, line 46: I would love some input on what this `help` should be. I got this wording from the learn guide for [Target Aware Workers](https://learn.hashicorp.com/tutorials/boundary/target-aware-workers?in=boundary/configuration) but I am not 💯 on it. The [description](https://github.com/hashicorp/boundary-ui/blob/main/addons/api/addon/generated/models/target.js#L39) from the target model sounded a bit too technical for the form. Thanks for any input.

randallmorey: I think the form field description on the model is perfectly fine.  One thing that might help (because there's a lot to understand) is to link out to the docs from that field.  We have a pattern for that now, right?

cameronperera: Ok. I will update it to use the model description instead.

I was not aware we were doing that. Do you mean what we have [here](https://boundary-ui.vercel.app/scopes/s_xk1i4dk50v2/host-catalogs/new?type=azure) when creating a dynamic host catalog?
If so, I can add that here as well.

DhariniJeeva:
LGTM!

DhariniJeeva (approved):

ZedLi (approved):
LGTM!

addons/core/translations/form/en-us.yaml
cameronperera: @randallmorey `help` text updated and the learn more link has been added (see screenshot above).

PR #1065, opened 2022-04-20
fix: 🐛 Fix radio card background color in dark mode

ZedLi:
✅ Closes: ICU-4021

:tickets: [Jira ticket](https://hashicorp.atlassian.net/browse/ICU-4021)

## Description

The background color for the radio cards was also set to white, even for dark mode. 
Also looks like we had an unncessary `display` field

:technologist: [Admin preview](https://boundary-8tg3crtmr-hashicorp.vercel.app/)

:rose: [Rose preview](https://boundary-ui-storybook-fu9l1o0o1-hashicorp.vercel.app/)

Before:
image

After:
image

Still looks normal in light mode:
image


addons/rose/addon/styles/rose/components/form/_radio-card.scss
```diff
@@ -75,7 +75,6 @@ $radio-svg: (
   }
   .rose-form-radio-label {
     @include type.type(s, bold);
-    display: flex;
```
ZedLi, line 78: It looks like this `display` was always getting overriden by another `display: block` on line 82

randallmorey (changes requested):
What about the title/label?  And is the paragraph text too dark?

addons/rose/addon/styles/rose/components/form/_radio-card.scss
```diff
@@ -75,7 +75,6 @@ $radio-svg: (
...
     color: var(--black);
```
randallmorey, line 78: We shouldn't use absolute colors like this except in very specific cases.  Use relative poles instead (stark/subtle).

randallmorey (changes requested):

addons/rose/addon/styles/rose/components/form/_radio-card.scss
ZedLi: Wow my eyes glazed right over that, I removed it and it looks like it defaults to the body's color of `var(--stark)`, should this be explicit for the `color` and `background-color` or is it ok to inherit it?

randallmorey: I prefer inherited.  Less code.

calcaide (approved):
LGTM! Nice work

DhariniJeeva (approved):
LGTM!

addons/rose/addon/styles/rose/variables/color/_css.scss
```diff
@@ -123,9 +123,9 @@
   --ui-border-subtler-1: #{color(ui-gray, 800)};
   --ui-border: #{color(ui-gray, 700)};
 
-  --ui-gray-subtler-6: #{color(ui-gray, 900)};
-  --ui-gray-subtler-5: #{color(ui-gray, 800)};
-  --ui-gray-subtler-4: #{color(ui-gray, 700)};
+  --ui-gray-subtler-6: #{color(ui-gray, 950)};
+  --ui-gray-subtler-5: #{color(ui-gray, 900)};
+  --ui-gray-subtler-4: #{color(ui-gray, 800)};
```
ZedLi, line 128: I only made these ones darker in dark mode, I couldn't find good examples of the lighter ones being used so left them as is. Let me know if you think I should also adjust them.

cameronperera (approved):
LGTM

PR #1066, opened 2022-04-21
refactor: 💡 Encapsulate desktop client window creation

calcaide:
## Description
Encapsulate desktop client window creation to reuse the creation of new windows.

What this PR is **NOT**:
This PR is not aiming for any code improvement or bug fixing.

What this PR is:
A refactor to could reuse the code that create the main window.

What is the need for this PR:
Mainly fixing the `mac os` Bug when opening/closing the main window.

## Testing
I tested this change within `linux (amd64)`, `mac (arm64)` and `windows (amd64)`. Opening Boundary desktop client and closing it again using the X icon (default behaviour).
Be aware the bug in Mac OS stills there, but I tested the desktop client opens properly after this change.

## Explanation of this approach

Notice (I already mentioned above) the goal of this approach it is NOT to provide any code improvement or fixing a bug, but encapsulate the creation of the main window. In the future, we need to review `index.js` since it start to be not easy to mantain and we can encapsulate other parts of that file, but this approach it is not looking for that.

`createWindow()`:
- Receives: `partition` and `closeWindowCb`.
- Returns: a window object with all the window handlers already set.
- Goal: have a pure function that creates a window, set the handlers and does not spread side effects (so everything happen within the function with the parameters received).
- Why the callback?
  - As mentioned earlier, we want `createWIndow` to be a pure function. If we set the `mainWindow.on('close')` event out of the `createWindow` function we will spread side effects. And this encapsulation will not be valid, i.e. the PR that fixes the mac window lifecycle window will use `mainWindow` without having that event handler, unless you duplicate code.

## Other approaches:
I am happy to discuss other approaches, if that's your desire, please make sure first you read the full description of this PR and understand that the goal of it is not to improve code yet.


ui/desktop/electron-app/src/index.js
```diff
@@ -184,33 +108,118 @@ app.on('ready', async () => {
...
+  // Configure dev tools menu
+  const menuTemplate = Menu.buildFromTemplate(menu.generateMenuTemplate());
+  if (isDev) {
+    const view = menuTemplate.getMenuItemById('view');
+    view.submenu.append(new MenuItem({ role: 'toggleDevTools' }));
+  }
+  Menu.setApplicationMenu(menuTemplate);
+
+  // Close window callback
+  const closeWindowCB = () => {
```
calcaide, line 215: **IMPORTANT:**
I created this callback to keep the same behaviour than before, when `mainWindow` receives the close event, it equals `mainWindow` to `null`.

ui/desktop/electron-app/src/index.js
```diff
@@ -152,28 +82,22 @@ app.on('ready', async () => {
       'icon.png'
     );
 
-  mainWindow = new BrowserWindow(browserWindowOptions);
+  const win = new BrowserWindow(browserWindowOptions);
```
gsusmi, line 85:
```suggestion
  const browserWindow = new BrowserWindow(browserWindowOptions);
```

ui/desktop/electron-app/src/index.js
```diff
@@ -184,33 +108,118 @@ app.on('ready', async () => {
...
+    view.submenu.append(new MenuItem({ role: 'toggleDevTools' }));
+  }
+  Menu.setApplicationMenu(menuTemplate);
+
+  // Close window callback
+  const closeWindowCB = () => {
+    mainWindow = null;
+  };
+
+  mainWindow = createWindow(partition, closeWindowCB);
```
gsusmi, line 219: Rather than passing in this callback, is it possible to create event handling on `mainWindow` itself?

ui/desktop/electron-app/src/index.js
calcaide: I did not want it to call it `browserWindow` to avoid confussion with Electron API `BrowserWindow`, but since that was just a personal thought, performing this change.

ui/desktop/electron-app/src/index.js
calcaide: @gsusmi 
I updated the description of this PR to explain why we are using a callback. Let me know if that explanation is sufficient and if you are ok with it or want to explore a different approach.
Thank you.

ui/desktop/electron-app/src/index.js
```diff
@@ -152,28 +82,22 @@ app.on('ready', async () => {
...
-  mainWindow = new BrowserWindow(browserWindowOptions);
+  const browserWindow = new BrowserWindow(browserWindowOptions);
 
   // If the user-specified origin changes, reload the page so that
   // the CSP can be refreshed with the this source allowed
-  runtimeSettings.onOriginChange(() => mainWindow.loadURL(emberAppURL));
-
-  // If you want to open up dev tools programmatically, call
-  // mainWindow.openDevTools();
+  runtimeSettings.onOriginChange(() => win.loadURL(emberAppURL));
```
cameronperera, line 89:
```suggestion
  runtimeSettings.onOriginChange(() => browserWindow.loadURL(emberAppURL));
```
Should this `win` should be `browserWindow`?

ui/desktop/electron-app/src/index.js
```diff
@@ -184,33 +108,118 @@ app.on('ready', async () => {
...
-  mainWindow.on('closed', () => {
-    mainWindow = null;
+  browserWindow.on('closed', () => {
+    closeWindowCB();
+  });
+
+  return browserWindow;
+};
+
+let mainWindow = null;
```
ZedLi, line 143: Do we still need to keep and set this variable? It looks like we don't do anything with this afterwards after creating the window. If we remove it, we don't have to return the `browserWindow` from `createWindow` and the `closed` event can still be as it was before. 

I think this still encapsulates setting the event handlers for the main window, though if the goal is to have this be a pure function and return the window even if we don't do anything with it currently, I'm fine with this.

ZedLi (approved):
Thanks for the detailed explanation of this PR!

cameronperera (approved):
LGTM!

ui/desktop/electron-app/src/index.js
calcaide: Hey @ZedLi 

This is declaration and initialization of `mainWindow` variable, we set it to null on initialization and when we destroy the window. So it is not undefined.

The purpose of having `mainWindow` is that we can re-create the window when handling event `active`. More details about this in this PR.

That said, happy to discuss your approach over a zoom call? Not sure I am getting 100% of it and interested on hear more about it :smiley:

ui/desktop/electron-app/src/index.js
calcaide: Nice catch!! Updated PR with that fix.

ui/desktop/electron-app/src/index.js
ZedLi: Had a quick conversation, I forgot that the objective was to fix the mac os window lifecycle and the other PR uses the `mainWindow`, makes sense to me!

gsusmi (approved):
LGTM. Ty!

PR #1069, opened 2022-04-25
feat: refresh sessions list view in desktop

DhariniJeeva:
:tickets: [Jira ticket](https://hashicorp.atlassian.net/browse/ICU-4023)

## Description
This pr adds a refresh button on the tool bar to allow users to manually refresh sessions data

:desktop_computer: [Desktop preview](https://boundary-ui-desktop-fjbmgr0u0-hashicorp.vercel.app/scopes/global/projects/sessions)

[storybook ](https://boundary-ui-storybook-n60p0s191-hashicorp.vercel.app/?path=/story/rose-loadingbutton--advanced)

Screen Shot 2022-04-25 at 11 10 18 AM


addons/rose/addon/components/rose/loading-button/index.js
```diff
@@ -0,0 +1,20 @@
...
+  @tracked isLoading = false;
+
+  @action
+  async toggleRefresh() {
+    this.isLoading = true;
+    try {
+      await this.args.onClick();
+      this.isLoading = false;
+    } catch (e) {
+      throw new Error('error refreshing sessions');
```
randallmorey, line 17: This button could be used to refresh any resource.  I'm also not sure if we should throw an error or just allow errors to occur.  Can we test what happens when network errors occur?

addons/rose/addon/styles/rose/components/toolbar/_toolbar.scss
```diff
@@ -1,16 +1,29 @@
...
+  display: flex;
   button {
     display: inline-block;
     font-weight: normal;
     vertical-align: middle;
   }
+  .rose-form {
+    display: flex;
+  }
+  .refresh-button {
```
randallmorey, line 21: If this lives in Rose, should it follow the same namespaced naming convention as other Rose components?

addons/rose/addon/styles/rose/components/toolbar/_toolbar.scss
```diff
@@ -1,16 +1,29 @@
...
+    color: var(--stark);
+    vertical-align: middle;
+    margin: auto ($xs + $xxxs) auto auto;
+  }
+  .loading {
```
randallmorey, line 26: Should this modifier class be nested under the associated component styles for clarity?

randallmorey (changes requested):
Since this is a Rose component, it should probably have an associated Storybook story.

addons/rose/addon/styles/rose/components/toolbar/_toolbar.scss
DhariniJeeva: moved!

addons/rose/addon/components/rose/loading-button/index.js
DhariniJeeva: Good point! I updated the `catch` to log the errors instead of throwing it as it made the state `isLoading` stuck and didn't let me refresh the sessions. This error(`transition aborted`) occurs when I repeatedly hit the refresh button. Now, when the error occurs, it logs to the console and the refresh functionality works as expected. Let me know if there's a better way to handle this!

cameronperera (approved):
Overall looks good. I did leave a couple of suggestions as well.

randallmorey:
Running prettier formatting might help.

addons/rose/addon/styles/rose/components/_loading-button.scss
```diff
@@ -0,0 +1,10 @@
+@use '../variables/sizing';
+
+.rose-refresh-button {
```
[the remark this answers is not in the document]
DhariniJeeva: ah, my bad! updated this to match the file name.

addons/rose/addon/components/rose/loading-button/index.js
```diff
@@ -0,0 +1,20 @@
...
+
+  @action
+  async toggleRefresh() {
+    this.isLoading = true;
+    try {
+      await this.args.onClick();
+    } catch (e) {
+      console.error('Error while loading data', e);
+    }
+    this.isLoading = false;
```
calcaide, line 18: Do we want to wait for line 14 code before flagging `this.isLoading` as `false?`

DhariniJeeva: yes! We have to wait for the onClick to happen. `this.isLoading` is what helps us indicate the state of the button - whether it is loading or ready to reload. And, if there's an error during the onClick event, we simply log the error and set the `this.isLoading` to false, which would allow the user to `refresh` the data, or it would be stuck at the `loading` state.  Hope this answers your question.

calcaide: Yes, this answer my question.

Then we will want the `this.loading=false` wrapped within a `.then()` statement, so we wait for the promise of `await this.args.onClick();` to be resolved.

cameronperera: I could be mistaken, but won't the `await` before `this.args.onClick()` prevent the code from continuing until that promise is resolved. I thought we would only need a `.then()` if this was not an `async` function.

calcaide: @cameronperera You are right. Being an `async` function behaves as you described.

@DhariniJeeva I was mistaken when reading this code. Resolving this comment and approving PR.

calcaide (approved):

addons/rose/addon/components/rose/loading-button/index.js
```diff
@@ -0,0 +1,20 @@
...
+  // =actions
+  @tracked isLoading = false;
+
+  @action
+  async toggleRefresh() {
+    this.isLoading = true;
+    try {
+      await this.args.onClick();
+    } catch (e) {
+      console.error('Error while loading data', e);
```
ZedLi, line 16: Do we generally do any sort of error handling for this? e.g. Displaying an alert that something failed?

PR #1078, opened 2022-04-28
feat: refresh sessions view in admin ui

DhariniJeeva:
:tickets: [Jira ticket](https://hashicorp.atlassian.net/browse/ICU-4136)

## Description

This pr adds a refresh button to manually refresh sessions data for admin ui

:technologist: [Admin preview](https://boundary-pc3uwxnt4-hashicorp.vercel.app/scopes/s_xk1i4dk50v2/sessions)


ZedLi (approved):
LGTM!

calcaide (approved):
LGTM! Good work :ok_hand:
Left a comment but not blocking :wink:

ui/admin/app/routes/scopes/scope/sessions.js
```diff
@@ -116,4 +117,13 @@ export default class ScopesScopeSessionsRoute extends Route {
   clearAllFilters() {
     this.status = [];
   }
+
+  /**
+   * refreshes session data.
+   */
+  @action
+  @loading
+  async refreshSessions() {
```
calcaide, line 126: We are defining `async` for a reason? no context on `ember-loading` so apologize is the question is too obvious.

cameronperera (approved):

ui/admin/app/routes/scopes/scope/sessions.js
DhariniJeeva: I thought this function is a good candidate to use the [@loader helper](https://github.com/hashicorp/boundary-ui/tree/main/addons/core#loading-helper) in case it consumes a lot of time.

calcaide: Awesome! thanks for the info :smile:

randallmorey: Not sure if it would work, but could you call `refresh` directly?  The only requirement would be that it be annotated as an action.

```js
@action
refresh() {
}
```

randallmorey, line 126: Looks like this does not need to be `async`?

DhariniJeeva: updated!

ui/admin/app/routes/scopes/scope/sessions.js
```diff
@@ -116,4 +117,14 @@ export default class ScopesScopeSessionsRoute extends Route {
   clearAllFilters() {
     this.status = [];
   }
+
+  /**
+   * refreshes session data.
+   */
+
+  @action
+  @loading
```
randallmorey, line 126: I don't think we need the loading decorator here.  The call to refresh the route should trigger the loading state.

randallmorey (changes requested):

PR #1080, opened 2022-04-28
feat: 🎸 implement live relative datetime component

randallmorey:
This PR introduces a component to the `core` addon to display locale-formatted relative date times (e.g. "30 seconds ago").  This component accepts a single argument and outputs text only (no markup).  The output is live-updated once per second.

Example usage:

```hbs
<RelativeDatetimeLive @date={{this.dateObject}} />
```

**Why not implement this in Rose?**
Rose consists of design-system related components and is relatively simple and static.  Due to the complexity of the live relative datetime component, its reliance on a clock service, and the simplicity of its output, it is not a good fit for inclusion in Rose.


addons/core/tests/integration/components/relative-datetime-live-test.js
```diff
@@ -0,0 +1,36 @@
+import { module, test } from 'qunit';
+import { setupRenderingTest } from 'ember-qunit';
+import { render } from '@ember/test-helpers';
+import { hbs } from 'ember-cli-htmlbars';
+
+module('Integration | Component | relative-datetime-live', function (hooks) {
+  setupRenderingTest(hooks);
+
+  test('it renders a relative datetime', async function (assert) {
```
cameronperera, line 9: Do we not always use `assert.expect` at the beginning our tests? For example, I would expect to see `assert.expect(6)` here.

cameronperera (approved):
One question but looks great. Nice work!

addons/core/tests/integration/components/relative-datetime-live-test.js
randallmorey: Ah.  We should.

randallmorey: Fixed!

PR #1081, opened 2022-04-28
chore: update github actions validator to depend on previous jobs

randallmorey:
This PR updates the monorepo validation job dependencies in GitHub Actions to ensure faster failure.  For example, if linting fails, there is no need to also run tests.


calcaide (approved):
Nice!! makes sense to check everything else before tests since is the most processing cost task.

Nice improvement! thank you

ZedLi (approved):
Makes sense to me!

PR #1082, opened 2022-04-28
feat: 🎸 switch to new token type parameter

johanbrandhorst:
## Description

Boundary is deprecating the use of the token_type parameter to the Authenticate endpoint, instead replacing it with the new type parameter.

See https://github.com/hashicorp/boundary/commit/80b2c9230fa84605d9aaf7c862f780b50690a36e for more context.

Note that this change will break compatibility with pre-0.8.0 controllers, so we may want to wait with merging this. I wanted to create a PR so we don't forget. I would appreciate advice on how to best migrate to the new variable.


randallmorey:
Let's wait on this until the binary eliminates support, so as not to break compatibility too soon.

cameronperera (approved):
LGTM!

PR #1083, opened 2022-04-28
chore: 🤖 move loading-button component to core addon

randallmorey:
Moves the loading button component from rose to core.


DhariniJeeva:
Ty for moving. What about the styles?

randallmorey:
> Ty for moving. What about the styles?

Addressed in #1086

PR #1089, opened 2022-04-29
style: 💄 Update Refresh Button to animate when loading

cameronperera:
## Description
Add animation styles to refresh button while loading

:desktop_computer: [Desktop preview](https://boundary-ui-desktop-git-refresh-button-spinner-styles-hashicorp.vercel.app/scopes/global/projects/targets)

correct-loading-spin

### DIFF
--- a/addons/rose/addon/styles/rose/components/_icon.scss
+++ b/addons/rose/addon/styles/rose/components/_icon.scss
@@ -17,3 +17,19 @@
     vertical-align: text-top;
   }
 }
+
+.is-loading .rose-icon {
+  @keyframes hds-flight-icon-animation-rotation {
+    to {transform: rotate(360deg);}
+  }
+
+  > svg {
+    animation: hds-flight-icon-animation-rotation 9s linear infinite;
+  }
+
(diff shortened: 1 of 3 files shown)


addons/rose/addon/styles/rose/components/_icon.scss
```diff
@@ -17,3 +17,19 @@
     vertical-align: text-top;
   }
 }
+
+.is-loading .rose-icon {
```
randallmorey, line 21: I'm thinking this is not well-scoped enough to live here.  Can we move this into the applications for now?  It's sort of temporary until we fully adopt HDS.

randallmorey (changes requested):

addons/rose/addon/styles/rose/components/_icon.scss
cameronperera: Absolutely. I will move it to applications.

cameronperera: Fixed.

ZedLi (approved):
LGTM!

randallmorey (approved):

PR #1090, opened 2022-04-29
fix: 🐛 Desktop client not correctly loading ember app url

ZedLi:
## Description
Fix desktop client not correctly loading ember app url


cameronperera (approved):
Nice catch!

calcaide (approved):
Awesome catch!!

PR #1091, opened 2022-04-29
fix: 🐛 Reduced extra padding in tables

ZedLi:
✅ Closes: ICU-4046

:tickets: [Jira ticket](https://hashicorp.atlassian.net/browse/ICU-4046)

## Description

Reduced some extra padding in our tables.

The ticket also included fixing the extra cell borders but it looks like a safari rendering bug which I couldn't pin down the reason.
I looked at the safari changelog/breaking changes and the only thing that looked promising was the removal of some `-webkit-margin-collapse` properties but that doesn't seem relevant. 

To reproduce it:
- Be in Safari 15.4.x (It didn't appear to me in Safari 14.x.x), could be a minor version in 15 that is fine?
- I could only reproduce it on my laptop display, for some reason I couldn't reproduce it on an external monitor. This was independent of any resolution my browser was using

Let me know if you have any ideas or suggestions I should look into!

Before Admin: 
image

After Admin:
image

Before Desktop:
image

After Desktop:
image


joshuaogle (approved):
Looks good!

cameronperera (approved):
LGTM

randallmorey (approved):

PR #1095, opened 2022-04-29
refactor: 💡 Avoid lint warning

calcaide:
Avoid lint warning on styles

PR to adress lint warning `Class should be nested within its parent Class  force-element-nesting`.


cameronperera (approved):
LGTM!

randallmorey (approved):

ui/admin/app/styles/app.scss
```diff
@@ -219,18 +219,20 @@
...
 // Spinning animation for refresh button icon
-.is-loading .rose-icon {
-  @keyframes hds-flight-icon-animation-rotation {
-    to {transform: rotate(360deg);}
-  }
-
-  > svg {
-    animation: hds-flight-icon-animation-rotation 9s linear infinite;
-  }
+.is-loading {
```
randallmorey, line 222: Should all of this go within the media query?

randallmorey (changes requested):

ZedLi (approved):
Makes sense, I'm fine with this rule!

ui/admin/app/styles/app.scss
calcaide: @randallmorey With all of this you mean:
```css
.is-loading {
  media-query {
    rest of the code
  }
}
```

Or you mean the media query wrapping the whole thing (including `.is-loading`) ?

randallmorey: The latter.  The media query can wrap everything, since the only thing it does is trigger animation.  Therefore it is only appropriate for users who have not requested reduced motion.

calcaide: @randallmorey ok, now I am following you. I was missing the `no-preference` within the `@media` query.

As per my understanding the `@media` query is unnecessary. The `@media` query value is defined for when the user has `no-preference` for `prefers-reduced-motion`. In that case I think we got the same result defining the `animation-duration` directly to `0.7s`.

Were I could se value for this `@media` query is eliminating the animation if the user has a `prefers-reduced-motion: reduce`. But the goal of this PR was to just refactor code to be linter compliant.

The question is, do we want to eliminate this animation for the users that expressed their desire of remove or replace animations?

@cameronperera mentioning you for visibility :wink:

ZedLi: I took it as by default, most people will have `no-preference` and will set it to the fast animation of `.7s` and if anyone has it set to `reduce`, the media query won't run and it will use the default value of `9s` which is much slower.  Perhaps it would be easier to understand if the default was `.7s` and we set the media query to `reduce` and use the `9s` value?

But don't we need a media query for how fast the animation speed is depending on your preference? And the `keyframes` is what triggers the animation?

cameronperera (approved):
LGTM. Thanks for digging into this.

randallmorey (approved):

ZedLi (approved):

PR #1099, opened 2022-05-03
chore: 🤖 bump binary version and update changelog

cameronperera:
## Description

bump binary version and update changelog


randallmorey (changes requested):
The specified CLI version is pulled directly from `releases.hashicorp.com`.  Thus this PR should not be merged until that version is published.  Ideally the changelog date would also match the actual desktop release date.

calcaide (changes requested):
Same comment as Randy. I would change the date to when we are suppose to release the DC.

Good job!

randallmorey:
@cameronperera I converted to draft for now.  Let's revisit when we have a better idea on release date.

cameronperera:
> The specified CLI version is pulled directly from `releases.hashicorp.com`. Thus this PR should not be merged until that version is published. Ideally the changelog date would also match the actual desktop release date.

Thanks for that feedback @randallmorey and @calcaide. That is extremely helpful. Just to make sure I am clear on my understanding, I wouldn't update the CLI version in the `VERSION` file until we have a final CLI version released to `releases.hashicorp.com`?

randallmorey:
> Just to make sure I am clear on my understanding, I wouldn't update the CLI version in the `VERSION` file until we have a final CLI version released to `releases.hashicorp.com`?

Correct.  It's ok to create the PR but it shouldn't merge until the specified version is released.

ZedLi (approved):

PR #1101, opened 2022-05-04
docs: ✏️ Update docs for yarn scripts

calcaide:
## Description

Update yarn scripts documentation for all the projects within Boundary UI monorepo.

**What this PR is not:**
- Will not improve code apart from yarn scripts.
- Will not bring any new features or changes aside from yarn scripts. And the goal for yarn scripts it is just enhance what we actually have.

### addons/api:
- Update yarn scripts docs.
- Add `precommit` script. **Why?** Running the precommit script for the whole monorepo it takes time an effort, if we are working just within API, this could be useful as a check. _This will have to be run manully, no pre-hook will trigger this._

### addons/auth:
- Update yarn scripts docs.
- Delete scripts and docs related to `hsb`, this repositoy does not have any templates. JUST the dummy app template.
- Add `precommit` script. **Why?** Running the precommit script for the whole monorepo it takes time an effort, if we are working just within Auth, this could be useful as a check. _This will have to be run manully, no pre-hook will trigger this._

### addons/core:
- Update yarn scripts docs.
- Add `lint:hbs` this project contains `.hbs` files but was missing the linter.
- Because added `lint:hbs` had to ignore a linting rule in one template to unlock. We will fix it later.
- Add `precommit` script. **Why?** Running the precommit script for the whole monorepo it takes time an effort, if we are working just within Auth, this could be useful as a check. _This will have to be run manully, no pre-hook will trigger this._

### addons/rose:
- Update yarn scripts docs.
- Add `precommit` script. **Why?** Running the precommit script for the whole monorepo it takes time an effort, if we are working just within Auth, this could be useful as a check. _This will have to be run manully, no pre-hook will trigger this._


addons/api/README.md
```diff
@@ -47,21 +47,28 @@ List of available project commands.  `yarn run <command-name>`
...
 | start | Runs the dummy app local server. |
 | test | Runs all tests. |
-| test:all | Runs tests across multiple Ember versions with ember-try. |
+| test:ember-compatibility | Runs tests across multiple Ember versions with ember-try. |
+| precommit | Runs all lint and format. |
 | doc:toc | Automatically generates a table of contents for this README file. |
 
 Additional commands in the monorepo package may affect this projects.
 
 ## Linting
```
calcaide, line 65: Not sure about the purpose of this section, but I just updated it

addons/auth/package.json
```diff
@@ -10,14 +10,14 @@
...
-    "lint:hbs:fix": "ember-template-lint . --fix",
     "lint:js": "eslint .",
+    "lint:js:fix": "eslint . --fix",
     "format": "npm run format:js",
     "format:js": "prettier --write '{addon,app,config,stories,tests}/**/*.js' *.js",
-    "lint:js:fix": "eslint . --fix",
     "start": "ember serve",
     "test": "ember test",
     "test:ember-compatibility": "ember try:each",
+    "precommit": "yarn lint && yarn format && yarn test",
```
calcaide, line 20: Beaware these are NOT run using `npm-run-all` and they are NOT running in parallel fashion.

Happy to discuss/change this.

ZedLi, line 20: How would this `precommit` work? 
I was under the impression our root `package.json` has a `precommit` script which is just an npm/yarn hook based solely on the name (side note, does anyone use yarn 2/will we need to migrate? It looks like yarn 2 [won't support these pre/post hooks](https://yarnpkg.com/getting-started/migration#explicitly-call-the-pre-and-post-scripts))

Another question is I notice `yarn format` would run `prettier --write` but would that stage the files even if this script ran? Have we considered using actual git hooks to run `eslint --fix` and `prettier --write` and staging them so we don't have to worry about fixing styles and failing a commit if it wasn't properly formatted?

calcaide: @ZedLi 

Excellent question!

- We use commitizen with default behaviour, this is what manages the `yarn commit`.
- The `precommit` script within root `package.json` it is the only one triggered by the hook (commitizen `git-cz` explained above).
- What `precommit` does you can follow the breadcrumbs but in a nutshell:
  - runs `compliance:license`.
  - runs yarn lint, which thanks to [npm-run-all](https://www.npmjs.com/package/npm-run-all) runs all the `lint:*` yarn scripts defined within the root `package.json`.
  - runs yarn test, as explained above, runs all the `test:` yarn scripts defined within the root `package.json`.

So the hook just exists within the root of the project. And the magic of it based on [cz-cli](https://github.com/commitizen/cz-cli.).

Any `precommit` script within the submodules will not be triggered by any hooks.

The goal of providing individual `precommit` scripts per project is to facilitate the developer work when just applying changes to a specific project. In example, you have been working within `addons/api` and before running the `yarn commit` from the root of the project , which takes a bit of time, you want to run such checks first just for `addons/api`, which will be way faster.

calcaide: @ZedLi 

As per your second question:
Feel free to bring this discussion to FE tech time. As per my understanding and usage of the `yarn format`. What it does is "fix" the format issues but does not stage files for you, just fix the issues. Is responsibility of the developer to stage and manage the changes with git.
If you run locally `yarn commit` or `yarn precommit` before pushing a branch the CI should run the exact same tools. But it is true, that if you commit with a classic `git commit` and does not run the `yarn commit` then no hooks will run at all.

Maybe the scripts and tools available for the dev should be better documented (working on it) but this is an interesting topic I would like to discuss over FE tech time

addons/core/addon/components/form/authenticate/index.hbs
```diff
@@ -1,3 +1,5 @@
+{{!-- //TODO fix the issue and delete disabling the rule --}}
+{{!-- template-lint-disable no-passed-in-event-handlers --}}
```
calcaide, line 2: I had to skip this rule in this file, it did not like we are passing `@submit` as arguments, [more infor about it here](https://github.com/ember-template-lint/ember-template-lint/blob/master/docs/rule/no-passed-in-event-handlers.md)

addons/auth/package.json
ZedLi: It depends how we choose to setup the hooks I think, I know the `cloud-ui` repo is manual but I've setup husky in the past which has a `prepare` script (this might be different for yarn and wasn't even needed with earlier versions) which automatically setsup the hooks when you do an `npm install`. 

Generally that was "good enough" in that any new contributor would have the hooks setup when they install, but any current developer would have to still go through that install step either by manually setting them up or doing another `npm install`

randallmorey: Ah that makes a lot of sense.  We could do a `postinstall` script!

calcaide: I think this is very interesting, despite this PR started this discussion, it goes a bit out of the scope.
For the sake of this PR and since we already start on FE tech time talks about it, I will resolve this conversation.
Feel free to reopen if you think so.

addons/auth/package.json
```diff
@@ -10,14 +10,14 @@
   "scripts": {
     "lint": "npm-run-all --aggregate-output --continue-on-error --parallel 'lint:!(fix)'",
     "lint:fix": "npm-run-all --aggregate-output --continue-on-error --parallel lint:*:fix",
-    "lint:hbs:fix": "ember-template-lint . --fix",
     "lint:js": "eslint .",
+    "lint:js:fix": "eslint . --fix",
```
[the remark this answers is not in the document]
calcaide: @randallmorey I am not understanding your comment.

We have 2 yarn script. One for linting: `lint:*` and other for format: `format:*`. Linting is done by Eslint and format by Prettier.

ZedLi (approved):

DhariniJeeva:
Great work! LGTM!

PR #1103, opened 2022-05-05
chore: 🤖 Add lzma-native and appdmg to resolutions

cameronperera:
It would be great if a few people could pull this down and verify this has no side effects for running the app locally.

This change is to fix my ability to run `yarn install` for this project.

### DIFF
--- a/package.json
+++ b/package.json
@@ -96,6 +96,7 @@
     "mout": "^1.2.3",
     "nanoid": "^3.1.31",
     "plist": "^3.0.5",
+    "lzma-native": "^8.0.6",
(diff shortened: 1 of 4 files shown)


ZedLi (approved):
Worked for me in Ubuntu and macOS (11.6.4)

randallmorey (approved):
Works for me.

calcaide (approved):
As discussed on slack thread, approving this PR but will be good to perform a test run in our CI before merging.

Worked for me on M1 with Mac os `12.3.1` and node `16.15.0`

PR #1104, opened 2022-05-11
Refactor UI generated artifacts creation.

calcaide:
:tickets: [Jira ticket](https://hashicorp.atlassian.net/browse/ICU-4219)

## Description

Refactor UI generated artifacts creation. Now we generate 2 artifacts: `admin-ui-oss` and `admin-ui-enterprise`.

As per deprecation procedure, we will keep generating `admin-ui` per 2 more weeks after the merge of this PR.

## Discard approach:
We wanted to keep it simple, both UI's oss and enterprise use the exact same dependencies, the only difference is the build procedure.
Because that we decided to refactor the current build job adding two more steps. 
- Clean the /dist folder after OSS built.
- Build enterprise

## Approach:
After working on the first approach and discard it, this approach embrace splitting each build in a separate job but running parallel, so no ime is wasted.

## Screenshots:
This screenshot represent the last test build on the refactor branch.
Screenshot from 2022-05-11 11-09-38

### DIFF
--- a/.github/workflows/build-admin-ui.yaml
+++ b/.github/workflows/build-admin-ui.yaml
@@ -34,13 +34,32 @@ jobs:
           key: node-modules-${{ hashFiles('**/yarn.lock') }}
           restore-keys: |
             node-modules-
-      - uses: actions/setup-node@v3
+      - name: Build OSS version
+        uses: actions/setup-node@v3
         with:
           node-version: '16.x'
         if: steps.node-modules-cache.outputs.cache-hit != 'true'
-      - run: yarn run build:ui:admin
-      - name: Upload artifact
+      - run: yarn run build:ui:admin:oss
+      - name: Upload artifact admin-ui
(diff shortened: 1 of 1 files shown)


calcaide:
> 


> Looks great! My only feedback is that there may be an opportunity to run the OSS and enterprise builds as separate jobs in parallel. Not sure if that would be faster overall.

@randallmorey I explained on the description the Alternative approach, but you know what? I did not think on applying such approach with paralellism! I think you are right we might have no time difference. And to be honest it is more elegant to have these separate in 2 jobs. 
So I will refactor this to the alternative approach + parallelism!
Thanks for the catch!

randallmorey (approved):

cameronperera (approved):
LGTM! Nice work breaking it up so they can run in parallel.

DhariniJeeva (approved):

PR #1105, opened 2022-05-11
Fix Users page not updating fields correctly

ZedLi:
:tickets: [Jira ticket](https://hashicorp.atlassian.net/browse/ICU-4189)

## Description

:technologist: [Admin preview](https://boundary-eorexoroq-hashicorp.vercel.app)

Taken from the ticket:
https://user-images.githubusercontent.com/6604151/166251178-fea3b1e4-bcbc-4c40-ab90-8a9d5c46da01.mov

After:
user

### DIFF
--- a/addons/api/addon/models/user.js
+++ b/addons/api/addon/models/user.js
@@ -40,7 +40,7 @@ export default class UserModel extends GeneratedUserModel {
    * There are many options, so we fallback in order of precedence.
    * @type {string}
    */
-  @computed('email,full_name,login_name,displayName')
+  @computed('email', 'full_name', 'login_name')
   get accountName() {
     const { email, full_name, login_name } = this;
     return email || full_name || login_name;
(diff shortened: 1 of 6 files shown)


addons/api/addon/models/user.js
```diff
@@ -40,7 +40,7 @@ export default class UserModel extends GeneratedUserModel {
    * There are many options, so we fallback in order of precedence.
    * @type {string}
    */
-  @computed('email,full_name,login_name,displayName')
+  @computed('email', 'full_name', 'login_name')
```
randallmorey, line 43: This computed decorator may not be required at all in Ember Octane:  https://github.com/ember-cli/eslint-plugin-ember/blob/master/docs/rules/no-computed-properties-in-native-classes.md

It should automagically work the same if the decorator is removed thanks to tracked properties (model attributes are automatically tracked).

randallmorey (approved):

cameronperera (approved):
LGTM!

calcaide (commented):
LGTM!

PR #1107, opened 2022-05-12
fix: 🐛 Update validateOrigin to throw error on bad request

cameronperera:
✅ Closes: ICU-4238

:tickets: [Jira ticket](https://hashicorp.atlassian.net/browse/ICU-4238)

## Description

If the result from netRequestPromise is successful but a not a 200
response, validateOrigin will throw an error

:desktop_computer: [Desktop preview](https://boundary-ui-desktop-git-icu-4238-enhancement-u-b49f83-hashicorp.vercel.app/origin) Note: I do not think this is able to be replicated with mirage.

#### Before:
dc-before

#### After:
dc-after

### DIFF
--- a/ui/desktop/electron-app/src/models/runtime-settings.js
+++ b/ui/desktop/electron-app/src/models/runtime-settings.js
@@ -31,7 +31,10 @@ class RuntimeSettings {
     // test to see if the origin is API-compatible with the desktop client.
     const scopesEndpoint = `${origin}/v1/scopes`;
     try {
-      await netRequestPromise(scopesEndpoint);
+      const result = await netRequestPromise(scopesEndpoint);
+      if (result.statusCode >= 300) {
+        throw new Error();
+      }
     } catch (e) {
       throw new Error(`Origin ${origin} could not be validated.`);
     }
--- a/ui/desktop/electron-app/src/models/runtime-settings.js
+++ b/ui/desktop/electron-app/src/models/runtime-settings.js
@@ -32,7 +32,7 @@ class RuntimeSettings {
     const scopesEndpoint = `${origin}/v1/scopes`;
     try {
       const result = await netRequestPromise(scopesEndpoint);
-      if (result.statusCode >= 300) {
+      if (result.statusCode >= 400) {
         throw new Error();
       }
     } catch (e) {
--- a/ui/desktop/electron-app/src/models/runtime-settings.js
+++ b/ui/desktop/electron-app/src/models/runtime-settings.js
@@ -32,7 +32,9 @@ class RuntimeSettings {
     const scopesEndpoint = `${origin}/v1/scopes`;
     try {
       const result = await netRequestPromise(scopesEndpoint);
-      if (result.statusCode >= 400) {
+      // Redirects (3xx) are handled in the netRequest Promist, so we only
+      // throw an error for 4xx or 5xx responses.
+      if (result.statusCode > 399) {
         throw new Error();
       }
     } catch (e) {
--- a/ui/desktop/electron-app/src/models/runtime-settings.js
+++ b/ui/desktop/electron-app/src/models/runtime-settings.js
@@ -34,7 +34,7 @@ class RuntimeSettings {
       const result = await netRequestPromise(scopesEndpoint);
       // Redirects (3xx) are handled in the netRequest Promist, so we only
       // throw an error for 4xx or 5xx responses.
-      if (result.statusCode > 399) {
+      if (result.statusCode >= 400) {
         throw new Error();
       }
     } catch (e) {
--- a/ui/desktop/electron-app/src/models/runtime-settings.js
+++ b/ui/desktop/electron-app/src/models/runtime-settings.js
@@ -32,7 +32,7 @@ class RuntimeSettings {
     const scopesEndpoint = `${origin}/v1/scopes`;
     try {
       const result = await netRequestPromise(scopesEndpoint);
-      // Redirects (3xx) are handled in the netRequest Promist, so we only
+      // Redirects (3xx) are handled in the netRequest Promise, so we only
       // throw an error for 4xx or 5xx responses.
       if (result.statusCode >= 400) {
         throw new Error();


ui/desktop/electron-app/src/models/runtime-settings.js
```diff
@@ -31,7 +31,10 @@ class RuntimeSettings {
     // test to see if the origin is API-compatible with the desktop client.
     const scopesEndpoint = `${origin}/v1/scopes`;
     try {
-      await netRequestPromise(scopesEndpoint);
+      const result = await netRequestPromise(scopesEndpoint);
+      if (result.statusCode >= 300) {
```
ZedLi, line 35: Is there a reason `3xx` status codes should be included?

cameronperera: I was actually thinking the same thing. I wanted to look into this endpoint to verify but my thought was `/scopes` should never return a 3** response while a different API might actually return a 3** response. Otherwise, I am happy to change it to `>= 400`.

randallmorey: Would a request resolve in a 300 on our end?  Or would the redirect be followed?  I honestly don't know the answer.  It seems to me that if we received a 300 that is not or can't be followed, it should be treated as an error.

cameronperera: I have only dealt with redirect request a few times. From what I remember they have to be handled directly, they don't just work. So that is one reason for treating it as an error. I looked over the endpoint for `/scopes` and it doesn't seem to do any redirects.

cameronperera: > I also don't see why we would get a 3xx response on our end (unless maybe to force an https connection?)

To answer this question, it is not that we would ever receive a 3xx on our end but that an incorrect origin URL might have an actual endpoint called `/scopes` that could return a 3xx. I wanted to protect against that. That might be too small of a chance to worry about though.

To the actual redirect topic. Thanks for that link, that makes we wonder if we would even see that it is a 3xx response before the redirect happened. @randallmorey what are you thoughts on this?

ZedLi: If the request has never been cached, the `3xx` has to resolve and another request has to be made. So if a legitimate `3xx` is returned by an origin, it would always make another `GET`.

I think this is actually still the case even if the `3xx` is cached, I tested this by hitting `google.com` and getting a `301` to `www.google.com` and it will still make the subsequent request even if the `301` was cached (though the following `200` will also usually be cached)

Either way, I think it makes sense to not guard against a `3xx` since if it was legitimate for any reason, we would error out before the actual request is made.

ZedLi, line 35:
```suggestion
      if (result.statusCode >= 400) {
```

cameronperera: @ZedLi and I tested hitting an endpoint that we know returns a 3xx response. The `result` we get back will never be the 3xx response, only a 2xx, 4xx or 5xx. The 3xx response is handled in the `net-request-promise.js` file. It will keep redirecting until the request responds with anything other than a 3xx. This does mean that on the off-chance that `http://random.url.com/v1/scopes` exists and returns a 2xx response, it would give the user the "Something went wrong" screen. I would imagine this chance is extremely low. 
This is why we decided to change the `>=300` to `400`.

ZedLi: One other interesting thing to note about the `net` request is that it will follow redirects by default and there's a [redirect event](https://www.electronjs.org/docs/latest/api/client-request#event-redirect) if you did want to see the actual `3xx` requests. But otherwise it's as @randallmorey says, we won't see the `3xx` response directly.

randallmorey: > This does mean that on the off-chance that http://random.url.com/v1/scopes exists and returns a 2xx response, it would give the user the "Something went wrong" screen

Why would an existing endpoint responding with success in the 200 range result in an error?

> This is why we decided to change the >=300 to 400.
> we won't see the 3xx response directly.

Doesn't this mean there is a gap in our handling on the application side?  If, say, an unhandled 3xx redirect were to reach the application code, the result would be... no result?  I agree it's unlikely, but should we handle it as an error to cover all possible responses?

randallmorey: And the only reason I suggest this is because I don't know what guarantees we have about the 3xx range from `net`.  I mean, if the redirect is bad then presumably we'd get a 400+ response from the redirect target.  So I don't actually know how we'd ever see 3xx.  But I also don't know if it's _impossible_.

cameronperera: The 2xx response from the `/scopes` endpoint would not cause the "Something went wrong" screen, but the following endpoints that we call will fail which would cause the "Something went wrong" screen to show up.

> Doesn't this mean there is a gap in our handling on the application side? If, say, an unhandled 3xx redirect were to reach the application code, the result would be... no result? I agree it's unlikely, but should we handle it as an error to cover all possible responses?

I unfortunately do not have a solid answer for that. We should set up some time to chat about this as a team though.

calcaide: As per my understanding of redirections, if the response comes with a 3xx, will be the browser (chromium in this particular case) who will request again to the new location, and it will be the response to that second request what we will be handling.
So I think this code is correct. 

Out of the scope of this PR, but I do also think we might want to test a redirection here to see it in action?

randallmorey (approved):

ui/desktop/electron-app/src/models/runtime-settings.js
```diff
@@ -31,7 +31,12 @@ class RuntimeSettings {
     // test to see if the origin is API-compatible with the desktop client.
     const scopesEndpoint = `${origin}/v1/scopes`;
     try {
-      await netRequestPromise(scopesEndpoint);
+      const result = await netRequestPromise(scopesEndpoint);
+      // Redirects (3xx) are handled in the netRequest Promist, so we only
```
ZedLi, line 35:
```suggestion
      // Redirects (3xx) are handled in the netRequest Promise, so we only
```

ZedLi (approved):
LGTM!

ui/desktop/electron-app/src/models/runtime-settings.js
cameronperera: Good catch. Thanks

PR #1109, opened 2022-05-13
refactor: 💡 Update language to use Sign In and Sign Out

cameronperera:
✅ Closes: ICU-3941

:tickets: [Jira ticket](https://hashicorp.atlassian.net/browse/ICU-3941)

## Description

Update authenticate/deauthenticate language to sign in and sign out
respectively

:technologist: [Admin preview](https://boundary-ui-git-icu-3941-refactor-authenticate-0d4a1d-hashicorp.vercel.app/scopes/global/authenticate/auth-method-id-0)

:desktop_computer: [Desktop preview](https://boundary-ui-desktop-git-icu-3941-refactor-auth-728899-hashicorp.vercel.app/scopes/global/authenticate/auth-method-id-0)

Screen Shot 2022-05-13 at 4 59 46 PM
Screen Shot 2022-05-13 at 4 59 29 PM

### DIFF
--- a/addons/core/addon/components/form/authenticate/oidc/index.hbs
+++ b/addons/core/addon/components/form/authenticate/oidc/index.hbs
@@ -7,7 +7,7 @@
 >
 
   <form.actions
-    @submitText={{t "actions.authenticate"}}
+    @submitText={{t "actions.sign-in"}}
     @showCancel={{false}} />
 
 </Rose::Form>
--- a/addons/core/addon/components/form/authenticate/oidc/index.hbs
+++ b/addons/core/addon/components/form/authenticate/oidc/index.hbs
@@ -7,7 +7,7 @@
 >
 
   <form.actions
-    @submitText={{t "actions.sign-in"}}
+    @submitText={{t "actions.authenticate"}}
     @showCancel={{false}} />
 
 </Rose::Form>
--- a/addons/core/addon/components/form/authenticate/password/index.hbs
+++ b/addons/core/addon/components/form/authenticate/password/index.hbs
@@ -22,7 +22,7 @@
     @label={{t "form.password.label"}} />
 
   <form.actions
-    @submitText={{t "actions.sign-in"}}
+    @submitText={{t "actions.authenticate"}}
     @showCancel={{false}} />
 
 </Rose::Form>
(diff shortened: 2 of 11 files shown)


cameronperera:
We use the language "authenticate" in a few places that I was not entirely sure if we should change them. So I will link them below and I would like some feedback on them. Some of these I am not entirely certain how to trigger to verify.
[Error: Authentication Failed](https://github.com/hashicorp/boundary-ui/blob/ICU-3941-refactor-authenticate-deauthenticate-to-sign-in-sign-out-within-the-u-is/addons/core/translations/errors/en-us.yaml#L19)
[Form: multiple places](https://github.com/hashicorp/boundary-ui/blob/ICU-3941-refactor-authenticate-deauthenticate-to-sign-in-sign-out-within-the-u-is/addons/core/translations/form/en-us.yaml#L70-L90)
[Core Translations: multiple places](https://github.com/hashicorp/boundary-ui/blob/ICU-3941-refactor-authenticate-deauthenticate-to-sign-in-sign-out-within-the-u-is/addons/core/translations/en-us.yaml#L4-L11)
[Resources: Many lines](https://github.com/hashicorp/boundary-ui/blob/ICU-3941-refactor-authenticate-deauthenticate-to-sign-in-sign-out-within-the-u-is/addons/core/translations/resources/en-us.yaml#L20)

randallmorey (commented):
Internally, we use `authenticate` and `deauthenticate` identifiers.  This is part of the API and ultimately became baked into the frontend code.  My question is, should we leave the translation key as it was and just change the text?  Keys do not necessarily need to map exactly into text.  `authenticate` -> `'Sign In'` may be acceptable.  It has the benefit of not introducing another name for an existing concept.

cameronperera:
> Internally, we use `authenticate` and `deauthenticate` identifiers. This is part of the API and ultimately became baked into the frontend code (in actions, tests, mocks, etc). My question is, should we leave the translation key as it was and just change the text? Keys do not necessarily need to map exactly into text. `authenticate` -> `'Sign In'` may be acceptable. It has the benefit of not introducing another name for an existing concept.

That is a good question. I am happy with leaving the key as `authenticate` to reduce adding more names for an existing concept.

randallmorey (approved):

ZedLi (approved):

DhariniJeeva (approved):
LGTM!

calcaide (approved):
LGTM! Nice work!!
Also I like the idea of leaving the keys being `authenticate` and `deauthenticate`. As Randy mentioned, siunce we inherit the language from BE this is very clear :smiley:

PR #1112, opened 2022-05-19
fix: 🐛 Use more common hasOwnProperty and don't modify payload

ZedLi:
## Description

Use `hasOwnProperty` instead of `hasOwn`. The latter only recently received support in some browsers such as Safari, take a look [here](https://caniuse.com/mdn-javascript_builtins_object_hasown). Probably too early to use as of right now.

Also stopped modifying the payload directly.

### DIFF
--- a/addons/api/addon/serializers/application.js
+++ b/addons/api/addon/serializers/application.js
@@ -291,21 +291,23 @@ export default class ApplicationSerializer extends RESTSerializer {
     id,
     requestType
   ) {
+    const newPayload = copy(payload, true);
+
     primaryModelClass.attributes.forEach((attribute) => {
       const {
         name,
         options: { readOnly },
       } = attribute;
 
-      if (readOnly && !Object.hasOwn(payload, name)) {
-        payload[name] = null;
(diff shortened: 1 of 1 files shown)


addons/api/addon/serializers/application.js
```diff
@@ -291,21 +291,23 @@ export default class ApplicationSerializer extends RESTSerializer {
...
+
     primaryModelClass.attributes.forEach((attribute) => {
       const {
         name,
         options: { readOnly },
       } = attribute;
 
-      if (readOnly && !Object.hasOwn(payload, name)) {
-        payload[name] = null;
+      if (readOnly && !Object.prototype.hasOwnProperty.call(newPayload, name)) {
```
ZedLi, line 302: If you're wondering why I don't just call `hasOwnProperty` on the object directly, it's to satisfy this [eslint rule](https://eslint.org/docs/rules/no-prototype-builtins)

randallmorey (approved):

cameronperera (approved):

DhariniJeeva (approved):

PR #1115, opened 2022-05-19
fix: 🐛 filter out scopes with no auth methods

DhariniJeeva:

## Description

### DIFF
--- a/ui/admin/app/controllers/scopes/scope/authenticate.js
+++ b/ui/admin/app/controllers/scopes/scope/authenticate.js
@@ -5,12 +5,22 @@ export default class ScopesScopeAuthenticateController extends Controller {
 
   /**
    * Moves the global scope to index 0.
+   * filter out the scopes which has no authenticatable auth methods
    * @type {Array}
    */
+
   get sortedScopes() {
-    return [
-      ...this.model.scopes.filter((scope) => scope.id === 'global'),
-      ...this.model.scopes.filter((scope) => scope.id !== 'global'),
-    ];
+    let filteredScopesIdsWithAuthMethods;
+    this.model.authenticatableAuthMethodsList.map((authMethod) => {
+      if (authMethod.content?.length > 0) {
+        filteredScopesIdsWithAuthMethods = this.model.scopesIdList.filter(
+          (i) => i === authMethod.content.query.scope_id
+        );
+      }
+    });
+
+    return this.model.scopes.filter(
+      (scope) => filteredScopesIdsWithAuthMethods?.indexOf(scope.id) > -1
+    );
   }
 }
--- a/ui/admin/app/routes/scopes/scope/authenticate.js
+++ b/ui/admin/app/routes/scopes/scope/authenticate.js
@@ -1,6 +1,6 @@
 import Route from '@ember/routing/route';
 import { inject as service } from '@ember/service';
-import { hash } from 'rsvp';
+import { hash, all } from 'rsvp';
 
 export default class ScopesScopeAuthenticateRoute extends Route {
   // =services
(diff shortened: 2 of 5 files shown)


ui/admin/app/templates/scopes/scope/authenticate.hbs
```diff
@@ -75,19 +78,6 @@
           {{#if @model.authMethods}}
             {{outlet}}
           {{/if}}
-          {{#unless @model.authMethods}}
```
DhariniJeeva, line 78: I removed the message which is displayed when there are no auth methods as it is not needed anymore.

ui/admin/app/controllers/scopes/scope/authenticate.js
```diff
@@ -5,12 +5,22 @@ export default class ScopesScopeAuthenticateController extends Controller {
...
    */
+
   get sortedScopes() {
-    return [
-      ...this.model.scopes.filter((scope) => scope.id === 'global'),
-      ...this.model.scopes.filter((scope) => scope.id !== 'global'),
-    ];
+    let filteredScopesIdsWithAuthMethods;
+    this.model.authenticatableAuthMethodsList.map((authMethod) => {
+      if (authMethod.content?.length > 0) {
```
DhariniJeeva, line 15: unsure if this can be accessed like this. suggestions are welcome!

ZedLi:
I know this is in draft but I noticed I'm getting a `500` on a `GET` to `http://localhost:9200/v1/auth-methods?filter[authorized_actions]=authenticate&scope_id=o_1234567890` when I visit the login screen

image

DhariniJeeva:
> http://localhost:9200/v1/auth-methods?filter[authorized_actions]=authenticate&scope_id=o_1234567890

ahh, thank you! `authenticate` should live in an array. fixed now!

ui/admin/app/routes/scopes/scope/authenticate.js
```diff
@@ -20,13 +20,34 @@ export default class ScopesScopeAuthenticateRoute extends Route {
...
+      })
+    );
+
     return hash({
       scope: this.modelFor('scopes.scope'),
       scopes: this.modelFor('scopes').filter((scope) => scope.isOrg),
       authMethods: this.store.query('auth-method', { scope_id }),
-
+      scopesIdList,
+      authenticatableAuthMethodsList,
```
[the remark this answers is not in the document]
DhariniJeeva: I have it on the controller side because I want to filter/sort the model data after it has been loaded, if this were to exist in the route, then the user would have to refresh the route to get the filtered data.

ui/admin/app/routes/scopes/scope/authenticate.js
```diff
@@ -20,13 +20,34 @@ export default class ScopesScopeAuthenticateRoute extends Route {
...
+              filter: { authorized_actions: 'authenticate' },
+            })
+          );
+        }
+      })
+    );
+
     return hash({
       scope: this.modelFor('scopes.scope'),
       scopes: this.modelFor('scopes').filter((scope) => scope.isOrg),
```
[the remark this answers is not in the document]
DhariniJeeva: removed!

ui/admin/app/routes/scopes/scope/authenticate.js
```diff
@@ -20,13 +20,27 @@ export default class ScopesScopeAuthenticateRoute extends Route {
...
+    let authMethodsScopeIds = [];
+    //filter the scopes with no auth-methods
+    authMethodsOfAllScopes.map(({ id: auth_id }) => {
+      authMethodsScopeIds.push(
+        this.store.peekRecord('auth-method', auth_id).scopeID
+      );
+    });
+
+    scopes = scopes.filter((scope) => authMethodsScopeIds.includes(scope.id));
+    const authMethods = this.store.query('auth-method', { scope_id });
```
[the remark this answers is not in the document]
DhariniJeeva: ah, I didn't know we could use `filterby` here! thank you!

randallmorey (approved):

ui/admin/app/routes/scopes/scope/authenticate.js
```diff
@@ -20,13 +20,26 @@ export default class ScopesScopeAuthenticateRoute extends Route {
    * scope and all scopes (for org navigation).
    * @return {Promise} `{scope, scopes, authMethods}`
    */
-  model() {
+  async model() {
     const { id: scope_id } = this.modelFor('scopes.scope');
+    // Fetch auth methods for the current scope
+    const authMethods = await this.store.query('auth-method', { scope_id });
+    // Preload all auth methods into the store
+    await this.store.query('auth-method', {
```
randallmorey, line 28: Forgot to filter this for authenticatable auth methods.

randallmorey (changes requested):

ui/admin/app/routes/scopes/scope/authenticate.js
```diff
@@ -20,13 +20,32 @@ export default class ScopesScopeAuthenticateRoute extends Route {
    * scope and all scopes (for org navigation).
    * @return {Promise} `{scope, scopes, authMethods}`
    */
-  model() {
+  async model() {
     const { id: scope_id } = this.modelFor('scopes.scope');
+    // Fetch auth methods for the current scope
+    const authMethods = await this.store.query('auth-method', { scope_id });
```
randallmorey, line 26: This should probably also be a resource filter call.

randallmorey (changes requested):
Added some final feedback.  How do you think we should handle the case where the currently selected scope has no auth'able auth methods?  Should we just allow it and show the no methods message?  Or should we redirect?  But then what happens if _no_ scopes have auth'able auth methods?

DhariniJeeva:
> Added some final feedback. How do you think we should handle the case where the currently selected scope has no auth'able auth methods? Should we just allow it and show the no methods message? Or should we redirect? But then what happens if _no_ scopes have auth'able auth methods?

1. we filter out the scopes with no authenticatable auth methods, so we shouldn't be seeing the scope on the list if it doesn't have any authenticatable auth methods. 
2.  If no scopes have authenticatable auth methods, then with the current setup, it will be just empty as there are no scopes or auth methods after the filtering is done. But I don't _think_ we will ever be in a situation such as this as we always generate a default global scope with auth methods in it. Please let me know if I understood this incorrectly.

PR #1126, opened 2022-06-07
Chore include terminated

randallmorey:
:tickets: [Jira ticket](https://hashicorp.atlassian.net/browse/ICU-4651)

## Description

Admin and Desktop now send a `include_terminated=true` query parameter when the terminated status filter is selected.


cameronperera (approved):
LGTM

PR #1127, opened 2022-06-07
create new image component

DhariniJeeva:
:tickets: [Jira ticket](https://hashicorp.atlassian.net/browse/ICU-4663)

## Description

This pr introduces a new image core component to display images on the onboarding screens.

Based on offline discussion added it to core components.

 

light theme:
Screen Shot 2022-06-07 at 3 50 52 PM

Dark theme:
Screen Shot 2022-06-07 at 3 50 59 PM


calcaide (commented):
Is this a component that just Admin UI wil use or also Desktop need it? 
In case just the Admin UI uses it, I will suggest to move the component within `/ui/admin/app/components` but I will not block this PR for that.

cameronperera (approved):
LGTM!

DhariniJeeva:
> Is this a component that just Admin UI wil use or also Desktop need it? In case just the Admin UI uses it, I will suggest to move the component within `/ui/admin/app/components` but I will not block this PR for that.

ty! for now, it makes sense to have it in the core, similar to other components, in case we introduce an onboarding process to the desktop as well. merging this to unblock other work, but will come back to it if that's not the case.

calcaide:
@DhariniJeeva gotcha! makes sense. Thanks for the clarification 😉

PR #1128, opened 2022-06-08
Fix rose peer dependencies

calcaide:
## Description:
Fix two warnings within `rose` about peerDependencies. We have to downgrade Peer dependencies to be compliance.

### Screenshots:

Screenshot with the two warnings before the fix:
Screen Shot 2022-06-07 at 1 04 06 PM

Screenshot with the two warning gone after the fix:
Screen Shot 2022-06-07 at 5 39 15 PM


cameronperera (approved):
LGTM!

DhariniJeeva (approved):

PR #1130, opened 2022-06-09
refactor: 💡 include types for host-catalog and target creation

cameronperera:
For resource creation, the host catalog and target creation needed to know what type each resource is. This will allow the logic to succeed in a real Boundary instance.


DhariniJeeva (approved):

randallmorey:
Ported this change into the `chore-move-onboarding` branch.

PR #1135, opened 2022-06-14
feat: 🎸 Add Controller URL component to Success screen

cameronperera:
✅ Closes: ICU-4785

:tickets: [Jira ticket](https://hashicorp.atlassian.net/browse/ICU-4785)

Screen Shot 2022-06-14 at 12 13 52 PM

### DIFF
--- a/ui/admin/app/templates/onboarding/quick-setup/create-resources/success.hbs
+++ b/ui/admin/app/templates/onboarding/quick-setup/create-resources/success.hbs
@@ -1,12 +1,12 @@
 {{page-title (t 'titles.success')}}
-<Rose::Layout::Centered class='success-centered'>
+<Rose::Layout::Centered class='onboarding-quick-setup-success'>
   <Rose::Frame>
     <:header>
       <h2>{{t 'onboarding.success.titles.connect'}}</h2>
     </:header>
     <:body>
-      <div class='success-boundary'>
-        <div class='success-sub-title'>
+      <div class='onboarding-quick-setup-success-boundary'>
+        <div class='onboarding-quick-setup-success-sub-title'>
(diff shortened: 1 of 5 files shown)


ui/admin/app/templates/onboarding/quick-setup/create-resources/success.hbs
```diff
@@ -35,34 +35,50 @@
           {{t 'onboarding.success.links.learn'}}
         </a>
 
+        <div class='controller'>
```
randallmorey, line 38: This class names feels a bit generic.  Should we namespace onboarding class names?  e.g. `onboarding-quick-start-`?  Open to ideas here.

randallmorey (changes requested):

ui/admin/app/templates/onboarding/quick-setup/create-resources/success.hbs
cameronperera: Yeah that is a good point. So something like `onboarding-quick-start-controller` and `onboarding-quick-start-developer`? This would prevent styles accidentally being applied to other "onboarding" work if we ever have more.

cameronperera: Class names updated.

randallmorey (changes requested):
It looks like the card stretches vertically to fill the window, which I don't think is expected.  Can it be centered vertically?
Screenshot 2022-06-14 at 16 33 32

cameronperera:
> It looks like the card stretches vertically to fill the window, which I don't think is expected. Can it be centered vertically? 

Fixed. Good catch.

randallmorey (approved):

DhariniJeeva (approved):
LGTM!!

PR #1142, opened 2022-06-16
Chore add onboarding role

cameronperera:
Screen Shot 2022-06-16 at 4 39 56 PM

### DIFF
--- a/addons/core/translations/en-us.yaml
+++ b/addons/core/translations/en-us.yaml
@@ -77,7 +77,7 @@ onboarding:
       desktop: Using Boundary Desktop
       developers: Developers
       controller: Controller URL
-      role: We have created your scopes and target
+      creation-success: We have created your scopes and target
     descriptions:
       download: Download the Boundary Desktop app to browse and connect to targets (available for Windows, Mac, and Linux)
       developers: You can also connect to targets using the Boundary CLI or API.
--- a/addons/core/translations/en-us.yaml
+++ b/addons/core/translations/en-us.yaml
@@ -77,10 +77,12 @@ onboarding:
       desktop: Using Boundary Desktop
       developers: Developers
       controller: Controller URL
+      role: We have created your scopes and target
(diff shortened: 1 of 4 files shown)


addons/core/translations/en-us.yaml
```diff
@@ -77,10 +77,12 @@ onboarding:
       desktop: Using Boundary Desktop
       developers: Developers
       controller: Controller URL
+      role: We have created your scopes and target
```
randallmorey, line 80: This key/value mapping doesn't make sense to me.  Why is the key "role"?

ui/admin/app/templates/onboarding/quick-setup/create-resources/success.hbs
```diff
@@ -5,53 +5,60 @@
       <h2>{{t 'onboarding.success.titles.connect'}}</h2>
     </:header>
     <:body>
-      <div class='onboarding-quick-setup-success-boundary'>
-        <div class='onboarding-quick-setup-success-sub-title'>
-          <Rose::Icon @name='logo-app-background' @size='extra-large' />
+      <div class='onboarding-quick-setup-success-role'>
+        <span>
```
randallmorey, line 9: What is the purpose of this `span`?

randallmorey (approved):

ui/admin/app/templates/onboarding/quick-setup/create-resources/success.hbs
cameronperera: To make the Icon and h3 tag inline with each other. That is what line 300 in `app.scss` is for.

addons/core/translations/en-us.yaml
cameronperera: Updated to `creation-success`. Even that seems not exactly right but more accurate than "role".

DhariniJeeva (approved):
ty! LGTM!

PR #1147, opened 2022-06-20
fix: grant scope field should be functional

randallmorey:
## Description

Fixes a bug with the role grant scope field where it would not allow changes to grant scope.

Screenshot 2022-06-20 at 12 12 53


gsusmi (approved):
LGTM. Ty!

PR #1150, opened 2022-06-23
Use precommit hooks

ZedLi:
## Description
Added pre-commit hook and disabled style rules that conflict with prettier.
Also ran the CLI tools for eslint and stylelint to confirm there were no conflicting rules, see [this](https://github.com/prettier/eslint-config-prettier#cli-helper-tool) and [this](https://github.com/prettier/eslint-config-prettier#cli-helper-tool)

You can test it by just doing another `yarn install`

### DIFF
--- a/addons/api/package.json
+++ b/addons/api/package.json
@@ -80,5 +80,9 @@
   },
   "ember-addon": {
     "configPath": "tests/dummy/config"
+  },
+  "lint-staged": {
+    "*.js": "eslint --fix",
+    "*.hbs": "ember-template-lint --fix"
   }
 }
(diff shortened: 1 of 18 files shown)


addons/api/package.json
```diff
@@ -80,5 +79,10 @@
   },
   "ember-addon": {
     "configPath": "tests/dummy/config"
+  },
+  "lint-staged": {
```
ZedLi, line 83: I didn't include these commands in our scripts because they're one off and not that useful out of the context of `lint-staged` (they get called with the files from lint-staged but they don't specify any by themselves)

Do you have any thoughts @randallmorey? Would you still prefer these commands be in scripts?

randallmorey (changes requested):
Thanks for putting this together!  Can you do a quick edit of the README to remove any mentions of `yarn commit`, and describe what to expect on commit going forward?

randallmorey (approved):

PR #1156, opened 2022-06-27
Refactor attributes fragments

randallmorey:
## Description

This PR eliminates ember-cli-model-fragments for select resources, instead relying on [the `isNestedAttribute` flag](https://github.com/hashicorp/boundary-ui/blob/main/addons/api/addon/serializers/application.js#L54).  The following resources are affected:

- `managed-group`
- `credential-store`
- `credential-library`
- `account`

This is a step closer to eliminating the dependency on model fragments in Boundary.  Please note that `auth-method` cannot yet be migrated, since it relies on fragment arrays for which we have no alternative.

As an added benefit this refactor resulted in a net reduction of LoC.


addons/api/addon/generated/models/account.js
```diff
@@ -43,4 +43,20 @@ export default class GeneratedAccountModel extends BaseModel {
...
+  // =attributes (password)
+  @attr('string', {
+    for: 'password',
+    isNestedAttribute: true,
+    description: 'The account login name',
+  })
+  login_name;
+
+  // =attributes (OIDC)
+  @attr('string', { for: 'oidc', isNestedAttribute: true }) subject;
```
calcaide, line 56: In other fields on this file, or in other files on this PR I see when defining `@attr` we use several lines rather than one liner.

I do not have a strong opinion here, but wondering if we should be consistent with it?

DhariniJeeva (approved):

addons/api/addon/generated/models/account.js
randallmorey: This is automated via prettier.  If the line exceeds the max line length, it gets split.

PR #1166, opened 2022-06-28
Refactor base scope fragment

randallmorey:
## Description

This PR converts the base model's `scope` field to a simple attribute from a model fragment.  This is a necessary step toward removing our dependency on [ember-data-model-fragments](https://github.com/adopted-ember-addons/ember-data-model-fragments), which approaches an abandoned state (it has been incompatible with recent Ember Data releases for about a year).


addons/api/addon/models/base.js
```diff
@@ -55,6 +53,10 @@ export default class BaseModel extends Model {
     if (model) {
       const json = model.serialize();
       json.scope_id = model.id;
+      json.type = model.type;
```
calcaide, line 56: These new attributes are coming from the API or we set them in UI?

ui/desktop/tests/acceptance/origin-test.js
```diff
@@ -11,11 +11,11 @@ import {
 } from '@ember/test-helpers';
 import { setupApplicationTest } from 'ember-qunit';
 import setupMirage from 'ember-cli-mirage/test-support/setup-mirage';
-//import { Response } from 'miragejs';
+// import { Response } from 'miragejs';
```
calcaide, line 14: If we do not need it anymore, can we get rid of the comment lines?

DhariniJeeva (approved):

PR #1169, opened 2022-06-29
fix: 🐛 unable to filter for terminated sessions

randallmorey:
## Description

This PR resolves a UX issue in admin and Desktop sessions views where it becomes impossible to filter for terminated sessions if sessions of other types do not exist, since the toolbar is hidden.  The solution is to display the filter toolbar at all times, allowing users to enable terminated sessions filtering.

Screenshot 2022-06-29 at 17 23 07

Screenshot 2022-06-29 at 15 46 32


DhariniJeeva (approved):
LGTM!

ZedLi (approved):

PR #1171, opened 2022-07-05
Chore update eslint and plugins update

calcaide:
## Description

The PR has so many changed files because the introduction of `no-assert-equal` rule with the update. [More info about the rule here](https://github.com/platinumazure/eslint-plugin-qunit/blob/master/docs/rules/no-assert-equal.md).

- Updated eslint and the eslint plugins we depend on.

#### How to test this change?
- Run `yarn run lint` and report 0 linting errors.
- Run `yarn run test` and report 0 failing tests.


ZedLi (approved):
LGTM, I assume we never have a use case for using loose equality?

DhariniJeeva (approved):

PR #1173, opened 2022-07-05
Update content security

DhariniJeeva:
:tickets: [Jira ticket](https://hashicorp.atlassian.net/browse/ICU-3593)

## Description
this pr updates the ember content security by following the [changelog](https://github.com/rwjblue/ember-cli-content-security-policy/blob/master/DEPRECATIONS.md) guide

### DIFF
--- a/addons/api/package.json
+++ b/addons/api/package.json
@@ -36,7 +36,7 @@
   "devDependencies": {
     "@babel/core": "^7.18.5",
     "@babel/eslint-parser": "^7.18.2",
-    "@babel/plugin-proposal-decorators": "^7.18.6",
+    "@babel/plugin-proposal-decorators": "^7.17.8",
     "@ember/optional-features": "^2.0.0",
     "@ember/test-helpers": "^2.8.1",
     "@glimmer/component": "^1.1.2",
--- a/addons/api/package.json
+++ b/addons/api/package.json
@@ -36,7 +36,7 @@
   "devDependencies": {
     "@babel/core": "^7.18.5",
     "@babel/eslint-parser": "^7.18.2",
-    "@babel/plugin-proposal-decorators": "^7.17.8",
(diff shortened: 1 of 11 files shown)


calcaide (changes requested):
There are two ways of applying CSP with this addon, as they mention on their readme: 
```
The policy can be delivered either via a Content-Security-Policy HTTP response header or as a meta tag in the index.html
```

As per my understanding of the actual code, we are delivering CSP with the response header, [code here](https://github.com/hashicorp/boundary-ui/blob/main/ui/admin/config/environment.js#L99).

This PR is introducing the usage of CSP as a meta tag in the `index.html`.

I would like to keep the delivery of CSP using HTTP response header. Unless there is team consensus to star deliverin it as meta tag.

DhariniJeeva:
> There are two ways of applying CSP with this addon, as they mention on their readme:
> 
> ```
> The policy can be delivered either via a Content-Security-Policy HTTP response header or as a meta tag in the index.html
> ```
> 
> As per my understanding of the actual code, we are delivering CSP with the response header, [code here](https://github.com/hashicorp/boundary-ui/blob/main/ui/admin/config/environment.js#L99).
> 
> This PR is introducing the usage of CSP as a meta tag in the `index.html`.
> 
> I would like to keep the delivery of CSP using HTTP response header. Unless there is team consensus to star delivering it as meta tag.

gotcha! I will look into using HTTP header

randallmorey (approved):

ui/admin/config/environment.js
```diff
@@ -97,18 +97,6 @@ module.exports = function (environment) {
     },
 
     contentSecurityPolicyHeader: 'Content-Security-Policy',
```
calcaide, line 99: Do we still In need of this?

DhariniJeeva: good catch! removed it

cameronperera (approved):

PR #1178, opened 2022-07-07
chore: 🤖 Update ember-auto-import

calcaide:
## Description

Update `ember-auto-import`.

Following the `ember-auto-import` documentation:

[ember-auto-import 2.0 guide](https://github.com/ef4/ember-auto-import/blob/main/docs/upgrade-guide-2.0.md#ember-auto-import-20).
[Usage from Addons](https://github.com/ef4/ember-auto-import#usage-from-addons).

- Apps (Admin and desktop in our case) must add `webpack` to their `devDependencies`.
- Apps that use addons with `ember-auto-import` >= 2 must has `ember-auto-import` on their `devDependencies`.
- [Addons](https://github.com/ef4/ember-auto-import#installing-ember-auto-import-in-an-addon)

Screen Shot 2022-07-07 at 12 42 37 PM


ZedLi:
Do we need to add webpack as a dev dependency in the other packages? I was looking at the upgrade guide [here](https://github.com/ef4/ember-auto-import/blob/main/docs/upgrade-guide-2.0.md#quick-summary)

cameronperera (approved):

ZedLi:
Reading that section says:

> Apps are required to depend on webpack. This gives the app control over the exact webpack version in use, and it makes it possible for the app to import directly from webpack for access to features like DefinePlugin that it may want to put into the webpack config.

and in the beginning

> apps must add webpack to their own dependencies (yarn add --dev webpack@5 or npm install --save-dev webpack@5)

How I took it as you still need a devDependency on webpack for our apps (admin and desktop) but i'm confused because it says it shouldn't have a peerDependency in addons.

calcaide:
> 

@ZedLi Exactly, running on the same confussion. I will be tweaking it and running some tests to double check and see if that brings some clarity. Thanks for the rubber duck debugging!!

ZedLi (approved):
LGTM, thanks for looking into this more!

PR #1180, opened 2022-07-08
refactor: 💡 Refactor default values for target creation

calcaide:
:tickets: [Jira ticket](https://hashicorp.atlassian.net/browse/ICU-5127)

## Description

Refactor default value of `session_connection_limit`.

New behaviour: if the user does not specify a limit, at creation time the backend will assign `-1` (unlimited) as default. So the UI does not deal with default values anymore for this field.

### How I test this PR?

With the latest `0.9.1` binary, I used `BOUNDARY_DEV_UI_PASSTHROUGH_DIR` to passthrough the UI in my local env to the binary running, [here more info how to use it](https://github.com/hashicorp/boundary#ui-development).

New target form before being send, no value specified for `session_connection_limit`:
Screen Shot 2022-07-08 at 11 58 42 AM

New target form after being created with the response from BE, notice now it shows `-1` in `session_connection_limit`:
Screen Shot 2022-07-08 at 11 59 02 AM

### DIFF
--- a/addons/api/addon/generated/models/target.js
+++ b/addons/api/addon/generated/models/target.js
@@ -30,7 +30,6 @@ export default class GeneratedTargetModel extends BaseModel {
 
   @attr('number', {
     description: '',
-    defaultValue: 1,
   })
   session_connection_limit;
 
--- a/addons/api/tests/unit/serializers/target-test.js
+++ b/addons/api/tests/unit/serializers/target-test.js
@@ -34,7 +34,7 @@ module('Unit | Serializer | target', function (hooks) {
       type: 'tcp',
       scope_id: 'org_1',
       session_max_seconds: 28800,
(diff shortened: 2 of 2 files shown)


DhariniJeeva (approved):
LGTM! 
minor query- do we need any mirage updates for this change?

calcaide:
> LGTM! minor query- do we need any mirage updates for this change?

@DhariniJeeva nice question!
The only reference to `session_connection_limit` within mirage I found is when we generate the data, and we alreaady account to generate some with `-1` as value, [code here](https://github.com/hashicorp/boundary-ui/blob/main/addons/api/mirage/factories/target.js#L30).

cameronperera (approved):

ZedLi (approved):

randallmorey (approved):

PR #1200, opened 2022-07-19
Icu 5230 credential serializer

lisbet-alvarez:
:tickets: [Jira ticket](https://hashicorp.atlassian.net/browse/ICU-5230)

## Description
Added credential serializer and unit tests.


cameronperera:
@lisbet-alvarez I think we need to merge main into this branch because we have work from our credential model branch showing up in the diff.

addons/api/addon/generated/models/credential.js
```diff
@@ -54,6 +54,13 @@ export default class GeneratedCredentialModel extends BaseModel {
   })
   password_hmac;
 
+  @attr('string', {
+    for: 'username_password',
+    isNestedAttribute: true,
```
randallmorey, line 59: Just want to confirm, are these nested under `attributes`?  e.g. `attributes.password`?

addons/api/addon/generated/models/credential.js
```diff
@@ -54,6 +54,13 @@ export default class GeneratedCredentialModel extends BaseModel {
...
+    description: 'The password for credential.',
+  })
+  password;
+
   @attr('string', {
     for: 'username_password',
```
randallmorey, line 65: Is the `type` of a user/pass cred store `username_password`?

addons/api/addon/generated/models/credential.js
lisbet-alvarez: Yes. Password, password_hmac, and username are all nested under attributes.

addons/api/addon/generated/models/credential.js
lisbet-alvarez: correct

randallmorey (approved):

DhariniJeeva:
@lisbet-alvarez looks like `it serializes correctly on create` unit test failed. Can you check on that?

cameronperera (approved):
LGTM!

DhariniJeeva (approved):
ty! LGTM!

PR #1203, opened 2022-07-20
Create static credentials tab 

DhariniJeeva:
this pr adds a `credentials` tab to a static credential store and corresponding new routes. 

templates related to that will be added in a separate pr to keep this one simple

credentials tab: https://boundary-o7euohz8l-hashicorp.vercel.app/scopes/s_1sheljmjgm/credential-stores/cs_20nu3mgl4y


ui/admin/app/components/credential-stores/credential-store/nav/index.hbs
```diff
@@ -2,12 +2,19 @@
...
+        {{t 'resources.credential-library.title_plural'}}
+      </nav.link>
+    {{/if}}
+  {{else}}
     <nav.link
-      @route='scopes.scope.credential-stores.credential-store.credential-libraries'
+      @route='scopes.scope.credential-stores.credential-store.credentials'
     >
-      {{t 'resources.credential-library.title_plural'}}
+      {{t 'resources.credential.title_plural'}}
```
cameronperera, line 17: Should this tab be wrapped in a feature flag? Also, should we show a tab for static credential stores while the feature is not enabled?

cameronperera (changes requested):
Looks great. I had one comment with potential changes.

ui/admin/app/components/credential-stores/credential-store/nav/index.hbs
DhariniJeeva: I don't think this requires feature-flag wrapper as we make the `static` non-clickable when the flag is disabled.  

Screen Shot 2022-07-20 at 2 54 11 PM

DhariniJeeva: regarding your second question - do you mean the `credentials` tab?

cameronperera: Yes that is what I meant. However, with your previous comment about static credential stores not being clickable right now. I see no need for a change. Apologies for the confusion.

cameronperera (approved):

PR #1207, opened 2022-07-21
Icu 5167 Create codeblock component

ZedLi:
:tickets: [Jira ticket](https://hashicorp.atlassian.net/browse/ICU-5167)

## Description

Added a code editor component which was taken from the [HCP core packages](https://github.com/hashicorp/cloud-ui/tree/master/packages/cloud-ui-core/addon/components/code-editor).

Some of the components and styles were switched over to use our rose components and styles.

:rose: [Rose preview](https://boundary-ui-storybook-git-icu-5167-ui-create-c-6dcbfb-hashicorp.vercel.app/?path=/story/rose-code-editor--readonly)

image


addons/rose/addon/components/rose/code-editor/toolbar/index.js
```diff
@@ -0,0 +1,22 @@
...
+import { action } from '@ember/object';
+import { later } from '@ember/runloop';
+
+export const COPY_ICON_TYPE = 'flight-icons/svg/clipboard-copy-16';
+export const COPIED_ICON_TYPE = 'flight-icons/svg/clipboard-checked-16';
+
+export default class RoseCodeEditorToolbarComponent extends Component {
+  @tracked copyIconType = COPY_ICON_TYPE;
+
+  @action
```
ZedLi, line 12: This [used to be a task](https://github.com/hashicorp/cloud-ui/blob/master/packages/cloud-ui-core/addon/components/code-editor/toolbar/index.js#L94) and they originally used ember concurrency.

I noticed we didn't have it installed and it seemed to be in essence a `setTimeout` to transition the button which seemed a bit excessive to install two packages for this. 

Let me know if there's any problems with not using ember-concurrency for this

addons/rose/addon/components/rose/code-editor/toolbar/index.hbs
```diff
@@ -0,0 +1,26 @@
...
+      ></div>
+    {{/if}}
+    <CopyButton
+      @clipboardText={{@copyText}}
+      @success={{this.copied}}
+      class='code-editor-toolbar__copy-button'
+      data-test-code-editor-toolbar-copy-button
+    >
+      <Rose::Icon @name={{this.copyIconType}} @size='medium' />
+      Copy code
```
ZedLi, line 23: I hardcoded this for now as I was a little unsure of how to proceed.

It looks like in `Rose`, we don't have any translations, I assume because they should be generic and the consumer would be usually supplying any text.

Should this component instead have been in `core`? It's still a little fuzzy for me when it should go to which project. Or should I just pass in another variable that sets the text for the copy button and let the consumer handle it?

calcaide: I think you brought very interesting points.

- Regarding where a component should leave, perhaps @randallmorey can help us? How I always see it is any shared component between UI's goes in Rose, but I also see we have some components within code, so a bit confused.
- About the hard-coded text I am completely fine with it. My humble opinion: let's take simplicity path, for now we just have the use case that require the button with such text. I will think about making that text a parameter of the component when we need it :)

addons/rose/package.json
```diff
@@ -112,6 +114,6 @@
   "lint-staged": {
     "*.js": "eslint --fix",
     "*.hbs": "ember-template-lint --fix",
-    "*.{js,scss}": "prettier --write"
+    "*.{js,scss,mdx}": "prettier --write"
```
calcaide, line 117: N-I-C-E!

addons/rose/ember-cli-build.js
```diff
@@ -7,6 +7,9 @@ module.exports = function (defaults) {
     // Add options here
   });
 
+  app.import('node_modules/codemirror/lib/codemirror.css');
```
calcaide, line 10: Not blocking:
I am talking without the context and the knowledge of using/consuming this addon. Perhaps this is the only way of making it work and I am totally fine with it but I did not see on any of our addons or UI's we import `.css` using `app.import`. There is a reason for it? maybe we can explore importing those within _index.scss, [link here](https://github.com/hashicorp/boundary-ui/blob/main/addons/rose/addon/styles/rose/_index.scss)

Or copy those styles within `addons/rose/addon/styles/rose/components/code-editor/_index.scss`?

Take this comment as I am just thinking loud :sweat_smile:

addons/rose/tests/integration/components/rose/code-editor-test.js
```diff
@@ -0,0 +1,27 @@
+import { module, test } from 'qunit';
+import { setupRenderingTest } from 'ember-qunit';
+import { render } from '@ember/test-helpers';
+import { hbs } from 'ember-cli-htmlbars';
+
+module('Integration | Component | rose/code-editor', function (hooks) {
+  setupRenderingTest(hooks);
+
+  test('it renders', async function (assert) {
+    let codeValue = 'export HCP_CLIENT_ID = {place your client_id here}';
```
calcaide, line 10: Not blocking:
Can `codeValue` be a `const` ?  Seems to me the variable does not change its state.

addons/rose/tests/integration/components/rose/code-editor/field-editor-test.js
```diff
@@ -0,0 +1,69 @@
...
+import { hbs } from 'ember-cli-htmlbars';
+
+module(
+  'Integration | Component | rose/code-editor/field-editor',
+  function (hooks) {
+    setupRenderingTest(hooks);
+    const editorSelector = '[data-test-code-editor-field-editor]';
+
+    test('it renders', async function (assert) {
+      let codeValue = `export HCP_CLIENT_ID = {place your client_id here}`;
```
calcaide, line 13: Not blocking:
Can `codeValue` be a `const` ? Seems to me the variable does not change its state.

addons/rose/tests/integration/components/rose/code-editor/field-editor-test.js
```diff
@@ -0,0 +1,69 @@
...
+      this.set('codeValue', codeValue);
+      await render(hbs`
+      <Rose::CodeEditor::FieldEditor @value={{this.codeValue}} />
+    `);
+      assert.dom(editorSelector).isVisible();
+      assert.dom(editorSelector).includesText(codeValue);
+
+      let myNewCode = 'NEW CODE';
```
calcaide, line 21: Same coment as above.

addons/rose/tests/integration/components/rose/code-editor/toolbar-test.js
```diff
@@ -0,0 +1,55 @@
...
+import { hbs } from 'ember-cli-htmlbars';
+
+module('Integration | Component | rose/code-editor/toolbar', function (hooks) {
+  setupRenderingTest(hooks);
+  const toolbarSelector = '[data-test-code-editor-toolbar]';
+  const copyButtonSelector = '[data-test-code-editor-toolbar-copy-button]';
+  const menuDividerSelector = '[data-test-code-editor-toolbar-menu-divider]';
+
+  hooks.beforeEach(function () {
+    let copyText = 'copyText';
```
calcaide, line 13: Not blocking:
Perhaps `copyText`a dn `onCopy` (line 41) can be `const`?

calcaide (approved):
LGTM!! Awesome job! left some comments but not blocking 😉

addons/rose/tests/integration/components/rose/code-editor-test.js
ZedLi: Yup good call, I updated the places I used `let`!

addons/rose/ember-cli-build.js
ZedLi: I had played around with trying to import it like that but it didn't seem to work (seems like the files are in the  root `node_modules` and not the `rose` one). 

I ended up following what cloud UI did and the [documentation](https://guides.emberjs.com/release/addons-and-dependencies/#toc_compiling-assets) made it seem like it was reasonable.

Let me know if there's any other ideas, I'm open to trying other things for this too.

calcaide: All good!! you already tried and followed best practices

addons/rose/addon/components/rose/code-editor/index.hbs
```diff
@@ -0,0 +1,8 @@
+<div data-test-code-editor>
+  {{yield
+    (hash
+      FieldEditor=(component 'rose/code-editor/field-editor' value=@codeValue)
```
randallmorey, line 4: Rose uses lowercase yielded component names.

addons/rose/addon/components/rose/code-editor/index.stories.mdx
```diff
@@ -0,0 +1,90 @@
...
+<Meta title="Rose|Code Editor" component="CodeEditor" />
+
+# Code Editor
+
+`CodeEditor` is a component that yields two presentational components which together provide an IDE-like UX.
+CodeEditor::FieldEditor is where the provided code is displayed with syntax highlighting.
+CodeEditor::Toolbar is intended to go on top of the FieldEditor and provides a 'Copy code' button as well as other optional user-provided actions.
+
+```html
+<CodeEditor @codeValue="{{this.codeIWantToSee}}" as |C|>
```
randallmorey, line 13: Boundary tends to use lowercase yielded names, e.g. `|c|` or `|editor|`

addons/rose/addon/components/rose/code-editor/index.stories.mdx
```diff
@@ -0,0 +1,90 @@
...
+  </Story>
+</Preview>
+
+<Preview>
+  <Story name="CodeEditorShell">
+    {{
+      template: hbs`
+      <Rose::CodeEditor @codeValue={{codeValue}} as |C|>
+        <C.Toolbar />
+        <C.FieldEditor class='autosize' @options={{options}} />
```
randallmorey, line 60: I think we'd want to make "autosize" an argument, rather than passing a classname like this.

addons/rose/addon/components/rose/code-editor/toolbar/index.hbs
```diff
@@ -0,0 +1,26 @@
...
+    {{#if (has-block 'menu')}}
+      <div data-test-code-editor-toolbar-menu class='code-editor-toolbar__menu'>
+        {{yield to='menu'}}
+      </div>
+      <div
+        class='code-editor-toolbar__menu-divider'
+        data-test-code-editor-toolbar-menu-divider
+      ></div>
+    {{/if}}
+    <CopyButton
```
randallmorey, line 16: Where is this coming from?

addons/rose/addon/components/rose/code-editor/toolbar/index.hbs
randallmorey: Rose components are as simple as they can be and never include human-readable text of any kind.  Also, hard-coded values are generally not used in Boundary UI.  We may want to port this to core.

addons/rose/addon/components/rose/code-editor/toolbar/index.js
```diff
@@ -0,0 +1,22 @@
...
+  copied() {
+    let { onCopy } = this.args;
+    let originalIconType = this.copyIconType;
+
+    this.copyIconType = COPIED_ICON_TYPE;
+    onCopy && onCopy();
```
randallmorey, line 18: Would an `if` statement be more consistent and/or clearer?

randallmorey (changes requested):
Good start.  I added some comments.  In general I'm wondering if the css class names can be changed to be consistent with Rose.

addons/rose/addon/components/rose/code-editor/toolbar/index.hbs
ZedLi: I can port this to core but thought that having it in storybook was really useful, any thoughts about that?

addons/rose/addon/components/rose/code-editor/toolbar/index.hbs
ZedLi: This is an import from a new package: [ember-cli-clipboard](https://github.com/jkusa/ember-cli-clipboard)

addons/rose/addon/components/rose/code-editor/toolbar/index.js
ZedLi: I'll update this to use an `if` statement if that's our convention but I personally think it's pretty clear, I would say it's quite idiomatic JS

addons/rose/addon/components/rose/code-editor/index.stories.mdx
ZedLi: I made it an argument but can you give a little explanation of what's the difference or why it's better?

addons/rose/addon/components/rose/code-editor/toolbar/index.js
randallmorey: It is a conditional yes, but I don't think we use this convention elsewhere at this time.

addons/rose/addon/components/rose/code-editor/toolbar/index.hbs
randallmorey: Ah.  As a housecleaning item after this PR is merged, we might update the existing copyable components to use this package instead of clipboard.js.

addons/rose/addon/components/rose/code-editor/toolbar/index.hbs
randallmorey: Yeah probably having it in storybook is more important for now.  Hopefully this is temporary and we eventually get an official code component.

randallmorey (approved):

PR #1208, opened 2022-07-21
Icu 5335 add doc link for credential store screens

cameronperera:
:tickets: [Jira ticket](https://hashicorp.atlassian.net/browse/ICU-5335)

## Note for discussion
I added tests to verify that the docs link is present and referencing the correct URL. I would like to not have to use the full URL in the future, maybe create a helper test method but I wanted to get feedback about including these tests from the team first. Let me know your thoughts.

## Description
Add DocLink component to new Credential Screen and the read Credential Store screen.

Screen Shot 2022-07-21 at 5 33 22 PM
Screen Shot 2022-07-21 at 5 33 38 PM

### DIFF
--- a/ui/admin/tests/acceptance/credential-store/read-test.js
+++ b/ui/admin/tests/acceptance/credential-store/read-test.js
@@ -81,4 +81,14 @@ module('Acceptance | credential-stores | read', function (hooks) {
       'Error 404'
     );
   });
+
+  test('Users can link to docs page for credential store', async function (assert) {
+    assert.expect(1);
+    await visit(urls.credentialStore);
+    assert.ok(
+      find(
+        `[href="https://boundaryproject.io/help/admin-ui/credential-stores"]`
+      )
+    );
+  });
 });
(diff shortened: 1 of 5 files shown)


ZedLi (approved):
LGTM, I'm personally fine with the tests!

DhariniJeeva (approved):
LGTM

lisbet-alvarez (approved):

ui/admin/tests/acceptance/credential-store/read-test.js
```diff
@@ -81,4 +81,14 @@ module('Acceptance | credential-stores | read', function (hooks) {
       'Error 404'
     );
   });
+
+  test('Users can link to docs page for credential store', async function (assert) {
+    assert.expect(1);
+    await visit(urls.credentialStore);
+    assert.ok(
+      find(
+        `[href="https://boundaryproject.io/help/admin-ui/credential-stores"]`
```
calcaide, line 90: A bit late to the party (I saw already merged).

I am fine with this test. If we want to test the user reach certain URL there is no other way but comparing that.

That said, since we already have a pattern to define urls (line 20 `const urls`) I will suggest to add this URL there, something like:
```javascript
const urls = {
  ...
  docsCredentialStore: 'https://boundaryproject.io/help/admin-ui/credential-stores'
}
```

Then on the test you just reference the url within the assertion.

I think that small change can facilitate mantainability of the code and facilitat (if need) re-using the code.

calcaide (commented):
LGTM

calcaide (commented):
LGTM

PR #1210, opened 2022-07-22
Refactor origin to cluster

calcaide:
## Description

Refactor `origin` to `clusterUrl`. Includes user facing strings, code and code comments.


ui/desktop/electron-app/src/models/runtime-settings.js
```diff
@@ -38,27 +38,27 @@ class RuntimeSettings {
...
+      throw new Error(`Cluster URL ${clusterUrl} could not be validated.`);
     }
   }
 
   /**
-   * Sets the origin to null.
+   * Sets the clusterUrl to null.
    */
-  async resetOrigin() {
-    this.origin = null;
```
calcaide, line 49: @hashicorp/boundary-ui I think this is a bug we fixed with this refactor.

Explanation:
- Within this file, the internal private var is declared `#origin`.
- Within all the methods that work or change that var is used as `#origin` but except this method, that was referencing it as `origin` (without the `#`).

On the refactor I changed this to add the `#` to the variable, which I think fixes a bug when reseting the cluster URL (former origin).

Thoughts?

ZedLi: I actually don't think this is a bug. It looks like it's using the setter to set the variable instead of directly accessing the private field. It depends if we want the `triggerOriginChanged` function to run when reset is called.

FYI I included a fiddle to show the private field setter difference for anyone who hasn't used this yet in JS: https://jsfiddle.net/6stq21k7/

ZedLi:
Should [this origin](https://github.com/hashicorp/boundary-ui/blob/refactor-origin-to-cluster/README.md?plain=1#L162) in the readme be updated?

A few other spots:
https://github.com/hashicorp/boundary-ui/blob/ffa09db928fa09d11f116553d1b2747ca94d4be8/ui/desktop/tests/end2end/test-helpers.js#L59
https://github.com/hashicorp/boundary-ui/blob/ffa09db928fa09d11f116553d1b2747ca94d4be8/ui/desktop/tests/end2end/authentication.spec.js#L16
https://github.com/hashicorp/boundary-ui/blob/ffa09db928fa09d11f116553d1b2747ca94d4be8/ui/desktop/tests/end2end/targets.spec.js#L16

ZedLi (approved):
LGTM, did some smoke testing and everything looked fine! Thanks for doing this big refactor!

cameronperera (approved):
LGTM! Nice work on this huge refactor. Thanks for doing it.

ui/desktop/electron-app/src/models/runtime-settings.js
calcaide: A-W-E-S-O-M-E @ZedLi I didn't though about that on that way and you are right. I though we were trying to access directly there. Thank you!!

calcaide:
> Should [this origin](https://github.com/hashicorp/boundary-ui/blob/refactor-origin-to-cluster/README.md?plain=1#L162) in the readme be updated?
> 
> A few other spots:
> 
> https://github.com/hashicorp/boundary-ui/blob/ffa09db928fa09d11f116553d1b2747ca94d4be8/ui/desktop/tests/end2end/test-helpers.js#L59
> 
> 
> https://github.com/hashicorp/boundary-ui/blob/ffa09db928fa09d11f116553d1b2747ca94d4be8/ui/desktop/tests/end2end/authentication.spec.js#L16
> 
> 
> https://github.com/hashicorp/boundary-ui/blob/ffa09db928fa09d11f116553d1b2747ca94d4be8/ui/desktop/tests/end2end/targets.spec.js#L16

@ZedLi Nice catch!!

PR #1214, opened 2022-07-25
feat: 🎸 Add doc link to Credential screens

cameronperera:
✅ Closes: ICU-5317

:tickets: [Jira ticket](https://hashicorp.atlassian.net/browse/ICU-5317)

## Description

The redirect for Credential does not work right now. This is a brand new redirect on the boundaryproject.io website as well so the redirect will not work until our next release when those changes are pushed to production.
Here is the [PR](https://github.com/hashicorp/boundary/pull/2272) that added the redirect to the boundaryproject.io website.

:technologist: [Admin preview](https://boundary-ui-git-icu-5317-create-a-new-doc-link-cccbec-hashicorp.vercel.app/scopes/s_1sheljmjgm/credential-stores/cs_20nu3mgl4y/credentials/cred_hzrxf5z5za)

Screen Shot 2022-07-25 at 5 43 52 PM
Screen Shot 2022-07-25 at 5 44 08 PM

### DIFF
--- a/ui/admin/app/templates/scopes/scope/credential-stores/credential-store/credentials/credential/index.hbs
+++ b/ui/admin/app/templates/scopes/scope/credential-stores/credential-store/credentials/credential/index.hbs
@@ -7,6 +7,7 @@
   <page.header>
     <h2>
       {{t 'resources.credential.title'}}
+      <DocLink @doc='credential' @iconSize='large' />
     </h2>
     <p>{{t 'resources.credential.description'}}</p>
     <Copyable
--- a/ui/admin/app/templates/scopes/scope/credential-stores/credential-store/credentials/new.hbs
+++ b/ui/admin/app/templates/scopes/scope/credential-stores/credential-store/credentials/new.hbs
@@ -9,6 +9,7 @@
   <page.header>
     <h2>
       {{t 'titles.new'}}
+      <DocLink @doc='credential' @iconSize='large' />
     </h2>
   </page.header>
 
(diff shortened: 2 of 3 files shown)


ZedLi:
How does the doc go from `https://boundaryproject.io/help/admin-ui` to `https://www.boundaryproject.io/docs/concepts/domain-model`? It looks like the other links in vercel work and not the new credential link but I'm not sure how this works.

Also is there a typo in [this line](https://github.com/hashicorp/boundary-ui/blob/5ab040d83e33de54aa46972cc082a15a6350bc62/ui/admin/config/environment.js#L94)? (looks like there's 3 Ts)

ZedLi (approved):

cameronperera:
> How does the doc go from `https://boundaryproject.io/help/admin-ui` to `https://www.boundaryproject.io/docs/concepts/domain-model`? It looks like the other links in vercel work and not the new credential link but I'm not sure how this works.

Sorry, I will add this explanation to the PR description but the redirect technically does not exist yet in production so the redirect will not work until our next release. That should not be an issue since we have feature flagged this work for now. Here is the [PR](https://github.com/hashicorp/boundary/pull/2272) that added the redirect to the docs website.

> Also is there a typo in [this line](https://github.com/hashicorp/boundary-ui/blob/5ab040d83e33de54aa46972cc082a15a6350bc62/ui/admin/config/environment.js#L94)? (looks like there's 3 Ts)

Yes, that is a typo. I will get that fixed. Thanks for catching that.

PR #1218, opened 2022-07-26
refactor: 💡 Edit a credential

lisbet-alvarez:
✅ Closes: ICU-5180

:tickets: [Jira ticket](https://hashicorp.atlassian.net/browse/ICU-5180)
:technologist: [Admin preview](https://boundary-ui-git-icu-5180-edit-a-credential-hashicorp.vercel.app/scopes/s_1sheljmjgm/credential-stores/cs_20nu3mgl4y/credentials/cred_hzrxf5z5za)

## Description
Password input field is now only viewable after user clicks on "Edit
Form"

### DIFF
--- a/ui/admin/app/components/form/credential/index.hbs
+++ b/ui/admin/app/components/form/credential/index.hbs
@@ -105,6 +105,29 @@
       {{/if}}
     </form.input>
   {{/if}}
+
+  {{#if form.isEditable}}
+    <form.input
+      @name='password'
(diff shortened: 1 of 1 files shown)


cameronperera:
Quick reminder that we will want an acceptance test for the password field rendering when the user goes to edit the credential. I would potentially just do it in a separate PR since this feature is added here but the credential acceptance tests are being added in another PR.

PR #1221, opened 2022-07-27
[CI-only] Add build job for HCPb variant

mdeggies:
## Description

This new job mirrors the enterprise build job, but lets us keep the existing logic laid out in https://github.com/hashicorp/boundary/blob/main/.github/workflows/build.yml#L136 to download the right UI artifacts based on the boundary variant/edition. This is needed for the release on August 1st. 🙏  Thanks

### DIFF
--- a/.github/workflows/build-admin-ui.yaml
+++ b/.github/workflows/build-admin-ui.yaml
@@ -94,8 +94,8 @@ jobs:
           node-version: '16.x'
         if: steps.node-modules-cache.outputs.cache-hit != 'true'
       - run: yarn run build:ui:admin:enterprise
-      - name: Upload artifact Admin UI Enterprise
+      - name: Upload artifact Admin UI Enterprise for HCP
         uses: actions/upload-artifact@v3
         with:
-          name: admin-ui-ent
+          name: admin-ui-hcp
           path: ./ui/admin/dist/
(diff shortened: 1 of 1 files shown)


.github/workflows/build-admin-ui.yaml
```diff
@@ -74,3 +74,28 @@ jobs:
...
+      - name: Build
+        uses: actions/setup-node@v3
+        with:
+          node-version: '16.x'
+        if: steps.node-modules-cache.outputs.cache-hit != 'true'
+      - run: yarn run build:ui:admin:enterprise
+      - name: Upload artifact Admin UI Enterprise
+        uses: actions/upload-artifact@v3
+        with:
+          name: admin-ui-ent
```
gsusmi, line 100: Should this be renamed to hcp? We already have [admin-ui-ent](https://github.com/hashicorp/boundary-ui/pull/1221/files#diff-9107bf357ecdbe3083ccc32162dfa2d2826ac1fac3468e413fdb65a5cf3829f8R75) variant.

mdeggies: Ahh thank you for the catch, that is definitely needed. I just pushed the update 🙏

cameronperera (approved):
LGTM

PR #1223, opened 2022-07-27
refactor: 💡 Deprecate faker to use @faker-js

calcaide:
## Description

Deprecate the usage of `faker` to use `@faker-js/faker`. More info about the reasons behind this refactor [on the jira ticket](https://hashicorp.atlassian.net/browse/ICU-5101).

Aside from the deprecation, this pr also changes the way `faker` is imported and used within files. Notice now we use faker in an API fashion. That could be changed if desired, but seemed to me helps to understand better the code.


randallmorey (approved):

ZedLi (approved):

DhariniJeeva (approved):

cameronperera (approved):
It WORKS!!!

PR #1227, opened 2022-07-28
Icu 5253 add query to include credentials

cameronperera:
:tickets: [Jira ticket](https://hashicorp.atlassian.net/browse/ICU-5253)

### Note about template linting rule
I updated the [template lint file](ui/admin/.template-lintrc.js) to allow for non-simple unless statements.

:technologist: [Admin preview](https://boundary-ui-git-icu-5253-add-query-to-include-c8da24-hashicorp.vercel.app/scopes/s_1sheljmjgm/targets/t_4uga1n9zdm/add-credential-sources)

Screen Shot 2022-07-28 at 4 28 51 PM


ui/admin/.template-lintrc.js
```diff
@@ -4,6 +4,7 @@ module.exports = {
   plugins: ['ember-template-lint-plugin-prettier'],
   extends: ['octane', 'ember-template-lint-plugin-prettier:recommended'],
   rules: {
+    'simple-unless': false,
```
DhariniJeeva, line 7: What does this do?

ui/admin/app/components/form/target/add-credential-sources/index.js
```diff
@@ -0,0 +1,68 @@
...
+    );
+    return [...notAddedCredentialLibraries, ...notAddedCredentials];
+  }
+
+  // =actions
+
+  /**
+   * Add/Remove credential source to current selection
+   * @param CredentialLibraryModel
+   */
```
DhariniJeeva, line 51: Can we include credential model here? 
 `@param {CredentialLibraryModel, Credential} credentialSource`

ui/admin/.template-lintrc.js
lisbet-alvarez: I was in a call with Cameron when he added this. Basically, the linter was getting upset because `{{#unless}}` was being used in conjunction with the `and` logical operator. The linter printed out an error saying that simple unless statements are recommended. So Cameron added that line in order to allow non-simple unless statements.

cameronperera: Exactly what Lis said. I would prefer to use an inline rule exception but I did not see any docs for that. I plan to look into it but I didn't want this to be blocker for this work.

ui/admin/app/components/form/target/add-credential-sources/index.js
cameronperera: Good catch. I will update that.

ui/admin/app/components/form/target/add-credential-libraries/index.js
```diff
@@ -1,65 +0,0 @@
-import Component from '@glimmer/component';
```
lisbet-alvarez, line 1: add-credential-libraries folder should be deleted too I think

ui/admin/app/components/form/target/add-credential-sources/index.js
```diff
@@ -0,0 +1,68 @@
+import Component from '@glimmer/component';
+import { computed, action } from '@ember/object';
+import { A } from '@ember/array';
+
+export default class FormTargetAddCredentialLibrariesIndexComponent extends Component {
```
lisbet-alvarez, line 5: I don't think this affects anything but for clarity the class name should be renamed to `FormTargetAddCredentialSourcesIndexComponent`

ui/admin/app/templates/scopes/scope/targets/target/add-credential-sources.hbs
```diff
@@ -14,16 +14,17 @@
...
         @model={{@model.target}}
         @credentialLibraries={{@model.credentialLibraries}}
+        @credentials={{@model.credentials}}
         @submit={{route-action 'save' @model.target}}
         @cancel={{route-action 'cancel'}}
       />
     {{/if}}
 
-    {{#unless @model.credentialLibraries}}
+    {{#unless (and @model.credentialLibraries @model.credentials)}}
```
ZedLi, line 27: I don't think this is quite right, if I'm understanding this correctly, we want to display the form if we have a credential library _or_ a credential.

I believe this is actually saying don't display it only if _both_ exists
i.e. `!(credentialLibrary && credential)` 

which would display this block if only one exists.

I think it would actually be 
```suggestion
    {{#unless (or @model.credentialLibraries @model.credentials)}}
```
which is `!(credentialLibrary || credential)` or `!credentialLibrary && !credential`

I actually agree with the template lint rule to only use simple `unless` now 😆

DhariniJeeva: I think we should move this logic to a js file and keep this form simple.  similar to [this](https://github.com/hashicorp/boundary-ui/blob/2d3c94b57eed313e8b16b4adabcdb8e0432651bc/ui/admin/app/components/form/target/add-credential-libraries/index.hbs#L1)

cameronperera: Great idea. Moving this to the controller!

ui/admin/app/components/form/target/add-credential-sources/index.js
cameronperera: Good catch. Fixed.

ui/admin/app/components/form/target/add-credential-sources/index.js
cameronperera: Removed

ui/admin/.template-lintrc.js
cameronperera: Removed.

ui/admin/app/controllers/scopes/scope/targets/target/add-credential-sources.js
```diff
@@ -8,6 +8,13 @@ export default class ScopesScopeTargetsTargetAddCredentialSourcesController exte
 
   // =attributes
 
+  get hasAvailableCredentialSources() {
+    return (
+      this.model.credentialLibraries.length > 0 ||
+      this.model.credentials.length > 0
+    );
```
randallmorey, line 15:
```suggestion
    return this.model.credentialLibraries.length || this.model.credentials.length;
```

ui/admin/app/routes/scopes/scope/targets/target/add-credential-sources.js
```diff
@@ -15,15 +15,26 @@ export default class ScopesScopeTargetsTargetAddCredentialSourcesRoute extends R
...
+    this.store.unloadAll('credential');
+  }
+
+  /**
+   * Checks for unassigned credential libraries.
+   * @param {[CredentialLibraryModel]} credentialLibraries
+   * @param {[CredentialModel]} credentials
+   * @type {boolean}
+   */
+  get hasAvailableCredentialSources() {
```
DhariniJeeva, line 31: I don't think this is necessary. What's this for?

randallmorey (approved):
Nice work!  Made a tiny suggestion.

lisbet-alvarez (approved):

ui/admin/app/controllers/scopes/scope/targets/target/add-credential-sources.js
```diff
@@ -8,6 +8,13 @@ export default class ScopesScopeTargetsTargetAddCredentialSourcesController exte
 
   // =attributes
```
DhariniJeeva, line 10: can we add a method signature here?

ui/admin/app/controllers/scopes/scope/targets/target/add-credential-sources.js
cameronperera: Much cleaner! Fixed.

DhariniJeeva (approved):

ui/admin/app/controllers/scopes/scope/targets/target/add-credential-sources.js
cameronperera: Added.

randallmorey (changes requested):
To keep the route class more readable, the post-fetch sorting/filtering/aggregating logic could be moved to a controller.

cameronperera:
> To keep the route class more readable, the post-fetch sorting/filtering/aggregating logic could be moved to a controller.

Fixed and pushed up.

ui/admin/app/components/form/target/add-credential-sources/index.js
```diff
@@ -0,0 +1,36 @@
...
+      this.selectedCredentialSourceIDs.removeObject(credentialSource.id);
+    }
+  }
+
+  /**
+   * Submit selected credential source ids
+   */
+  @action
+  submit(fn) {
+    fn(this.selectedCredentialSourceIDs);
```
ZedLi, line 34: Do we need to pass another function through the submit action? Could the passed in submit have been accessed with `this.args.submit`?

Or on the template use `@onSubmit={{fn @submit this.selectedCredentialSourceIDs}}`?

Not sure if there's a best practice but it feels a little weird to have to pass in another anonymous function like this, but maybe that's just me

ui/admin/tests/acceptance/targets/credential-sources-test.js
```diff
@@ -104,13 +119,51 @@ module('Acceptance | targets | credential-libraries', function (hooks) {
     assert.strictEqual(currentURL(), urls.addCredentialSources);
   });
 
+  test('displays list of credential sources with only credential libraries available', async function (assert) {
```
ZedLi, line 122: Is there a test for when there's both credential libraries and credentials available?

ZedLi (approved):
LGTM, great work!

ui/admin/app/components/form/target/add-credential-sources/index.js
cameronperera: That is a good question. I think this might just be the existing practice. This might be one of those things were we evaluate this as a team to decide on the best practice going forward. Any thoughts on this @randallmorey?

ui/admin/tests/acceptance/targets/credential-sources-test.js
cameronperera: 🤦‍♂️ I was too focused on edge cases I forgot to test the normal behavior. Adding now.

ui/admin/app/components/form/target/add-credential-sources/index.js
randallmorey: You can try without the action and see if it works.

randallmorey (approved):

ui/admin/app/components/form/target/add-credential-sources/index.js
cameronperera: I will try this in a different branch so I am not holding up my team with this PR.

PR #1235, opened 2022-08-03
Icu 5476 refactor target model serializer and mirage

DhariniJeeva:
Jira - https://hashicorp.atlassian.net/browse/ICU-5476

 
Screen Shot 2022-08-03 at 10 01 47 AM

[admin](https://boundary-fsgl0oaf5-hashicorp.vercel.app/scopes/s_1sheljmjgm/targets/t_4uga1n9zdm/credential-sources)


addons/core/translations/resources/en-us.yaml
```diff
@@ -378,6 +378,7 @@ credential-library:
   types:
     vault: Vault
     static: Static
+    username_password: Username & Password
```
lisbet-alvarez, line 381: I'm not sure if its necessary to define username_password type again here since it's already defined on line 394.

lisbet-alvarez, line 381: I think this might not be necessary since it is already defined on line 394.

cameronperera (approved):
LGTM

lisbet-alvarez (approved):

addons/api/mirage/config.js
```diff
@@ -464,19 +464,31 @@ export default function () {
...
         updatedAttrs.credentialLibraryIds = target.credentialLibraryIds;
+        updatedAttrs.credentialIds = target.credentialIds;
         attrs.applicationCredentialSourceIds.forEach((id) => {
-          if (!updatedAttrs.credentialLibraryIds.includes(id)) {
-            updatedAttrs.credentialLibraryIds.push(id);
+          if (
+            !updatedAttrs.credentialLibraryIds.includes(id) ||
+            !updatedAttrs.credentialIds.includes(id)
+          ) {
+            if (id.includes('cred')) {
```
ZedLi, line 473: Is it safe to assume that other id's won't have `cred` as part of them?

ZedLi (approved):
LGTM, great work

addons/api/mirage/config.js
cameronperera: For now it is safe. Static `username_password` credentials are prefixed with `cred_` while `vault` credential libraries are prefixed with `cl_`. I am fairly certain that this will need another refactor for future credential injection work.

addons/api/mirage/config.js
```diff
@@ -464,19 +464,31 @@ export default function () {
       // If adding credential sources, push them into the array
       if (method === 'add-credential-sources') {
         updatedAttrs.credentialLibraryIds = target.credentialLibraryIds;
+        updatedAttrs.credentialIds = target.credentialIds;
         attrs.applicationCredentialSourceIds.forEach((id) => {
-          if (!updatedAttrs.credentialLibraryIds.includes(id)) {
-            updatedAttrs.credentialLibraryIds.push(id);
+          if (
+            !updatedAttrs.credentialLibraryIds.includes(id) ||
```
randallmorey, line 470: Should this be `&&`?

randallmorey (changes requested):

addons/api/mirage/config.js
randallmorey: You might use `startsWith` to be more specific.

randallmorey (changes requested):

addons/api/mirage/config.js
cameronperera: Good suggestion. Fixed!

randallmorey (approved):

PR #1244, opened 2022-08-10
feat: 🎸 Add create worker page

ZedLi:
:tickets: [Jira ticket](https://hashicorp.atlassian.net/browse/ICU-5246)

## Description
The create page for workers. Still need to add tests.

Based on the newer [figma mockups](https://www.figma.com/file/ukvUqtTU2Bc2iULxZxw26k/Self-managed-workers-(BYOW)?node-id=1559%3A18841) that include input fields for the code editors

The code editor is slightly different from the mockups, mainly the copy button is on a toolbar above the editor instead of inline. I can bring over the `code blocks` component which is very simple and looks like the mockup but doesn't have the syntax highlighting which I kind of liked, would love to hear people's thoughts on this. 

As a consequence of having the toolbar, I put some of our text in the toolbar area otherwise the whitespace looks off. 

:technologist: [Admin preview](https://boundary-ui-git-icu-5246-create-worker-view-hashicorp.vercel.app/)

image


ui/admin/ember-cli-build.js
```diff
@@ -34,6 +34,9 @@ module.exports = function (defaults) {
     app.import('node_modules/clipboard/dist/clipboard.min.js');
   }
 
+  app.import('node_modules/codemirror/lib/codemirror.css');
```
ZedLi, line 37: I unfortunately needed to import these files again. Should I just copy the contents and put them inside the rose styles?

addons/core/translations/resources/en-us.yaml
```diff
@@ -420,9 +420,40 @@ worker:
...
+      help: The path on your server where you will store your configuration file
+    worker_tags:
+      label: Worker Tags
+      help: Boundary can use tags to define key-value pairs which targets can use to determine where they should route connections
+    steps:
+      1:
+        title: 1. Create a config file
+        create_directory: Create a configuration file in a new directory on your server
+        create_config: 'Open the file with a text editor, such as Vi. Paste the following configuration into the worker config file:'
+        save_config: Save the configuration in your `pki-worker.hcl` file.
```
ZedLi, line 449: I kind of wanted to wrap the `pkie-worker.hcl` words in a `<code></code>` block but wasn't sure how to accomplish this cleanly without using the `htmlSafe` [option](https://ember-intl.github.io/ember-intl/docs/helpers/t#format-html-message).

Any recommendations, are we opposed to using this for wrapping HTML that doesn't have user input? I know that it's not recommended in general to render arbitrary HTML but in this case it's just HTML tags we're passing along. 

Let me know your thoughts.

ui/admin/app/components/form/worker/index.hbs
```diff
@@ -0,0 +1,135 @@
...
+            @name='flight-icons/svg/check-circle-fill-16'
+            @size='medium'
+          />
+          <p>{{t 'resources.worker.form.steps.3.registered'}}</p>
+        </div>
+      </div>
+    {{/if}}
+  </div>
+
+  <Rose::LinkButton @route='scopes.scope.workers' @style='secondary'>
```
ZedLi, line 132: This link will go back to the `workers` page but won't refresh the list, is there a recommended way I should accomplish this? Normally we call `this.refresh()` after transitioning somewhere after a save but in this case I don't think we transition immediately.

joshuaogle:
Thanks for getting the inputs in! That's such a nice experience.

So we can detect that they are running this on HCP, we just don't get the cluster ID? We don't currently show the UUID on their overview page, so we may need to add that or add instructions for getting it from the URL. What do you think?

addons/rose/addon/styles/rose/components/layout/_page.scss
```diff
@@ -32,6 +32,7 @@
 
   .rose-layout-page-body {
     grid-area: body;
+    width: 100%;
```
ZedLi, line 35: Is there any concern with me adding an explicit `width: 100%` here?

I needed to do this because some of the elements (such as the code editors and token input field) will grow to the size of input, and it looks like we only have a relative `66%` width on the form but nothing on the parents.

image
image


The alternative is I set a fixed `width` or `max-width` on the `.worker-create-section` elements instead of a percentage but I don't think we normally do that?

I'd love to hear any alternatives that I'm not thinking of.

randallmorey (changes requested):
Noticed a few UI details that could be improved:

Icon/text alignment:
Screenshot 2022-08-12 at 11 27 35

Copy button wrapping may be undesirable here:
Screenshot 2022-08-12 at 11 27 47

Once a worker is added, the add field and button remain disabled so that no further workers may be added:
Screenshot 2022-08-12 at 11 28 06

ZedLi:
> For read-only code blocks, should we disable the cursor and line highlighting? We do want to enable text selection though.

@randallmorey I updated the editors to not show a cursor nor the active line highlighting. I also changed the two editors that are meant to display shell commands to not have line numbers, I thought it made more sense since it's meant to be in the terminal vs being a file.

Let me know what you think.

dvonanderson (approved):
LGTM

ui/admin/app/components/form/worker/index.hbs
ZedLi: Any suggestions on how to accomplish this @randallmorey?

ui/admin/app/components/form/worker/create-worker-led/index.hbs
```diff
@@ -0,0 +1,129 @@
...
+          <p>{{t 'resources.worker.form.steps.3.registered'}}</p>
+        </div>
+      </div>
+    {{/if}}
+  </div>
+
+  <Rose::LinkButton
+    @route='scopes.scope.workers'
+    @style='secondary'
+    {{on 'click' (route-action 'refreshWorkers')}}
```
ZedLi, line 125: This works but for some reason it seems to trigger two requests, is this expected behavior?

example

randallmorey (approved):
Approving for now.  Will circle back to the layout styles.

PR #1251, opened 2022-08-11
fix: 🐛 fixes host set attributes field types

randallmorey:
:tickets: [Jira ticket](https://hashicorp.atlassian.net/browse/ICU-5616)

## Description

Fixes normalization intended to support both string and array typed `filters` fields.

Closes hashicorp/boundary#2103

### DIFF
--- a/addons/api/mirage/factories/host-set.js
+++ b/addons/api/mirage/factories/host-set.js
@@ -35,21 +35,24 @@ export default factory.extend({
       return faker.datatype.number();
     }
   },
-  // AWS specific
-  filters() {
+
+  attributes() {
+    // AWS specific
     if (this.plugin?.name === 'aws') {
       const filtersAmount = faker.datatype.number({ min: 1, max: 5 });
-      let result = [];
+      let filters = [];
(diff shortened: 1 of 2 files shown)


cameronperera (approved):
LGTM!

addons/api/mirage/factories/host-set.js
```diff
@@ -35,21 +35,24 @@ export default factory.extend({
       return faker.datatype.number();
     }
   },
-  // AWS specific
-  filters() {
+
+  attributes() {
+    // AWS specific
     if (this.plugin?.name === 'aws') {
```
ZedLi, line 41: The azure plugin will have the filter in just `filter` and not `attributes.filters`?

ZedLi (approved):

addons/api/mirage/factories/host-set.js
gsusmi: Yup. Azure filter supports a single string value whereas aws supports multiple filters as an array.

DhariniJeeva (approved):

PR #1254, opened 2022-08-12
feat: 🎸 Update credential model with ssh private key attributes

cameronperera:
✅ Closes: ICU-5645

:tickets: [Jira ticket](https://hashicorp.atlassian.net/browse/ICU-5645)

### DIFF
--- a/addons/api/addon/serializers/credential.js
+++ b/addons/api/addon/serializers/credential.js
@@ -5,6 +5,10 @@ export default class CredentialSerializer extends ApplicationSerializer {
   serialize() {
     const serialized = super.serialize(...arguments);
     delete serialized['attributes']['password_hmac'];
+    if (serialized.type === 'username_password') {
+      delete serialized['attributes']['private_key'];
+      delete serialized['attributes']['passphrase'];
+    }
     return serialized;
   }
 
--- a/addons/api/addon/generated/models/credential.js
+++ b/addons/api/addon/generated/models/credential.js
@@ -46,6 +46,14 @@ export default class GeneratedCredentialModel extends BaseModel {
   })
   version;
 
+  // =attributes (username_password, ssh_private_key)
+  @attr('string', {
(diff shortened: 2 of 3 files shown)


addons/api/addon/serializers/credential.js
```diff
@@ -5,6 +5,10 @@ export default class CredentialSerializer extends ApplicationSerializer {
   serialize() {
     const serialized = super.serialize(...arguments);
     delete serialized['attributes']['password_hmac'];
+    if (serialized.type === 'username_password') {
```
ZedLi, line 8: Should this possibly be 
```suggestion
    if (serialized.type !== 'ssh_private_key') {
```
that way the check is flipped and if you add more types you don't have to keep adding to the if statement?

ZedLi (approved):
LGTM!

addons/api/addon/serializers/credential.js
cameronperera: This will likely change to a switch statement because we will have to do a similar delete on fields. Which, I believe we have an existing pattern of checking the type, then calling another serializer just for that type. I believe host catalogs do this. I am going to work on the serializer for this next and that should fix us this logic. This was mostly done to satisfy tests that failed due to the new attributes.

DhariniJeeva:
Looks good! Could we add a related serializer test for this change?

cameronperera:
> 

I was going to add the serializer work in another PR but I can add it here if you'd like. It will still be a small PR.

DhariniJeeva:
> > 
> 
> I was going to add the serializer work in another PR but I can add it here if you'd like. It will still be a small PR.

ahh ok! Noticed some serializer changes being included in this pr. But feel free to include the tests based on your preference. totally ok to have in another pr!  👍

DhariniJeeva (approved):

lisbet-alvarez (approved):

PR #1257, opened 2022-08-15
Icu 5646 serializer

cameronperera:
:tickets: [Jira ticket](https://hashicorp.atlassian.net/browse/ICU-5646)

## Description
Updated serialize/normalize for static credentials.
Username & Password: After creation, we don't keep track of the `password` field. We only display the `username` when reading.
SSH Private Key: After creation, we don't keep track of the `private key` or `passphrase` field. We only display the `username` when reading.

Mirage updates will be in the next PR.

### DIFF
--- a/addons/api/addon/serializers/credential.js
+++ b/addons/api/addon/serializers/credential.js
@@ -19,8 +19,7 @@ export default class CredentialSerializer extends ApplicationSerializer {
 
   serializeUsernamePassword() {
     const serialized = super.serialize(...arguments);
-    // Remove password_hmac and non-username_password type attributes
-    delete serialized['attributes']['password_hmac'];
+    // Remove non-username_password type attributes
     delete serialized['attributes']['private_key'];
     delete serialized['attributes']['passphrase'];
     return serialized;
@@ -28,8 +27,7 @@ export default class CredentialSerializer extends ApplicationSerializer {
 
   serializeSSHPrivateKey() {
     const serialized = super.serialize(...arguments);
-    // Remove private_key_hmac and non-ssh_private_key type attributes
-    delete serialized['attributes']['private_key_hmac'];
+    // Remove non-ssh_private_key type attributes
     delete serialized['attributes']['password'];
     return serialized;
   }
--- a/addons/api/addon/serializers/credential.js
+++ b/addons/api/addon/serializers/credential.js
@@ -2,13 +2,35 @@ import ApplicationSerializer from './application';
 import { copy } from 'ember-copy';
 
 export default class CredentialSerializer extends ApplicationSerializer {
-  serialize() {
-    const serialized = super.serialize(...arguments);
-    delete serialized['attributes']['password_hmac'];
-    if (serialized.type === 'username_password') {
-      delete serialized['attributes']['private_key'];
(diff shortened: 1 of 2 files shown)


lisbet-alvarez (approved):

addons/api/addon/serializers/credential.js
```diff
@@ -2,20 +2,43 @@ import ApplicationSerializer from './application';
...
+        return this.serializeUsernamePassword(...arguments);
+      case 'ssh_private_key':
+      default:
+        return this.serializeSSHPrivateKey(...arguments);
+    }
+  }
+
+  serializeUsernamePassword() {
     const serialized = super.serialize(...arguments);
+    // Remove password_hmac and non-username_password type attributes
```
ZedLi, line 22: I'm wondering if it makes sense to pick the fields we actually want for each case rather than deleting the ones we don't want? Doing the latter puts us in the situation of having to add more things to delete for each case when a new credential type is added

addons/api/addon/serializers/credential.js
```diff
@@ -2,20 +2,43 @@ import ApplicationSerializer from './application';
...
+    // Remove private_key_hmac and non-ssh_private_key type attributes
+    delete serialized['attributes']['private_key_hmac'];
+    delete serialized['attributes']['password'];
     return serialized;
   }
 
   normalize(typeClass, hash, ...rest) {
     const normalizedHash = copy(hash, true);
     const normalized = super.normalize(typeClass, normalizedHash, ...rest);
-    normalized['data']['attributes']['password'] = '';
```
ZedLi, line 18: Was password ever returned from the API? Was this not needed?

cameronperera: The password is not returned from the API. This was added when we created this for Username & Password credentials with a slight misunderstanding and what exactly was coming back.

addons/api/addon/serializers/credential.js
cameronperera: I had a similar thought but not sure the best way to accomplish that. My biggest issue is it would be a lot of duplication for the fields like name, desc, type, and username. After some testing, we actually don't need to delete the hmac attributes because they don't exist on the model. I will remove those shortly. For the other attributes, I am not entirely sure the best method.

ZedLi: That's true, it would be pretty annoying as well.

I was going to suggest that maybe we should use an attribute like `isSecret` to not serialize them like we do with `readOnly` fields but I notice that we have a similar [one that exists](https://github.com/hashicorp/boundary-ui/blob/ICU-5646-serializer/addons/api/addon/serializers/application.js#L60) already to prevent normalization, but I don't see it preventing serialization (not sure what happens the secrets object when serializing). 

Perhaps this is something we could use or do something similar on the serialization side?

ZedLi (approved):

DhariniJeeva (approved):

addons/api/addon/serializers/credential.js
cameronperera: This sounds like more great content for our FE Techtime. Thanks for the feedback though. It is great to discuss this stuff.

PR #1258, opened 2022-08-15
chore: 🤖 simplify importing of codemirror styles

randallmorey:
## Description

Simplifies codemirror style import so that consumer apps do not need to reimport it.

### DIFF
--- a/addons/rose/index.js
+++ b/addons/rose/index.js
@@ -30,6 +30,10 @@ module.exports = {
         destDir: this.name,
         include: ['**/*'],
       }),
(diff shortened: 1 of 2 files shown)


addons/rose/index.js
```diff
@@ -30,6 +30,10 @@ module.exports = {
         destDir: this.name,
         include: ['**/*'],
       }),
+      new Funnel(stylePath, {
```
ZedLi, line 33: How does this work, does the monokai theme need to be imported too? It  doesn't look like it worked for me:
image

PR #1259, opened 2022-08-16
refactor: 💡 convert brokered cred IDs array to simple attribute

randallmorey:
## Description

Replaces fragment array usage on `target.brokered_credential_source_ids` to a simple attribute.  This moves us one step closer to eliminating `ember-data-model-fragments`.


addons/api/addon/transforms/string-array.js
```diff
@@ -0,0 +1,13 @@
+import Transform from '@ember-data/serializer/transform';
+
+export default class StringArrayTransform extends Transform {
+  deserialize(serialized) {
+    const stringValues = serialized || [];
+    return stringValues.map((value) => ({ value }));
+  }
+
+  serialize(deserialized) {
```
cameronperera, line 9: I don't see these being used outside of the tests include in the PR. Are these being added for future use?

randallmorey: You're right.  This is here just for completeness.  Not used yet.

randallmorey: But it might be used by other fragment string arrays.  Won't know until they are all refactored.

cameronperera (approved):

ZedLi (approved):
LGTM, but is there a reason we needed to have the arrays of strings become arrays of objects with a `value`?

randallmorey:
> LGTM, but is there a reason we needed to have the arrays of strings become arrays of objects with a `value`?

@ZedLi this is the minimal approach, avoiding downstream changes.  The interface was already an object with `value` attribute, rather than raw strings.  The interface could still be changed later, but the point of this PR is to eliminate the dependency on model fragments first.

addons/api/tests/unit/models/target-test.js
```diff
@@ -249,11 +249,6 @@ module('Unit | Model | target', function (hooks) {
       2,
       'Target has two entires in brokered_credential_source_ids'
     );
-    assert.strictEqual(
```
DhariniJeeva, line 252: Why are we removing this?

DhariniJeeva (approved):

lisbet-alvarez (approved):

addons/api/addon/models/target.js
```diff
@@ -53,7 +53,11 @@ export default class TargetModel extends GeneratedTargetModel {
    * instances).  Unresolvable instances are excluded from the array.
    * @type {[CredentialLibraryModel, CredentialModel]}
    */
-  @computed('brokered_credential_source_ids.[]', 'store')
+  @computed(
+    'brokered_credential_source_ids',
+    'brokered_credential_source_ids.[]',
```
ZedLi, line 58: It seems like if we were to use ember arrays it's would be automatically tracked according to the [docs](https://guides.emberjs.com/release/upgrading/current-edition/tracked-properties/#toc_arrays)

addons/api/tests/unit/models/target-test.js
randallmorey: The field is now a simple array rather than a fragment array, therefore `credentialSources` which depends on it is no longer recomputed when the IDs array changes.  This computed field is recomputed only on a model reload/refresh/rollback.  So this initial zero length check computes `credentialSources`, but later checks would not result in recomputation.  Arguably this might be avoided by resetting the array to its current value (we've done this elsewhere).  I can make that update.

PR #1261, opened 2022-08-16
feat: 🎸 Add SSH Private Key mirage support

cameronperera:
✅ Closes: ICU-5647

:tickets: [Jira ticket](https://hashicorp.atlassian.net/browse/ICU-5647)

## Description
Update mirage to mock SSH Private key credential types

Screen Shot 2022-08-16 at 4 44 52 PM

### DIFF
--- a/addons/api/mirage/factories/credential.js
+++ b/addons/api/mirage/factories/credential.js
@@ -13,8 +13,6 @@ export default factory.extend({
         return generatedId('credspk_');
       case 'username_password':
         return generatedId('credup_');
-      default:
-        return generatedId('cred_');
     }
   },
   authorized_actions: () =>
--- a/addons/api/mirage/factories/credential.js
+++ b/addons/api/mirage/factories/credential.js
@@ -3,9 +3,20 @@ import { faker } from '@faker-js/faker';
 import permissions from '../helpers/permissions';
 import generatedId from '../helpers/id';
 
+const types = ['username_password', 'ssh_private_key'];
+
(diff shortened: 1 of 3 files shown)


addons/api/mirage/factories/credential.js
```diff
@@ -3,9 +3,20 @@ import { faker } from '@faker-js/faker';
...
-  type: 'username_password',
+  type: (i) => types[i % types.length],
+  id() {
+    switch (this.type) {
+      case 'ssh_private_key':
+        return generatedId('credspk_');
+      case 'username_password':
+        return generatedId('credup_');
+      default:
+        return generatedId('cred_');
```
ZedLi, line 17: Are there types that return this currently? What else is there besides the `username_password` and `ssh_private_key`?

ZedLi (approved):

addons/api/mirage/factories/credential.js
cameronperera: Versions <0.10 will only have `cred_` so I included it as the default option. However, mirage doesn't care about the version so I can remove it. I typically add a default case for lint purposes but I don't think our lint rules care about that.

lisbet-alvarez (approved):

addons/api/mirage/factories/credential.js
cameronperera: Lint has no issue with a lack of default so I just removed it.

DhariniJeeva (approved):

PR #1265, opened 2022-08-17
fix: 🐛 disable scope serialization for username_password credential type

DhariniJeeva:
tested against boundary instance
Screen Shot 2022-08-17 at 3 51 12 PM


randallmorey (approved):

lisbet-alvarez (approved):

cameronperera (approved):
Thanks for making this change @DhariniJeeva!!!

PR #1266, opened 2022-08-17
Icu 5651 create spk type

lisbet-alvarez:
:tickets: [Jira ticket](https://hashicorp.atlassian.net/browse/ICU-5651)

:technologist: [Admin preview](https://boundary-ui-git-icu-5651-create-spk-type-hashicorp.vercel.app/scopes/s_awxc4f4cye/credential-stores/cs_48mihor9zp/credentials/new)

## Description
Add functionality for creating a ssh private key credential and add corresponding acceptance tests.


ui/admin/app/routes/scopes/scope/credential-stores/credential-store/credentials/new.js
```diff
@@ -19,15 +19,24 @@ export default class ScopesScopeCredentialStoresCredentialStoreCredentialsNewRou
...
+   * creating another, but reuse name/description if available.
    * @return {CredentialModel}
    */
   model({ type = 'username_password' }) {
     const { id: credential_store_id } = this.modelFor(
       'scopes.scope.credential-stores.credential-store'
     );
+    let name, description;
+    if (this.currentModel?.isNew) {
+      ({ name, description } = this.currentModel);
```
ZedLi, line 32: This is clever syntax, I haven't seen destructuring to existing variables before. 🤯

ZedLi (changes requested):
There's a few things that I know was present/copied from elsewhere but I think should be cleaned up as we're adding new code.

I commented on some things for one form but it should be applied for both forms.

ZedLi (approved):
LGTM!

cameronperera (changes requested):

cameronperera (approved):

cameronperera:
@lisbet-alvarez current, the creation of the SPK does not work because we are including the passphrase even when we don't need to and the API does not like that. I am ok with merging this work in because we are going to work on that refactor for all static credential types next.

ui/admin/app/routes/scopes/scope/credential-stores/credential-store/credentials/new.js
```diff
@@ -19,17 +19,39 @@ export default class ScopesScopeCredentialStoresCredentialStoreCredentialsNewRou
...
+   * Set query param to default type
+   * before exiting route
+   * @param {Controller} controller
+   * @param {boolean} isExiting
+   */
+  resetController(controller, isExiting) {
+    if (isExiting) {
+      controller.set('type', 'username_password');
+    }
+  }
```
lisbet-alvarez, line 53: `queryParam` values are saved even after a user exits the route. Added the above method to reset the queryParam value so that the correct card is selected when user re-enters the route to create a new credential. @cameronperera @DhariniJeeva thoughts on this.

cameronperera: Nice. I dig this approach. I didn't realize this is a problem we have but all of our routes that use a queryParam have the same issue. It even persist upon successful creation of said resource. When you go back to create a new resource, it will still be the same type of the previously created resource.

cameronperera (approved):

ui/admin/app/routes/scopes/scope/credential-stores/credential-store/credentials/new.js
lisbet-alvarez: Thank you for noticing this issue!

addons/api/addon/serializers/credential.js
```diff
@@ -22,27 +22,33 @@ export default class CredentialSerializer extends ApplicationSerializer {
...
   serializeSSHPrivateKey() {
     const serialized = super.serialize(...arguments);
     // Remove non-ssh_private_key type attributes
     delete serialized['attributes']['password'];
+    if (!serialized['attributes']['passphrase']) {
+      delete serialized['attributes']['passphrase'];
+    }
+    if (!serialized['attributes']['private_key']) {
+      delete serialized['attributes']['private_key'];
+    }
```
randallmorey, line 41:
```suggestion
    if (!serialized?.attributes?.passphrase) delete serialized.attributes.passphrase;
    if (!serialized?.attributes?.private_key) delete serialized.attributes. private_key;
```

addons/api/addon/serializers/credential.js
```diff
@@ -22,27 +22,33 @@ export default class CredentialSerializer extends ApplicationSerializer {
...
   }
 
   normalize(typeClass, hash, ...rest) {
     const normalizedHash = copy(hash, true);
     const normalized = super.normalize(typeClass, normalizedHash, ...rest);
     // Remove passphrase as we don't track it after being created/updated
-    delete normalized['data']['attributes']['passphrase'];
+    normalized['data']['attributes']['passphrase'] = '';
+    normalized['data']['attributes']['password'] = '';
+    normalized['data']['attributes']['private_key'] = '';
```
randallmorey, line 51:
```suggestion
    normalized.data.attributes.passphrase = '';
    normalized.data.attributes.password = '';
    normalized.data.attributes.private_key = '';
```

ui/admin/app/routes/scopes/scope/credential-stores/credential-store/credentials/new.js
randallmorey: Stickiness may or may not be desirable.  For example, if a user wants to create a bunch of items of the same type having to change it each time might actually be undesirable.  Before we implement this UX here, lets discuss.  If we decide to do it, it ought to be holistic (there are other places where this change would need to be made).

randallmorey (changes requested):

ui/admin/app/routes/scopes/scope/credential-stores/credential-store/credentials/new.js
lisbet-alvarez: Good point, I will remove this.

addons/api/addon/serializers/credential.js
```diff
@@ -22,27 +22,29 @@ export default class CredentialSerializer extends ApplicationSerializer {
...
     delete serialized['attributes']['passphrase'];
     return serialized;
   }
-
   serializeSSHPrivateKey() {
     const serialized = super.serialize(...arguments);
     // Remove non-ssh_private_key type attributes
     delete serialized['attributes']['password'];
+    if (!serialized?.attributes?.passphrase) delete serialized.attributes.passphrase;
+    if (!serialized?.attributes?.private_key) delete serialized.attributes. private_key;
```
ZedLi, line 37:
```suggestion
    if (!serialized?.attributes?.private_key) delete serialized.attributes.private_key;
```

addons/api/addon/serializers/credential.js
```diff
@@ -22,27 +22,29 @@ export default class CredentialSerializer extends ApplicationSerializer {
         return this.serializeSSHPrivateKey(...arguments);
     }
   }
-
   serializeUsernamePassword() {
     const serialized = super.serialize(...arguments);
     // Remove non-username_password type attributes
     delete serialized['attributes']['private_key'];
```
ZedLi, line 28: Is there a reason some fields use square bracket notation while some use dot notation?

ZedLi (approved):

addons/api/addon/serializers/credential.js
lisbet-alvarez: Originally the code that was added in this PR used bracket notation to match the existing code but then was changed through a suggestion. Would you like me to change the rest of the serializer to use dot notation as well?

ZedLi: Yeah I think it'd be good to be consistent!

randallmorey (approved):

DhariniJeeva (approved):
LGTM!

PR #1269, opened 2022-08-18
Icu 5578 - introduce injected application credential on the api 

DhariniJeeva:
https://hashicorp.atlassian.net/browse/ICU-5578

this pr introduces injected application credential sources on the API

https://boundary-jzpbsj2iq-hashicorp.vercel.app/scopes/s_1sheljmjgm/targets/t_2p8vm6smug/brokered-credential-sources

https://boundary-jzpbsj2iq-hashicorp.vercel.app/scopes/s_1sheljmjgm/targets/t_2p8vm6smug/host-sources


addons/api/addon/models/target.js
```diff
@@ -31,6 +31,16 @@ export default class TargetModel extends GeneratedTargetModel {
   })
   brokered_credential_source_ids;
 
+  /**
+   * Injected Application Credential source ids are read only and can be
+   * persisted via a dedicated call to `addInjectedApplicationCredentialSources()`.
+   */
+  @fragmentArray('fragment-string', {
```
ZedLi, line 38: Does this depend on randy's PR that changed these `fragmentArray`s?

DhariniJeeva: Yes, this part of the code has been changed with that pr. I'll pull from the Main when those changes are merged

addons/api/mirage/route-handlers/target.js
```diff
@@ -0,0 +1,109 @@
...
+  const updatedAttrs = {
+    version: attrs.version,
+  };
+
+  const addToSourcesList = (originalSourcesIds, selectedSourcesIds) => {
+    const listOfSources = new Set(originalSourcesIds);
+    for (const elem of selectedSourcesIds) {
+      listOfSources.add(elem);
+    }
+    return Array.from(listOfSources);
```
randallmorey, line 27:
```suggestion
    const sourceIds = new Set([...originalSourceIds, ...selectedSourceIds);
    return [...sourceIds];
```

addons/api/mirage/route-handlers/target.js
```diff
@@ -0,0 +1,109 @@
...
+    if (selectedBrokeredCredentials) {
+      const listOfBrokeredCredentialSources = addToSourcesList(
+        originalBrokeredCredentials,
+        selectedBrokeredCredentials
+      );
+      if (listOfBrokeredCredentialSources?.length) {
+        updatedAttrs.brokeredCredentialSourceIds =
+          listOfBrokeredCredentialSources;
+        target.attributes.brokeredCredentialSourceIds =
+          listOfBrokeredCredentialSources;
```
randallmorey, line 71: Why both?  The `target.update()` call below is using the `updatedAttrs`, which commits the values to the mock record.

randallmorey (changes requested):
Looking good!  I just wanted to check on whether or not the cred source IDs are nested attributes?  They are not specified as nested in the Ember model, and it looks like they weren't nested previously.  But in a number of places they are treated as if they are nested.

addons/api/mirage/route-handlers/target.js
cameronperera: They are not nested attributes.

DhariniJeeva: It is not nested in the actual target model. But when mocking this target model, adding cred source ID fields directly to the target threw some errors in mirage as these fields are not actual models, hence I had it add those to target attributes. https://github.com/hashicorp/boundary-ui/pull/1269/files#diff-cbf1b6ba7d0d7fca2c5312169053b033bfc23b1663dc415b04e7dc30b60e2051L72

addons/api/mirage/config.js
```diff
@@ -463,32 +465,89 @@ export default function () {
...
-          if (
-            !updatedAttrs.credentialLibraryIds.includes(id) ||
-            !updatedAttrs.credentialIds.includes(id)
-          ) {
-            if (id.startsWith('cred')) {
-              updatedAttrs.credentialIds.push(id);
-            } else {
-              updatedAttrs.credentialLibraryIds.push(id);
+        if (
+          target.attributes.brokeredCredentialSourceIds?.length &&
```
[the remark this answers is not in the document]
DhariniJeeva: this change is outdated and not relevant anymore.

addons/api/mirage/route-handlers/target.js
randallmorey: The new cred source fields can't be expressed as relationships in the Mirage models the way they previously were.  As long as you omit the relationships in the Mirage model file, you should be able to treat them as regular fields without nesting them.

DhariniJeeva: actually I did try this and it didn't work because I somehow did not commit the code which removes the existing target model relationship! thanks Randy, it seems to work without the nested structure and I will update the pr

DhariniJeeva: done!

cameronperera (approved):
LGTM

PR #1271, opened 2022-08-23
feat: 🎸 Add code editor to api response display

ZedLi:
✅ Closes: ICU-5767

:tickets: [Jira ticket](https://hashicorp.atlassian.net/browse/ICU-5767)

## Description
Replaced the raw api output with code editor for some syntax highlighting

image


lisbet-alvarez (approved):

ZedLi:
We use the `monokai` theme by default but we can use other ones seen [here](https://codemirror.net/5/demo/theme.html#monokai), let me know if you think we should use a different one since this is just for JSON display and the keys and values look the same for strings.

DhariniJeeva (approved):

PR #1274, opened 2022-08-25
test: 💍 Add acceptance tests for deleting an spk credential

lisbet-alvarez:
✅ Closes: ICU-5654

:tickets: [Jira ticket](https://hashicorp.atlassian.net/browse/ICU-5654)

## Description
Acceptance tests for spk type credential added to delete-test.js


ui/admin/tests/acceptance/credential-store/credentials/delete-test.js
```diff
@@ -44,73 +46,183 @@ module(
...
+      assert.notOk(
+        find('.rose-layout-page-actions .rose-dropdown-button-danger')
+      );
+      assert.strictEqual(
+        getUsernamePasswordCredentialCount(),
+        usernameKeyPairCredentialCount
+      );
+    });
+
+    test('can accept delete username & password credential via dialog', async function (assert) {
```
ZedLi, line 143: What is the difference between this test and `can delete username & key pair credential`?

ui/admin/tests/acceptance/credential-store/credentials/delete-test.js
```diff
@@ -44,73 +46,183 @@ module(
...
+      await visit(urls.usernameKeyPairCredential);
+      await click('.rose-layout-page-actions .rose-dropdown-button-danger');
+      assert.strictEqual(
+        getUsernameKeyPairCredentialCount(),
+        usernameKeyPairCredentialCount - 1
+      );
+      assert.ok(confirmService.confirm.calledOnce);
+    });
+
+    test('cannot cancel delete username & password credential via dialog', async function (assert) {
```
ZedLi, line 175: Should these be  
```suggestion
    test('can cancel delete username & password credential via dialog', async function (assert) {
```

ui/admin/tests/acceptance/credential-store/credentials/delete-test.js
```diff
@@ -44,73 +46,183 @@ module(
...
+      assert.expect(2);
+      const confirmService = this.owner.lookup('service:confirm');
+      confirmService.enabled = true;
+      confirmService.confirm = sinon.fake.returns(reject());
```
ZedLi, line 179: I don't think this needs to be handled in this PR but should we perhaps not mock out the response and actually click through the dialog? Is there a a reason we need to do this?

ui/admin/tests/acceptance/credential-store/credentials/delete-test.js
```diff
@@ -44,73 +46,183 @@ module(
...
+        getUsernameKeyPairCredentialCount();
+      instances.usernameKeyPairCredential.authorized_actions =
+        instances.usernameKeyPairCredential.authorized_actions.filter(
+          (item) => item !== 'delete'
+        );
+      await visit(urls.usernameKeyPairCredential);
+      assert.strictEqual(currentURL(), urls.usernameKeyPairCredential);
+      assert.notOk(
+        find('.rose-layout-page-actions .rose-dropdown-button-danger')
+      );
```
cameronperera, line 136: Same suggestion here and on line 115 in regards to using `assert.dom('selector').doesNotExist();`.

cameronperera (changes requested):

ui/admin/tests/acceptance/credential-store/credentials/delete-test.js
lisbet-alvarez: There seems to be a pattern of clicking through the dialog in the update-test to make sure it works and using mocks in the delete-test file. Not sure if this needs further discussion but I'm happy to change it to using try catch blocks to actually click through the dialog.

ui/admin/tests/acceptance/credential-store/credentials/delete-test.js
lisbet-alvarez: Since this test just mocks out the dialog box, the regular delete button is used. So, the main difference is that the dialog mock is used in this test. I see why this test is confusing, so I agree with actually clicking through the buttons instead of using mocks. I will change this.

ui/admin/tests/acceptance/credential-store/credentials/delete-test.js
lisbet-alvarez: Nvm, try catch blocks are not needed.

cameronperera (approved):

PR #1276, opened 2022-08-25
chore: 🤖 create relative date helper

dvonanderson:
Create a relative date helper that outputs "time" ago format.

✅ Closes: ICU-5847

## Description
We needed a way to display a relative date in “time” ago format, so instead of building logic for that one instance in `admin/ui` , we are moving the logic from current `<RelativeDatetimeLive/>` component and make it into a helper that can be used in any template throughout the application if needed.

Screen Shot 2022-08-25 at 10 25 43
Screen Shot 2022-08-25 at 10 26 20


addons/core/addon/helpers/relative-datetime-live.js
```diff
@@ -0,0 +1,67 @@
...
+      },
+      unitMappings[0]
+    );
+    const scaledDelta =
+      delta < 0
+        ? Math.ceil(delta / greatestMeaningfulUnit.value)
+        : Math.floor(delta / greatestMeaningfulUnit.value);
+    const unit = greatestMeaningfulUnit.unit;
+    return this.intl.formatRelative(scaledDelta, {
+      unit: unit,
```
ZedLi, line 63: This variable could either be inlined or just 
```suggestion
      unit,
```

addons/core/tests/integration/helpers/relative-datetime-live-test.js
```diff
@@ -0,0 +1,38 @@
+import { module, test } from 'qunit';
+import { setupRenderingTest } from 'ember-qunit';
+import { render } from '@ember/test-helpers';
+import { hbs } from 'ember-cli-htmlbars';
+
+module('Integration | Helper | relative-datetime-live', function (hooks) {
+  setupRenderingTest(hooks);
+
+  test('it renders relative date in "time" ago format', async function (assert) {
```
ZedLi, line 9: Maybe show a few examples of a date in the future as well?

ZedLi (approved):
LGTM

randallmorey (approved):
Great work refactoring this into a helper!

addons/core/tests/integration/helpers/relative-datetime-live-test.js
dvonanderson: @ZedLi meaning?

ZedLi: As in some tests with the date in the future to show it works not just with dates in the past

dvonanderson: @ZedLi fixed

PR #1277, opened 2022-08-25
refactor: 💡 update credential serializer to delete password

cameronperera:
✅ Closes: ICU-5731

:tickets: [Jira ticket](https://hashicorp.atlassian.net/browse/ICU-5731)

## Description

update credential serializer to delete password attribute if empty
string or null allowing the user to update just the username if needed.

### DIFF
--- a/addons/api/addon/serializers/credential.js
+++ b/addons/api/addon/serializers/credential.js
@@ -28,6 +28,9 @@ export default class CredentialSerializer extends ApplicationSerializer {
     // Remove non-username_password type attributes
     delete serialized.attributes.private_key;
     delete serialized.attributes.passphrase;
+    // Remove password from the payload if null
+    if (!serialized?.attributes?.password)
+      delete serialized.attributes.password;
     return serialized;
   }
 
@@ -35,6 +38,7 @@ export default class CredentialSerializer extends ApplicationSerializer {
     const serialized = super.serialize(...arguments);
     // Remove non-ssh_private_key type attributes
     delete serialized.attributes.password;
+    // Remove private_key and/or passphrase from the payload if null
     if (!serialized?.attributes?.passphrase)
       delete serialized.attributes.passphrase;
     if (!serialized?.attributes?.private_key)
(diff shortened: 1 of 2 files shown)


addons/api/tests/unit/serializers/credential-test.js
```diff
@@ -33,6 +33,34 @@ module('Unit | Serializer | credential', function (hooks) {
     });
   });
 
+  test('it serializes username_password type correctly on create no password change', function (assert) {
```
DhariniJeeva, line 36: Can we rephrase this to improve readability?

addons/api/tests/unit/serializers/credential-test.js
```diff
@@ -64,6 +92,35 @@ module('Unit | Serializer | credential', function (hooks) {
     });
   });
 
+  test('it serializes ssh_private_key type correctly on create no private key or passphrase change', function (assert) {
```
DhariniJeeva, line 95: same comment as above

randallmorey (approved):

lisbet-alvarez (approved):

addons/api/addon/serializers/credential.js
```diff
@@ -28,13 +28,17 @@ export default class CredentialSerializer extends ApplicationSerializer {
     // Remove non-username_password type attributes
     delete serialized.attributes.private_key;
     delete serialized.attributes.passphrase;
+    // Remove password from the payload if null
+    if (!serialized?.attributes?.password)
```
ZedLi, line 32: Is this safe? Can someone be allowed to set an empty string as their password? Can the field both be null or undefined or always just one? Curious if we can make these check be a little more specific.

ZedLi (approved):

addons/api/addon/serializers/credential.js
randallmorey: Users cannot unset password.  Probably could add a comment here to that effect.

addons/api/tests/unit/serializers/credential-test.js
cameronperera: Yes. I will do that now.

addons/api/addon/serializers/credential.js
cameronperera: Correct. I can add a comment to explain that here. If the user wants to remove that credential password, they have to delete the whole thing.

DhariniJeeva (approved):

PR #1282, opened 2022-08-29
Add injected application credential support

DhariniJeeva:
This pr adds new translation strings for injected application cred sources and new routes

### DIFF
--- a/addons/core/translations/resources/en-us.yaml
+++ b/addons/core/translations/resources/en-us.yaml
@@ -329,8 +329,6 @@ target:
       add:
         title: Add Injected Application Credential Sources
         description: Select injected Application credential sources to assign to this target.
-
-
 credential-store:
   title: Credential Store
   title_plural: Credential Stores
--- a/addons/core/translations/resources/en-us.yaml
+++ b/addons/core/translations/resources/en-us.yaml
@@ -275,6 +275,7 @@ target:
     delete: Delete Target
     add-host-sources: Add Host Sources
     add-brokered-credential-sources: Add Brokered Credentials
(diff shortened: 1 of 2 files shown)


cameronperera (approved):

addons/core/translations/resources/en-us.yaml
```diff
@@ -311,6 +312,23 @@ target:
       add:
         title: Add Brokered Credential Sources
         description: Select brokered credential sources to assign to this target.
+  injected-application-credential-source:
+    title: Injected Application Credential
+    title_plural: Injected Application Credentials
+    types:
+      vault: Vault
+      username_password: Username & Password
```
ZedLi, line 320: Not blocking, but I notice this and the ssh string appears more than a few times in this file, should it maybe live in a shared spot?

ZedLi (approved):

addons/core/translations/resources/en-us.yaml
DhariniJeeva: I like that idea!  For now, it makes sense to me to keep it within resources and not use the 
[root](https://github.com/hashicorp/boundary-ui/blob/main/addons/core/translations/en-us.yaml) as these are just credential sources specific. 
One thing to note is that some of the occurrences have a `help` associated with them. Perhaps, we can add a new file within resources to move some of the most commonly used resource translations there? Open to suggestions!

DhariniJeeva: Merging this pr for now. But added this topic to our tech time to hear others' take on this.

PR #1285, opened 2022-08-29
chore: 🤖 add abilities for injected app cred sources

DhariniJeeva:
This pr adds the `ability` to add and remove injected application cred sources ad related tests.

### DIFF
--- a/addons/api/addon/abilities/target.js
+++ b/addons/api/addon/abilities/target.js
@@ -45,14 +45,7 @@ export default class TargetAbility extends ModelAbility {
   /**
    * @type {boolean}
    */
-  get canRemoveBrokeredCredentialSources() {
-    return this.hasAuthorizedAction('remove-credential-sources');
-  }
-
-  /**
-   * @type {boolean}
-   */
-  get canRemoveInjectedApplicationCredentialSources() {
+  get canRemoveCredentialSources() {
     return this.hasAuthorizedAction('remove-credential-sources');
   }
 }
--- a/addons/api/addon/abilities/target.js
+++ b/addons/api/addon/abilities/target.js
@@ -39,7 +39,7 @@ export default class TargetAbility extends ModelAbility {
    */
   get canAddInjectedApplicationCredentialSources() {
     return (
-      !this.model.isTCP && this.hasAuthorizedAction('add-credential-sources')
+      this.model.isSSH && this.hasAuthorizedAction('add-credential-sources')
     );
   }
   /**
--- a/addons/api/addon/abilities/target.js
+++ b/addons/api/addon/abilities/target.js
@@ -37,7 +37,22 @@ export default class TargetAbility extends ModelAbility {
   /**
    * @type {boolean}
    */
+  get canAddInjectedApplicationCredentialSources() {
+    return (
+      !this.model.isTCP && this.hasAuthorizedAction('add-credential-sources')
+    );
+  }
+  /**
+   * @type {boolean}
+   */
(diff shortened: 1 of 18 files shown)


addons/api/addon/abilities/target.js
```diff
@@ -34,10 +34,25 @@ export default class TargetAbility extends ModelAbility {
     return this.hasAuthorizedAction('add-credential-sources');
   }
 
+  /**
+   * @type {boolean}
+   */
+  get canAddInjectedApplicationCredentialSources() {
+    return (
+      !this.model.isTCP && this.hasAuthorizedAction('add-credential-sources')
```
cameronperera, line 42: Can we update this logic to the below suggestion? As of right now, the business rules state only SSH targets can inject credentials. `!this.model.isTCP...` is slightly confusing. It could imply we only want to exclude TCP targets from credential injection.
```suggestion
      this.model.isSSH && this.hasAuthorizedAction('add-credential-sources')
```

addons/api/addon/abilities/target.js
```diff
@@ -34,10 +34,25 @@ export default class TargetAbility extends ModelAbility {
...
    * @type {boolean}
    */
   get canRemoveBrokeredCredentialSources() {
     return this.hasAuthorizedAction('remove-credential-sources');
   }
+
+  /**
+   * @type {boolean}
+   */
+  get canRemoveInjectedApplicationCredentialSources() {
```
ZedLi, line 55: Do we need different abilities that check the same permissions? Can we call them something more generic?

cameronperera (changes requested):
Looks good. Just one suggestion.

addons/api/addon/abilities/target.js
DhariniJeeva: Good point. I'll try making it generic

DhariniJeeva: We are adding the ability to check the target type as we currently have no way of differentiating between brokered vs. application credentials. And making it generic will prevent me from checking the type such as `@model.isSSH` and I believe it is preferred to keep this logic within the ability instead of bringing it to the template/components.

ZedLi: I might just not know the context but if they both check for the same permission why do you need to differentiate between the types? 

And if we do need the type, is there a reason why `canRemoveInjectedApplicationCredentialSources` isn't also only for ssh targets as well? Shouldn't the add and remove be symmetrical?

randallmorey: Add and remove are asymmetrical.  The reason is that only SSH targets support _adding_ injected credentials in the first place.  So you can't have injected credentials on a target you're not supposed to.  Or stated in the positive:  if you have the remove action AND a credential source is assigned to a target, you can definitely remove it regardless of type.  Thus, you can express remove generically, but not add.  The question then becomes should we _treat as symmetrical_ add and remove here.  It's more of a style/philosophical question.

DhariniJeeva: The backend has the same permission for both brokered and injected credentials. But currently, only SSH targets can have both brokered and injected.  Adding a type check to `remove` seems unnecessary as the user can't delete a resource they never added(in the case of injected credentials for TCP targets). 

Had an offline discussion with Randy; I can make `remove` generic as we don't care about types for that action.

ZedLi: Makes sense, thanks for clarifying!

ZedLi (approved):
LGTM!

PR #1291, opened 2022-08-31
refactor: 💡 update password input field to be a password type

cameronperera:
## Description

update password input field to be a password type

### DIFF
--- a/ui/admin/app/components/form/credential/username_password/index.hbs
+++ b/ui/admin/app/components/form/credential/username_password/index.hbs
@@ -107,12 +107,14 @@
   {{#if (or @model.isNew form.isEditable)}}
     <form.input
       @name='password'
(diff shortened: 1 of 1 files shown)


ZedLi (approved):
LGTM, not blocking but should we have a toggle to display the field so a user can see what the password is?

cameronperera:
> 

I actually thought about that this morning. I will look at how the desktop client is doing it and see if it is something we can easily port over (or just use) in the admin UI.

cameronperera:
> LGTM, not blocking but should we have a toggle to display the field so a user can see what the password is?

I actually thought about that this morning. I will look at how the desktop client is doing it and see if it is something we can easily port over (or just use) in the admin UI.

lisbet-alvarez (approved):

PR #1293, opened 2022-08-31
chore: 🤖 Remove "ember-template-lint-plugin-prettier"

ZedLi:
## Description
Removed the `ember-template-lint-plugin-prettier` package. Enabled stylistic linting from `ember-template-lint`. Will lint all the hbs files in a different PR.


addons/core/.template-lintrc.js
```diff
@@ -1,5 +1,5 @@
 'use strict';
 
 module.exports = {
-  extends: 'octane',
```
ZedLi, line 4: Changed all of them from `octane` to `recommended` because it's just [an alias](https://github.com/ember-template-lint/ember-template-lint/blob/v3.12.0/lib/config/octane.js) for `recommended`. It's also removed in `4.0`.

dvonanderson (approved):
LGTM

cameronperera (approved):

DhariniJeeva:
I think it needs a `yarn format` within addons. Any reason why it is erroring now?

ZedLi:
> I think it needs a `yarn format` within addons. Any reason why it is erroring now?

Yeah I'm working on it, it turns out I'm not gonna be able to enable `stylistic` rules right off the bat and even turning on the `eol-last` is a headache as some of the versions aren't on the right version to enable automatic fixing. This PR will have to include quite a few format changes

PR #1295, opened 2022-08-31
Add injected credentials tab

DhariniJeeva:
This pr introduces injected app credentials on the UI. 

[jira](https://hashicorp.atlassian.net/jira/software/c/projects/ICU/boards/93?modal=detail&selectedIssue=ICU-5564&assignee=60e49552c0db53006aaf642b)

NOTE: acceptance tests will be added in a separate pr 

SSH:
Screen Shot 2022-09-01 at 3 53 22 PM
Screen Shot 2022-09-01 at 3 52 59 PM

empty message:
Screen Shot 2022-09-01 at 3 57 00 PM

TCP: 
Screen Shot 2022-09-01 at 3 53 13 PM

empty message:
Screen Shot 2022-09-01 at 3 56 47 PM


ui/admin/app/components/targets/target/actions/index.hbs
```diff
@@ -15,6 +15,16 @@
       {{t 'resources.target.actions.add-brokered-credential-sources'}}
     </dropdown.link>
   {{/if}}
+
+  {{#if (can 'addInjectedApplicationCredentialSources target' @model)}}
```
ZedLi, line 19: Has the convention always been to use the method name and not the string form (i.e. `can 'add injected application credential sources for target' @model`) that the [ember can](https://github.com/minutebase/ember-can#looking-up-abilities) convention describes?

ui/admin/app/controllers/scopes/scope/targets/target/add-injected-application-credential-sources.js
```diff
@@ -0,0 +1,51 @@
...
+  /**
+   * Filter out credential sources not already added to the target.
+   * @type {[CredentialLibraryModel, CredentialModel]}
+   */
+  @computed(
+    'model.{target.injected_application_credential_source_ids.[],credentialLibraries.[],credentials.[]}'
+  )
+  get filteredCredentialSources() {
+    // Get IDs for credential sources already added to the current target
+    const currentCredentialSourceIDs =
```
ZedLi, line 29: Not blocking: I personally prefer using a `Set` for these operations but this is a topic I added to the tech time to talk about. Curious what other's thoughts are.

In this case there's performance optimization here since a `filter` with `includes` can be slow _if_ there's a decent amount of values in either list, which is unlikely the case here. I still prefer set in these scenarios because it seems like we're using it like one and it's always faster.

ui/admin/app/templates/scopes/scope/targets/target/injected-application-credential-sources.hbs
```diff
@@ -0,0 +1,146 @@
...
+        <table.body as |body|>
+          {{#each
+            @model.injectedApplicationCredentialSources
+            as |injectedApplicationCredentialSource|
+          }}
+            <body.row as |row|>
+              <row.headerCell>
+                {{#if
+                  (eq
+                    injectedApplicationCredentialSource.type 'username_password'
```
ZedLi, line 44: Is there a property on the model to determine the type (e.g. isUsernamePassword)? I feel like we've been using a property for some models like `isSsh`

ui/admin/app/components/targets/target/actions/index.hbs
DhariniJeeva: not sure why we chose this. Randy would know more about this. I personally like the camel case convention that we currently follow.

ui/admin/app/templates/scopes/scope/targets/target/injected-application-credential-sources.hbs
DhariniJeeva: here the model is `target`, and its types are `TCP and SSH`. `username_password ` is a credential type and we can't use the credential model property here.

ZedLi: Are these not credentials models? Is the issue that it could be different type of models or are these just completely different objects? I thought we would be getting models back from this [property](https://github.com/hashicorp/boundary-ui/blob/609295c776b2716853c5bf77cd56e7fdf26c776a/addons/api/addon/models/target.js#L81)

DhariniJeeva: `injectedApplicationCredentialSources` doesn't return credentials or cred lib models. It updates the `injected_application_credential_source_ids` which Is an array field on the target model.

DhariniJeeva: You are right, this [array](https://github.com/hashicorp/boundary-ui/blob/609295c776b2716853c5bf77cd56e7fdf26c776a/addons/api/addon/models/target.js#L81) contains the credential models which are associated with the target. So I can add isUsernamePassword property to the cred model and should have access within that target

ZedLi (approved):
LGTM

ui/admin/app/routes/scopes/scope/targets/target/injected-application-credential-sources.js
```diff
@@ -0,0 +1,58 @@
...
+  model() {
+    return this.modelFor('scopes.scope.targets.target');
+  }
+
+  // =actions
+
+  /**
+   * Remove a credential library from the current target.
+   * @param {TargetModel} target
+   * @param {CredentialLibraryModel} credentialLibrary
```
cameronperera, line 47: Can we update `credential library` here as well?

cameronperera (changes requested):
Overall LGTM. Just a few small changes.

ui/admin/app/routes/scopes/scope/targets/target/injected-application-credential-sources.js
DhariniJeeva: ty for that catch! updated

lisbet-alvarez (approved):

cameronperera (approved):

PR #1297, opened 2022-09-02
test: 💍 acceptance tests for injected credentials

DhariniJeeva:
[Jira](https://hashicorp.atlassian.net/browse/ICU-5572)


cameronperera (approved):

ui/admin/tests/acceptance/targets/injected-application-credential-sources-test.js
```diff
@@ -0,0 +1,437 @@
...
+      instances.credential = instances.credentials[0];
+      instances.target = this.server.create('target', {
+        scope: instances.scopes.project,
+        type: 'ssh',
+      });
+
+      randomlySelectedCredentials = this.server.schema.credentials
+        .all()
+        .models.filter(Boolean)
+        .map((cred) => cred.id);
```
ZedLi, line 87: What does the `.models` do? Doesn't `.all()` return models?

Would this work? 
```suggestion
      randomlySelectedCredentials = this.server.schema.credentials
        .where(Boolean)
        .map((cred) => cred.id);
```

ZedLi (approved):
LGTM, just had one question

ui/admin/tests/acceptance/targets/injected-application-credential-sources-test.js
DhariniJeeva: ah, I believe this is a copy/paste error. let me clean this up

DhariniJeeva: I removed the filter part as that is not needed

PR #1298, opened 2022-09-02
Chore/enable stylistic lint rule

ZedLi:
## Description
Upgraded ember-template-lint to latest version to enable some automatic fixing. Check out migration guide [here](https://github.com/ember-template-lint/ember-template-lint/blob/v4.0.0/docs/migration/v4.md), there didn't seem to be any issues.

Enabled stylistic rules and update some of our linting issues.

### DIFF
--- a/addons/core/.template-lintrc.js
+++ b/addons/core/.template-lintrc.js
@@ -1,7 +1,7 @@
 'use strict';
 
 module.exports = {
-  extends: ['recommended'],
+  extends: ['recommended', 'stylistic'],
   rules: {
     quotes: 'single',
     'eol-last': 'editorconfig',
--- a/addons/core/.template-lintrc.js
+++ b/addons/core/.template-lintrc.js
@@ -5,5 +5,6 @@ module.exports = {
   rules: {
     quotes: 'single',
     'self-closing-void-elements': false,
+    'block-indentation': false,
   },
 };
--- a/addons/core/.template-lintrc.js
+++ b/addons/core/.template-lintrc.js
@@ -4,6 +4,6 @@ module.exports = {
   extends: ['recommended', 'stylistic'],
   rules: {
     quotes: 'single',
-    'eol-last': 'editorconfig',
+    'self-closing-void-elements': false,
   },
 };
--- a/addons/core/addon/helpers/resource-filter.js
+++ b/addons/core/addon/helpers/resource-filter.js
@@ -11,9 +11,9 @@ import { action } from '@ember/object';
  *
  * @example
  *
- *   {{#with resource-filter 'status' as |param|}}
+ *   {{#let resource-filter 'status' as |param|}}
  *     {{param.name}} / {{param.value}} / {{param.selectedValue}}
- *   {{/with}}
+ *   {{/let}}
  */
 export default class ResourceFilterHelper extends Helper {
   // =services
--- a/addons/core/translations/en-us.yaml
+++ b/addons/core/translations/en-us.yaml
@@ -53,6 +53,7 @@ themes:
   light: Light
   dark: Dark
 onboarding:
+  title: Onboarding
   quick-start:
     titles:
       choose-path: Welcome to Boundary
--- a/addons/rose/tests/dummy/app/templates/index.hbs
+++ b/addons/rose/tests/dummy/app/templates/index.hbs
@@ -1,18 +1,21 @@
 <Rose::Toolbar as |toolbar|>
-  <Rose::Button
-    @style='inline-link-action'
-    @iconLeft='flight-icons/svg/x-16'
-  >Clear somthing</Rose::Button>
+  <Rose::Button @style='inline-link-action' @iconLeft='flight-icons/svg/x-16'>
+    Clear something
+  </Rose::Button>
   <toolbar.primary>
     <Rose::Form>
       <Rose::Button
         @style='inline-link-action'
         @iconLeft='flight-icons/svg/x-16'
-      >Clear somthing</Rose::Button>
+      >
+        Clear something
+      </Rose::Button>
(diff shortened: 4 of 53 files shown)


addons/core/.template-lintrc.js
```diff
@@ -1,9 +1,10 @@
 'use strict';
 
 module.exports = {
-  extends: ['recommended'],
+  extends: ['recommended', 'stylistic'],
   rules: {
     quotes: 'single',
-    'eol-last': 'editorconfig',
+    'self-closing-void-elements': false,
+    'block-indentation': false,
```
ZedLi, line 8: Turned off the `block-indentation` rule as it seems like the `--fix` option is buggy and will sometimes crash the linter. This might better be manually handled with prettier.

addons/core/.template-lintrc.js
```diff
@@ -1,9 +1,10 @@
 'use strict';
 
 module.exports = {
-  extends: ['recommended'],
+  extends: ['recommended', 'stylistic'],
   rules: {
     quotes: 'single',
-    'eol-last': 'editorconfig',
+    'self-closing-void-elements': false,
```
ZedLi, line 7: This [rule](https://github.com/ember-template-lint/ember-template-lint/blob/v3.12.0/docs/rule/self-closing-void-elements.md) didn't seem that consequential and it conflicted with prettier so I turned it off.

Let me know if you think we should keep it on.

ui/admin/app/components/form/group/add-members/index.hbs
```diff
@@ -1,3 +1,4 @@
+{{!-- template-lint-disable no-whitespace-for-layout  --}}
```
ZedLi, line 1: Had to add turn off the individual rule here and in `add-principals` as it seems we're using layout whitespace as part of our design and have a TODO to come back and fix this.

addons/core/addon/helpers/resource-filter.js
```diff
@@ -11,9 +11,9 @@ import { action } from '@ember/object';
  *
  * @example
  *
- *   {{#with resource-filter 'status' as |param|}}
+ *   {{#let resource-filter 'status' as |param|}}
```
ZedLi, line 14: These changes from `with` to `let` are the most interesting and probably could use the most scrutiny in this PR. 

I actually couldn't find documentation myself before and didn't know the difference between the two (except that `with` is a base handlebars helper) but this [rule](https://github.com/ember-template-lint/ember-template-lint/blob/v3.12.0/docs/rule/no-with.md) and the accompanying [RFC](https://github.com/emberjs/rfcs/blob/master/text/0445-deprecate-with.md) that outlines the deprecation of `with` goes into much more detail.

The way we use `with` should be able to just be replaced with `let` and when I tested it looked fine but definitely could use some extra eyes on this.

ui/admin/.template-lintrc.js
```diff
@@ -19,6 +18,8 @@ module.exports = {
       ],
     },
     quotes: 'single',
-    'eol-last': 'editorconfig',
+    'self-closing-void-elements': false,
+    'block-indentation': false,
+    'no-route-action': false,
```
ZedLi, line 23: This is a new [rule](https://github.com/ember-template-lint/ember-template-lint/blob/v3.12.0/docs/rule/no-route-action.md) that actually discourages the use of `ember-route-action-helper` in place of just using controllers. 

I had to disable this as we use it everywhere but I added a topic to our tech time to discuss if we should migrate off of this.

addons/core/translations/en-us.yaml
```diff
@@ -53,6 +53,7 @@ themes:
   light: Light
   dark: Dark
 onboarding:
+  title: Onboarding
```
ZedLi, line 56: Had to add this as it seems we were using a bare string in `onboarding.hbs`

ui/admin/.template-lintrc.js
calcaide: AMAZING!!! I love the way to managed this and bring it to FE tech time, Thanks!

addons/rose/tests/dummy/app/templates/index.hbs
```diff
@@ -1,26 +1,31 @@
 <Rose::Toolbar as |toolbar|>
-  <Rose::Button
-    @style='inline-link-action'
-    @iconLeft='flight-icons/svg/x-16'
-  >Clear somthing</Rose::Button>
+  <Rose::Button @style='inline-link-action' @iconLeft='flight-icons/svg/x-16'>
```
calcaide, line 2: To keep consitency with lines below, Do you think we can move the `@style` and `@iconLeft` to new nested lines?

ui/admin/app/components/form/group/add-members/index.hbs
calcaide: Feel free to create a Jira task under tech debt 😉

calcaide (approved):
LGTM!! Amazing work @ZedLi

DhariniJeeva (approved):

ui/admin/app/components/form/group/add-members/index.hbs
ZedLi: Good point, I'll create tickets for these!

addons/rose/tests/dummy/app/templates/index.hbs
ZedLi: I had used prettier to format it and I think it didn't move them into newlines because it didn't go past the character length limit we have of 80 (it was at 79). 

The other ones have whitespace for indentation which pushed them past 80 so it nested them. 
It is a little weird that it doesn't look consistent but that's what the formatter does 🤷

PR #1302, opened 2022-09-06
chore: 🤖 enable ssh-target for production

DhariniJeeva:
:tickets: [Jira ticket](https://hashicorp.atlassian.net/browse/ICU-5951)


cameronperera (approved):
LGTM! We might want to go a head and remove the references to `ssh-target` for the other ENV since line 17 will set it to true for everything.

ZedLi (approved):

lisbet-alvarez (approved):

PR #1305, opened 2022-09-07
feat: 🎸 Feature flag for `oss` and `hcp` binary

dvonanderson:
✅ Closes: ICU-5903

:tickets: [Jira ticket](https://hashicorp.atlassian.net/browse/ICU-5903)

## Description
Create feature flag for `oss` and `hcp` to change specific input fields based on binary editions

[Vercel boundary-ui](https://boundary-ui-git-icu-5903-create-a-new-feature-c074d5-hashicorp.vercel.app/)

:technologist: [Admin preview](https://www.figma.com/file/ukvUqtTU2Bc2iULxZxw26k/Self-managed-workers-(BYOW)?node-id=1801%3A19663)

- Added two additional feature flags called `byow-pki-cluster-id` and `byow-pki-upstreams` to the environment configuration.  If the user has `oss` binary, they will get the `initial_upstreams` field and if they are using `hcp` binary they will get the `hcp_cluster_id` field.
-  Also added conditional logic to the `workerConfigText` property  to recognize what binary edition the user is using and change the output value of the `<Rose::CodeEditor>`
- Wrote test to check for input label values if either feature flag has a 'true' value. 

- `initial_upstreams` for `oss`
image
- 'boundary-cluster_id` for `hcp`
image


ui/admin/app/components/form/worker/create-worker-led/index.js
```diff
@@ -1,27 +1,44 @@
...
+   * @type {string}
+   */
   get createConfigText() {
     return `mkdir ${this.configFilePath || '<path>'}/ ;\\
 touch ${this.configFilePath || '<path>'}/pki-worker.hcl`;
   }
 
+  /**
+   * Returns config creation text for `<Rose::CodeEditor>`. The user will see
+   * different ouptuts based on if the are using `hcp` or `oss` binaries.
```
ZedLi, line 30:
```suggestion
   * different outputs based on if they are using `hcp` or `oss` binaries.
```

ui/admin/app/components/form/worker/create-worker-led/index.js
```diff
@@ -1,27 +1,44 @@
...
+   * @type {string}
+   */
   get workerConfigText() {
-    return `hcp_boundary_cluster_id = "${this.clusterId || '<config_id>'}"
-
-listener "tcp" {
+    return `${
+      this.features.isEnabled('byow-pki-hcp-cluster-id')
+        ? `hcp_boundary_cluster_id = "${this.clusterId || '<config_id>'}"`
+        : `listener "tcp" {
```
ZedLi, line 37: Why is the listener part of the config being removed for the hcp edition?

ui/admin/app/components/form/worker/create-worker-led/index.js
```diff
@@ -35,18 +52,39 @@ worker {
...
+  },
+  ${
+    this.features.isEnabled('byow-pki-upstream')
+      ? `initial_upstreams = [${
+          this.initialUpstreams
+            ? this.initialUpstreams
+                .split(',')
+                .filter((tag) => tag.trim())
+                .map((tag) => `"${tag?.trim()}"`)
+                .join(', ')
```
ZedLi, line 65: These aren't tags anymore, I think we should just extract out a utility function to handle this since it's the exact same code.

ui/admin/app/components/form/worker/create-worker-led/index.js
```diff
@@ -35,18 +52,39 @@ worker {
...
+            : '\n\t"<upstream1>",\n\t"<upstream2>",\n\t"<upstream3>"\n'
```
ZedLi, line 66: We shouldn't use tabs here since we use spaces everywhere else and users are going to be copying this config.

ui/admin/app/components/form/worker/create-worker-led/index.js
```diff
@@ -35,18 +52,39 @@ worker {
...
+        }  ]`
+      : ''
   }
 }`;
   }
-
+  /**
+   * Returns boundar installation command, and worker worker server config
+   * for for `<Rose::CodeEditor>`.
```
ZedLi, line 74:
```suggestion
   * Returns boundary installation command and start worker server command
   * for `<Rose::CodeEditor>`.
```

addons/core/translations/en-us.yaml
```diff
@@ -15,6 +15,7 @@ titles:
   home: Home
   iam: IAM
   new: New
+  pki: New PKI Worker
```
ZedLi, line 18: I think this probably should live in the specific resource file instead of the root translation?

ui/admin/tests/acceptance/workers/create-test.js
```diff
@@ -38,6 +39,28 @@ module('Acceptance | workers | create', function (hooks) {
     assert.strictEqual(getWorkersCount(), workersCount + 1);
   });
 
+  test('cluster id input field is visible for `hcp` binary', async function (assert) {
+    assert.expect(1);
+    const config = getOwner(this).resolveRegistration('config:environment');
+    config.featureFlags['byow-pki-hcp-cluster-id'] = true;
+    config.featureFlags['byow-pki-upstream'] = false;
+    await visit(newWorkerURL);
+    assert
```
ZedLi, line 48: Can you also assert the field in each test doesn't exist?

ui/admin/app/components/form/worker/create-worker-led/index.js
dvonanderson: @zed yea that was mistake coming in the next commit

dvonanderson: @ZedLi inline conditional wasn't working for the `workerConfigText` getter, so I chose to do a simple `if and else` statement.  Adds a bit more code, but it allows control of spacing that is needed for the text output.

ui/admin/app/components/form/worker/create-worker-led/index.js
dvonanderson: @ZedLi fixed

addons/core/translations/en-us.yaml
dvonanderson: @ZedLi  fixed

ui/admin/tests/acceptance/workers/create-test.js
dvonanderson: @ZedLi what do you think of the `doesNotIncludeText` assertion?

ui/admin/app/components/form/worker/create-worker-led/index.js
```diff
@@ -1,22 +1,48 @@
...
   @tracked configFilePath;
   @tracked workerTags;
+  @tracked initialUpstreams;
+
+  // =methods
+
+  textFilter(val) {
+    return val
+      .split(',')
+      .filter((itm) => itm.trim())
```
ZedLi, line 22: what does `itm` stand for?

ui/admin/app/components/form/worker/create-worker-led/index.js
```diff
@@ -1,22 +1,48 @@
...
   @tracked generatedWorkerAuthToken;
   @tracked clusterId;
   @tracked ipAddress;
   @tracked configFilePath;
   @tracked workerTags;
+  @tracked initialUpstreams;
+
+  // =methods
+
+  textFilter(val) {
```
ZedLi, line 19: `textFilter` isn't very descriptive, maybe something like `convertCommaSeparatedValuesToArray` or something similar?

ui/admin/app/components/form/worker/create-worker-led/index.js
```diff
@@ -1,22 +1,48 @@
...
   }
 
+  /**
+   * Returns config creation text for `<Rose::CodeEditor>`. The user will see
+   * different outputs based on if they are using `hcp` or `oss` binaries.
+   * @type {string}
+   */
   get workerConfigText() {
-    return `hcp_boundary_cluster_id = "${this.clusterId || '<config_id>'}"
+    if (this.features.isEnabled('byow-pki-hcp-cluster-id')) {
```
ZedLi, line 44: Rather than wrap everything in an if statement and duplicate sections, I think you could just have a variable that builds up the config and do the if statement for the sections that might be there and concatenate them if they are. 
e.g 
```
let config;
if (flag) {
  config += `${conditonal field || 'field'}`
}
...
```


Alternatively you could pull each section out to a variable and just have an if statement (or switch) and build the configs with the variables.
e.g.
```
const listener = `
listener "tcp" {
  ...
`

if (flag) {
  return `
  ${boundaryClusterId}
  ${listener}
  ${worker}
  ...
`
}
```

ui/admin/tests/acceptance/workers/create-test.js
ZedLi: I think instead of looking for a specific `nth-child` that for something doesn't exist you should do a `find` for the label that shouldn't exist and confirm it's not found

addons/core/translations/en-us.yaml
```diff
@@ -89,4 +89,3 @@ onboarding:
       learn: Learn about Boundary Desktop app
       cli: Learn about Boundary command-line (CLI)
       api: Learn about Boundary API
-
```
ZedLi, line 92: I think this newline got removed

ZedLi (changes requested):

ui/admin/app/components/form/worker/create-worker-led/index.js
calcaide: I do agree with @ZedLi. Can we use a more descriptive name for what this function is doing?

ui/admin/app/components/form/worker/create-worker-led/index.hbs
```diff
@@ -5,42 +5,54 @@
...
-    @name='cluster_id'
-    @value={{this.clusterId}}
-    @label={{t 'resources.worker.form.cluster_id.label'}}
-    @helperText={{t 'resources.worker.form.cluster_id.help'}}
-    @linkText={{t 'actions.learn-more'}}
-    @link={{doc-url 'worker.manage-workers'}}
-    placeholder='69b6ddb3-ffec-42ab-a994-f43a6519470a'
-  />
+  <h3>{{t 'resources.worker.form.steps.1.title'}}</h3>
+  <div class='worker-create-section'>
```
calcaide, line 9: I am wondering what is the purpose of this div and the new introduced class. Are we going to use it to give specific styles?

ZedLi (changes requested):

ui/admin/app/components/form/worker/create-worker-led/index.hbs
ZedLi: It's just some custom CSS to handle the custom styles for this form since it's a bit different from our `fieldset` component and we had some custom styling for things like icons

ZedLi (approved):

calcaide (approved):
👍

PR #1306, opened 2022-09-09
chore: 🤖 Add feature flags for dev and test environments

calcaide:
Add feature flags for development and test environments.

### Context:
Feature flags gives us the avility to turn `ON` or `OFF` different feature flags depending on the Boundary UI edition (OSS, enterprise and HCP) we want to build/deploy.

### Problem:
Within development and test environments not all the feature flags are turned `ON`.

This makes more difficult some of the development tasks. 

Also makes Admin UI Vercel deployment to not show all the features. Remember Vercel deployment reflects a development environment.

### Solution:
Override all the new feature flags: `static-credentials`, `byow` and `ssh-target` within `development` and `test` environments.

This way ALL features will be available at development and test time. Also Admin UI Vercel deployment will have all these features available.

### Screenshots:

On the screenshots you can see the `featureFlags` values:

Vercel deployment BEFORE this PR (so NOT all featureFlags are turned on):
vercel-before-changes

Vercel deployment AFTER this PR (so ALL featureFlags are turned on):
Screen Shot 2022-09-09 at 2 17 32 PM

### DIFF
--- a/ui/admin/config/environment.js
+++ b/ui/admin/config/environment.js
@@ -125,7 +125,10 @@ module.exports = function (environment) {
     };
 
     // Enable features in development
+    ENV.featureFlags['search'] = true;
+    ENV.featureFlags['static-credentials'] = true;
     ENV.featureFlags['byow'] = true;
+    ENV.featureFlags['ssh-target'] = true;
   }
 
   if (environment === 'test') {
@@ -145,7 +148,10 @@ module.exports = function (environment) {
     ENV.enableConfirmService = false;
 
     // Enable tests for development features
+    ENV.featureFlags['search'] = true;
+    ENV.featureFlags['static-credentials'] = true;
     ENV.featureFlags['byow'] = true;
+    ENV.featureFlags['ssh-target'] = true;
   }
 
   if (environment === 'production') {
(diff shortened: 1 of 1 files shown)


ui/admin/config/environment.js
```diff
@@ -125,7 +125,10 @@ module.exports = function (environment) {
     };
 
     // Enable features in development
+    ENV.featureFlags['search'] = true;
```
cameronperera, line 128: I don't think we need `search`. When I search the codebase I don't see it being used at all. Should we just remove it entirely?

cameronperera (approved):
I left a non-blocking comment. LGTM and thanks for looking into this for the team.

ZedLi (approved):

ui/admin/config/environment.js
calcaide: You are right @cameronperera , I assumed we still using it. I will remove it. Nice catch!! :raised_hands:

DhariniJeeva:
Ty! Can we add a comment explaining why we are adding to dev and test for enterprise? since we generally remove the dev and test flags after moving to production, it will be helpful to add a note on why we don't follow that for the enterprise features.

lisbet-alvarez (approved):
LGTM, I also appreciate the detailed description!

calcaide:
@DhariniJeeva  I am not sure I am understanding your question.

I am not understanding what you mean by "adding to dev and test for enterprise" in your question.

I think in the description of the PR is explained why we are overriding the variables within test and development environments.

The reason why usually we remove feature flags after moving to production is because then are not feature flags anymore.

If you can clarify your question, I will try to answer the best I can.

PR #1318, opened 2022-09-13
Icu 5991 UI add worker filter form field for vault type credential stores

DhariniJeeva:
:tickets: [Jira ticket](https://hashicorp.atlassian.net/browse/ICU-5991)
 
[admin-ui](https://boundary-ui-git-icu-5991-ui-add-worker-filter-9fe9ca-hashicorp.vercel.app/scopes/s_1sheljmjgm/credential-stores/cs_ucau082vov)

doc link is added [here](https://github.com/hashicorp/boundary/pull/2447)

### DIFF
--- a/addons/core/translations/resources/en-us.yaml
+++ b/addons/core/translations/resources/en-us.yaml
@@ -373,6 +373,9 @@ credential-store:
       help: A PEM-encoded private key matching the client certificate from Client Certificate.
     client_certificate_key_hmac:
       label: Client Certificate Key HMAC
+    worker_filter:
+      label: Worker Filter
+      help: Optional boolean expression to filter the workers that are allowed to satisfy this request.
 credential-library:
   title: Credential Library
   title_plural: Credential Libraries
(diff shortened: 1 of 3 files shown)


cameronperera (approved):
LGTM!

addons/core/translations/resources/en-us.yaml
```diff
@@ -373,6 +373,9 @@ credential-store:
       help: A PEM-encoded private key matching the client certificate from Client Certificate.
     client_certificate_key_hmac:
       label: Client Certificate Key HMAC
+    worker_filter:
```
ZedLi, line 376: It looks like we have an identical translation in the `forms` [translations](https://github.com/hashicorp/boundary-ui/blob/main/addons/core/translations/form/en-us.yaml#L56), should we just use that instead?

ZedLi (approved):
LGTM, just left one comment

addons/core/translations/resources/en-us.yaml
DhariniJeeva: Done!

PR #1325, opened 2022-09-14
Icu 6034 update tags in create page to support multi input fields

ZedLi:
:tickets: [Jira ticket](https://hashicorp.atlassian.net/browse/ICU-6034)

## Description
Updated creation flow to use multi input fields to create tags. This supports adding keys as part of the tags as well.

:technologist: [Admin preview](https://boundary-ui-git-icu-6034-update-tags-in-create-1eef31-hashicorp.vercel.app/)

image


randallmorey (changes requested):
Wow great working getting this together so quick.  What do you think about representing key/value pairs as and array of objects rather than arrays?  That might make the code read a little nicer than using indices.  e.g. `@value={{tags.key}}`

ui/admin/app/components/form/worker/create-worker-led/index.hbs
```diff
@@ -45,13 +45,72 @@
...
+      <:body>
+        <Rose::Table @style='condensed' as |table|>
+          <table.body as |body|>
+            {{#each this.workerTags as |tags index|}}
+              <body.row as |row|>
+                <row.cell>
+                  <form.input
+                    @type='text'
+                    @value={{get tags 0}}
+                    {{on 'input' (fn this.updateWorkerTag index 0)}}
```
randallmorey, line 62: If the tag is an object with tracked properties, manually handling the event might be unnecessary.

ZedLi: > Wow great working getting this together so quick. What do you think about representing key/value pairs as and array of objects rather than arrays? That might make the code read a little nicer than using indices. e.g. `@value={{tags.key}}`

I had originally tried using an object (for the entire structure, in place of an array of arrays) but handling tracking/binding of the keys was a massive pain and I couldn't figure out a good solution to doing it nicely.

When you say an object, do you mean making a new class that represents the tag (basically a tuple)? If so, I definitely could do that but where should the file live? I originally was thinking of doing something like that but wasn't sure what the convention was.

calcaide: Interesting stuff!! I will subscribe to this thread to follow up what solution finally implement, since this specific topic is very interesting 😄

ZedLi: I made a tag class inside the component class file (temporarily, not sure where else to put it, though it is quite specific to this use case).

It definitely does make it easier to read and is cleaner as I don't have to manually handle the tracking now.

Let me know what you think.

ui/admin/app/components/form/worker/create-worker-led/index.js
```diff
@@ -46,10 +50,19 @@ touch ${this.configFilePath || '<path>'}/pki-worker.hcl`;
...
+    ${
+      this.workerTags.length
+        ? this.workerTags
+            .map((tag) => {
+              return `${tag.key} = [${this.convertCommaSeparatedValuesToArray(
+                tag.value
+              )}]`;
+            })
+            .join('\n    ')
+        : 'key = ["<tag1>", "<tag2>"]'
```
randallmorey, line 62: There's quite a bit going on here for being directly embedded in a string.  Can this be separated out somehow?

ui/admin/app/components/form/worker/create-worker-led/index.js
```diff
@@ -121,4 +134,26 @@ boundary server -config="${this.configFilePath || '<path>'}/pki-worker.hcl"`;
...
+
+  @action
+  addWorkerTag() {
+    this.workerTags.pushObject(new Tag(this.newWorkerKey, this.newWorkerValue));
+    this.newWorkerKey = '';
+    this.newWorkerValue = '';
+  }
+
+  @action
+  removeWorkerTag(index) {
```
randallmorey, line 156: Maybe rename to `removeWorkerTagByIndex`?

randallmorey (changes requested):

ui/admin/app/components/form/worker/create-worker-led/index.js
ZedLi: I extracted it out to a function, let me know if that reads better

ui/admin/app/components/form/worker/create-worker-led/index.js
```diff
@@ -121,4 +129,36 @@ boundary server -config="${this.configFilePath || '<path>'}/pki-worker.hcl"`;
...
+      .split(',')
+      .filter((value) => value.trim())
+      .map((value) => `"${value?.trim()}"`)
+      .join(', ');
+  }
+
+  getTagConfigString() {
+    return this.workerTags
+      .map((tag) => {
+        return `${tag.key} = [${this.convertCommaSeparatedValuesToArray(
```
lisbet-alvarez, line 146: Perhaps this return keyword and brackets is not necessary since an arrow function expression is being used ? Could be changed to just 
.map((tag) => `${tag.key} = [${this.convertCommaSeparatedValuesToArray(tag.value)}]`)

lisbet-alvarez (approved):

ui/admin/app/components/form/worker/create-worker-led/index.js
ZedLi: Good call, updated it!

ui/admin/app/components/form/worker/create-worker-led/index.js
```diff
@@ -121,4 +129,35 @@ boundary server -config="${this.configFilePath || '<path>'}/pki-worker.hcl"`;
...
+      .join(', ');
+  }
+
+  getTagConfigString() {
+    return this.workerTags
+      .map(
+        (tag) =>
+          `${tag.key} = [${this.convertCommaSeparatedValuesToArray(tag.value)}]`
+      )
+      .join('\n    ');
```
DhariniJeeva, line 149: would`\t` work for the spacing here? feel like that might read better

DhariniJeeva (approved):
LGTM. left a minor note

ui/admin/app/components/form/worker/create-worker-led/index.js
ZedLi: I kept it all spaces since that's what the rest of the config is using. I tried not to mix tabs and spaces while building up the config

cameronperera (approved):

ui/admin/app/components/form/worker/create-worker-led/index.js
```diff
@@ -1,6 +1,18 @@
...
+
+class Tag {
+  @tracked key;
+  @tracked value;
+
+  constructor(key, value) {
+    this.key = key;
+    this.value = value;
+  }
+}
```
randallmorey, line 15: Since the tags UI is a client-side convenience, this seems like a great solution!  Tags aren't used by any model or other code outside of this component.  There are a handful of other examples where a class is needed, and they live as close to their uses as possible.

PR #1326, opened 2022-09-15
feat: 🎸 update the tags badge in workers list view

ZedLi:
:tickets: [Jira ticket](https://hashicorp.atlassian.net/browse/ICU-6035)

## Description
:technologist: [Admin preview](https://boundary-ui-git-icu-6035-update-tags-in-list-view-hashicorp.vercel.app/)

Instead of displaying all the tags, we are now displaying just the number of tags as there's not enough room to include both the key and value of the tags in the limited space.

image


ui/admin/app/routes/scopes/scope/workers/index.js
```diff
@@ -28,6 +28,16 @@ export default class ScopesScopeWorkersIndexRoute extends Route {
 
   model() {
     const workers = this.modelFor('scopes.scope.workers');
+    workers.forEach((worker) => {
```
ZedLi, line 31: Do we generally modify the model data? 

I couldn't figure out a good way of calculating this data (I didn't realize I couldn't just call a function from a template). I think otherwise I'd have to make a template helper for this? 

Let me know if there's a better way to accomplish this.

addons/core/translations/resources/en-us.yaml
```diff
@@ -465,6 +465,7 @@ worker:
     session_count: Session Count
     ip_address: IP Address
     release_version: Release Version
+    tag_count: '{tagCount, plural, =1 {# tag} other {# tags}}'
```
calcaide, line 468: @ZedLi 
WoW I had no idea we could something like this.

NOT BLOCKING: Not sure if we want this logic here, I always understood translation files as plain text. Would you mind to bring this to FE tech time?

addons/api/mirage/factories/worker.js
```diff
@@ -15,10 +15,30 @@ export default factory.extend({
...
+      return { os: ['ubuntu'] };
+    }
+    if (i % 2 === 0) {
+      return {
+        type: [faker.random.word(), faker.random.word()],
+        os: ['ubuntu'],
+        env: ['dev', 'local'],
+      };
+    }
+    return { type: [] };
```
randallmorey, line 29: Does the real API allow this?

addons/api/mirage/factories/worker.js
```diff
@@ -15,10 +15,30 @@ export default factory.extend({
...
+  },
+  config_tags: (i) => {
+    if (i % 3 === 0) {
+      return { os: ['ubuntu'] };
```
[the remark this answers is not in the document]
randallmorey: As it's mock data, it's just for internal use.

ui/admin/app/routes/scopes/scope/workers/index.js
randallmorey: No.  I'd suggest adding a getter on the model itself to return the tag count with some provision to cache the value, since it is unlikely to change often.

ZedLi: Yes that would make more sense, I think I was tunnel vision'd on the `workers`

ZedLi: I don't think it should be `normalizer`, it looks like those cases are dealing still with actual data from the API and just updating it to the correct format?

In this case this is an extra field that I'm calculating, I think having a getter on the model makes sense! I think I was tunnel visioning on the `workers` list and didn't think to just have it on the `worker` model so I'll update this

addons/api/mirage/factories/worker.js
ZedLi: It would be more likely the field doesn't exist, would you prefer I have the field be undefined?

addons/api/mirage/factories/worker.js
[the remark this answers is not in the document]
ZedLi: Yeah it was just an example to show key/values, it could have been anything

addons/core/translations/resources/en-us.yaml
ZedLi: I can add a topic to tech time but I think this makes the most sense here. This is meant to handle localization differences as some languages have different rules for how pluralization work (in the same way you might have it format different dates).

ui/admin/app/routes/scopes/scope/workers/index.js
calcaide: Aha!! that makes way more sense @randallmorey 

Sorry for the confusion @ZedLi !! 

Thanks for the learning 😃

addons/core/translations/resources/en-us.yaml
calcaide: Sure!!

100% Agreed on the differences between languages.

Mentioned to bring it to tech time because this is an interesting topic 😃

addons/api/mirage/factories/worker.js
ZedLi: I changed it to undefined as that's more likely to happen from the API

addons/api/addon/models/worker.js
```diff
@@ -21,6 +21,21 @@ export default class WorkerModel extends GeneratedWorkerModel {
     return this.type === 'pki';
   }
 
+  /**
+   * Returns the number of config tags present on the worker.
+   * @type {number}
+   */
+  get tagCount() {
```
calcaide, line 28: @ZedLi Nice!! 😎

calcaide (approved):

ui/admin/app/routes/scopes/scope/workers/index.js
ZedLi: Also re: caching the getter, I'm leaning towards not caching as it shouldn't be that computationally expensive, and judging from this [response](https://stackoverflow.com/a/68185081) and the official docs for [cached](https://api.emberjs.com/ember/4.6/functions/@glimmer%2Ftracking/cached) it might be worse. 

With that in mind, it actually looks like our version of ember doesn't have access to the `cached` decorator yet. If we did want to implement caching, do we normally prefer using the polyfill or just doing the more manual method by using the [cache primitives](https://api.emberjs.com/ember/3.28/modules/@glimmer%2Ftracking%2Fprimitives%2Fcache)?

lisbet-alvarez (approved):

PR #1327, opened 2022-09-15
feat: 🎸 Update Detail Page to show Tags

dvonanderson:
:tickets: [Jira ticket](https://hashicorp.atlassian.net/browse/ICU-6037)

## Description

The read-only worker tags in detailed view will now be displayed in a `key:value` pair format.
- (note) I used [ICU-6035-update-tags-in-list-view](https://github.com/hashicorp/boundary-ui/pull/1326) locally to build this out, so current view may not look the same due to changes to mirage data.

Screen Shot 2022-09-15 at 16 52 40


addons/core/translations/form/en-us.yaml
```diff
@@ -51,6 +51,9 @@ worker_last_seen:
   help: Last time the worker reported status.
 worker_address:
   help: URL and port for this worker.
+worker_tags:
```
ZedLi, line 54: I think we already have a translation for [worker tags](https://github.com/hashicorp/boundary-ui/blob/main/addons/core/translations/resources/en-us.yaml#L481) and the help text.

Maybe we should check in with what text we should stick with and be consistent.

ui/admin/app/components/form/worker/index.hbs
```diff
@@ -52,7 +52,36 @@
...
+                  @type='text'
+                  @value={{key}}
+                  @disabled={{true}}
+                  @readOnly={{true}}
+                />
+              </row.cell>
+              <row.cell>
+                <form.input
+                  @type='text'
+                  @value={{value}}
```
ZedLi, line 74: Can we display the tags with spaces between the commas rather than just outputting the array?

ZedLi (changes requested):

ui/admin/app/components/form/worker/index.hbs
```diff
@@ -52,7 +52,36 @@
       </field.errors>
     {{/if}}
   </form.textarea>
-
+  <form.fieldset>
+    <:title>{{t 'form.worker_tags.label'}}</:title>
+    <:description>{{t 'form.worker_tags.help'}}</:description>
```
ZedLi, line 57: I think these translation lookups need to be updated

lisbet-alvarez (approved):

PR #1334, opened 2022-09-16
Update fieldset

DhariniJeeva:
this pr updates fieldset to include helpertext component. This is needed to display `helpertext` in preferred-endpoints field. https://boundary-ui.vercel.app/scopes/s_1sheljmjgm/host-catalogs/hc_r9jfw7e4kg/host-sets/hs_cy5mmuagv6

### DIFF
--- a/addons/rose/addon/components/rose/form/fieldset/index.hbs
+++ b/addons/rose/addon/components/rose/form/fieldset/index.hbs
@@ -19,6 +19,16 @@
     </div>
   {{/if}}
 
+  {{#if @helperText}}
+      <Rose::Form::HelperText
+        @id='helper-text-{{this.id}}'
+        @error={{@error}}
+        @link={{@link}}
+        @linkText={{@linkText}}
+      >
+        {{@helperText}}
+      </Rose::Form::HelperText>
+    {{/if}}
   <div class='rose-form-fieldset-body'>
     {{yield to='body'}}
   </div>
--- a/addons/rose/addon/components/rose/form/fieldset/index.stories.mdx
+++ b/addons/rose/addon/components/rose/form/fieldset/index.stories.mdx
@@ -7,7 +7,7 @@ import { text, boolean, select } from '@storybook/addon-knobs';
 
 # Form Fieldset
 
-Renders a simple fieldset with title, description and body.
+Renders a simple fieldset with title, description and body and an optional helperText
 
 ```html
 <Rose::Form::Fieldset @helperText={{helperText}}>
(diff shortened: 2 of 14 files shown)


addons/rose/addon/components/rose/form/fieldset/index.stories.mdx
```diff
@@ -10,20 +10,25 @@ import { text, boolean, select } from '@storybook/addon-knobs';
 Renders a simple fieldset with title, description and body.
```
ZedLi, line 10: Should this comment also be updated?

ZedLi (approved):
LGTM! Just left one comment

One other question that's not blocking, this seems unrelated to your PR but it looks like the title, description, and body don't yield or display in the storybook docs/example, anyone know why?

addons/rose/addon/components/rose/form/fieldset/index.hbs
```diff
@@ -19,6 +19,16 @@
...
+  {{#if @helperText}}
+      <Rose::Form::HelperText
+        @id='helper-text-{{this.id}}'
+        @error={{@error}}
+        @link={{@link}}
+        @linkText={{@linkText}}
+      >
+        {{@helperText}}
+      </Rose::Form::HelperText>
+    {{/if}}
```
cameronperera, line 31: I don't think this will work exactly how we need it. Currently, `error`, `link`, and `linkText` are undefined and we just get a second line under the description. It seems like we should be updating the description block to include capabilities to handle use cases like a "learn more" link. Thoughts?

cameronperera (changes requested):

addons/rose/addon/components/rose/form/fieldset/index.hbs
ZedLi: Wouldn't those fields be passed in by the consumer of the component?
I think you wouldn't really use a description if you go with this route since the new `HelperText` includes text for that.

Perhaps I need more context but what capability does the description block even need? Couldn't you just have yielded the link yourself?

Is this to try and reuse the `HelperText` component?

DhariniJeeva: yes, the `helperText` will be used in place of the description as it includes the text and link. This practice is similar to how we handle `learn more` link in other fields. for ex: [input](https://boundary-ui-storybook.vercel.app/?path=/docs/rose-form-input--basic-field). 

Screen Shot 2022-09-19 at 12 33 17 PM

cameronperera: Yes. I was thinking about this wrong. Thanks for walking me through it with more info!

DhariniJeeva:
merging this to unblock the upcoming pr! posted the question on slack cc: @ZedLi

PR #1335, opened 2022-09-17
feat: 🎸 Make checkbox equality checks more general

ZedLi:
## Description
Changed the checkbox group component to be able to take in a function to determine equality of the items in the scenario that they're not just primitive types (e.g. objects).

This is currently needed in the filter tags PR. The tags are now not just simple strings but objects that contain a `key` and `value` field.


addons/rose/addon/components/rose/form/checkbox/group/index.js
```diff
@@ -25,13 +24,51 @@ export default class RoseFormCheckboxGroupComponent extends Component {
     const selectedItems = this.args.selectedItems || [];
     const currentItems = [...selectedItems];
 
-    if (currentItems.includes(item)) {
-      const i = currentItems.indexOf(item);
+    if (this.doesItemExist(currentItems, item)) {
```
ZedLi, line 27: Are we opposed to pulling in a package like lodash's `isEqual` to handle these equality checks? This component currently assumes the items are primitives and can't handle complex types (e.g. arrays or objects).

This would remove the need for the consumer to pass in a custom equality check (though it is more flexible in the scenario where you might want nearly similar objects to be considered equal).

randallmorey: If the arrays/objects are equal by reference, then an equality test would evaluate to true.  Is there a case where we are comparing copies of complex types?

randallmorey (changes requested):
Since the changes allow passing optional arguments, I might keep the existing stories as is (they should still work).  Maybe add one or two _new_ stories to show how to work with complex types?

addons/rose/addon/components/rose/form/checkbox/group/index.js
ZedLi: Yeah, currently I'm using the `resource-filter` and unfortunately the `allowedValues` will be different from the `selectedValue` as it deserializes the `selectedValue` from query strings which won't be the same as the ones created from `allowedValues`

addons/rose/addon/components/rose/form/checkbox/group/index.stories.mdx
```diff
@@ -17,38 +17,87 @@ Renders a form checkboxfield group.
...
+  </Story>
+</Preview>
+
+```htmlbars
+<Rose::Form::Checkbox::Group
+  @name='checkbox-group'
+  @items={{this.items}}
+  @selectedItems={{this.selectedItems}}
+  @onChange={{fn this.onChange}}
+  @itemEqualityFunc={{this.isEqual}}
```
randallmorey, line 63: This isn't necessary, right?

randallmorey (changes requested):

addons/rose/addon/components/rose/form/checkbox/group/index.stories.mdx
ZedLi: What do you mean by necessary? This is the example documentation for the new story that includes the `@itemEqualityFunc`. The old one is still [above it](https://github.com/hashicorp/boundary-ui/blob/c90d1e933d942b5a44a12934e0a4dc39329580f0/addons/rose/addon/components/rose/form/checkbox/group/index.stories.mdx?plain=1#L12)

DhariniJeeva (approved):
LGTM!

cameronperera (approved):

PR #1337, opened 2022-09-19
Icu 6036 update filter dropdown for tags in list view

ZedLi:
:tickets: [Jira ticket](https://hashicorp.atlassian.net/browse/ICU-6036)

## Description
:technologist: [Admin preview](https://boundary-ui-git-icu-6036-update-filter-dropdow-cff8ec-hashicorp.vercel.app/)

Change the dropdown filter to support filtering down by key/value pairs for the tags.
Still depends on #1335. Will rebase once it makes it in.

image

### DIFF
--- a/addons/api/addon/models/worker.js
+++ b/addons/api/addon/models/worker.js
@@ -36,6 +36,20 @@ export default class WorkerModel extends GeneratedWorkerModel {
     );
   }
 
+  /**
+   * Returns the config tags as an array of key/value pair objects.
+   * @type {[object]}
+   */
+  getConfigTagList() {
+    if (!this.config_tags) {
+      return null;
+    }
+
+    return Object.entries(this.config_tags).flatMap(([key, value]) =>
+      value.map((tag) => ({ key, value: tag }))
+    );
+  }
+
   @attr({
     description:
       'The deduplicated union of the tags reported by the worker ' +
--- a/ui/admin/app/routes/scopes/scope/workers/index.js
+++ b/ui/admin/app/routes/scopes/scope/workers/index.js
@@ -1,8 +1,6 @@
 import Route from '@ember/routing/route';
 import { action } from '@ember/object';
 import { resourceFilter } from 'core/decorators/resource-filter';
-import { tracked } from '@glimmer/tracking';
-import { A } from '@ember/array';
 
 export default class ScopesScopeWorkersIndexRoute extends Route {
   // =attributes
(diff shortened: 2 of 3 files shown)


ZedLi:
Should I update the form section component in rose to have the icon include a little bit of right-margin?

image
vs with a little extra margin that lines up the text
image

dvonanderson:
> Should I update the form section component in rose to have the icon include a little bit of right-margin?
> 
> image
> 
> vs with a little extra margin that lines up the text and doesn't make the icon/label feel as close
> image

I think either one are fine, I personally like the group things together

addons/rose/addon/components/rose/form/checkbox/group/index.js
```diff
@@ -25,13 +24,51 @@ export default class RoseFormCheckboxGroupComponent extends Component {
...
+
+  /**
+   * Checks if the item is in the current items list.
+   * If @itemEqualityFunc is not specified, will do a simple `includes` on the list.
+   * @param currentItems
+   * @param item
+   * @returns {boolean}
+   */
+  @action
+  doesItemExist(currentItems, item) {
```
randallmorey, line 45: Should this be `has` or `includes` or something similar?  The semantics are about inclusion in the list.

randallmorey: Since the selected items are an argument to the component already, we shouldn't need to pass them again here.

addons/rose/addon/components/rose/form/checkbox/group/index.js
```diff
@@ -25,13 +24,51 @@ export default class RoseFormCheckboxGroupComponent extends Component {
...
+  }
+
+  /**
+   * Finds the index of the item in the current items list
+   * If @itemEqualityFunc is not specified, will use a simple `indexOf`.
+   * @param currentItems
+   * @param item
+   * @returns {number}
+   */
+  getItemIndex(currentItems, item) {
```
randallmorey, line 65: Since the selected items are an argument to the component already, we shouldn't need to pass them again here.

randallmorey (changes requested):

addons/rose/addon/components/rose/form/checkbox/group/index.js
ZedLi: Good point. I updated these in #1335.

DhariniJeeva (approved):

PR #1338, opened 2022-09-19
refactor: 💡 translation fix

lisbet-alvarez:
## Description
Just a small translation typo for credential store that is visible when user has neither list nor create permissions.


DhariniJeeva (approved):
good catch!

PR #1344, opened 2022-09-20
Icu 6153 unknown credential support

DhariniJeeva:

## Description

### DIFF
--- a/addons/core/translations/resources/en-us.yaml
+++ b/addons/core/translations/resources/en-us.yaml
@@ -305,6 +305,7 @@ target:
       vault: Vault
       username_password: Username & Password
       ssh_private_key: Username & Key Pair
+      unknown: Unknown
     messages:
       welcome:
         title: Welcome to brokered credentials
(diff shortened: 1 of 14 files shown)


cameronperera (approved):
LGTM!
Thanks for completing this work!

addons/core/translations/resources/en-us.yaml
```diff
@@ -418,6 +409,7 @@ credential:
   types:
     username_password: Username & Password
     ssh_private_key: Username & Key Pair
+    unknown: Unknown
```
ZedLi, line 412: Should this maybe be in a more generic place since it looks like both `credential-library` and `credential` uses this and I assume more in the future?

ZedLi (approved):
LGTM, just left one comment

addons/core/translations/resources/en-us.yaml
cameronperera: Yes I agree. I don't think it is blocking for this PR since we have a large translation refactor coming up soon. We need a more generic term for referring to both credential libraries and static credentials.

DhariniJeeva: yes, we need to come up with a generic term and a spot where we can add all cred/cred library-related types.

PR #1350, opened 2022-09-23
Fix bug dc token env

ZedLi:
## Description
This sets the `env` variable for the token on the spawned child process when we connect. We shouldn't need to clean up as it's only scoped to the child process.


ui/desktop/electron-app/src/models/session.js
```diff
@@ -48,13 +48,18 @@ class Session {
...
-    });
+    return spawnAsyncJSONPromise(command)
+      .then((spawnedSession) => {
+        this.#process = spawnedSession.childProcess;
+        this.#proxyDetails = spawnedSession.response;
+        this.#process = spawnedSession.childProcess;
+        this.#id = this.#proxyDetails.session_id;
+        return this.#proxyDetails;
+      })
+      .finally(() => {
```
ZedLi, line 59: I cleaned it up when the command finishes running instead of when the session stops because it shouldn't be needed after the command runs so doesn't need to live until the user closes the session to clean up.

ui/desktop/electron-app/src/helpers/spawn-promise.js
```diff
@@ -25,9 +25,13 @@ module.exports = {
    * @param {string} command
    * @return {Promise}
    */
-  spawnAsyncJSONPromise(command) {
+  spawnAsyncJSONPromise(command, token) {
     return new Promise((resolve, reject) => {
-      const childProcess = spawn(path(), command);
+      const childProcess = spawn(path(), command, {
+        env: {
+          BOUNDARY_TOKEN: token
```
ZedLi, line 32: We should probably spread the original `...process.env` so the child processes have the same environment variables

calcaide: Good catch!

calcaide (approved):

cameronperera (approved):
LGTM

PR #1352, opened 2022-09-27
fix: 🐛 Sort workers when filtering is enabled as well

ZedLi:
## Description
Quick change to make sure workers are also sorted when filtering. Currently only unfiltered workers are sorted. 
This change will allow filtered workers to also have a stable order when refreshing workers, the filter should preserve order.


calcaide (approved):
LGTM! Thanks for the work

PR #1355, opened 2022-09-29
Chore/update linter

ZedLi:
## Description
Added back `ember-template-lint-plugin-prettier` and continue to use `ember-template-lint` in our commits (which runs prettier under the hood). Standardize on removing newlines in `hbs` files and suggest this through `editorconfig` but will also be enforced on commit by `ember-template-lint`.
I personally think it's easier at this point to just work with prettier rather than against it.

The reason for this change is:
- `ember-template-lint` by itself wasn't catching some formatting issues such as indentation
  - Its `block-indentation` rule was buggy and couldn't be turned on
- Using `yarn format` which calls prettier was unusable as it stripped out the newlines out of all files again
- Some of the formatting issues required us to now manually run prettier beforehand

Will reformat everything if we're fine with this change at the end.


cameronperera (approved):
LGTM

DhariniJeeva (approved):
Ty for taking care of this!

calcaide (approved):
Awesome! thaks for caring 🙌

calcaide (approved):
LGTM!! Thanks for the work!

Just a question, I understand on `.hbs` files we do not want the empty line at the end, is that right?

ZedLi:
> LGTM!! Thanks for the work!
> 
> Just a question, I understand on `.hbs` files we do not want the empty line at the end, is that right?

Yup correct, we're just not including the new line at the end for `hbs` files to work with prettier

PR #1356, opened 2022-09-30
chore: 🤖 Allow retrieval of arm64 cli

ZedLi:
## Description
This will allow using the arm64 cli for m1 macs. This can sometimes be a painpoint as using the `amd64` cli on m1 macs can sometimes return a rosetta error and if you're using the desktop client it will silently fail and you won't know what the issue is. 

See this [thread](https://hashicorp.slack.com/archives/C012GTH1C4E/p1664381734699799) for more info.

There was some hesitancy as this might enable building for arm64 but it looks like we're _already_ doing that with `electron-forge`. I was originally going to flag the arch for the boundary CLI based off of maybe a production env variable or flag but the default setting in `electron-forge` won't let us set the `arch` anyways so didn't see much point, see this [documentation](https://www.electronforge.io/configuration#packager-config). I can manually override it by setting it through the CLI but not through config it seems.

I would be interested in hearing why we use `electron-forge` over something like `electron-builder` which shouldn't require us to build on a specific platform in our CI and just generate all of the exectuables. I added this to tech time already.

### DIFF
--- a/ui/desktop/electron-app/config/cli.js
+++ b/ui/desktop/electron-app/config/cli.js
@@ -11,7 +11,7 @@ const downloadArtifact = (version) => {
   const archivePlatform = {};
   if (isMac()) {
     archivePlatform.name = 'darwin';
-    archivePlatform.arch = 'amd64';
+    archivePlatform.arch = os.arch();
   }
 
   if (isWindows()) {


calcaide:
@ZedLi 
You are right, seems like `electron-forge` was passing the OS Arch already. Good thing, as you mentioned last week, the Circle CI runners are running just `amd64`, so CI will not deliver any `arm64` build as it is. Good catch!

Regarding your question of `electron-forge` vs `electron-builder`. When I started on the project `electron-forge` was already in place and never checked `electron-builder`.

Are you suggesting changing the builder? Or just for this specific use case will be more convenient? In case we want to suggest a change, let's bring the topic to FE tech time.

ZedLi:
> Are you suggesting changing the builder? Or just for this specific use case will be more convenient? In case we want to suggest a change, let's bring the topic to FE tech time.

I was asking for any historical context for our choice, as I was wondering if we had just chosen it to start with but never re-evaluated. In this case, if we changed it it would be a change completely and not just for this specific use case.

lisbet-alvarez (approved):

cameronperera (approved):

DhariniJeeva (approved):

PR #1359, opened 2022-10-03
docs: ✏️ Add debug desktop client documentation

calcaide:
Add how to debug desktop client documentation.

Changes on the table of contents are result of running `doc:toc`

### DIFF
--- a/ui/desktop/README.md
+++ b/ui/desktop/README.md
@@ -6,19 +6,22 @@ The desktop client UI for Boundary.
 <!-- DON'T EDIT THIS SECTION, INSTEAD RE-RUN doctoc TO UPDATE -->
 **Table of Contents**  *generated with [DocToc](https://github.com/thlorenz/doctoc)*
 
-- [Prerequisites](#prerequisites)
-- [Installation](#installation)
-- [Yarn Commands](#yarn-commands)
-- [Running / Development](#running--development)
-  - [Developing Using Non-Release Versions of Boundary](#developing-using-non-release-versions-of-boundary)
-  - [Environment Variables](#environment-variables)
-  - [Building for Production](#building-for-production)
-    - [Environment Variables](#environment-variables-1)
-  - [Running Tests](#running-tests)
-  - [Running end to end Tests](#running-end-to-end-tests)
-  - [Deploying](#deploying)
-- [Debug desktop client](#debug-desktop-client)
-- [Further Reading / Useful Links](#further-reading--useful-links)
+- [Desktop client UI](#desktop-client-ui)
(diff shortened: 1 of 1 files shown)


ui/desktop/README.md
```diff
@@ -176,6 +179,24 @@ Steps:
 
 TBD
 
+## Debug desktop client
+
+To debug the desktop client we recommend using node debugging client. 
+
+To start the electron app with the inspector adding a breakpoint before code starts:
+
+`$ yarn start:desktop --- --inspect-brk`.
```
ZedLi, line 188: I've never seen `---`, only `--` what's that mean?

ui/desktop/README.md
```diff
@@ -176,6 +179,24 @@ Steps:
 
 TBD
 
+## Debug desktop client
+
+To debug the desktop client we recommend using node debugging client.
```
ZedLi, line 184: Could you possibly give a little more context on _when_ we would want to actually attach a debugger, maybe a scenario? I haven't come across the need yet

ZedLi (approved):
LGTM, just left two comments!

ui/desktop/README.md
calcaide: @ZedLi yeah, the documentation shows just `--`. I am not sure why we use the `---` in front of the `--inspect` but if you do not use the `---` does not work.
Sorry I can not answer the question.

ui/desktop/README.md
calcaide: @ZedLi Good Idea I will bring this topic to tech time.

That is correct, this instructions are just to debug the `main` process. I will update the documentation to reference that.

Absoulutley right, you do not need to use the `chrome` inspector. You can use different ones, as you mentioned you can use the one on webstorms. [More info about inspector clients here](https://nodejs.org/en/docs/guides/debugging-getting-started/#inspector-clients).

calcaide: I would also update the documentation with and scenario that shows the need to debug the `main` process.

I used the `main process` debug to help on the fix for use `env://` on the token. Not saying you need to the debug the main process to catch that bug, but it helped me.

ui/desktop/README.md
ZedLi: Found the answer, it's for some reason `ember electron` specific, see [docs here](https://ember-electron.js.org/docs/guides/development-and-debugging#command-line-arguments).

DhariniJeeva (approved):

ui/desktop/README.md
calcaide: @ZedLi Awesome!!! :raised_hands:

PR #1361, opened 2022-10-03
refactor: 💡 Update IPC mock for desktop client

ZedLi:
## Description
The goal of this PR is to enable the IPC mock for desktop client as right now if you run the DC locally with electron, there's no way to use the IPC mock and have target connection be mocked.

However, there is some interesting behavior I found a bit perplexing which I changed to be consistent across desktop and admin:
- The `ENABLE_MIRAGE` env variable isn't actually being used at all in admin, it's solely the `API_HOST` which determines if we will use mirage. I found this a bit misleading as our README implies that `ENABLE_MIRAGE` needs to be false alongside `API_HOST`.
  - This is half true as there's no actual way to turn off mirage in admin, we just have requests passthrough
- In development for both admin and desktop, I parse the `ENABLE_MIRAGE` env variable and if there's no variable set, default it to true
  - I want  `ember-cli-mirage`  to always have a value of `true` or `false` in dev so we can check it correctly to determine if we're using mirage, i.e. no undefined or null from not setting `ENABLE_MIRAGE`
- I removed the passthroughs completely
  - Using an `API_HOST` should require mirage to be off and therefore no passthroughs should be needed. 

Nothing should change for production environments.

### DIFF
--- a/addons/api/mirage/config.js
+++ b/addons/api/mirage/config.js
@@ -18,11 +18,6 @@ export default function () {
   // delay for each request, automatically set to 0 during testing
   this.timing = 1;
 
-  // Allow any configured API host to passthrough, which is useful in
-  // development for testing against a locally running backend.
-  if (config.api.host) this.passthrough(`${config.api.host}/**`);
-  this.passthrough();
-
   // Scope resources
 
   this.get(
--- a/addons/api/mirage/scenarios/ipc.js
+++ b/addons/api/mirage/scenarios/ipc.js
@@ -147,9 +147,9 @@ export default function initializeMockIPC(server, config) {
    * main process, and routes messages to it in a way similar to our
    * preload.js script.
    *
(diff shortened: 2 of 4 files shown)


addons/api/mirage/config.js
```diff
@@ -18,11 +18,6 @@ export default function () {
   // delay for each request, automatically set to 0 during testing
   this.timing = 1;
 
-  // Allow any configured API host to passthrough, which is useful in
-  // development for testing against a locally running backend.
-  if (config.api.host) this.passthrough(`${config.api.host}/**`);
-  this.passthrough();
```
ZedLi, line 24: Was there a reason we needed this `passthrough`? I don't understand the reasoning why we would want to have requests passthrough to our current domain, i.e. localhost?

I would actually expect if we're hitting an endpoint that's not being mocked, that it _should_ automatically fail.

PR #1363, opened 2022-10-03
chore: 🤖 clean form errors

DhariniJeeva:
jira- https://hashicorp.atlassian.net/browse/ICU-6269


calcaide (approved):
@DhariniJeeva 
From the code POC looks good to me. But I am missing some context on the changes. I understand we are stopping to use `{{@model.errors.base}}`, or we did in the past (which I might miss) and these are leftovers? 
Thank you for the woek!!

ZedLi (approved):

cameronperera (approved):

DhariniJeeva:
> @DhariniJeeva From the code POC looks good to me. But I am missing some context on the changes. I understand we are stopping to use `{{@model.errors.base}}`, or we did in the past (which I might miss) and these are leftovers? Thank you for the woek!!

@calcaide  yes, that's right. We don't need base errors anymore. maybe @zed can add more context on the history here?

ZedLi:
Ah yeah for some reason we've been carrying over the `@error` argument for our forms but it doesn't actually use it anymore if you look at the [form](https://github.com/hashicorp/boundary-ui/blob/main/addons/rose/addon/components/rose/form/index.hbs) component.

The theory is that perhaps it was vestigial when we were passing our base errors to the form which handled the displaying.

calcaide:
@DhariniJeeva @ZedLi 
Thank you for the context!!! makes all the sense and thank you for this work 😃

PR #1364, opened 2022-10-04
fix: 🐛 Fix SCSS lint issues

calcaide:
## Description

Fix SCSS lint issues

CI failing because the lint issues before the changes:
Screen Shot 2022-10-04 at 11 48 41 AM

### DIFF
--- a/ui/admin/app/styles/app.scss
+++ b/ui/admin/app/styles/app.scss
@@ -398,51 +398,53 @@
 .worker-create-section {
   border: sizing.rems(xxxxs) solid var(--ui-gray-subtler-3);
   border-radius: sizing.rems(xxs);
-  padding: sizing.rems(l);
   margin-bottom: sizing.rems(xl);
+  padding: sizing.rems(l);
 
   > div {
     margin-bottom: sizing.rems(l);
   }
 
   > code {
-    color: var(--failure);
     background-color: var(--ui-gray-subtler-5);
     border: sizing.rems(xxxxs) solid var(--ui-gray-subtler-2);
+    color: var(--failure);
   }
 
   .worker-auth-token {
-    display: flex;
-    gap: sizing.rems(m);
     align-items: flex-end;
+    display: flex;
+    gap: sizing.rems(m); // sass-lint:disable-line no-misspelled-properties
   }
 
   .worker-success {
-    display: flex;
     align-items: center;
+    display: flex;
     justify-content: space-between;
 
     > div {
-      display: flex;
-      gap: sizing.rems(xs);
       align-items: center;
+      display: flex;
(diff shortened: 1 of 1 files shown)


ui/admin/app/styles/app.scss
```diff
@@ -398,51 +398,53 @@
...
     border: sizing.rems(xxxxs) solid var(--ui-gray-subtler-2);
+    color: var(--failure);
   }
 
   .worker-auth-token {
-    display: flex;
-    gap: sizing.rems(m);
     align-items: flex-end;
+    display: flex;
+    gap: sizing.rems(m); // sass-lint:disable-line no-misspelled-properties
```
calcaide, line 417: @ZedLi linter says `gap` is misspelled, suggests to use `row-gap`. I have no context with the implementation of these styles, so I wanted to check with you before changing the `gap` to `row-gap` or to discuss possible solutions.

dvonanderson: I think that `gap` allows you to set both spacing for the `column`  and `row` without having to specifically say `column-gap` or `row-gap`

ZedLi: Yes that's right, it's just shorthand for both `row-gap` and `column-gap`, I wonder if our linter needs to just be updated to pick the syntax?

ZedLi (approved):
LGTM, thanks for fixing these warnings, totally missed them!

cameronperera (approved):
LGTM. I did a little digging and it looks like this is an issue for `row-gap` as well. Unfortunately, no activity on [this issue](https://github.com/sasstools/sass-lint/issues/1220).

ZedLi:
> LGTM. I did a little digging and it looks like this is an issue for `row-gap` as well. Unfortunately, no activity on [this issue](https://github.com/sasstools/sass-lint/issues/1220).

Ahh this is an unmaintained repo since 2017...We should probably migrate off of this perhaps to something like `stylelint`?

cameronperera:
> > LGTM. I did a little digging and it looks like this is an issue for `row-gap` as well. Unfortunately, no activity on [this issue](https://github.com/sasstools/sass-lint/issues/1220).
> 
> Ahh this is an unmaintained repo since 2017...We should probably migrate off of this perhaps to something like `stylelint`?

Potentially. I can try to migrate locally and see how that goes.

randallmorey (approved):

ui/admin/app/styles/app.scss
calcaide: All good! 😄 In this case I just add the one liner to deactivate such rule. Not a fan of this practice but the code makes sense, so better to wait for the linter to be updated (they are already working on it).
Thanks for the info guys 👍

PR #1370, opened 2022-10-11
Chore update ember modifier

calcaide:
:tickets: [Jira ticket](https://hashicorp.atlassian.net/browse/ICU-6353)

## Description

Update `ember-modifier` to the latest v3 `3.2.7` in preparation for `ember-modifier` v4 and ember 4 upgrade.

Refactored `code-mirror` modifier to a functional approach following [migration notes](https://github.com/ember-modifier/ember-modifier/blob/master/MIGRATIONS.md#migrations) from [ember-modifier project](https://github.com/ember-modifier/ember-modifier).

For testing checkout creating a new worker page and validate still working as before.

### DIFF
--- a/addons/rose/addon/modifiers/code-mirror.js
+++ b/addons/rose/addon/modifiers/code-mirror.js
@@ -42,14 +42,16 @@ const _PRESET_DEFAULTS = {
  */
 export default class CodeMirrorModifier extends Modifier {
   _editor;
+  didSetup = false;
 
-  didInstall() {
-    this._setup();
-  }
+  modify(element, positional, named) {
+    if (!this.didSetup) {
+      this._setup(element, named);
+      this.didSetup = true;
+    }
 
-  didUpdateArguments() {
-    if (this._editor?.getValue() !== this.args.named.value) {
-      this._editor?.setValue(this.args.named.value);
+    if (this._editor?.getValue() !== named.value) {
+      this._editor?.setValue(named.value);
     }
   }
 
@@ -58,9 +60,10 @@ export default class CodeMirrorModifier extends Modifier {
    * @method CodeMirror#_onChange
    * See documentation for the CodeMirror editor events, specifically 'change' event here https://codemirror.net/doc/manual.html#events.
    */
-  _onChange(editor) {
-    const { onInput } = this.args.named;
+  _onChange(editor, named) {
+    const { onInput } = named;
     const newVal = editor.getValue();
+
     onInput && onInput(newVal);
   }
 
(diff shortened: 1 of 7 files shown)


addons/rose/addon/modifiers/on-click-inside.js
```diff
@@ -5,16 +5,19 @@ import { modifier } from 'ember-modifier';
...
  * when the user clicks an item.
  */
-export default modifier((element, [callback]) => {
-  function handleClick(event) {
-    if (element.contains(event.target)) {
-      callback(element, event);
+export default modifier(
+  (element, [callback]) => {
+    function handleClick(event) {
+      if (element.contains(event.target)) {
```
ZedLi, line 11: I know this was present previously and unrelated to your change, but why is this modifier needed?

How is this different from the default `{{on "click"}}` modifier? Why do we need to check if the element contains the target, are there scenarios where we could click an element that isn't under the DOM hierarchy of where we attached this modifier?

ZedLi (approved):
LGTM, left a few non blocking comments but good work on the upgrade!

cameronperera (approved):

addons/rose/addon/modifiers/on-click-inside.js
calcaide: Unfortunately I can't answer your question @ZedLi .

As you said the modifiers were there, the goal and the scope for this PR was purely to just update the ember addon that allow those modifiers to work.

That said, I agree that these questions should be answered. So [I created this tech debt ticket for it](https://hashicorp.atlassian.net/browse/ICU-6387) (feel free to take it).

The files that have those modifiers have comments mentioning is useful for dropdowns and such, maybe we can start checking there.

ZedLi: Cool! And I was just asking for the `on-click-inside` modifier, the `on-click-outside` modifier makes perfect sense to me and is what I would expect

addons/rose/addon/modifiers/code-mirror.js
```diff
@@ -42,14 +42,16 @@ const _PRESET_DEFAULTS = {
  */
 export default class CodeMirrorModifier extends Modifier {
   _editor;
```
lisbet-alvarez, line 44: Are all the instance variables and methods that start with an underscore, basically custom properties not part of outside apis/libraries?

lisbet-alvarez (approved):

DhariniJeeva (approved):

addons/rose/addon/modifiers/on-click-inside.js
calcaide: @ZedLi Awesome!! Updated the ticket making reference to check just `on-click-inside` 😄 Thanks for the catch!

addons/rose/addon/modifiers/code-mirror.js
calcaide: Yes and No.

The `public` / `private` vars in js is a huge topic lol. This code was already there I just refactored so I am talking as I understood the code, but I could be wrong.

Long story short: the `_editor` yes, I understand this var as private (not outside of the API). The parameter `_` within `modifier()` methods it just express that we are receiving a parameter but we are not going to use it, but we have to declare it because we are receiving other parameters in the same function.

i.e.:

```javascript
testMethod(foo, bar, func){
  // Within the testMethod we will use the 3 vars foo, bar, func
}

testMethod(foo, _, func){
  // Within the testMethod we will use JUST 2 vars, foo and func
}
```

Not sure this answers your question, let me know 😉

PR #1371, opened 2022-10-11
Chore rose storybook deprecations

calcaide:
## Description

More context on this [work here](https://hashicorp.atlassian.net/browse/ICU-6367)

This PR fixes 4 deprecations for rose/storybook:
- [Deprecated implicit PostCSS loader](https://github.com/storybookjs/storybook/blob/next/MIGRATION.md#deprecated-implicit-postcss-loader).
- [Correct globs in mainjs](https://github.com/storybookjs/storybook/blob/next/MIGRATION.md#correct-globs-in-mainjs).
- [Deprecated hierarchy separator (hierarchy-separator)](https://github.com/storybookjs/eslint-plugin-storybook/blob/main/docs/rules/hierarchy-separator.md)
- [Deprecated scoped blocks imports](https://github.com/storybookjs/storybook/blob/next/MIGRATION.md#deprecated-scoped-blocks-imports)

PostCSS deprecation message before changes:
postcss-before

No PostCSS deprecation message after changes:
postcss-after

Glob in main deprecation message before changes:
glob-before

No Glob in main deprecation message after changes:
glob-after

Hierarchy separators deprecation message before:
hierarchy-separators-before

No Hierarchy separators deprecation message after:
hierarchy-separators-after

**As a side effect of Hierarchy separators, the menu has changed:**

Old version:
Screen Shot 2022-10-11 at 2 43 16 PM

New version:
Screen Shot 2022-10-11 at 2 42 15 PM

### DIFF
--- a/addons/rose/addon/components/rose/dropdown/index.stories.mdx
+++ b/addons/rose/addon/components/rose/dropdown/index.stories.mdx
@@ -1,17 +1,18 @@
-import { Meta, Story, Preview } from '@storybook/addon-docs/blocks';
+import { Meta, Story, Preview } from '@storybook/addon-docs';
 import { hbs } from 'ember-cli-htmlbars';
 import { number, boolean, object } from '@storybook/addon-knobs';
 import { action } from '@storybook/addon-actions';
 
-<Meta title="Rose|Dropdown" component="Dropdown" />
+<Meta title="Rose/Dropdown" component="Dropdown" />
 
 # Dropdown
 
 A simple dropdown component with support for links and buttons.
 
 <Preview>
-  <Story name="Basic Dropdown" height="7rem">{{
-    template: hbs`
+  <Story name="Basic Dropdown" height="7rem">
+    {{
+      template: hbs`
       <Rose::Dropdown @text="Choose One&hellip;" @count={{count}} @icon="flight-icons/svg/org-16" @showCaret={{showCaret}} as |dropdown|>
         <dropdown.link @route="index">Menu item</dropdown.link>
         <dropdown.link @route="about">Menu item</dropdown.link>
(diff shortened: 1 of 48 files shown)


cameronperera (approved):
LGTM! Thanks Carlos

addons/rose/addon/components/rose/dropdown/index.stories.mdx
```diff
@@ -148,30 +169,31 @@ form fields.
...
+      @items="{{this.items}}"
+      @selectedItems="{{this.selectedItems}}"
+      @onChange="{{fn"
+      this.onChange}}
+      as
+      |group|
+    >
+      <dropdown.item>
+        <group.checkbox
+          @label="{{group.item.name}}"
```
ZedLi, line 184: Unfortunately I think prettier doesn't know how to handle this, it thinks it's html and that the `@label` is a normal html attribute and is wrapping with it in quotes instead of leaving it alone.

I couldn't find a handlebars format but right now prettier is messing up the double curly braces so we might need to undo the formatting or figure out how to handle the handlebars format

calcaide: @ZedLi All the format was automatically applied it was nothing I did.

I need this change ASAP to keep working on storybook upgrades (this is blocking). It is possible you take care of formating issues in another PR? or what we can do to not add formating noise to this PR?

Thank you.

ZedLi: I think the easiest would be to reset to the changes you had before the commit through reflog and just skip the formatting, I think we have to figure out how to handle the html or htmlbars templating as unfortunately it doesn't work well right now

ZedLi: I was able to figure out how to get syntax highlighting correctly with handlebars in storybook but unfortunately can't undo your formatting changes, you'll have to undo those and I can cleanup the formatting in a different PR.

ZedLi: I can fix the errors in the next formatting PR

ZedLi (approved):

PR #1374, opened 2022-10-12
fix: 🐛 Session setup method deprecation message

calcaide:
:tickets: [Jira ticket](https://hashicorp.atlassian.net/browse/ICU-6390)

## Description

Fix session setup method deprecation message. More info within the Jira ticket

Deprecation message in `addons/auth` before the changes:
ember-simple-auth-before

No deprecation message in `addons/auth` after the changes:
ember-simple-auth-after

Deprecation message in `ui/admin` before changes:
adminui-simpleauth-before

NO deprecation message in `ui/admin` after changes:
desktop-simpleauth-after

Deprecation message in `ui/desktop` before changes:
desktop-simpleauth-before

NO deprecation message in `ui/desktop` after changes:

desktop-simpleauth-after


addons/auth/ember-cli-build.js
```diff
@@ -4,7 +4,10 @@ const EmberAddon = require('ember-cli/lib/broccoli/ember-addon');
 
 module.exports = function (defaults) {
   let app = new EmberAddon(defaults, {
-    // Add options here
+    'ember-simple-auth': {
+      // https://github.com/mainmatter/ember-simple-auth/blob/master/guides/upgrade-to-v4.md#call-sessionsetup-method-in-your-applicationroute
```
ZedLi, line 8: Don't we have to call `this.session.setup();` manually now in our application routes according to the docs to turn this feature on?

calcaide:
@ZedLi @randallmorey 

Sorry I did not explain the plan for this PR.

My initial plan was to change JUST the auth addon. Then consumers (admin UI and DC) will start to show deprecation message, and later on I was going to provide a PR fix those consumers.

Following the plan this PR is working NP. The consumers start show those deprecation message. 

Be aware:
- The change on `addon/auth/ember-cli-build` was to tell `ember-simple-auth` we are going to setup session method ourselves. This forces `ember-simple-auth` to expect us manually call `session.setup()` aka prepare for the upgrade and because that deprecation message stops showing within `addon/auth`.
- Within `addon/auth` we can not call `this.session.setup()` because we do not even have basic routing defined, therefore no need to do so. We do not have an app that needs run either.

**That said your comments made me realize my strategy was too cumbersome. I updated this PR to achieve next:**
- Update `addon/auth `(done in the first iteration of this PR).
- Update consumers of auth Admin UI and Desktop Client UI to manually call `session.setup()` and change `ember-cli-build` to tell `ember-simple-auth` we will set it manually.

I did also updated screenshots, now showing consumers before/after changes.

addons/auth/ember-cli-build.js
calcaide: Answered this on a comment below 😉  thank you for the feedback!

cameronperera (approved):

ZedLi (approved):

PR #1379, opened 2022-10-13
fix: 🐛 Disable add worker tag if there's no key

ZedLi:
## Description
Fixed an issue where adding a new worker tag with an empty key causes an error and the user can't add or remove tags anymore. Changed it so that the add button is disabled unless a user adds a tag key. Empty tag values are allowed as they would just be `key = []`

:technologist: [Admin preview](https://boundary-ui-git-bugfix-fix-worker-tag-error-hashicorp.vercel.app/scopes/global/workers/new)

image


calcaide (approved):
LGTM!! good catch!

cameronperera (approved):

PR #1380, opened 2022-10-17
fix: 🐛 Forgotten session setup method in core

calcaide:
## Description

Forgotten ember auth session setup method

Deprecation message in browser console before changes:
Screen Shot 2022-10-17 at 6 43 36 PM

NO Deprecation message in browser console after changes:
Screen Shot 2022-10-17 at 6 47 49 PM

### DIFF
--- a/addons/core/ember-cli-build.js
+++ b/addons/core/ember-cli-build.js
@@ -5,6 +5,9 @@ const EmberAddon = require('ember-cli/lib/broccoli/ember-addon');
 module.exports = function (defaults) {
   let app = new EmberAddon(defaults, {
     // Add options here
+    'ember-simple-auth': {
+      useSessionSetupMethod: true,
+    },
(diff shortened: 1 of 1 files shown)


ZedLi (approved):

lisbet-alvarez (approved):

cameronperera (approved):

randallmorey (commented):
Should this also call the setup method?  Or is it already called?

calcaide:
> Should this also call the setup method? Or is it already called?

@randallmorey As per my understanding [of the documentation](https://github.com/mainmatter/ember-simple-auth/blob/master/guides/upgrade-to-v4.md#call-sessionsetup-method-in-your-applicationroute), we should call `this.session.setup()` within the application route. Because core has no routes, there is no place to do such call.

PR #1381, opened 2022-10-17
feat: 🎸 remove animation when user specifies reduced motion

lisbet-alvarez:
✅ Closes: ICU-4330

:tickets: [Jira ticket](https://hashicorp.atlassian.net/browse/ICU-4330)

## Description
Refresh spinning animation is removed when user specifies reduced motion in local settings.

:technologist: [Admin preview](https://boundary-ui-git-icu-4330-refactor-prefers-redu-3a79ae-hashicorp.vercel.app/scopes/global/workers)

:desktop_computer: [Desktop preview](https://boundary-ui-desktop-git-icu-4330-refactor-pref-42d767-hashicorp.vercel.app/scopes/global/projects/targets)

Demo in Admin UI: 

https://user-images.githubusercontent.com/107949262/196240761-fbe29ce3-0fdf-4efc-a016-9210993f6849.mov

Demo in Desktop UI:

https://user-images.githubusercontent.com/107949262/196240891-0733d108-4175-4fad-a86f-a2d699084c98.mov

### DIFF
--- a/ui/admin/app/styles/app.scss
+++ b/ui/admin/app/styles/app.scss
@@ -237,6 +237,15 @@
   }
 }
 
+// No spinning animation for refresh button when user specifies reduced motion
+@media (prefers-reduced-motion: reduce) {
+  .is-loading {
+    .rose-icon {
+      animation: none;
+    }
+  }
+}
+
 //onboarding quick setup styles
 .choose-path {
   display: block;
--- a/ui/desktop/app/styles/app.scss
+++ b/ui/desktop/app/styles/app.scss
@@ -499,3 +499,12 @@ $chevron-down: url("data:image/svg+xml;utf-8,<svg viewBox='0 0 24 24' fill='%231
     }
   }
 }
+
+// No spinning animation for refresh button when user specifies reduced motion
+@media (prefers-reduced-motion: reduce) {
+  .is-loading {
+    .rose-icon {
+      animation: none;
+    }
+  }
+}


cameronperera (approved):

ui/admin/app/styles/app.scss
```diff
@@ -237,6 +237,15 @@
   }
 }
 
+// No spinning animation for refresh button when user specifies reduced motion
+@media (prefers-reduced-motion: reduce) {
```
ZedLi, line 241: Do you think we should remove the media query for `@media (prefers-reduced-motion: no-preference)` since that's the default and there's only 2 options for this media query?

calcaide: Good point!! +1 to that question!

ZedLi (approved):

calcaide (approved):
Awesome!!!  I leave a none blocking comment 😉 

I love when a11y works this good 😃

DhariniJeeva (approved):

ui/admin/app/styles/app.scss
lisbet-alvarez: I agree, really good point. I tested this, like mentioned, by removing the the "no-preference" media query but it doesn't work as expected. Not sure why but specifying "reduce motion" doesn't stop the animation when the media query is removed for "no-preference".

ZedLi: Yup I think it's fine, I realized it's most likely some of the other CSS properties are leftover even with the new media query as they don't override the same ones. 

Thanks for investigating!

randallmorey: I think this may need more investigation?  [According to MDN, `no-preference` should work across modern browsers, including mobile browsers](https://developer.mozilla.org/en-US/docs/Web/CSS/@media/prefers-reduced-motion).

lisbet-alvarez: This is late response, but below is a vid of what I was referring to. Perhaps my syntax is off somewhere?

https://user-images.githubusercontent.com/107949262/196768735-8671f7f0-8c8e-4b60-a463-0284f621832b.mov

ZedLi: Yeah this is my bad, I didn't look too closely and I was thinking we didn't need two queries to accomplish this but forgot this was already accomplished because we had used the `no-preference` option to add animation so by default there was none. We didn't need another query to handle the no animation state.

The issue when you removed the other one was that it was applying it by default now and I think the new selector wasn't targeting the right animation in the first place so it wasn't overriding it with `none`

lisbet-alvarez: No worries! 
Oh okay. That makes sense thank you!

PR #1384, opened 2022-10-18
chore: 🤖 auto-dismiss auth error notifications

randallmorey:
:tickets: [Jira ticket](https://hashicorp.atlassian.net/browse/ICU-6369)

## Description

Adds auto-dismissing to login errors, since they do not contain actionable information.

:technologist: [Admin preview](https://boundary-ui-git-chore-reduce-auth-error-noise-hashicorp.vercel.app):  submit `error` to the login name field

:desktop_computer: [Desktop preview](https://boundary-ui-desktop-git-chore-reduce-auth-error-noise-hashicorp.vercel.app/cluster-url):  submit `error` to the login name field

https://user-images.githubusercontent.com/8390120/196560113-f7b1130c-ad0d-437b-96a3-cb5dc7fda21b.mov

https://user-images.githubusercontent.com/8390120/196560122-6829f604-b0ee-4a75-b8a9-3b7463567810.mov


joshuaogle:
This looks great! I like the instant removal over animating out 👍

dvonanderson (approved):
Awesome, this is just to make sure the build up of notifications don't happen?

addons/core/addon/decorators/notify.js
```diff
@@ -67,9 +70,11 @@ export function notifyError(notification, options = { catch: false }) {
           ? intlService.t(candidateKey)
           : candidateKey;
 
+        const sticky = options.sticky === undefined ? true : options.sticky;
```
ZedLi, line 73:
```suggestion
        const sticky = options.sticky ?? true;
```

Should we also add a default to the parameter since we already do it for catch as well? i.e. `options = { catch: false, sticky: true }`

lilic:
Thank you! 🙇‍♀️

addons/core/addon/decorators/notify.js
calcaide: I like the idea of adding `sticky` to the default parameter.

addons/core/addon/decorators/notify.js
```diff
@@ -45,7 +45,10 @@ export function notifySuccess(notification) {
  *
  * @param {string|function} notification
  * @param {object} options
- * @param {object} options.catch - defaults to false
+ * @param {object} options.sticky - defaults to true, whether or not to persist
+ *                                  the notification until user dismissal
+ * @param {object} options.catch - defaults to false, whether ot not to catch
```
cameronperera, line 50:
```suggestion
 * @param {object} options.catch - defaults to false, whether or not to catch
```
Small typo here.

addons/core/addon/decorators/notify.js
cameronperera: I agree. It could be confusing to see the comment that says it defaults to true but not see that in the signature.

cameronperera (changes requested):
Looks good overall. Small typo fix suggested.

addons/core/addon/decorators/notify.js
ZedLi: Well I mean he is technically defaulting to `true` currently, I think it needs to be done since someone can still always send in an object with just the `catch` and it won't default the sticky to true so we need to check it.

DhariniJeeva (approved):

cameronperera (approved):
LGTM

calcaide (approved):
LGTM!! Nice looking!! Thanks for the comment!

addons/core/addon/decorators/notify.js
```diff
@@ -45,7 +45,10 @@ export function notifySuccess(notification) {
  *
  * @param {string|function} notification
  * @param {object} options
- * @param {object} options.catch - defaults to false
+ * @param {object} options.sticky - defaults to true, whether or not to persist
+ *                                  the notification until user dismissal
+ * @param {object} options.catch - defaults to false, whether or not to catch
+ *                                 and squelch the error
  */
 export function notifyError(notification, options = { catch: false }) {
```
randallmorey, line 53:
```suggestion
export function notifyError(notification, options = { catch: false, sticky: false }) {
```

addons/core/addon/decorators/notify.js
randallmorey: Added additional explanation and signature.

PR #1385, opened 2022-10-21
refactor: 💡 Deprecate renderTemplate in header-nav

calcaide:
:tickets: [Jira ticket](https://hashicorp.atlassian.net/browse/ICU-6017)

## Description

Deprecate the usage of `renderTemplate` in the application `header-nav`.

Created a `header-nav` component to get rid of named outlets and `renderTemplate`. Add Organizations list and Projects list to `scope` service, so the new `header-nav` component can consume the data.

[Link to Vercel deployment of this PR](https://boundary-ui-git-refactor-deprecate-rendertempl-db50f8-hashicorp.vercel.app/) free of `renderTemplate` deprecation messages in browser console.

### DIFF
--- a/addons/core/addon/services/scope.js
+++ b/addons/core/addon/services/scope.js
@@ -16,5 +16,15 @@ export default class ScopeService extends Service {
   /**
    * @type {ScopeModel}
    */
+  @tracked orgsList;
+
+  /**
+   * @type {ScopeModel}
+   */
   @tracked project;
+
+  /**
+   * @type {ScopeModel}
+   */
+  @tracked projectsList;
 }
--- a/addons/core/addon/services/scope.js
+++ b/addons/core/addon/services/scope.js
@@ -3,7 +3,7 @@ import { tracked } from '@glimmer/tracking';
 
 /**
  * This simple non-functional service is used to retain references to
- * the current active scope(s).
+ * lists of scopes (orgs and projects) and selected scopes (orgs and projects).
  */
 export default class ScopeService extends Service {
   // =attributes
--- a/addons/core/tests/unit/services/scope-test.js
+++ b/addons/core/tests/unit/services/scope-test.js
@@ -12,7 +12,7 @@ module('Unit | Service | scope', function (hooks) {
 
   test('it can contain orgs list scope', function (assert) {
     let service = this.owner.lookup('service:scope');
-    service.orgsList = {};
+    service.orgsList = [];
     assert.ok(service.orgsList, 'Service contains orgs list scope');
   });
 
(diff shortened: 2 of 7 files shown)


dvonanderson (approved):

DhariniJeeva (approved):

addons/core/addon/services/scope.js
```diff
@@ -13,8 +13,18 @@ export default class ScopeService extends Service {
    */
   @tracked org;
 
+  /**
+   * @type {ScopeModel}
+   */
+  @tracked orgsList;
```
ZedLi, line 19: I assume this is needed because the component could now be in any route and you need access to the list of scopes? 

Perhaps we should also update the comment for this service to also include now the lists instead of just the current active scope.

ZedLi (approved):

addons/core/addon/services/scope.js
randallmorey: Yes that is the reason.

addons/core/tests/unit/services/scope-test.js
```diff
@@ -10,9 +10,21 @@ module('Unit | Service | scope', function (hooks) {
     assert.ok(service.org, 'Service contains org scope');
   });
 
+  test('it can contain orgs list scope', function (assert) {
+    let service = this.owner.lookup('service:scope');
+    service.orgsList = {};
```
randallmorey, line 15: I sort of expect orgs list to be an array type.  Why is it an object in the tests?

ui/admin/app/components/header-nav/index.js
```diff
@@ -0,0 +1,10 @@
+import Component from '@glimmer/component';
+import { inject as service } from '@ember/service';
+
+export default class HeaderNavComponent extends Component {
+  // =services
+
+  @service session;
+  @service scope;
+  @service router;
```
randallmorey, line 9: Is this service used by the component?

randallmorey (approved):
Awesome work!

addons/core/addon/services/scope.js
calcaide: Yes, that is the reason behind this implementation, have a list of scopes (orgs + projects) available under any route.

Indeed, I miss the comment I will update it.

addons/core/tests/unit/services/scope-test.js
calcaide: @randallmorey 
`projects` and `orgs` are an object because when querying them, within `scopes/scope` route it returns the model class which is an object.

Even accessing the class such as `orgs.content` returns an `object` type.

Happy to discuss this over slack or zoom if you like.

ui/admin/app/components/header-nav/index.js
calcaide: You are right, this component does not need the service router, I delete it.

cameronperera (approved):

addons/core/tests/unit/services/scope-test.js
calcaide: @randallmorey Changed

PR #1386, opened 2022-10-21
Icu 6424 deprecate @storybook/addon knobs

ZedLi:
:tickets: [Jira ticket](https://hashicorp.atlassian.net/browse/ICU-6424)

## Description
Apologies for the large PR. This is to deprecate the usage of storybook's `addon-knobs` in favor of `addon-controls` as outlined [here](https://github.com/storybookjs/storybook/discussions/15060).

For the most part, this is a direct migration to controls from knobs. There were a few scenarios where I added some more controls for more control over the component as part of the story. 
For the form/radio/card component, I also simplified the wide vs tile views into one as they can be toggled with a control.

For testing, it might be helpful to use our current [rose](https://boundary-ui-storybook.vercel.app/) as a comparison. 

One unrelated change I noticed is that I'm not sure that storybook's webpack can handle our yielded named blocks syntax (i.e. `<:header>`). I occasionally got errors complaining about it from webpack and this would explain why some components aren't showing the yielded named components. If this is the case, it might be just a limitation of using ember with storybook as we use broccoli while storybook uses webpack under the hood.

:rose: [Rose preview](https://boundary-ui-storybook-git-icu-6424-deprecate-s-402098-hashicorp.vercel.app/)

### DIFF
--- a/addons/rose/.storybook/main.js
+++ b/addons/rose/.storybook/main.js
@@ -10,6 +10,7 @@ module.exports = {
     '@storybook/addon-actions',
     '@storybook/addon-docs',
     '@storybook/addon-a11y',
+    '@storybook/addon-controls',
   ],
   emberOptions: {
     polyfills: [namedBlockPolyfill],
--- a/addons/rose/.storybook/main.js
+++ b/addons/rose/.storybook/main.js
@@ -7,9 +7,9 @@ module.exports = {
   ],
   addons: [
     '@storybook/addon-knobs',
-    '@storybook/addon-actions',
     '@storybook/addon-docs',
     '@storybook/addon-a11y',
+    '@storybook/addon-actions',
     '@storybook/addon-controls',
   ],
   emberOptions: {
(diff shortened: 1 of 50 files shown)


calcaide (approved):
Awesome work Zed!!! this is very need!!

As per your comment 

> I noticed is that I'm not sure that storybook's webpack can handle our yielded named blocks syntax (i.e. <:header>)

When working on storybook deprecations + ember 4 I notice next related to what you mention:
- Indeed, specially in dev mode are some occasional weird stuff going on. But seems to me is related to stuff I mentioned below and I as soon the builds go through and work I will not worry much.
- Storybook is huuuuuge, just upgrade a minor version can have many implications such as deprecations or requirement within versions of other dependencies.
I personally think (nothing I prove) once we upgrade Storybook to more recent versions + Ember 4 these occasional problems will go away.

Many thanks for this.

DhariniJeeva (approved):
thanks!

cameronperera (approved):
Nice work!

PR #1389, opened 2022-10-24
feat: 🎸 Auto-populate hcp cluster id

ZedLi:
✅ Closes: ICU-6200

:tickets: [Jira ticket](https://hashicorp.atlassian.net/browse/ICU-6200)

## Description
This PR attempts at a front end only solution for auto-populating the HCP cluster ID. Ideally we would have an API we could hit that could return what the current cluster is and any other meta data but in the absence of that, we can attempt a best effort solution. This currently just pulls the guid from the URL and auto-populates the cluster ID.

Some other ideas I considered was passing a query string or saving the ID from the cluster in local storage from the HCP side but that requires navigation from the portal or a specific link; this is the simplest way to auto populate the ID from just a normal url. 

There was some talk in the future (see [this thread](https://hashicorp.slack.com/archives/C03FT474K1C/p1666277925120279)) that we might allow CNAME-ing so customers will have more friendly URLs. This wouldn't work in those scenarios but it would just revert to the current functionality and the user would have to manually fill it out.

Cleaned up the form: there was an unused `@cancel` and I had to pull the route action up to a parameter so it can be integration-tested easier.

Unfortunately this can't be easily tested but I included some integration tests to show what kind of URLs it will populate. Take a look [here](https://regex101.com/r/FpYBNN/1) if you want to play around with the regex yourself or want an explanation of it.

### DIFF
--- a/addons/core/addon/services/browser-object.js
+++ b/addons/core/addon/services/browser-object.js
@@ -1,0 +1,23 @@
+import Service from '@ember/service';
+
+/**
+ * A simple service that wraps around the global window and
+ * document variables that allows us to mock them in testing
+ */
+export default class BrowserObjectService extends Service {
+  /**
+   * The global window object
+   * @returns {Window}
+   */
+  get window() {
+    return window;
+  }
+
+  /**
+   * The global document object
+   * @returns {Document}
+   */
+  get document() {
+    return document;
+  }
+}
--- a/addons/core/addon/services/browser-object.js
+++ b/addons/core/addon/services/browser-object.js
@@ -18,6 +18,14 @@ export default class BrowserObjectService extends Service {
    * @returns {Document}
    */
   get document() {
-    return document;
+    return this.window?.document;
+  }
+
+  /**
+   * The hostname of the current location of the URL
+   * @returns {string}
+   */
+  get hostname() {
+    return this.window?.location?.hostname;
   }
 }
--- a/ui/admin/app/components/form/worker/create-worker-led/index.js
+++ b/ui/admin/app/components/form/worker/create-worker-led/index.js
@@ -3,7 +3,6 @@ import { tracked } from '@glimmer/tracking';
 import { inject as service } from '@ember/service';
 import { action } from '@ember/object';
 import { A } from '@ember/array';
-import window from 'ember-window-mock';
 
 class Tag {
   @tracked key;
@@ -18,6 +17,7 @@ class Tag {
 export default class FormWorkerCreateWorkerLedComponent extends Component {
   // =services
   @service features;
+  @service browserObject;
 
   // =attributes
   @tracked generatedWorkerAuthToken;
@@ -150,7 +150,7 @@ boundary server -config="${this.configFilePath || '<path>'}/pki-worker.hcl"`;
   }
 
   getClusterIdFromURL() {
-    const hostname = window.location.hostname;
+    const hostname = this.browserObject.window.location.hostname;
 
     // Match against a guid with either the prod, int, or dev hcp domain
     const matcher =
--- a/ui/admin/app/components/form/worker/create-worker-led/index.js
+++ b/ui/admin/app/components/form/worker/create-worker-led/index.js
@@ -3,6 +3,7 @@ import { tracked } from '@glimmer/tracking';
 import { inject as service } from '@ember/service';
 import { action } from '@ember/object';
 import { A } from '@ember/array';
+import window from 'ember-window-mock';
 
 class Tag {
   @tracked key;
@@ -20,7 +21,7 @@ export default class FormWorkerCreateWorkerLedComponent extends Component {
 
   // =attributes
   @tracked generatedWorkerAuthToken;
-  @tracked clusterId;
+  @tracked clusterId = this.getClusterIdFromURL();
   @tracked ipAddress;
   @tracked configFilePath;
   @tracked initialUpstreams;
@@ -29,7 +30,6 @@ export default class FormWorkerCreateWorkerLedComponent extends Component {
   @tracked newWorkerValue;
 
   // =properties
-
   /**
    * Returns directory creation text for `<Rose::CodeEditor>`.
    * @type {string}
@@ -147,6 +147,21 @@ boundary server -config="${this.configFilePath || '<path>'}/pki-worker.hcl"`;
           `${tag.key} = [${this.convertCommaSeparatedValuesToArray(tag.value)}]`
       )
       .join('\n    ');
+  }
+
+  getClusterIdFromURL() {
+    const hostname = window.location.hostname;
+
+    // Match against a guid with either the prod, int, or dev hcp domain
+    const matcher =
+      /^([0-9a-f]{8}-[0-9a-f]{4}-[1-5][0-9a-f]{3}-[89ab][0-9a-f]{3}-[0-9a-f]{12})\.boundary\.(hashicorp\.cloud|hcp\.to|hcp\.dev)$/;
+
+    if (matcher.test(hostname)) {
+      // Grab the captured guid
+      return matcher.exec(hostname)[1];
+    }
+
+    return undefined;
   }
 
   @action
--- a/ui/admin/app/components/form/worker/create-worker-led/index.js
+++ b/ui/admin/app/components/form/worker/create-worker-led/index.js
@@ -21,7 +21,7 @@ export default class FormWorkerCreateWorkerLedComponent extends Component {
 
   // =attributes
   @tracked generatedWorkerAuthToken;
-  @tracked clusterId = this.clusterIdFromURL;
+  @tracked clusterID = this.clusterIDFromURL;
   @tracked ipAddress;
   @tracked configFilePath;
   @tracked initialUpstreams;
@@ -29,16 +29,16 @@ export default class FormWorkerCreateWorkerLedComponent extends Component {
   @tracked newWorkerKey;
   @tracked newWorkerValue;
 
-  get clusterIdFromURL() {
-    const hostname = this.browserObject.window.location.hostname;
+  get clusterIDFromURL() {
+    const hostname = this.browserObject.hostname;
 
     // Match against a guid with either the prod, int, or dev hcp domain
     const matcher =
-      /^([0-9a-f]{8}-[0-9a-f]{4}-[1-5][0-9a-f]{3}-[89ab][0-9a-f]{3}-[0-9a-f]{12})\.boundary\.(hashicorp\.cloud|hcp\.to|hcp\.dev)$/;
+      /^(?<clusterID>[0-9a-f]{8}-[0-9a-f]{4}-[1-5][0-9a-f]{3}-[89ab][0-9a-f]{3}-[0-9a-f]{12})\.boundary\.(?:hashicorp\.cloud|hcp\.to|hcp\.dev)$/;
 
     if (matcher.test(hostname)) {
       // Grab the captured guid
-      return matcher.exec(hostname)[1];
+      return matcher.exec(hostname).groups.clusterID;
     }
 
     return undefined;
@@ -61,7 +61,7 @@ touch ${this.configFilePath || '<path>'}/pki-worker.hcl`;
    */
   get workerConfigText() {
     const clusterText = `hcp_boundary_cluster_id = "${
-      this.clusterId || '<config_id>'
+      this.clusterID || '<config_id>'
     }"`;
 
     const tagsText = `tags {
--- a/ui/admin/app/components/form/worker/create-worker-led/index.js
+++ b/ui/admin/app/components/form/worker/create-worker-led/index.js
@@ -33,15 +33,11 @@ export default class FormWorkerCreateWorkerLedComponent extends Component {
     const hostname = this.browserObject.hostname;
 
     // Match against a guid with either the prod, int, or dev hcp domain
-    const matcher =
+    const clusterIDMatcher =
       /^(?<clusterID>[0-9a-f]{8}-[0-9a-f]{4}-[1-5][0-9a-f]{3}-[89ab][0-9a-f]{3}-[0-9a-f]{12})\.boundary\.(?:hashicorp\.cloud|hcp\.to|hcp\.dev)$/;
 
-    if (matcher.test(hostname)) {
-      // Grab the captured guid
-      return matcher.exec(hostname).groups.clusterID;
-    }
-
-    return undefined;
+    // Return the cluster ID otherwise return null if it doesn't match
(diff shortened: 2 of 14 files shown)


cameronperera (commented):
Overall this looks great. I do have a question though.
Is there a specific reason you used this package to get the hostname versus how it is done for [desktop](https://github.com/hashicorp/boundary-ui/blob/main/ui/desktop/app/routes/cluster-url.js#L30-L35) and [onboarding](https://github.com/hashicorp/boundary-ui/blob/main/ui/admin/app/routes/onboarding/quick-setup/create-resources/success.js#L15-L20)?
Just curious is adding this new package is worth it for accessing the hostname. It does seem like setting `window` in our test is really easy. I will have to look at how we are setting the clusterUrl for desktop test.

ZedLi:
> Just curious is adding this new package is worth it for accessing the hostname.

Not sure I'm understanding the question, the package I added is only used to help mock out the global `window`. In non-testing environments it's just a passthrough to the normal global `window` object which I could access without any package.

I wasn't aware of the wrapper we've previously used with window but I'm not sure I see the benefit using it here. In those scenarios it accesses the same global `window` which you could now mock out in a route or controller.

I'm accessing the window from a component and just want to test that, I don't see much benefit to indirectly wrapping the global `window` when I'd have to still pass in a `window` to the component most likely from a route or controller to correctly mock it out. I brought in the package to simplify this aspect.

cameronperera (approved):

calcaide (approved):
Terrific work @ZedLi !!! :raised_hands: 👏 

This small details are the ones that make the UI shine.

randallmorey (changes requested):

addons/core/addon/services/browser-object.js
```diff
@@ -0,0 +1,23 @@
...
+   */
+  get window() {
+    return window;
+  }
+
+  /**
+   * The global document object
+   * @returns {Document}
+   */
+  get document() {
```
randallmorey, line 20: Should this be expressed in terms of `this.window`?

ui/admin/app/components/form/worker/create-worker-led/index.js
```diff
@@ -149,6 +149,21 @@ boundary server -config="${this.configFilePath || '<path>'}/pki-worker.hcl"`;
       .join('\n    ');
   }
 
+  getClusterIdFromURL() {
+    const hostname = this.browserObject.window.location.hostname;
```
randallmorey, line 153: Perhaps the location/hostname should also be part of the browser service, if not this entire method.

ui/admin/app/components/form/worker/create-worker-led/index.js
```diff
@@ -149,6 +149,21 @@ boundary server -config="${this.configFilePath || '<path>'}/pki-worker.hcl"`;
       .join('\n    ');
   }
 
+  getClusterIdFromURL() {
```
randallmorey, line 152:
```suggestion
  get clusterID() {
```

ui/admin/app/components/form/worker/create-worker-led/index.js
```diff
@@ -17,10 +17,11 @@ class Tag {
 export default class FormWorkerCreateWorkerLedComponent extends Component {
   // =services
   @service features;
+  @service browserObject;
 
   // =attributes
   @tracked generatedWorkerAuthToken;
-  @tracked clusterId;
+  @tracked clusterId = this.getClusterIdFromURL();
```
randallmorey, line 24: The value wouldn't change, so it can probably just be a getter.
```suggestion

```

randallmorey (changes requested):
[Let's use the test helpers provided by the feature flag addon](https://github.com/kategengler/ember-feature-flags) rather than manipulating the configuration in tests.

ui/admin/app/components/form/worker/create-worker-led/index.js
ZedLi: I can't make this directly just `clusterId` as I still need the tracked value for the input field on the form. I still wanted the user to be able to modify it if they wanted to, but just auto-populate it.

ui/admin/app/components/form/worker/create-worker-led/index.js
```diff
@@ -149,6 +149,21 @@ boundary server -config="${this.configFilePath || '<path>'}/pki-worker.hcl"`;
...
+    // Match against a guid with either the prod, int, or dev hcp domain
+    const matcher =
+      /^([0-9a-f]{8}-[0-9a-f]{4}-[1-5][0-9a-f]{3}-[89ab][0-9a-f]{3}-[0-9a-f]{12})\.boundary\.(hashicorp\.cloud|hcp\.to|hcp\.dev)$/;
+
+    if (matcher.test(hostname)) {
+      // Grab the captured guid
+      return matcher.exec(hostname)[1];
+    }
+
+    return undefined;
```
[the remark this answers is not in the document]
ZedLi: I was doing it to be explicit, but if we don't mind this, I can omit this in functions for the future.

I'm changing it to a getter though and need to keep it to satisfy the [getter-return](https://eslint.org/docs/latest/rules/getter-return) eslint rule (needs to have an explicit return)

addons/core/addon/services/browser-object.js
ZedLi: Any particular reason? They reference the same object so I don't have a preference

ZedLi:
> [Let's use the test helpers provided by the feature flag addon](https://github.com/kategengler/ember-feature-flags) rather than manipulating the configuration in tests.

Have we used this before? It can't find the import and the docs say it depends on `ember-cli-qunit` which is now deprecated

ui/admin/app/components/form/worker/create-worker-led/index.js
ZedLi: I can make the hostname as another getter in the service

ui/admin/app/components/form/worker/create-worker-led/index.js
```diff
@@ -17,19 +17,34 @@ class Tag {
 export default class FormWorkerCreateWorkerLedComponent extends Component {
   // =services
   @service features;
+  @service browserObject;
 
   // =attributes
   @tracked generatedWorkerAuthToken;
-  @tracked clusterId;
+  @tracked clusterId = this.clusterIdFromURL;
```
randallmorey, line 24: This wouldn't change, so it doesn't need to be tracked.  In the template, you can access `this.clusterIdFromURL` directly.
```suggestion

```

ui/admin/app/components/form/worker/create-worker-led/index.js
```diff
@@ -17,19 +17,34 @@ class Tag {
...
   @tracked ipAddress;
   @tracked configFilePath;
   @tracked initialUpstreams;
   @tracked workerTags = A([]);
   @tracked newWorkerKey;
   @tracked newWorkerValue;
 
-  // =properties
+  get clusterIdFromURL() {
```
randallmorey, line 32:
```suggestion
  get clusterIDFromURL() {
```

ui/admin/app/components/form/worker/create-worker-led/index.js
```diff
@@ -17,19 +17,34 @@ class Tag {
...
+    const hostname = this.browserObject.window.location.hostname;
```
randallmorey, line 33: Perhaps location and hostname should be part of the browser service.

ui/admin/app/components/form/worker/create-worker-led/index.js
```diff
@@ -17,19 +17,34 @@ class Tag {
...
+    // Match against a guid with either the prod, int, or dev hcp domain
+    const matcher =
+      /^([0-9a-f]{8}-[0-9a-f]{4}-[1-5][0-9a-f]{3}-[89ab][0-9a-f]{3}-[0-9a-f]{12})\.boundary\.(hashicorp\.cloud|hcp\.to|hcp\.dev)$/;
+
+    if (matcher.test(hostname)) {
+      // Grab the captured guid
+      return matcher.exec(hostname)[1];
+    }
 
+    return undefined;
```
randallmorey, line 44: Undefined is the default return value.
```suggestion

```

addons/core/addon/services/browser-object.js
randallmorey: The goal is the fewest possible global references.  You could even express this using a computed `alias` to be idiomatic:  https://api.emberjs.com/ember/release/functions/@ember%2Fobject%2Fcomputed/alias

ui/admin/app/components/form/worker/create-worker-led/index.js
randallmorey: https://api.emberjs.com/ember/release/functions/@ember%2Fobject%2Fcomputed/alias

ui/admin/app/components/form/worker/create-worker-led/index.js
```diff
@@ -17,19 +17,34 @@ class Tag {
...
+    const hostname = this.browserObject.window.location.hostname;
+
+    // Match against a guid with either the prod, int, or dev hcp domain
+    const matcher =
+      /^([0-9a-f]{8}-[0-9a-f]{4}-[1-5][0-9a-f]{3}-[89ab][0-9a-f]{3}-[0-9a-f]{12})\.boundary\.(hashicorp\.cloud|hcp\.to|hcp\.dev)$/;
+
+    if (matcher.test(hostname)) {
+      // Grab the captured guid
+      return matcher.exec(hostname)[1];
+    }
```
randallmorey, line 42:
```suggestion
  const clusterIDMatcher =
    /^(?<clusterID>[0-9a-f]{8}-[0-9a-f]{4}-[1-5][0-9a-f]{3}-[89ab][0-9a-f]{3}-[0-9a-f]{12})\.boundary\.(hashicorp\.cloud|hcp\.to|hcp\.dev)$/;

  return hostname.match(clusterIDMatcher)?.groups?.clusterID;
```

randallmorey (changes requested):

ui/admin/app/components/form/worker/create-worker-led/index.js
randallmorey: Conversely, if hostname is provided as a getter here, then you only have to stub/mock `service.hostname` in your tests.

ui/admin/app/components/form/worker/create-worker-led/index.js
randallmorey: Oooh.  Got it.  The cluster ID from URL is just an initial value.

randallmorey:
> > [Let's use the test helpers provided by the feature flag addon](https://github.com/kategengler/ember-feature-flags) rather than manipulating the configuration in tests.
> 
> Have we used this before? ~It can't find the import and the docs says it depends on `ember-cli-qunit` which is now deprecated~
> 
> Edit: Uh so I think the docs are just both outdated and documenting unreleased work but the version we use (which is the most recent version) doesn't include the `disableFeature` as seen [here](https://github.com/kategengler/ember-feature-flags/blob/v6.0.0/addon-test-support/index.js). The codebase [currently](https://github.com/kategengler/ember-feature-flags/blob/master/addon-test-support/index.js) has it as it's part of the docs but it's never been released it looks like...

You are absolutely right.  The `disableFeature` test helper hasn't been released yet.  I looked and it seems the test helper just calls `service.disable(feature)` under the hood.  You could try that, just so we're going through the service rather than modifying the config directly:  https://github.com/kategengler/ember-feature-flags/blob/v6.0.0/addon/services/features.js#L35

cameronperera:
> You are absolutely right. The `disableFeature` test helper hasn't been released yet. I looked and it seems the test helper just calls `service.disable(feature)` under the hood. You could try that, just so we're going through the service rather than modifying the config directly: https://github.com/kategengler/ember-feature-flags/blob/v6.0.0/addon/services/features.js#L35

In regards to the `disableFeature` functionality, I had to do the same thing for ssh targets when we implemented it. You can see that [here](https://github.com/hashicorp/boundary-ui/blob/main/ui/admin/tests/acceptance/targets/create-test.js#L46), then on line 142 I call `featuresService.disable('ssh-target');`. The plan was to build a custom feature flag helper since ember-feature-flags hasn't released that feature yet. Now might be the time to create the helper (separate of this PR).

ui/admin/app/components/form/worker/create-worker-led/index.js
ZedLi: I'll update this but it looks like we're pretty inconsistent about `Id` vs `ID` as part of camel casing, should we standardize?

ui/admin/app/components/form/worker/create-worker-led/index.js
ZedLi: Again this is here to mainly satisfy the eslint rule.

ui/admin/app/components/form/worker/create-worker-led/index.js
ZedLi: But again, don't I need this tracked value to still keep the input working? We still want the input especially in cases where this fails to extract a value. This is to just to set an initial value for the input.

addons/core/addon/services/browser-object.js
ZedLi: Updated it to just use `this.window`

ui/admin/app/components/form/worker/create-worker-led/index.js
ZedLi: Sure, it's not much different either way so I included it

ui/admin/app/components/form/worker/create-worker-led/index.js
ZedLi: I can use a named capturing group but I'm actually more of a fan of using the `if` since if you're not familiar with regex, it's much more obvious what's happening

cameronperera:
> > 
> 
> @cameronperera Totally down to work on those helpers. Not necessary but wondering, if we work on those test helpers, do we need to keep using `ember-feature-flags`? or we will be able to get rid of that package?

If we created the helper we would be able to remove that package entirely. It would be relatively easy to handle both `enableFeature` and `disableFeature`. Plus one less package!

ui/admin/app/components/form/worker/create-worker-led/index.js
ZedLi: Actually I updated to without the if since it's clear once there's a comment IMO

ZedLi:
> If we created the helper we would be able to remove that package entirely. It would be relatively easy to handle both `enableFeature` and `disableFeature`. Plus one less package!

Just so I'm clear, are we suggesting just writing our own test helpers for those two functions, which currently are very thin wrappers, or are we suggesting writing our own _template_ helper and service for `ember-feature-flags` to replace the package?

PR #1390, opened 2022-10-25
feat: 🎸 Re-enabling footer for Boundary Admin UI

dvonanderson:
Re-enable global footer and fix styling

✅ Closes: ICU-6454

:tickets: [Jira ticket](https://hashicorp.atlassian.net/browse/ICU-6454)

## Description
The global footer should be visible to bring Boundary into alignment with other products

image
image


DhariniJeeva (approved):

cameronperera (approved):

ZedLi:
I have a general question regarding the layout of the footer

currently it looks like this: 
image

should it be more like this:
image

We can still align to the center of the screen but I guess it's a question of what we should do when the screen shrinks like this and whether it should go over the sidebar:
image

dvonanderson:
> I have a general question regarding the layout of the footer
> 
> currently it looks like this: image
> 
> should it be more like this: image
> 
> We can still align to the center of the screen but I guess it's a question of what we should do when the screen shrinks like this and whether it should go over the sidebar: image

@ZedLi your bottom example is the original way that I wanted to correct the display system for `global-layout`, but because the footer is part of the `global-layout` itself and the sidebar is part of the `global-layout-body` and the main content is yielded through the `sidebar-layout`  so couldn't really target the side-sidebar specifically, but what I can do is change the media query to provide a different view based on screen size.

dvonanderson:
@ZedLi what you think? 
image

calcaide:
I do agree with @ZedLi 

I think the approach we should follow is 
197864133-ebcfed04-fd32-4fc0-9917-2323a555ba36

To me, in this case makes more sense a change on the layout to follow a natural flow. 

What do you think?

dvonanderson:
I am also wondering if we really need a footer for the login in page?  Looking at HCP and they seem to ditch it for the login and password reset screen.  Also if we do need it should the content of the footer be centered in those two layouts?


Screenshot 2022-10-26 181957
Screenshot 2022-10-26 182031

joshuaogle:
@dvonanderson I'm pretty sure HCP is only hiding theirs because of the layout on that page. I don't think the footer looks bad in the corner, but centered might be less distracting. Which one feels better to you?

dvonanderson:
> @dvonanderson I'm pretty sure HCP is only hiding theirs because of the layout on that page. I don't think the footer looks bad in the corner, but centered might be less distracting. Which one feels better to you?

I think it looks great there!

ui/admin/app/styles/app.scss
```diff
@@ -18,6 +18,13 @@
   }
 }
 
+//fix for sidebar wrappper around centered content
+[class*='sidebar-content-centered'] {
```
dvonanderson, line 22: Not 100% about this solution.  Probably could conditionally change this layout somewhere else.  thoughts?  Challenge is not create a layout that will take care of the `login` and `change-password` views.  Both of those views are wrapped in the `sidebar-layout`,

PR #1392, opened 2022-10-26
Use hds badge in sessions admin

DhariniJeeva:
jira - https://hashicorp.atlassian.net/browse/ICU-6550

[admin ui](https://boundary-q6et4wqe5-hashicorp.vercel.app/scopes/s_1sheljmjgm/sessions)

figma: https://www.figma.com/file/gc0RIXxaknS36CsQreeCM8/HDS-Adoption-project

Screen Shot 2022-10-26 at 2 23 46 PM


calcaide (approved):
Awesome work Dharini!!! very excited for the HDS integration :stuck_out_tongue_closed_eyes:

addons/core/addon/components/session-status/index.js
```diff
@@ -0,0 +1,20 @@
...
+
+export default class SessionStatus extends Component {
+  get statusColor() {
+    return STATUS_MAPPING[this.args.status][0];
+  }
+
+  get statusIcon() {
+    return STATUS_MAPPING[this.args.status][1];
+  }
+}
```
cameronperera, line 20: Non-blocking question
Should we set some kind of default here? Just thinking from the aspect of handling unknown credential types, we could try for the same here. Since we rely on the status from the BE, technically they could add a new session status that we have not accounted for. However, this may not be an issue since we will release everything in parody now. I included a status of `stale` and it shows no badge as an example.
Screen Shot 2022-10-26 at 9 25 51 AM

cameronperera (approved):

addons/core/addon/components/session-status/index.js
DhariniJeeva: Ah, Good point. I'm unsure what sort of `default` status we should set. There is scope for adding more types with credentials, but I think there may not be new types for `status.` Either way, I'll raise this question in the channel

addons/core/addon/components/session-status/index.js
```diff
@@ -0,0 +1,20 @@
+import Component from '@glimmer/component';
+
+const STATUS_MAPPING = {
+  //map session status to hds text color and icon style
+  //more info - https://design-system-components-hashicorp.vercel.app/components/badge
+  active: ['success', 'check'],
+  pending: ['neutral', 'delay'],
+  canceling: ['critical', 'x'],
```
randallmorey, line 8: Should cancelling be a warning color with a distinct icon from terminated?  Potentially makes it easier to scan a list with these.

ui/admin/app/styles/app.scss
```diff
@@ -3,6 +3,7 @@
 
 @import 'rose';
 @import 'notify';
+@import '@hashicorp/design-system-components';
```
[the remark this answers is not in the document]
randallmorey: The style import will be required in consuming apps and won't come through Rose directly.

addons/core/addon/components/session-status/index.js
```diff
@@ -0,0 +1,20 @@
...
+  terminated: ['critical', 'x'],
```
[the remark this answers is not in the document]
DhariniJeeva: Agree!

addons/core/addon/components/session-status/index.js
randallmorey: We should show status strings even when we don't have styles for them, and default them to a neutral badge with no icon.

ui/admin/app/styles/app.scss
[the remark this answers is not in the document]
ZedLi: Why is that? Is there a reason it's not or can't be handled by `rose` if we're importing everything else through `rose`?

ui/admin/app/templates/scopes/scope/sessions/index.hbs
```diff
@@ -114,7 +114,18 @@
...
+                {{else}}
+                  <SessionStatus
+                    @status={{t
+                      (concat
+                        'resources.session.status.'
+                        sessionAggregate.session.status
+                      )
+                    }}
+                  />
+                {{/if}}
```
randallmorey, line 128: This is potentially unsafe, since the translates strings differ by locale.  Let's pass in the session itself and do the known/unknown switching and translation within the component itself.  This simplifies any call sites so that we aren't performing this switch in multiple places.

randallmorey (changes requested):

addons/api/addon/models/session.js
```diff
@@ -102,6 +102,13 @@ export default class SessionModel extends GeneratedSessionModel {
     return this.isActive || this.isPending;
   }
 
+  /**
+   * True if status is an unknown type.
+   * @type {boolean}
+   */
+  get isUnknown() {
```
randallmorey, line 109: If sessions come in different types, this is ambiguous.  Let's clarify:
```suggestion
  get isUnknownStatus() {
```

randallmorey (changes requested):

addons/api/addon/models/session.js
randallmorey: Recommending this because we use `isUnknown` elsewhere to mean "unknown type".  But in this context we mean status.  Arguably we should be more specific about both.  For now let's change this instance.

addons/core/addon/components/session-status/index.js
```diff
@@ -0,0 +1,32 @@
...
+    return STATUS_MAPPING[this.args.status]
+      ? STATUS_MAPPING[this.args.status]['color']
+      : 'neutral';
+  }
+
+  get statusBadgeIcon() {
+    return (
+      STATUS_MAPPING[this.args.status] &&
+      STATUS_MAPPING[this.args.status]['icon']
+    );
```
randallmorey, line 24: Would this work?
```suggestion
    return STATUS_MAPPING[this.args.status]?.icon;
```

addons/core/addon/components/session-status/index.js
```diff
@@ -0,0 +1,32 @@
...
+  }
+
+  get statusBadgeType() {
+    return STATUS_MAPPING[this.args.status]
+      ? STATUS_MAPPING[this.args.status]['type']
+      : 'outlined';
```
randallmorey, line 30: Would this work?
```suggestion
    return STATUS_MAPPING[this.args.status]?.type || 'outlined';
```

randallmorey (changes requested):
Looks good.  Made some suggestions.  Can this take a single argument for session instead of both session and its status?  The component will essentially become "status aware", which is fine because it is a SessionStatus component.  Also, can this include integration tests for each known status type, as well as some for unknown types?

ui/admin/app/templates/scopes/scope/sessions/index.hbs
```diff
@@ -114,7 +114,10 @@
                 </time>
               </row.cell>
               <row.cell>
-                <Rose::Badge>{{sessionAggregate.session.status}}</Rose::Badge>
+                <SessionStatus
+                  @status={{sessionAggregate.session.status}}
+                  @model={{sessionAggregate.session}}
+                />
```
randallmorey, line 120: What do you think about passing only session, making the component session-aware?  We're already calling it a SessionStatus component and it lives inside `core`, so I think it's ok.

randallmorey (changes requested):

ui/admin/app/templates/scopes/scope/sessions/index.hbs
DhariniJeeva: yup makes sense!

randallmorey (approved):

PR #1396, opened 2022-10-27
refactor: 💡 existing read credential store acceptance tests

lisbet-alvarez:
✅ Closes: ICU-5517

:tickets: [Jira ticket](https://hashicorp.atlassian.net/browse/ICU-5517)

## Description
Refactor read credential store acceptance tests to simulate user clicking in the UI


ZedLi (approved):

ui/admin/tests/acceptance/credential-store/read-test.js
```diff
@@ -54,53 +56,68 @@ module('Acceptance | credential-stores | read', function (hooks) {
     urls.staticCredentialStore = `${urls.credentialStores}/${instances.staticCredentialStore.id}`;
     urls.vaultCredentialStore = `${urls.credentialStores}/${instances.vaultCredentialStore.id}`;
     urls.unknownCredentialStore = `${urls.credentialStores}/foo`;
+    
     authenticateSession({});
+    await visit(urls.projectScope);
```
cameronperera, line 61: Love the change of adding the initial `visit()` here since all test need it. I would recommend changing this to `await visit(urls.credentialStores);` though. The list of credential stores is where each of these test should start and removes the initial `click()` you have in each test.
```suggestion
    await visit(urls.credentialStores);
```

ui/admin/tests/acceptance/credential-store/read-test.js
```diff
@@ -54,53 +56,68 @@ module('Acceptance | credential-stores | read', function (hooks) {
...
       instances.staticCredentialStore.authorized_actions.filter(
         (item) => item !== 'read'
       );
-    await visit(urls.staticCredentialStore);
-    assert.notOk(find('main tbody .rose-table-header-cell:nth-child(1) a'));
+
+    await click(`[href="${urls.credentialStores}"]`);
+
+    assert.dom(`[href="${urls.staticCredentialStore}"]`).doesNotExist();
+    assert.dom(`[href="${urls.vaultCredentialStore}"]`).isVisible();
```
cameronperera, line 97: I would recommend updating `.isVisible()` to `.exists()`. `isVisible()` might imply that it can be hidden which is not the case in the context of this test.
```suggestion
    assert.dom(`[href="${urls.vaultCredentialStore}"]`).exists();
```

ui/admin/tests/acceptance/credential-store/read-test.js
```diff
@@ -54,53 +56,68 @@ module('Acceptance | credential-stores | read', function (hooks) {
...
+    assert.expect(2);
+    await visit(urls.credentialStores);
+    assert.dom(`[href="${urls.unknownCredentialStore}"]`).doesNotExist();
+
     await visit(urls.unknownCredentialStore);
     await a11yAudit();
+
     assert.strictEqual(
       find('.rose-message-subtitle').textContent.trim(),
       'Error 404'
```
cameronperera, line 123: While this `strictEqual` works, I think the `.hasText()` is more clear and easier to read.
```suggestion
    assert.dom(find('.rose-message-subtitle'))
      .hasText('Error 404');
```

cameronperera (changes requested):
Looks good. I left some feedback for some minor tweaks.

ui/admin/tests/acceptance/credential-store/read-test.js
ZedLi: I think this makes sense if you're testing the credential stores as we should be able to start there but i think we have some tests where we test if we have permissions to goto the credential stores page and it makes sense to start on the project page. I think we either have to start on the project page initially or just have the initial navigation handled in each test.

cameronperera: I think I am leaning towards removing this from the beforeEach and going back to a visit at the beginning of each test.

ui/admin/tests/acceptance/credential-store/read-test.js
ZedLi: I think the `find` here is unnecessary as `assert.dom` will do the finding.

cameronperera: Right. Thanks for pointing that out. Another reason I like `assert.dom`

cameronperera (approved):
LGTM

PR #1397, opened 2022-10-27
Refactor scss

randallmorey:
Updates sass dependencies and moves rose into addon's app/styles folder.

Preparation for #1398


addons/rose/index.js
```diff
@@ -9,38 +9,14 @@ module.exports = {
   included(app) {
     this._super.included.apply(this, arguments);
 
-    this.includeStyles(app);
+    app.import('node_modules/codemirror/lib/codemirror.css');
+    app.import('node_modules/codemirror/theme/monokai.css');
```
ZedLi, line 13: Nice, this is much cleaner!

ZedLi (approved):
LGTM but would love some context on this change

dvonanderson (approved):

addons/rose/index.js
cameronperera: Does this remove the need to import these css files in our other `ember-cli-build` files as well?

cameronperera (approved):

addons/rose/index.js
randallmorey: Yup.  Ideally Rose brings all the styles and assets along with it, with no additional configuration.  It's been a journey.

randallmorey: Note that this PR handles the removal of these imports from admin and desktop.

PR #1399, opened 2022-10-27
chore: 🤖 Update electron to latest version

ZedLi:
:tickets: [Jira ticket](https://hashicorp.atlassian.net/browse/ICU-3857)

## Description
Upgraded electron to `21.2.0`
Went through the [stable releases](https://www.electronjs.org/releases/stable?version=21) and [breaking changes](https://www.electronjs.org/docs/latest/breaking-changes) to make sure there's nothing that affects us.

The original ticket was originally for testing each major version but with electron's current versioning system (a new major version bump ~2 months), it would be prohibitive to do full testing for each major version right now, especially as most major versions don't have that many significant changes. We can jump to the latest and make sure everything works and discuss how we would like to handle keeping electron up to date for the future (maybe use dependabot?)

Tested with windows and m1 mac builds that all the normal functionality works.


lisbet-alvarez (approved):

cameronperera (approved):

calcaide (approved):
LGTM!! Awesome work!!

Did you test the `boundary-ui-release` (DC release procedure) CI running this branch? If you have any questions about it, ping me on Slack and I will help with that 😉

calcaide:
Nice pragmatism finding an unnecessary fine grain task but converted to 1 only step.

My five cents about how we check for Electron update:
I have no idea if we can set dependabot for such purpose, as per my understanding (I can be wrong, never really went deep on it) dependabot checks for security issues.

As per regular basis I check repository deps updates running yarn outdated command within boundary-ui repo and tells me what is being outdated.

Other option is set a reminder on a calendar?

calcaide:
@ZedLi 
Awesome!! I did not know we can configure it that way. Perhaps something we can discuss in tec time with the rest of the team? I would be down to leverage dependabot to check important deps, like electron, node, ember and such.

PR #1400, opened 2022-10-27
refactor: 💡 host-catalogs read acceptance test

lisbet-alvarez:
✅ Closes: ICU-6544

:tickets: [Jira ticket](https://hashicorp.atlassian.net/browse/ICU-6544)

## Description
Refactor host catalog read acceptance tests to simulate user clicking in the UI


ui/admin/tests/acceptance/host-catalogs/read-test.js
```diff
@@ -71,14 +74,33 @@ module('Acceptance | host-catalogs | read', function (hooks) {
...
+    assert.dom(`[href="${urls.unknownHostCatalog}"]`).doesNotExist();
+
     await visit(urls.unknownHostCatalog);
     await a11yAudit();
-    assert.ok(find('.rose-message-subtitle').textContent.trim(), 'Error 404');
+
+    assert.strictEqual(
+      find('.rose-message-subtitle').textContent.trim(),
+      'Error 404'
+    );
```
ZedLi, line 94: I think this also could have been: 

```suggestion
    assert.dom('.rose-message-subtitle').hasText('Error 404');
```

ui/admin/tests/acceptance/host-catalogs/read-test.js
```diff
@@ -71,14 +74,33 @@ module('Acceptance | host-catalogs | read', function (hooks) {
       instances.hostCatalog.authorized_actions.filter(
         (item) => item !== 'read'
       );
-    await visit(urls.hostCatalog);
-    assert.notOk(find('main tbody .rose-table-header-cell:nth-child(1) a'));
+
+    await click(`[href="${urls.hostCatalogs}"]`);
+
+    assert.dom(`[href="${urls.hostCatalog}"]`).doesNotExist();
```
ZedLi, line 80: We have a test that we can't navigate to a `host catalog` inside `host catalogs` but I think we should also have a test we can't navigate to the `host-catalogs` without proper permissions as well

ZedLi (changes requested):
Left a few small comments! Looks great otherwise

ui/admin/tests/acceptance/host-catalogs/read-test.js
lisbet-alvarez: Maybe I'm misunderstanding the suggestion but since host-catalogs is a route and not a model I can't set permissions.

ui/admin/tests/acceptance/host-catalogs/read-test.js
```diff
@@ -60,25 +61,45 @@ module('Acceptance | host-catalogs | read', function (hooks) {
...
+
+  test('users can link to docs page for host catalog', async function (assert) {
+    assert.expect(1);
+    await visit(urls.projectScope);
+
+    await click(`[href="${urls.hostCatalogs}"]`);
+
+    assert
+      .dom(`[href="https://boundaryproject.io/help/admin-ui/host-catalogs"]`)
+      .isVisible();
```
lisbet-alvarez, line 103:
```suggestion
      .exists();
```

ui/admin/tests/acceptance/host-catalogs/read-test.js
ZedLi: I think there might be a little confusion regarding the permissions. `host-catalogs` permission would be the `list` which doesn't live on the individual model, it lives one layer above in the scope and you'd have to mock `authorized_collection_actions` instead of `authorized_actions`. This is true for any of the collections. 

I'm suggesting we test this permission check [here](https://github.com/hashicorp/boundary-ui/blob/main/ui/admin/app/templates/scopes/scope.hbs#L108).

You can see we mock the permission [here](https://github.com/hashicorp/boundary-ui/blob/main/addons/api/mirage/factories/scope.js#L27)

cameronperera: @ZedLi are you suggesting a test that verifies if the user can (or cannot) navigate to the `urls.hostCatalogs` route?

cameronperera: I just needed to wait one more second for your comment to come through 😆 
@lisbet-alvarez here is an [example](https://github.com/hashicorp/boundary-ui/blob/6f8afbb7af5927e45a45756150951fc1c63272ab/ui/admin/tests/acceptance/sessions-test.js#L89-L100) of what @ZedLi is suggesting.

Great suggestion Zed!

lisbet-alvarez: Ahh okay makes sense! Thxx

ui/admin/tests/acceptance/host-catalogs/read-test.js
```diff
@@ -60,25 +61,56 @@ module('Acceptance | host-catalogs | read', function (hooks) {
...
+
+    assert.dom(`[href="${urls.hostCatalog}"]`).doesNotExist();
   });
 
-  test('visiting an unknown host catalog displays 404 message', async function (assert) {
+  test('cannot navigate to host catalogs tab without proper authorization', async function (assert) {
     assert.expect(1);
+    await visit(urls.orgScope);
+    instances.scopes.project.authorized_collection_actions['host-catalogs'] =
+      [];
```
lisbet-alvarez, line 88: is there a preference between using the filter method or setting the attribute equal to empty list ?

cameronperera: I think the preference would be filter. By setting it to an empty list, we could get a false positive. Using the filter method allows us to be exact in what is being tested.

lisbet-alvarez: I didn't realize this before, but authorized collections actions is already tested in the list-test.js files. So I can instead refactor those files to use the more appropriate methods.

cameronperera: 🤦🏼‍♂️ Thanks for noticing that before we added a bunch of duplicate tests.

cameronperera (approved):
LGTM!

DhariniJeeva (approved):

ZedLi (approved):

PR #1410, opened 2022-11-03
Icu 6705 add stylelint to rose

ZedLi:
:tickets: [Jira ticket](https://hashicorp.atlassian.net/browse/ICU-6705)

## Description
Added stylelint to rose. Disabled a few rules with comments explaining why.


addons/rose/.stylelintrc.js
```diff
@@ -0,0 +1,19 @@
+'use strict';
+
+module.exports = {
+  extends: ['stylelint-config-standard-scss', 'stylelint-config-prettier-scss'],
```
ZedLi, line 4: Used `stylelint-config-standard-scss` over `stylelint-config-sass-guidelines` which we're using in desktop. This one seems to be better supported now and is recommended by the official docs.

addons/rose/app/styles/rose/components/form/_radio-card.scss
```diff
@@ -180,18 +196,20 @@ $radio-svg: (
 
   &.error {
     .rose-form-radio-field {
-      &:not(:disabled):not(:focus) {
-        &:not(:disabled):not(:focus) {
+      &:not(:disabled, :focus) {
+        &:not(:disabled, :focus) {
```
randallmorey, line 200: Are these duplicated?  Not sure why they were in the first place, but probably unnecessary?

addons/rose/package.json
```diff
@@ -96,12 +97,16 @@
     "jest": "^27.4.7",
     "loader.js": "^4.7.0",
     "npm-run-all": "^4.1.5",
+    "postcss": "^8.4.18",
```
randallmorey, line 100: Why is PostCSS added?

randallmorey (changes requested):

addons/rose/package.json
calcaide: Curious about this too.

ZedLi: Forgot to comment about this, stylelint 14 [does not support any non-css](https://stylelint.io/migration-guide/to-14#syntax-option-and-automatic-inferral-of-syntax) syntax out of the box anymore. 

Both [stylelint-config-standard-scss](https://github.com/stylelint-scss/stylelint-config-standard-scss/blob/main/package.json#L64) and [stylelint-config-sass-guidelines](https://github.com/bjankord/stylelint-config-sass-guidelines/blob/main/package.json#L38) have a peer dependency on `postcss` so they won't work without it.

addons/rose/app/styles/rose/components/form/_radio-card.scss
calcaide: Indeed, seems duplicate and unneccesary states. I tested within storybook and I couldn't reproduce to show this specific not diabled and not focus.

addons/rose/package.json
calcaide: Ahh! that makes all the sense then 😉

calcaide (approved):

addons/rose/app/styles/rose/components/form/_radio-card.scss
ZedLi: Looks like it, removed it

randallmorey (approved):

cameronperera (approved):

PR #1420, opened 2022-11-04
Icu 6600 update badges in admin UI 2

DhariniJeeva:
:tickets: [Jira ticket](https://hashicorp.atlassian.net/browse/ICU-6600)

Screen Shot 2022-11-04 at 12 19 06 PM

Screen Shot 2022-11-04 at 1 01 08 PM

Screen Shot 2022-11-04 at 12 59 56 PM

**badgeCount**:
Screen Shot 2022-11-04 at 2 07 54 PM

### DIFF
--- a/addons/rose/addon/components/rose/dropdown/index.hbs
+++ b/addons/rose/addon/components/rose/dropdown/index.hbs
@@ -20,7 +20,7 @@
     <span class='rose-dropdown-trigger-text'>
       {{@text}}
       {{#if @count}}
-        <Rose::Badge>{{@count}}</Rose::Badge>
+        <Hds::BadgeCount @text={{@count}} @size='small' />
       {{/if}}
     </span>
 
--- a/ui/admin/app/components/host-catalog-type-badge/index.hbs
+++ b/ui/admin/app/components/host-catalog-type-badge/index.hbs
@@ -1,10 +1,4 @@
-{{#if this.isStaticOrUnknownHostCatalogType}}
-  <Hds::Badge
-    @text={{t (concat 'resources.host-catalog.types.' @model.compositeType)}}
-  />
-{{else}}
-  <Hds::Badge
-    @icon={{concat @model.compositeType '-color'}}
-    @text={{t (concat 'resources.host-catalog.types.' @model.compositeType)}}
-  />
-{{/if}}
+<Hds::Badge
+  @icon={{this.icon}}
+  @text={{t (concat 'resources.host-catalog.types.' @model.compositeType)}}
+/>
--- a/ui/admin/app/components/host-catalog-type-badge/index.hbs
+++ b/ui/admin/app/components/host-catalog-type-badge/index.hbs
@@ -1,15 +1,15 @@
 {{#if @model.isStatic}}
-  <Rose::Badge>
-    {{t (concat 'resources.host-catalog.types.' @model.compositeType)}}
-  </Rose::Badge>
+  <Hds::Badge
+    @text={{t (concat 'resources.host-catalog.types.' @model.compositeType)}}
+  />
 {{else if @model.isUnknown}}
-  <Rose::Badge>
-    {{t (concat 'resources.host-catalog.types.' @model.compositeType)}}
-  </Rose::Badge>
+  <Hds::Badge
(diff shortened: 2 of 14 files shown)


addons/rose/addon/components/rose/dropdown/index.hbs
```diff
@@ -20,7 +20,7 @@
     <span class='rose-dropdown-trigger-text'>
       {{@text}}
       {{#if @count}}
-        <Rose::Badge>{{@count}}</Rose::Badge>
+        <Hds::BadgeCount @text={{@count}} @size='small' />
```
DhariniJeeva, line 23: HDS has a [badgeCount](https://design-system-components-hashicorp.vercel.app/components/badge-count) component that I thought fit here. Added an image of this in the description.

randallmorey (approved):

ui/admin/app/components/host-catalog-type-badge/index.hbs
```diff
@@ -1,15 +1,15 @@
 {{#if @model.isStatic}}
-  <Rose::Badge>
-    {{t (concat 'resources.host-catalog.types.' @model.compositeType)}}
-  </Rose::Badge>
+  <Hds::Badge
```
calcaide, line 2: No blocker:
Code block from line 2 to 4 is the same as 6 to 8.

To avoid repeat code, what do you think of refactoring the conditional to use two conditions? since we already have [ember-trut-helpers](https://github.com/jmurphyau/ember-truth-helpers#readme) something like:

```html
{{#if (or @model.isStatic @model.isUnknown)}}
  ...code
{{else}}
  ...code
{{/if}}
```

calcaide (approved):
LGTM! Thank you for this, HDS looking good 💯

ui/admin/app/components/host-catalog-type-badge/index.hbs
cameronperera: I believe we discussed not using helper methods like or or and in templates to keep it more readable. In my case, I moved the logic into the controller. We could do the same here.

cameronperera (approved):

ui/admin/app/components/host-catalog-type-badge/index.hbs
ZedLi: I think a simple `or` or `and` are fine, I do think if there's negation involved with the `or` or `and` (it gets nested) then it might be prone to logic errors or hard to read.

I'd be fine with this `or` personally but I'd like to hear when we should use `ember-truth-helpers` if not

DhariniJeeva: Good point! I moved this conditional check to the component js file

DhariniJeeva: I prefer to keep the logic out of the component/template and keep it simple. And this would also work out perfectly if we add more host catalog types.

calcaide: Moving the logic to the controller works for me 😄  That's a valid solution, and the de-facto solution if you are not using ember-truth-helpers.

randallmorey: Taking a step back from the code, the logic seems to be something like "if this is a known plugin, display the badge with an icon otherwise show a text badge".  If that's the case, perhaps the model itself should have an `isKnownPlugin` getter.  You could then flip this else/if order and you won't need a component JS file.

I'd also reexamine the existing getters related to types on host catalogs, sets, and hosts in general.  Do they all make sense?

randallmorey (changes requested):

ui/admin/app/components/host-catalog-type-badge/index.hbs
randallmorey: However, there won't necessarily be a 1-to-1 mapping between the plugin name (even if known) and Flight icons.  Perhaps this could be represented using a similar mapping strategy as what was used for session badges?  i.e. keep a hash of the mappings in the component.

DhariniJeeva: that's a valid point! I'll investigate more and update the pr

randallmorey:
Closing in favor of basing this from a different branch.

PR #1425, opened 2022-11-07
Chore upgrade ember intl

randallmorey:
:tickets: [Jira ticket](https://hashicorp.atlassian.net/browse/icu-6771)

## Description

This PR migrates boundary to ember-intl 6.x, which is a necessary step toward Ember 4.

[Admin preview](https://boundary-ui-git-chore-upgrade-ember-intl-hashicorp.vercel.app)
[Desktop preview](https://boundary-ui-desktop-git-chore-upgrade-ember-intl-hashicorp.vercel.app)

### DIFF
--- a/addons/api/addon/models/base.js
+++ b/addons/api/addon/models/base.js
@@ -1,6 +1,7 @@
 import Model from '@ember-data/model';
 import { attr } from '@ember-data/model';
 import { computed } from '@ember/object';
+import { inject as service } from '@ember/service';
 
 /**
  * Base model class provides common functionality for all models.
@@ -25,6 +26,10 @@ import { computed } from '@ember/object';
  *   2) Or as an ID:  `record.scopeID`.
  */
 export default class BaseModel extends Model {
+  // =services
+
+  @service store;
+
   // =relationships
 
   /**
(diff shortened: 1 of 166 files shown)


cameronperera (commented):
I was trying to find the 6.0.1 release and changelog but I cannot seem to find it. NPM and their GitHub repo says v6.0.0 is still a beta release. Do you have a link with info on migrating from v5 to v6?

calcaide (approved):
LGTM!! Thank you for the work

randallmorey:
> I was trying to find the 6.0.1 release and changelog but I cannot seem to find it. NPM and their GitHub repo says v6.0.0 is still a beta release. Do you have a link with info on migrating from v5 to v6?

No I couldn't find it either.  Yet it exists.

cameronperera (approved):

ZedLi (changes requested):
> No I couldn't find it either. Yet it exists.

I don't think this should be `6.0.1`. I see the `6.x` beta versions in both github and npm but the `6.0.1` [version is deprecated](https://www.npmjs.com/package/ember-intl/v/6.0.1), which is why you don't see documentation for it. You can also confirm this by visiting their npm page and selecting see deprecated versions. 

Their latest is still pointed at `5.7.0` but they've marked the `6.x` release as beta which means we should explicitly grab the beta version if we need to (probably `6.0.0-beta.4`?), you can normally just do `yarn add ember-intl@beta`

PR #1432, opened 2022-11-08
Chore ember 4 upgrade

calcaide:
**DO NOT MERGE YET**

## Description

Upgrade Addons and UI's to Ember 4

### DIFF
--- a/addons/api/package.json
+++ b/addons/api/package.json
@@ -54,11 +54,10 @@
     "ember-disable-prototype-extensions": "^1.1.3",
     "ember-export-application-global": "^2.0.1",
     "ember-load-initializers": "^2.1.2",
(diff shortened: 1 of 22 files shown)


addons/api/package.json
```diff
@@ -41,6 +40,7 @@
     "@babel/plugin-proposal-decorators": "^7.18.6",
     "@ember/optional-features": "^2.0.0",
     "@ember/test-helpers": "^2.8.1",
+    "@embroider/test-setup": "^0.48.1",
```
calcaide, line 43: `@embroider/test-setup` is using at development time (`ember-try`) not in production. That's why I think makes more sense to have it as a `devDependency` than as `dependency`.

PR #1434, opened 2022-11-08
feat: 🎸 Add links to Host Sets assigned to targets

cameronperera:
## Description

The target host sources view can now link to the host sets associated with that target

✅ Closes: ICU-6525

:tickets: [Jira ticket](https://hashicorp.atlassian.net/browse/ICU-6525)

:technologist: [Admin preview](https://boundary-ui-git-icu-6525-linkify-hosts-under-t-e27737-hashicorp.vercel.app/scopes/s_1sheljmjgm/targets/t_2p8vm6smug/host-sources)

Screen Shot 2022-11-08 at 5 49 12 PM


dvonanderson (approved):

ui/admin/tests/acceptance/targets/host-sources-test.js
```diff
@@ -75,6 +87,24 @@ module('Acceptance | targets | host-sources', function (hooks) {
...
+  });
+
+  test('cannot navigate to an unknown host set type', async function (assert) {
+    assert.expect(1);
+    instances.target.update({
+      hostSets: instances.hostCatalogPlugin.hostSets,
+    });
+
+    await visit(urls.targetHostSources);
+    assert.dom(`[href="${urls.unknownHostSet}"]`).doesNotExist();
```
ZedLi, line 105: Is it possible to have the check be similar as the known host set type by looking for the actual `a` link (and it doesn't exist)?

I find it hard to verify the accuracy of tests like these because it could either be the link legitimately doesn't exist or we made a typo/got the link wrong for the `href` and we're serendipitously thinking the test is passing correctly (looks right in this case to be clear 😆).

Any thoughts on this?

ZedLi (approved):
Left one non-blocking thought

ui/admin/tests/acceptance/targets/host-sources-test.js
cameronperera: Pushed a better assert. Updated the click for the known test as well for more clarity on where we expect the user to be clicking.

DhariniJeeva (approved):
LGTM!

lisbet-alvarez (approved):

PR #1436, opened 2022-11-09
chore: 🤖 install HDS and associated assets

randallmorey:
Closes icu-6711 icu-6712 icu-6709

Establishes auto-importing of HDS styles and assets through Rose.  Admin and desktop receive HDS automatically with no additional configuration.


addons/core/tests/integration/helpers/relative-datetime-live-test.js
```diff
@@ -35,7 +35,7 @@ module('Integration | Helper | relative-datetime-live', function (hooks) {
     await render(hbs`{{relative-datetime-live this.date}}`);
     assert.dom(this.element).hasText('23 years ago');
 
-    this.date = now + 8 * 31 * 24 * 60 * 60 * 1000;
+    this.date = now + 8 * 30 * 24 * 60 * 60 * 1000;
```
DhariniJeeva, line 38: I noticed this test was failing for me as well, but when I reran it, it passed 🤯

dvonanderson: @DhariniJeeva I think it may be related to the time change

addons/rose/index.js
```diff
@@ -12,11 +12,43 @@ module.exports = {
...
   },
 
+  /**
+   * Due to a limitation in how ember treats nested addons (see https://github.com/ember-cli/ember-cli/issues/4475)
+   * this is neeeded to reach down into @hashicorp/ember-flight-icons' contentFor hook to run the logic
+   * that injects the sprite into the DOM
+   */
+  contentFor(type, config) {
+    return this.findOwnAddonByName('@hashicorp/design-system-components')
+      .findOwnAddonByName('@hashicorp/ember-flight-icons')
```
DhariniJeeva, line 28: This is awesome! Didn't know we could chain this which removes the extra import of flight icons

DhariniJeeva (approved):

addons/rose/index.js
randallmorey: I just discovered this.  One less direct dependency!

PR #1437, opened 2022-11-09
feat: 🎸 badge migration in sessions admin ui

DhariniJeeva:


joshuaogle:
These colors look great! :chef-kiss:

The badges look like they might be a little tall compared to the designs (24px tall), but looks like that might not be in scope for this PR

randallmorey:
> These colors look great! :chef-kiss:
> 
> The badges look like they might be a little tall compared to the designs (24px tall), but looks like that might not be in scope for this PR

Resolved.  We didn't have `*, *::before, *::after { box-sizing: border-box; }` in our reset as required by HDS.

randallmorey (approved):

PR #1447, opened 2022-11-10
refactor: migrate to HDS select form fields

randallmorey:
:tickets: [Closes icu-6597](https://hashicorp.atlassian.net/browse/icu-6597)

## Description

:technologist: [Admin preview](https://boundary-ui-git-refactor-hds-select-hashicorp.vercel.app/scopes/s_id6ozx3wue/auth-methods/am_e4nm52bkf2)
:rose: [Rose preview](https://boundary-ui-storybook-git-refactor-hds-select-hashicorp.vercel.app/?path=/docs/rose-form-select-removed--basic)

This PR replaces rose/form/select with hds/form/select.  The Rose select, and its emberx-select dependency, have been removed entirely, rather than deprecated.  This was possible because selects were used in production in only one form:  OIDC auth method.

Experimental dark mode for HDS is enabled.

Screenshot 2022-11-11 at 10 59 42

Screenshot 2022-11-09 at 20 13 23

Screenshot 2022-11-10 at 09 47 18


addons/rose/addon/components/rose/form/index.hbs
```diff
@@ -22,7 +22,6 @@
         'rose/form/checkbox/group' disabled=this.disabled
       )
       input=(component 'rose/form/input' disabled=this.disabled)
-      select=(component 'rose/form/select' disabled=this.disabled)
```
DhariniJeeva, line 25: Shouldn't this be replaced with the HDS component?

randallmorey: It may not need to be.  The only reason to do this was to get the disabled state piped through.  However, this is easy to do without yielding.  As long as the form itself yields its state, we can do this:

```
<Hds::Form::Select::Field disabled={{form.disabled}} as |F|>
```

joshuaogle:
Would it be possible to change `--token-form-control-base-border-color-default` one step darker to use `--token-color-palette-neutral-200`? It would match the other inputs a tiny bit closer and doesn't feel quite so stark

addons/rose/addon/helpers/set-from-event.js
```diff
@@ -0,0 +1,9 @@
+import { helper } from '@ember/component/helper';
+import { set } from '@ember/object';
+
```
DhariniJeeva, line 3: can we add a comment here explaining why we need this helper?

ui/admin/app/components/form/auth-method/oidc/index.hbs
```diff
@@ -155,22 +155,27 @@
...
+                  {{on 'change' (set-from-event algorithm 'value')}}
+                  as |F|
                 >
-                  <field.label>{{t 'form.algorithm.label'}}</field.label>
-                  <field.field as |select|>
-                    {{#each this.signingAlgorithms as |algorithm|}}
-                      <select.option
-                        @value={{algorithm}}
-                      >{{algorithm}}</select.option>
+                  <F.Label class='sr-only'>{{t
```
DhariniJeeva, line 166: Is this class necessary here?

ui/admin/app/components/form/auth-method/oidc/index.hbs
```diff
@@ -155,22 +155,27 @@
...
+                <Hds::Form::Select::Field
                   @value={{algorithm.value}}
-                  @error={{@model.errors.attributes_signing_algorithms}}
-                  @contextual={{true}}
-                  @onChange={{fn (mut algorithm.value)}}
-                  as |field|
+                  @isInvalid={{@model.errors.attributes_signing_algorithms}}
+                  disabled={{form.disabled}}
+                  {{on 'change' (set-from-event algorithm 'value')}}
+                  as |F|
```
DhariniJeeva, line 164: personally, I find something like this `as |field|` more readable. Thoughts on this?

DhariniJeeva:
👏  LGTM! just left a couple of minor comments

DhariniJeeva (approved):

ui/admin/app/components/form/auth-method/oidc/index.hbs
randallmorey: I agree.  However, [HDS uses this syntax](https://design-system-components-hashicorp.vercel.app/components/form/select).  For consistency with the HDS docs and for consistency across products and time, we should probably align with HDS.

ui/admin/app/components/form/auth-method/oidc/index.hbs
randallmorey: Yes.  This hides the label visually.  It's necessary for accessibility, but visually redundant because of the fields.

addons/rose/addon/helpers/set-from-event.js
randallmorey: Done

PR #1448, opened 2022-11-10
Icu 6800 api is secret data modeling

cameronperera:
## Description

Move secret attribute logic to application serializer that will remove the key/value pair if the value is null or empty

✅ Closes: ICU-6800

:tickets: [Jira ticket](https://hashicorp.atlassian.net/browse/ICU-6800)


addons/api/addon/serializers/credential.js
```diff
@@ -13,45 +13,8 @@ export default class CredentialSerializer extends ApplicationSerializer {
...
-    // Users cannot unset the private key or passphrase, they would have
-    // to delete the credential entirely.
-    if (!serialized?.attributes?.private_key_passphrase)
-      delete serialized.attributes.private_key_passphrase;
-    if (!serialized?.attributes?.private_key)
-      delete serialized.attributes.private_key;
-    return serialized;
+  serialize() {
+    return super.serialize(...arguments);
   }
```
randallmorey, line 18: This can be removed:
```suggestion
```

randallmorey (changes requested):

addons/api/addon/serializers/application.js
```diff
@@ -39,6 +39,8 @@ export default class ApplicationSerializer extends RESTSerializer {
   serializeAttribute(snapshot, json, key, attribute) {
     const { type, options } = attribute;
     let value = super.serializeAttribute(...arguments);
+    // Remove secret attributes that are null or empty
+    if (options.isSecret && !json[key]) delete json[key];
```
ZedLi, line 43: Since this is in the application serializer, should we be more specific than checking a falsy value? It's unlikely but in the future, I wouldn't want to get caught accidentally stripping out something like `0`.

cameronperera: Yes, I can add a test case for that.

cameronperera: I updated this to check the type. We will add another type check once JSON creds are implemented so this would have been needed.

ZedLi (approved):
LGTM!

randallmorey (commented):
Are there other models or serialisers that have similar secrets handling logic that should be updated?

cameronperera:
> Are there other models or serialisers that have similar secrets handling logic that should be updated?

Looking at our models and serializers the one model/serializer that had similar logic was `fragment-auth-method-attributes`. The serializer has a check similar the `isSecret` logic.
`if (key === 'client_secret' && !json[key]) delete json[key];`
With that be a fragment of auth-method, is it handled differently by the application serialize/normalize? I noticed other parts of the `fragment-auth-method-attributes` serializer is a copy of other logic like `readOnly`.

randallmorey:
> > Are there other models or serialisers that have similar secrets handling logic that should be updated?
> 
> Looking at our models and serializers the one model/serializer that had similar logic was `fragment-auth-method-attributes`. The serializer has a check similar the `isSecret` logic. `if (key === 'client_secret' && !json[key]) delete json[key];` With that be a fragment of auth-method, is it handled differently by the application serialize/normalize? I noticed other parts of the `fragment-auth-method-attributes` serializer is a copy of other logic like `readOnly`.

Ah.  Let's come back to this when model fragments is removed.

randallmorey (approved):

addons/api/tests/unit/serializers/credential-test.js
```diff
@@ -97,7 +97,7 @@ module('Unit | Serializer | credential', function (hooks) {
     const store = this.owner.lookup('service:store');
     const serializer = store.serializerFor('credential');
     const record = store.createRecord('credential', {
-      private_key_passphrase: '',
+      private_key_passphrase: null,
```
randallmorey, line 100: What's the reason for this change?

cameronperera: That was just to verify that if the value was `null` the `isSecret` logic would still work as expected in the test. If the user somehow set the value to null, we still don't want that sent to the API.

lisbet-alvarez (approved):
LGTM

dvonanderson (approved):
Awesome

PR #1461, opened 2022-11-14
refactor: 💡 migrate set-password form to HDS text input

randallmorey:
:tickets: [✅ Closes: icu-6918](https://hashicorp.atlassian.net/browse/icu-6918)

## Description

:technologist: [Admin preview](https://boundary-ui-git-refactor-hds-form-set-password-hashicorp.vercel.app/scopes/global/auth-methods/am_3enprm7y2t/accounts/acct_ldyksmlz79/set-password)

Migrates the account set-password form to use HDS text input.

Screenshot 2022-11-14 at 13 10 08
Screenshot 2022-11-14 at 13 10 03


DhariniJeeva (approved):

ui/admin/app/components/form/account/password/set-password/index.hbs
```diff
@@ -1,28 +1,19 @@
...
-      <field.errors as |errors|>
-        {{#each @model.errors.password as |error|}}
-          <errors.message>{{error.message}}</errors.message>
-        {{/each}}
-      </field.errors>
-    {{/if}}
-  </form.input>
+    <F.Label>{{t 'form.set_password.label'}}</F.Label>
+    <F.HelperText>{{t 'form.set_password.help'}}</F.HelperText>
+  </Hds::Form::TextInput::Field>
```
cameronperera, line 16: Sorry if I missed it but is there builtin handling for errors in HDS so that we do not have to use the `{{#if @model.errors.password}}` pattern?

ui/admin/app/components/form/account/password/set-password/index.hbs
```diff
@@ -1,28 +1,19 @@
...
-  as |form|
->
+<Rose::Form @onSubmit={{fn this.submit @submit this.password}} as |form|>
 
-  <form.input
-    @name='password'
-    @type='password'
+  <Hds::Form::TextInput::Field
+    @isRequired={{true}}
+    @isOptional={{false}}
```
cameronperera, line 5: Do we need to explicitly set `isOptional` as false if it is false by default?

ui/admin/app/components/form/account/password/set-password/index.hbs
randallmorey: Errors for this form never worked at the field level.  What was there previously was copy pasta and was actually incorrect 😬 .

PR #1464, opened 2022-11-14
Refactor host set fragment arrays

randallmorey:
This removes the usage of fragments from host sets. The main workaround we use here is whenever a user clicks the edit button, we recreate a deep copy of the array fields on the model so ember knows they're a different field and can correctly mark them as dirty and we can roll them back when canceling.


addons/api/addon/models/host-set.js
```diff
@@ -71,23 +71,12 @@ export default class HostSetModel extends GeneratedHostSetModel {
...
    */
-  @fragmentArray('fragment-string', {
-    readOnly: true,
-    emptyArrayIfMissing: true,
-  })
-  host_ids;
+  @attr({ readOnly: true, emptyArrayIfMissing: true }) host_ids;
 
-  @fragmentArray('fragment-string', {
-    emptyArrayIfMissing: true,
```
ZedLi, line 81: @randallmorey Do you know the order of transforms and serializers? 

Does it go transform --> normalizer --> model and then model --> serializer --> transform?

I ask because we removed the `emptyArrayIfMissing` because I assume the transformer handles it for `string-array` but a serializer test was broken because the normalizer wasn't aware of the transform

ui/desktop/app/routes/scopes/scope/projects/targets/target/hosts.js
```diff
@@ -27,9 +27,7 @@ export default class ScopesScopeProjectsTargetsTargetHostsRoute extends Route {
     );
 
     // Extract host ids from all host sets
-    const hostIds = hostSets.map(({ host_ids }) =>
-      host_ids.content.map((hostId) => hostId.value)
```
ZedLi, line 31: What was the `content` here for previously? When I was testing it's just a normal array of host IDs now but I didn't expect the `content`

addons/api/addon/models/host-set.js
randallmorey: It's possible that if a field is wholly absent, a transform isn't activated.  However, [normalisation of empty arrays based on that attribute option are independent of the transform](https://github.com/hashicorp/boundary-ui/blob/main/addons/api/addon/serializers/application.js#L191).  It's a judgement call to determine if you need/want an empty array.

calcaide (approved):
Interesting solution! Thanks for this work

addons/api/addon/models/host-set.js
ZedLi: Alright makes sense, I added back the attribute option as I do think it makes sense to have an empty array

ui/admin/tests/acceptance/host-catalogs/host-sets/create-test.js
```diff
@@ -79,6 +79,67 @@ module('Acceptance | host-catalogs | host sets | create', function (hooks) {
...
+      type: 'plugin',
+      plugin: {
+        id: `plugin-id-1`,
+        name: 'aws',
+      },
+    });
+
+    const count = getHostSetCount();
+    await visit(
+      `${urls.hostCatalogs}/${instances.hostCatalog.id}/host-sets/new`
```
lisbet-alvarez, line 95:
```suggestion
      `${urls.newHostSet}`
```

ui/admin/tests/acceptance/host-catalogs/host-sets/create-test.js
```diff
@@ -79,6 +79,67 @@ module('Acceptance | host-catalogs | host sets | create', function (hooks) {
...
+      type: 'plugin',
+      plugin: {
+        id: `plugin-id-1`,
+        name: 'azure',
+      },
+    });
+
+    const count = getHostSetCount();
+    await visit(
+      `${urls.hostCatalogs}/${instances.hostCatalog.id}/host-sets/new`
```
lisbet-alvarez, line 126:
```suggestion
      `${urls.newHostSet}`
```

lisbet-alvarez (approved):
LGTM

PR #1469, opened 2022-11-15
fix: 🐛 fix broken relative datetime live render tests

randallmorey:
Fixes flakey relative datetime render tests.


DhariniJeeva (approved):

cameronperera (approved):

ZedLi (approved):
LGTM! We could've also split each one of these into their own test case and had the date set in the beforeEach

PR #1472, opened 2022-11-16
fix: 🐛 Fix design system peer dep

calcaide:
## Description

After the last update of `@hashicorp/design-system-components` we got a peerDeependency warning. This PR fix it.

Terminal warning before the fix:
Screen Shot 2022-11-16 at 8 28 38 AM

Termina showing NO warning after the fix:
Screen Shot 2022-11-16 at 9 34 08 AM

### DIFF
--- a/addons/rose/package.json
+++ b/addons/rose/package.json
@@ -89,6 +89,7 @@
     "ember-source": "~3.28.8",
     "ember-template-lint": "^4.14.0",
     "ember-template-lint-plugin-prettier": "^4.0.0",
+    "ember-test-selectors": "^6.0.0",
     "eslint": "^8.18.0",
     "eslint-config-prettier": "^8.5.0",
     "eslint-plugin-ember": "^10.6.1",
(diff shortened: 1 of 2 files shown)


ZedLi:
Hmm you're right, it is a [peer dependency](https://github.com/hashicorp/design-system/blob/main/packages/ember-flight-icons/package.json#LL82). I wonder if it actually should be one, why wouldn't it be just a `dependency` as noted in its [documentation](https://github.com/mainmatter/ember-test-selectors#usage-in-ember-addons)?

calcaide:
> Hmm you're right, it is a [peer dependency](https://github.com/hashicorp/design-system/blob/main/packages/ember-flight-icons/package.json#LL82). I wonder if it actually should be one, why wouldn't it be just a `dependency` as noted in its [documentation](https://github.com/mainmatter/ember-test-selectors#usage-in-ember-addons)?

@ZedLi Sorry I am not understading your question 😅

ZedLi:
> @ZedLi Sorry I am not understading your question 😅

I was just asking if it needs to be a peer dependency, should it be a normal dependency and if so, should we raise it with the design team rather than us adding the dependency as the consuming application

randallmorey:
It is already a dependency of the addon:  https://github.com/hashicorp/design-system/blob/main/packages/components/package.json#L40.  So

DhariniJeeva (approved):
Based on the discussion [here](https://hashicorp.slack.com/archives/C7KTUHNUS/p1668718311716599?thread_ts=1668629216.870039&cid=C7KTUHNUS) seems like it will be fixed on the HDS side to remove this dependency.  But approving this, if you want to remove the warning noise while you deal with other deprecations

lisbet-alvarez (approved):
LGTM

PR #1473, opened 2022-11-16
Remove rose badge styles

DhariniJeeva:


ui/desktop/tests/end2end/targets.spec.js
```diff
@@ -110,7 +110,7 @@ test.describe('Targets end to end test suite', async () => {
       'div.rose-layout-page-body >> table >> tbody >> tr >> nth=0 >> td >> nth=1 >> .copyable-content'
     );
     const currentStatus = await boundaryWindow.innerText(
-      'div.rose-layout-page-body >> table >> tbody >> tr >> nth=0 >> td >> nth=3 >> .rose-badge-body'
+      'div.rose-layout-page-body >> table >> tbody >> tr >> nth=0 >> td >> nth=3 >> .hds-badge'
```
DhariniJeeva, line 113: looks like we don't use this right now, but I updated this anyway to use the right class

calcaide: Thank you for caring!! Hopefully this quarter we might have some time to add some smoke test to run at development stage.

calcaide (approved):
LGTM!! Thanks

PR #1477, opened 2022-11-17
fix: 🐛 Fix accounts not displaying correctly in auth methods

ZedLi:
## Description

As part of the auth method model fragment refactor, I noticed that accounts and members are currently not correctly displaying their info. It looks like we refactored the account `attributes` and hoisted them up with `isNestedAttribute` but never updated the usage of accounts in auth methods.

:technologist: [Admin preview](https://boundary-ui-git-bugfix-fix-account-displays-in-cf579e-hashicorp.vercel.app/scopes/s_id6ozx3wue/auth-methods/am_wtbbeizy0q/accounts)

Befeore:
image

After:
image


dvonanderson (approved):
LGTM

calcaide (approved):
Good Catch!!

randallmorey (approved):

PR #1478, opened 2022-11-17
chore: 🤖 Upgrade ember-browser-services

calcaide:
## Description

Upgrade `ember-browser-services`.

ember-browser-services brings 2 key dependencies for Ember 4 work: `ember-cli-htmlbars` and `ember-cli-babel`. Updating 2 major versions of `ember-browser-services` brings fresh updated `ember-cli-htmlbars` and `ember-cli-babel`.


ZedLi (approved):
I wasn't aware we were using this package, I had created our own simple service to handle this earlier when I needed to use `window`. Perhaps we should make a ticket to standardize and use this package instead?

calcaide:
@ZedLi 

My 2 cents:
- Me neither! I discover this package while debugging and following breadcrumbs of dependencies (ember-cli-htmlbars and ember-cli-babel). Since I suspect this is one of the multiple causes addons/core is having a hard time upgrading to Ember 4.
I did not explore why we have this package and what we use it for, just wanted to move forward with the roadblock.

- Sure!! I think that's a good idea. After checking a bit more I think is good to use a library that encapsulates all this hassle for us (dealing with browser API's). Also the usage of `window` seems very straigt forward with this library. Let's bring it to FE tech time? I am all down for it, but to let the team know 😃

cameronperera (approved):

PR #1485, opened 2022-11-18
chore: 🤖 JSON credential data modeling

dvonanderson:
✅ Closes: ICU-6653, ICU-6684, ICU-6685, ICU-6802, ICU-6803

:tickets: [Jira ticket](https://hashicorp.atlassian.net/browse/ICU-6653) [Jira ticket](https://hashicorp.atlassian.net/browse/ICU-6684) [Jira ticket](https://hashicorp.atlassian.net/browse/ICU-6685) [Jira ticket](https://hashicorp.atlassian.net/browse/ICU-6802) [Jira ticket](https://hashicorp.atlassian.net/browse/ICU-6803) 

## Description
Added json type to credential model, created `json_object` attribute for `json` type, updated credential serializer to handle `json_object` attribute data being sent to backend and updated normalizer to handlle secret untracked 'json_object'.

### DIFF
--- a/addons/api/addon/models/credential.js
+++ b/addons/api/addon/models/credential.js
@@ -3,7 +3,7 @@ import GeneratedCredentialModel from '../generated/models/credential';
 /**
  * Supported Credential types.
  */
-export const types = ['username_password', 'ssh_private_key'];
+export const types = ['username_password', 'ssh_private_key', 'json'];
 
 export default class CredentialModel extends GeneratedCredentialModel {
   // =attributes
--- a/addons/api/addon/models/credential.js
+++ b/addons/api/addon/models/credential.js
@@ -3,7 +3,7 @@ import GeneratedCredentialModel from '../generated/models/credential';
 /**
  * Supported Credential types.
  */
-export const types = ['username_password', 'ssh_private_key', 'json'];
+export const types = ['username_password', 'ssh_private_key'];
 
 export default class CredentialModel extends GeneratedCredentialModel {
   // =attributes
--- a/addons/api/addon/serializers/credential.js
+++ b/addons/api/addon/serializers/credential.js
@@ -13,6 +13,20 @@ export default class CredentialSerializer extends ApplicationSerializer {
    */
   serializeScopeID = false;
 
+  serialize() {
+    const serialized = super.serialize(...arguments);
+
+    if (serialized.type === 'json') {
+      delete serialized.attributes.username;
+
+      serialized.attributes.object = serialized.attributes.json_object;
+      delete serialized.attributes.json_object;
+      let parsedSecret = JSON.parse(serialized.attributes.object);
+      serialized.attributes.object = parsedSecret;
+    }
+
+    return serialized;
+  }
 
   normalize(typeClass, hash, ...rest) {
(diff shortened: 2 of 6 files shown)


addons/api/addon/models/credential.js
```diff
@@ -3,7 +3,7 @@ import GeneratedCredentialModel from '../generated/models/credential';
 /**
  * Supported Credential types.
  */
-export const types = ['username_password', 'ssh_private_key'];
+export const types = ['username_password', 'ssh_private_key', 'json'];
```
cameronperera, line 6: We can leave JSON credentials unknown until the we have the forms created.
```suggestion
export const types = ['username_password', 'ssh_private_key'];
```

addons/api/addon/serializers/credential.js
```diff
@@ -13,6 +13,20 @@ export default class CredentialSerializer extends ApplicationSerializer {
...
+  serialize() {
+    const serialized = super.serialize(...arguments);
+
+    if (serialized.type === 'json') {
+      delete serialized.attributes.username;
+
+      serialized.attributes.object = serialized.attributes.json_object;
+      delete serialized.attributes.json_object;
+      let parsedSecret = JSON.parse(serialized.attributes.object);
+      serialized.attributes.object = parsedSecret;
```
cameronperera, line 25:
```
      if (serialized?.attributes?.json_object) {
            const parsedSecret = JSON.parse(serialized.attributes.json_object);
            delete serialized.attributes.json_object;
            serialized.attributes.object = parsedSecret;
      }
```
I just remembered we need a check to make sure `json_object` has value before we try to parse it. In some cases like editing, they may not have updated the secret and the `json_object` key/value pair won't exist anymore.

addons/core/translations/resources/en-us.yaml
```diff
@@ -436,10 +436,12 @@ credential:
   types:
     username_password: Username & Password
     ssh_private_key: Username & Key Pair
+    json: JSON
     unknown: Unknown
   help:
     username_password: This host requires a username and password to connect.
     ssh_private_key: This host requires a username, public key, and private key to connect.
+    json: This host requires JSON blob to connect.
```
cameronperera, line 444: I talked with Joshua about what the help text should be and he agreed we should have all Credential type help text match what is in the wireframe. We should go ahead and update all the help text.
Screen Shot 2022-11-18 at 10 47 30 AM

cameronperera (changes requested):

addons/api/addon/serializers/credential.js
ZedLi, line 25: You could've just parsed this first and stored it in `serialized.attributes.object` rather than saving to `serialized.attributes.object` twice.

e.g. `serialized.attributes.object = JSON.parse(serialized.attributes.json_object)`

dvonanderson: thansk @ZedLi fixed

dvonanderson: @cameronperera take a look when you get a chance

cameronperera (approved):
Beautiful. Looks great!

ZedLi (approved):

PR #1489, opened 2022-11-21
Ember upgrade 4.4 LTS addons/auth

calcaide:
## Description

**Be aware this merges to `refactor-upgrade-ember-4` long living branch.**

Upgrade from Ember 3.x to Ember 4.4 LTS within `addons/auth`.

### DIFF
--- a/addons/auth/.ember-cli
+++ b/addons/auth/.ember-cli
@@ -5,5 +5,11 @@
 
     Setting `disableAnalytics` to true will prevent any data from being sent.
   */
-  "disableAnalytics": false
+  "disableAnalytics": false,
+
+  /**
+  Setting `isTypeScriptProject` to true will force the blueprint generators to generate TypeScript
+  rather than JavaScript by default, when a TypeScript version of a given blueprint is available.
+  */
+  "isTypeScriptProject": false
 }
--- a/addons/auth/package.json
+++ b/addons/auth/package.json
@@ -8,7 +8,7 @@
     "ember-addon"
   ],
   "scripts": {
-    "lint": "npm-run-all --aggregate-output --continue-on-error --parallel 'lint:!(fix)'",
+    "lint": "npm-run-all --aggregate-output --continue-on-error --parallel \"lint:!(fix)\"",
     "lint:fix": "npm-run-all --aggregate-output --continue-on-error --parallel lint:*:fix",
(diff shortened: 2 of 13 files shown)


addons/auth/package.json
```diff
@@ -46,11 +46,11 @@
     "ember-disable-prototype-extensions": "^1.1.3",
     "ember-export-application-global": "^2.0.1",
     "ember-load-initializers": "^2.1.2",
-    "ember-maybe-import-regenerator": "^1.0.0",
     "ember-page-title": "^7.0.0",
     "ember-qunit": "^6.0.0",
     "ember-resolver": "^8.0.3",
-    "ember-source": "~3.28.8",
+    "ember-source": "~4.4.0",
+    "ember-source-channel-url": "^3.0.0",
```
calcaide, line 53: I add `ember-source-channel-url` as a devDependency.

The cause is the linter complaining about [this rule](https://github.com/mysticatea/eslint-plugin-node/blob/master/docs/rules/no-extraneous-require.md).

It is used JUST by `ember-try`, so an alternative to this will be ignoring the especific rule mentioned above for `config/ember-try`.

I already add to `.npmignore` the `config/ember-try` but still complaining about the issue mentioned above. Happy to discuss other solutions.

addons/auth/.ember-cli
```diff
@@ -5,5 +5,11 @@
...
     Setting `disableAnalytics` to true will prevent any data from being sent.
   */
-  "disableAnalytics": false
+  "disableAnalytics": false,
+
+  /**
+  Setting `isTypeScriptProject` to true will force the blueprint generators to generate TypeScript
+  rather than JavaScript by default, when a TypeScript version of a given blueprint is available.
+  */
+  "isTypeScriptProject": false
```
calcaide, line 14: Even tho this is the default value, ember-cli added this change when running the upgrade script, I think is best to keep it.

addons/auth/package.json
randallmorey: I didn't have any issues locally.  I pushed up an updated yarn.lock so perhaps that will help.

calcaide: Yes, I validate your comment. Pull latest changes, re-install deps, run the addon and everything seems fine to me.

ZedLi (approved):
LGTM!

PR #1497, opened 2022-11-23
fix: 🐛 Fix claims map dropdown not having a default value

ZedLi:
## Description
This fixes the claims map `to` field not having a default value when adding a claim map even though it looks like it's selected


DhariniJeeva (approved):
ty!

calcaide (approved):
LGTM

PR #1498, opened 2022-11-23
feat: 🎸 create form for json credentials

lisbet-alvarez:
✅ Closes: ICU-6682

:tickets: [Jira ticket](https://hashicorp.atlassian.net/browse/ICU-6682)
:technologist: [Admin preview](https://boundary-ui-git-icu-6682-create-json-credentials-hashicorp.vercel.app/scopes/s_s13jlc148p/credential-stores/cs_o2wxelugtt/credentials/new?type=json)

## Description
Form for creating json credentials
Just a draft for now because I need to add code editor, info field comp., and hds radio cards.


addons/api/addon/models/credential.js
```diff
@@ -3,7 +3,7 @@ import GeneratedCredentialModel from '../generated/models/credential';
 /**
  * Supported Credential types.
  */
-export const types = ['username_password', 'ssh_private_key'];
+export const types = ['username_password', 'ssh_private_key', 'json'];
```
cameronperera, line 6: FYI, we need to add a feature flag check along with this change. We want to make sure in a production env we cannot access JSON credentials still. I would look at how [credential store abilities](https://github.com/hashicorp/boundary-ui/blob/main/addons/api/addon/abilities/credential-store.js#L18) is using the feature service and do something similar in the [credential ability](https://github.com/hashicorp/boundary-ui/blob/main/addons/api/addon/abilities/credential.js#L12-L14) `canRead`

lisbet-alvarez: Thank you for catching that! I will look into that.

ui/admin/tests/acceptance/credential-store/credentials/create-test.js
```diff
@@ -188,17 +242,49 @@ module(
...
+                  description: 'Field required for creating a json credential.',
+                },
+              ],
+            },
+          }
+        );
+      });
+
+      await click(`[href="${urls.newCredential}"]`);
+      await click('[value="ssh_private_key"]');
```
lisbet-alvarez, line 282:
```suggestion
      await click('[value="json"]');
```

ui/admin/app/routes/scopes/scope/credential-stores/credential-store/credentials/new.js
```diff
@@ -29,15 +29,19 @@ export default class ScopesScopeCredentialStoresCredentialStoreCredentialsNewRou
       'scopes.scope.credential-stores.credential-store'
     );
     let name, description;
+    const json_object = '';
```
cameronperera, line 32:
```suggestion
    const json_object = '{}';
```
This should be set to '{}' so the user isn't shown a linter error right out of the gate.

addons/api/package.json
```diff
@@ -33,6 +33,8 @@
     "ember-cli-mirage": "^2.2.0",
     "ember-copy": "^2.0.1",
     "ember-data": "^3.20.0",
+    "ember-data-model-fragments": "5.0.0-beta.2",
```
cameronperera, line 36: I think this is actually gone now. @calcaide is that correct?

addons/api/addon/models/credential.js
```diff
@@ -1,12 +1,15 @@
 import GeneratedCredentialModel from '../generated/models/credential';
+import { inject as service } from '@ember/service';
 
 /**
  * Supported Credential types.
  */
-export const types = ['username_password', 'ssh_private_key'];
+export let types = ['username_password', 'ssh_private_key'];
```
randallmorey, line 7: What about json?

addons/api/package.json
calcaide: @cameronperera that is correct is gone and merged to main [in this PR](https://github.com/hashicorp/boundary-ui/pull/1507/files#diff-3e27e6039dc4ff0f8e02759ce553da2ec6691658e4e0f1244483f5d564c8ff44L35) (kudos to @ZedLi !!)

@lisbet-alvarez perhaps this branch needs to be updated from main to get latest changes?

ui/admin/app/components/form/credential/json/index.hbs
```diff
@@ -0,0 +1,116 @@
...
+  >
+    {{#if @model.errors.description}}
+      <field.errors as |errors|>
+        {{#each @model.errors.description as |error|}}
+          <errors.message>{{error.message}}</errors.message>
+        {{/each}}
+      </field.errors>
+    {{/if}}
+  </form.textarea>
+  {{#if (feature-flag 'json-credentials')}}
```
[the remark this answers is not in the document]
randallmorey: The JSON cred feature flag doesn't make sense within the JSON cred form itself.  If this form is exposed, it means the feature is definitely enabled.

randallmorey (changes requested):
Looking great.  Made some change requests.

As a rule of thumb, we should avoid testing translated strings _directly_.  These are subject to change over the long term.  We discussed providing a helper to do this based on translation keys rather than translated strings, but this doesn't exist yet.

Note:  when the string contents of error messages come directly from the test case itself, it's fine to match the actual output strings.  These are _self contained_ within the test and not subject to change by outside concerns.

ui/admin/tests/acceptance/credential-store/credentials/create-test.js
```diff
@@ -188,17 +244,70 @@ module(
...
+          'create'
+        )
       );
+      assert.dom('.rose-form-radio-card').exists({ count: 2 });
+      assert
+        .dom('.rose-form-fieldset-body div:first-child label')
+        .includesText('Username & Password');
+      assert
+        .dom('.rose-form-fieldset-body div:nth-child(2) label')
+        .includesText('Username & Key Pair');
```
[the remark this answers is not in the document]
cameronperera: What if we assert against the `name` or `value` of the input element? Those will not change like translations might.
Screen Shot 2022-12-06 at 11 37 29 AM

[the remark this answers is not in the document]
randallmorey: SGTM

ui/admin/app/routes/scopes/scope/credential-stores/credential-store/credentials/new.js
calcaide: I understand we are doing this for linting purposes, but is quite confusing to set a variable with the word `object` on it as a string.

Linter is a tool to help developers, perhaps I am missing context but I would preffer to skip the linting issue here rather than setting an object as a string.

As a quick test, I checkout this branch, set `const json_object = {}` and got no errors ?

calcaide (changes requested):
Left some comments.

- This PR should be updated within main, `ember-data-model-fragments` is gone and should not be updated here.
- I am fine upgrading `ember-feature-flags` here and not asking for a change, but usually, for such a delicate piece (this is very important for Boundary UI) I will suggest an specific separate PR where we can validate the upgrade does brake nothing without adding extra noise.

addons/api/package.json
```diff
@@ -32,7 +32,7 @@
     "ember-cli-htmlbars": "^6.1.0",
     "ember-cli-mirage": "^2.2.0",
     "ember-copy": "^2.0.1",
-    "ember-data": "^3.28.12",
+    "ember-feature-flags": "^6.0.0",
```
ZedLi, line 35: Was `ember-data` accidentally removed here?

And were we previously using `ember-feature-flags` in API without actually installing it in our deps?

addons/api/addon/models/credential.js
```diff
@@ -23,6 +26,9 @@ export default class CredentialModel extends GeneratedCredentialModel {
    * @type {boolean}
    */
   get isUnknown() {
+    types = this.features.isEnabled('json-credentials')
```
ZedLi, line 29: Are we intentionally re-setting the `types` variable if `isUnknown` accessed? That seems like it could be prone to unexpected behavior since we're exporting it for use outside the model and it won't change until the `isUnknown` getter is called.

Can't we create the right types on initialization based off the feature flag rather than re-assigning? The consumers of the types downstream then don't need to re-filter or check with the same feature flag

ZedLi (changes requested):

addons/api/package.json
calcaide: Actually that's true. I do not see a usage of `ember-feature-flgas` within API 🤔

addons/api/addon/models/credential.js
randallmorey: The API addon can and probably should know about any available types.  Let's not worry about feature flags within API addon.  This would be separate from any UI related concerns where we hide or show CRUD features based on a feature flag.  So, let's list `json` here with no special handling in the addon.

ZedLi: That's fine too, I would also have expected all the types to be available but my main concern was the re-assigning of the `types` variable so either way works for me.

ui/admin/app/components/form/credential/json/index.hbs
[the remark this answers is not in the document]
cameronperera: Removed.

addons/api/package.json
cameronperera: Removed.

ui/admin/app/components/form/credential/index.js
```diff
@@ -0,0 +1,16 @@
...
+
+export default class FormCredentialComponent extends Component {
+  // =services
+  @service features;
+
+  // =attributes
+  /**
+   * @type {object}
+   */
+  credentialTypes = this.features.isEnabled('json-credentials')
```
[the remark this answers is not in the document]
cameronperera: Done.

ui/admin/tests/acceptance/credential-store/credentials/create-test.js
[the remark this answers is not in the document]
cameronperera: Going to address this once the Radio Card migration PR for Credential forms is merged in.

ui/admin/app/components/form/credential/json/index.hbs
```diff
@@ -0,0 +1,116 @@
...
+        <F.HelperText>{{t (concat 'resources.credential.types.' @model.type)}}
+        </F.HelperText>
+      </InfoField>
+
+      <form.fieldset>
+        <:title>{{t 'resources.credential.form.json.label'}}</:title>
+        <:description>{{t 'resources.credential.form.json.help'}}</:description>
+        <:body>
+          <Form::Field::JsonSecret
+            @disabled={{form.disabled}}
```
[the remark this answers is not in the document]
cameronperera: Forms have been updated.

addons/api/package.json
cameronperera: This has been resolved.

addons/api/addon/models/credential.js
cameronperera: This is getting a refactor based on the discussion I had with @randallmorey yesterday. I will ping the team once it is pushed up.

calcaide:
@cameronperera 

I just notice this PR shows changes within `yarn.lock` but has no changes within any `package.json`. Make sure you update this branch with `main` and sync those dependencies 😉  If you have any question regarding that let me know

randallmorey (changes requested):
Let's remove any feature flagging from the API addon.

cameronperera:
> Let's remove any feature flagging from the API addon.

Doing that now and creating a credential ability override that will utilize the feature flag in admin ui.

ui/admin/tests/acceptance/credential-store/credentials/create-test.js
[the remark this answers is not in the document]
cameronperera: Updated.

ui/admin/app/components/form/credential/json/index.hbs
```diff
@@ -0,0 +1,116 @@
...
+  as |form|
+>
+
+  <form.input
+    @name='name'
+    @value={{@model.name}}
+    @label={{t 'form.name.label'}}
+    @error={{@model.errors.name}}
+    @helperText={{t 'form.name.help'}}
+    disabled={{form.disabled}}
```
DhariniJeeva, line 15: I don't think both the disabled/readonly do anything here. `@disabled={{@model.isSaving}}` should take care of disabling the input fields when the form is disabled.

calcaide:
Cameron and I reviewed changes within `yarn.lock`. As you can see there are no changes within `package.json` and the only changes reflected within `yarn.lock` is the integrity sha. Before `sha1` now `sha512`. As per my understanding ([this post made me realize](https://github.com/yarnpkg/yarn/issues/7658)) this change reflects an upstream change not related to us but to the registry.
Anyway is better to have `sha512` rather than `sha1` so we do not think this will cause any issues.

ui/admin/app/components/form/credential/json/index.hbs
cameronperera: Removed those attributes.

ui/admin/app/components/form/credential/json/index.hbs
```diff
@@ -0,0 +1,112 @@
...
+      <F.HelperText>{{t (concat 'resources.credential.types.' @model.type)}}
+      </F.HelperText>
+    </InfoField>
+
+    <form.fieldset>
+      <:title>{{t 'resources.credential.form.json.label'}}</:title>
+      <:description>{{t 'resources.credential.form.json.help'}}</:description>
+      <:body>
+        <Form::Field::JsonSecret
+          @disabled={{form.disabled}}
```
DhariniJeeva, line 95: Do we still need this one?

ui/admin/app/abilities/credential.js
```diff
@@ -0,0 +1,14 @@
...
+import { inject as service } from '@ember/service';
+
+export default class OverrideCredentialAbility extends CredentialAbility {
+  // =service
+  @service features;
+
+  get canRead() {
+    return this.features.isEnabled('json-credentials') ||
+      this.model.type !== 'json'
+      ? super.canRead
```
ZedLi, line 11: Shouldn't this be:
```suggestion
      ? super.canRead()
```

Do we have tests for these different scenarios?

ui/admin/app/abilities/credential.js
```diff
@@ -0,0 +1,14 @@
+import CredentialAbility from 'api/abilities/credential';
+import { inject as service } from '@ember/service';
+
+export default class OverrideCredentialAbility extends CredentialAbility {
+  // =service
+  @service features;
+
+  get canRead() {
+    return this.features.isEnabled('json-credentials') ||
```
ZedLi, line 9: Does this make sense as the first check? Couldn't this ability be used for other credentials which would always return `true`? I would have expected us to check for `type === 'json' && features.isEnabled`

ZedLi (changes requested):

ui/admin/app/components/form/credential/json/index.hbs
cameronperera: Yes. This form field is not actually connected to rose so the `@disabled={{@model.isSaving}}` does not reach those component.

ui/admin/app/components/form/credential/json/index.hbs
```diff
@@ -0,0 +1,112 @@
...
+          @value={{@model.json_object}}
+          @onInput={{fn (mut @model.json_object)}}
```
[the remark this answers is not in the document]
cameronperera: This was suggestion by @randallmorey and removed the need for creating an `onChange` action. I was not aware of the linting rule. We have been using `mut` in a few places in our app so I didn't think it was discouraged.

ui/admin/app/abilities/credential.js
cameronperera: `canRead` is a getter so you don't invoke it like a function. I actually added the `()` at first and it was causing errors.

cameronperera: As for the test question, I asked the team about that. I will update this accordingly.

ui/admin/app/components/form/credential/json/index.hbs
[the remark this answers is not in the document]
ZedLi: Ah it turns out it's not turned on by default. We can discuss this more in tech time but I would prefer not using `mut` and since we don't use it in too many places, it wouldn't be too bad to remove its usage.

Would definitely be interested to hear others' opinions

ui/admin/app/abilities/credential.js
ZedLi: Good point, totally thought it was a function and not a getter

ui/admin/app/abilities/credential.js
cameronperera: This is being used by all credentials.

My thought behind this was if the flag is enabled then we don't care about the right side of the `||` and we carry on with the original credential ability. If the flag is disabled we then move the right side of the `||` and we check the credential type. If it is not a `json` credential we let the original credential ability do it's job. Everything else should be a `json` credential while the feature is disabled and we do not want it to be able to read.

ZedLi: > My thought behind this was if the flag is enabled then we don't care about the right side of the `||`

Why is this? If for example we were in a non `json` credential _and_ the flag is on, they would all be able to read now because the boolean check will short circuit and we would never get the `super.canRead` check. Shouldn't we only be checking the flag in conjunction with the `json` type?

ui/admin/tests/acceptance/credential-store/credentials/create-test.js
```diff
@@ -188,17 +244,70 @@ module(
...
+      await click(`[href="${urls.newCredential}"]`);
+      await click('[value="json"]');
+      await click('[type="submit"]');
+
+      assert
+        .dom('.rose-notification-body')
+        .hasText('Error in provided request.');
+    });
+
+    test('cannot navigate to json credential when feature is disabled', async function (assert) {
```
cameronperera, line 292: @ZedLi this is the test that verifies that when the flag is disabled, the `canRead` ability is working properly.

ZedLi: Can we actually add some unit tests for the `ability`? 

I don't think this makes it clear that non `json` types _can't_ read if they don't have permissions and the feature flag is enabled.

ui/admin/app/abilities/credential.js
ZedLi: Talked with @cameronperera, turns out I was misled by the formatting of this, the check short circuits against the `this.model.type !== 'json'` and not against the rest of the ternary so it works fine!

ui/admin/app/components/form/credential/json/index.hbs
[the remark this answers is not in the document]
randallmorey: We'll need a custom helper to replace this along the lines of `set-from-codemirror @model 'json_object'`.  This is a special case where `onInput` doesn't pass an event as you would expect were this a true form field.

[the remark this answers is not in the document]
cameronperera: I would like to discuss this topic as a team. I will add it to our agenda. If we decide to move away from `mut` we can turn that lint option on and update our uses of `mut`.

DhariniJeeva (approved):

ZedLi (approved):

randallmorey (approved):

PR #1499, opened 2022-11-23
Icu 6680 UI code editor for secrets

cameronperera:
:tickets: [Jira ticket](https://hashicorp.atlassian.net/browse/ICU-6680)

## Description
This is the code editor that will be used for editing the `json_object` attribute for a JSON credential. This is my first time creating a component so please don't hold back on the feedback. Below is a video walking through the 4 different states this editor can be in.

1. Actual code editor during initial creation of credential
2. Detail screen view with message confirming secret was saved
3. Edit screen view with button to take the user back to the code editor
4. Editing screen view but the actual code editor now

This SecretEditor is currently only going to be used within the JSON Credential form component so it lives very close to it for that reason.

Usage: 
JSONSecret
```
<Form::Field::JsonSecret
  @disabled={{form.disabled}}
  @value={{@model.json_object}
  @onInput={{fn (mut @model.json_object)}}
  @showEditButton={{true}}
/>
```
JSONSecret::Skeleton
```
<Form::Field::JsonSecret::Skeleton>
  <Rose::Icon @name='flight-icons/svg/eye-off-16' @size='medium' />
  {{t 'form.secret-editor.details'}}
</Form::Field::JsonSecret::Skeleton>
```

_NOTE: I had to mock the JSON form to test this component but it does not exist in this branch. So pulling this branch and testing locally won't work without some mocking._

https://user-images.githubusercontent.com/29386339/203649712-5f25a53d-e633-4d66-bb64-48617397e437.mp4


ui/admin/app/components/form/credential/json/secret-editor/index.js
```diff
@@ -0,0 +1,29 @@
...
+export default class FormCredentialJsonSecretEditorComponent extends Component {
+  // =actions
+  @tracked allowEditing = false;
+
+  @action
+  allowEdit() {
+    this.allowEditing = true;
+  }
+
+  @computed('args.{isNew,isEditing}')
```
ZedLi, line 15: Are these `computed` needed? `allowEditing` is already tracked and I thought `args` were automatically tracked with glimmer components?

ui/admin/ember-cli-build.js
```diff
@@ -30,6 +30,8 @@ module.exports = function (defaults) {
     app.import('node_modules/clipboard/dist/clipboard.min.js');
   }
 
+  app.import('node_modules/jsonlint/lib/jsonlint.js');
```
ZedLi, line 33: Is it possible to have this import handled in rose so the consumer doesn't need to to install it?

ui/admin/app/components/form/credential/json/secret-editor/index.hbs
```diff
@@ -0,0 +1,50 @@
+<div class='secret-editor'>
+  {{#if this.canEdit}}
+    <Rose::CodeEditor as |c|>
+      <c.fieldEditor
+        class='json-secret-editor'
+        @onInput={{@onInput}}
```
ZedLi, line 6: Not related to this PR, but I realized it looks like we don't actually pass `onInput` to the editor in the `code-editor`

ui/admin/app/styles/app.scss
```diff
@@ -452,3 +454,103 @@
...
 }
+
+// secret editor
+.json-secret-editor {
+  .CodeMirror {
+    height: 210px;
+  }
+
+  .CodeMirror-gutters {
+    background: #31332b;
```
ZedLi, line 465: Do we need these hard coded colors? Does our current color scheme not work?

ui/admin/app/styles/app.scss
```diff
@@ -452,3 +454,103 @@
...
+  }
+}
+
+.secret-editor {
+  background: #272822;
+  border-radius: 4px;
```
ZedLi, line 471: Don't we normally use `sizing.rems` instead of direct pixels?

ui/admin/app/styles/app.scss
```diff
@@ -452,3 +454,103 @@
...
+  position: absolute;
+  top: 50%;
+  z-index: 1;
+
+  .rose-button-body {
+    font-size: 14px;
+  }
+}
+
+.rose-button-secret-editor-button {
```
ZedLi, line 546: Where is this class being used?

ui/admin/app/styles/app.scss
```diff
@@ -452,3 +454,103 @@
...
+  background: #272822;
+  border-radius: 4px;
+  height: 210px;
+  margin-bottom: sizing.rems(l);
+}
+
+.skeleton-wrapper {
+  color: var(--ui-gray);
+  display: flex;
+  flex-direction: row;
```
ZedLi, line 479: This is technically not needed since it's `row` by default but also fine to be explicit

ui/admin/app/styles/app.scss
```diff
@@ -452,3 +454,103 @@
...
+  }
+}
+
+.rose-button-secret-editor-button {
+  background: #272822;
+  border: 1px solid #6f7682;
+  border-radius: 2px;
+  box-shadow: 0 4px 4px #000004;
+  height: 36px;
+  left: 50%;
```
ZedLi, line 552: Are these `left` and `top` necessary? Aren't they already being handled by this button's wrapper?

ui/admin/app/styles/app.scss
```diff
@@ -452,3 +454,103 @@
...
+    &.xl {
+      width: 20%;
+    }
+  }
+}
+
+.details-view {
+  font-size: 14px;
+  left: 50%;
+  margin: -12px 0 0 -165.5px;
```
ZedLi, line 524: Why are these negative margins needed, is this to account for the size of the element itself? Could we have used `transform: translate(-50%, -50%)` instead?

ZedLi (changes requested):

ui/admin/app/styles/app.scss
```diff
@@ -452,3 +454,103 @@
...
+  position: absolute;
+  top: 50%;
+  z-index: 1;
+
+  .rose-icon {
+    #{--size}: sizing.rems(m);
```
ZedLi, line 530: Is this class and work around even needed, how come we couldn't have just set passed a `@size='medium'` to the rose icon?

Also are we continuing to use rose icon or are we swapping to the new hds flight icons?

ui/admin/app/styles/app.scss
randallmorey: My preference is to avoid arbitrary pixel values like this.  Is there a way to accomplish the desired positioning in a way that doesn't depend on the element size?  If pixels are pragmatic, perhaps include some comments for context.

randallmorey (changes requested):
Couple of requests:
- Can we get generous comments to explain what's going on in the CSS?  It's not immediately clear.
- Can the class names use a consistent prefix for this feature?

ui/admin/app/components/form/credential/json/secret-editor/index.js
cameronperera: Yep that is right. I was having issues with `allowEditing` triggering a render after it was changed but I think it was due to using computed vs leaving it off and letting tracked do its job. Fixed now.

ui/admin/app/styles/app.scss
cameronperera: 😮‍💨 I got lost in the weeds of editing styles and just did it here, forgetting I can just set `@size`. Fixed now.

cameronperera:
> Couple of requests:
> 
> * Can we get generous comments to explain what's going on in the CSS?  It's not immediately clear.
> * Can the class names use a consistent prefix for this feature?

Absolutely. I am going through the feedback from Zed and I will do this as well.

ui/admin/app/styles/app.scss
cameronperera: This color was added based off the Monokai color scheme. It was needed primarily for creating the editor skeleton. I was going to ask the team and @joshuaogle if I should created a color variable for this which I am happy to do. I was not exactly sure where best to declare that though.

ui/admin/app/styles/app.scss
cameronperera: Fixed.

ui/admin/app/styles/app.scss
cameronperera: Thanks. I thought I cleared up all the unnecessary styling. Removed.

ui/admin/app/styles/app.scss
cameronperera: You are correct. Removed.

ui/admin/app/styles/app.scss
cameronperera: Updated to use transform.

addons/rose/addon/modifiers/code-mirror.js
```diff
@@ -9,6 +9,8 @@ import 'codemirror/addon/edit/matchbrackets';
 import 'codemirror/addon/edit/closebrackets';
 import 'codemirror/addon/selection/active-line';
 import 'codemirror/addon/display/autorefresh';
+import 'codemirror/addon/lint/lint'
```
calcaide, line 12: Asking for curisotiy: what are these linters for? for the editor itself?

ui/admin/app/styles/app.scss
```diff
@@ -452,3 +457,100 @@
     }
   }
 }
+
+// secret editor
```
calcaide, line 461: Commenting with no context:
In [this file](https://github.com/hashicorp/boundary-ui/pull/1499/files#diff-3b83e0b7d2d1b6c3989f81f6450624eb7a3b70a7cf5c65d886a0449a266b5816R1) I understand `<div class='secret-editor'>` as the root markup of our editor. Is that right?

If is not right, ignore the rest of the comment.

If so, for dev experience and facilitate understanding of the code, what do you think of wrapping all the css regarding the editor within `.secret-editor` ?

Something like:
```css
.secret-editor {
  ...css code

  .json-secret-editor { ... }

and so on
}
```

This way helps to understand all the code within `.json-secret-editor` pertains to the editor and helps isolating that css to avoid override collisions.

What do you think?

ui/admin/ember-cli-build.js
calcaide: +1 to @ZedLi comment.

I imagine the editor consumption as any other of our Rose components, just using it on the markup.

ui/admin/package.json
```diff
@@ -35,7 +35,8 @@
     "posttest": "ember coverage-merge && yarn clean:coverage"
   },
   "dependencies": {
-    "ember-named-blocks-polyfill": "^0.2.5"
+    "ember-named-blocks-polyfill": "^0.2.5",
```
calcaide, line 38: Can you explain why we are adding these 2 dependencies to admin?

calcaide:
@cameronperera looking good but left some comments ;)

ui/admin/ember-cli-build.js
cameronperera: It is possible. Moved this import to rose.

addons/rose/addon/modifiers/code-mirror.js
cameronperera: Yes. I saw that CodeMirror allows for a live linter within the editor itself. I had to add this base lint import to also use the jsonlint that we are using for this component.

ui/admin/app/styles/app.scss
cameronperera: That is actually how I originally had it before I ever opened the PR but I wasn't sure if that was desired or not. I think we could discuss that during tech time to get the whole team's input.

ui/admin/package.json
cameronperera: I originally added `jsonlint` which is a package needed for the linter built into CodeMirror. However, this has been removed from `admin/package.json` and moved to `rose/package.json` to keep all CodeMirror related imports as close to Rose as possible.

ui/admin/app/styles/app.scss
```diff
@@ -452,3 +459,110 @@
...
+
+  // skeleton text rows for detail and edit views
+  // applies general styles to child div of skeleton-body
+  // then adds dynamic width to each child
+  // depending on the class (e.g., s, m, l)
+  > div {
+    background: $secret-editor-gutter;
+    border-radius: sizing.rems(m);
+    height: sizing.rems(xs);
+    margin: 11px;
```
DhariniJeeva, line 509:
```suggestion
```

ui/admin/app/components/form/credential/json/secret-editor/index.js
```diff
@@ -0,0 +1,25 @@
...
+  allowEdit() {
+    this.allowEditing = true;
+  }
+
+  get isDetailView() {
+    return !this.args.isNew && !this.args.isEditing;
+  }
+
+  get isEditView() {
+    return !this.args.isNew && this.args.isEditing && !this.allowEditing;
```
DhariniJeeva, line 19: So it is an edit view if it's not new and in editing mode and can't edit?

ui/admin/app/styles/app.scss
calcaide: Good point! this is not like a code rule, more developer experience so good call to discuss it on FE tech time 😉

DhariniJeeva:
Awesome work! I left a couple of questions!

ui/admin/app/styles/app.scss
cameronperera: Oops. Thanks for catching that.

ui/admin/app/components/form/credential/json/secret-editor/index.js
cameronperera: Yes, this is confusing a bit. I will add comments here to give more context. To answer your question though, yes.
When editing a JSON credential, the secret editor can have 2 states. It can be in the edit state for making changes to the name and description. This keeps the `json_object` value an empty string. This is important for when the data is serialized. Since `json_object` is marked as a secret, it will remove that attribute from the API request. 
The second state is editing everything. This will "activate" the editor for the user and they can change the secret at this point.
Please let me know if you have more questions. I will add those comments now.

DhariniJeeva: Ty for clarifying. I think comments and renaming to an elaborate variable name might help!

ui/admin/app/styles/app.scss
```diff
@@ -452,3 +459,110 @@
...
+    &.m {
+      width: 10%;
+    }
+
+    &.l {
+      width: 15%;
+    }
+
+    &.xl {
+      width: 20%;
```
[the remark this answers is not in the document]
cameronperera: I could but I wanted a more dynamic width for varying screen sizes. Thoughts on that?

ZedLi (approved):
Great job!

dvonanderson (approved):
Looks great, even if you decide to switch to a different folder structure

randallmorey (approved):

joshuaogle (approved):
Thanks for the runthrough in Zoom and the small tweak to the gutter padding. Looks great! 👍

DhariniJeeva (approved):
LGTM!

PR #1500, opened 2022-11-24
Chore/cleanup filter attribute

ZedLi:
## Description
This PR aims to cleanup the last usage of `attributes_{field}` for error handling. See this [thread](https://hashicorp.slack.com/archives/C012GTH1C4E/p1668703662585969?thread_ts=1668703520.160539&cid=C012GTH1C4E) for more info.

The issue with using `filter` as the field name is it conflicts with the `filter` function for arrays. Therefore I renamed them to `filter_string` to avoid the collision and serialize it back to `filter`.

Also open to different naming suggestions.

### DIFF
--- a/addons/api/addon/serializers/host-set.js
+++ b/addons/api/addon/serializers/host-set.js
@@ -3,7 +3,7 @@ import { copy } from 'ember-copy';
 
 const fieldByType = {
   aws: ['preferred_endpoints', 'filters', 'sync_interval_seconds'],
-  azure: ['preferred_endpoints', 'filter', 'sync_interval_seconds'],
+  azure: ['preferred_endpoints', 'filter_string', 'sync_interval_seconds'],
 };
 export default class HostSetSerializer extends ApplicationSerializer {
   // =properties
@@ -69,6 +69,15 @@ export default class HostSetSerializer extends ApplicationSerializer {
         delete json.attributes[key];
       }
     }
+
+    if (key === 'filter_string') {
+      const { filter_string } = snapshot.record;
+      if (filter_string) {
+        json.attributes.filter = json.attributes.filter_string;
+      }
+      delete json.attributes.filter_string;
+    }
+
     return value;
   }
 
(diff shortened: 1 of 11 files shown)


addons/api/addon/serializers/managed-group.js
```diff
@@ -7,4 +8,29 @@ export default class ManagedGroupSerializer extends ApplicationSerializer {
...
    */
   serializeScopeID = false;
+
+  serialize() {
+    const serialized = super.serialize(...arguments);
+
+    if (serialized.attributes.filter_string) {
+      serialized.attributes.filter = serialized.attributes.filter_string;
+    } else {
+      serialized.attributes.filter = null;
```
ZedLi, line 18: Is this needed? There's a test called `it serializes correctly on create with no filter provided` that explicitly tests that filter is set to `null` when it's not present, I actually couldn't find where it was getting set to null so wasn't sure of the intention.

addons/api/addon/serializers/host-set.js
```diff
@@ -69,6 +69,15 @@ export default class HostSetSerializer extends ApplicationSerializer {
         delete json.attributes[key];
       }
     }
+
+    if (key === 'filter_string') {
+      const { filter_string } = snapshot.record;
+      if (filter_string) {
+        json.attributes.filter = json.attributes.filter_string;
+      }
+      delete json.attributes.filter_string;
```
randallmorey, line 78: Why inspect the snapshot while subsequently operating on `json`?

addons/api/addon/serializers/managed-group.js
randallmorey: By default, empty strings are serialised to `null`.  On serialisation, you might want to check that `filter_string` exists as a key on `attributes` (or check that it is strictly not `undefined`), rather than simply falsy.  That should eliminate the need for an else case, I believe?

randallmorey (changes requested):

addons/api/addon/serializers/host-set.js
ZedLi: Ah I was just pattern matching what was previously done, is the implication I should just check the json directly and not bother interrogating the snapshot?

addons/api/addon/serializers/managed-group.js
ZedLi: That makes sense, I'll update it

addons/api/addon/serializers/host-set.js
randallmorey: If the value comes from `json.attributes`, I'd check that instead.

ZedLi: Yup I updated it!

calcaide (approved):
LGTM! Awesome work

PR #1502, opened 2022-11-28
chore: 🤖 radio card migration - target and Credentials

DhariniJeeva:
[Jira](https://hashicorp.atlassian.net/browse/ICU-6925)

[targets](https://boundary-qwtcuck8o-hashicorp.vercel.app/scopes/s_1sheljmjgm/targets/new?type=tcp)

[credentials](https://boundary-qwtcuck8o-hashicorp.vercel.app/scopes/s_1sheljmjgm/credential-stores/cs_20nu3mgl4y/credentials/credup_hzrxf5z5za)


ui/admin/app/components/form/credential/index.js
```diff
@@ -0,0 +1,10 @@
+import Component from '@glimmer/component';
+import { types } from 'api/models/credential';
+
+export default class FormAuthMethodOidcComponent extends Component {
```
cameronperera, line 4:
```suggestion
export default class FormCredentialComponent extends Component {
```

cameronperera (changes requested):

cameronperera (approved):

dvonanderson (approved):
Awesome

addons/api/addon/models/target.js
```diff
@@ -2,6 +2,11 @@ import GeneratedTargetModel from '../generated/models/target';
 import { attr } from '@ember-data/model';
 import { inject as service } from '@ember/service';
 
+export const types = {
```
ZedLi, line 5: It wasn't obvious to me that the array was names for the icons at first, could we maybe rename the variable to something more descriptive to indicate that?

Another question: is there a reason it's an array?

randallmorey: It's not really just for icons.  It could be for other stuff and we use a similar concept elsewhere.  I think we need some standardisation around the naming and structure of type hashes.

ZedLi: Right, I think it's a little misleading to just call it `types` when usually our other `types` are just arrays for this. 

I think we can keep the types variable just for the array and have a new one for icons that builds off that variable.

DhariniJeeva: I agree. I'll look into standardizing the name and the structure

ui/admin/app/components/form/target/details/index.js
```diff
@@ -0,0 +1,29 @@
...
+//Note: this is a temporary solution till we have resource type helper in place
+const icons = ['network', 'terminal'];
+
+export default class FormTargetComponent extends Component {
+  // =properties
+  /**
+   * maps resource type with icon
+   * @type {object}
+   */
+  get mapResourceTypewithIcon() {
```
ZedLi, line 13:
```suggestion
  get mapResourceTypeWithIcon() {
```

ZedLi (approved):

ui/admin/app/components/form/target/details/index.js
DhariniJeeva: ty! updated in a few places!

PR #1504, opened 2022-11-29
refactor: 💡 Refactor Mirage server configuration

calcaide:
:tickets: [Jira ticket](https://hashicorp.atlassian.net/browse/ICU-7056)

## Description

Mirage Refactor to accomadate new changes within `ember-cli-mirage`, `miragejs` and Ember 4.

More info about it, [here](https://www.ember-cli-mirage.com/docs/advanced/server-configuration)

### DIFF
--- a/addons/api/mirage/config.js
+++ b/addons/api/mirage/config.js
@@ -1,4 +1,6 @@
-import config from '../config/environment';
+import { discoverEmberDataModels } from "ember-cli-mirage";
+import { createServer } from 'miragejs';
+import EnvironmentConfig from '../config/environment';
 import { authHandler, deauthHandler } from './route-handlers/auth';
 import { targetHandler } from './route-handlers/target';
 import { pickRandomStatusString } from './factories/session';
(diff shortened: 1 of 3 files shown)


addons/api/mirage/config.js
```diff
@@ -1,20 +1,35 @@
-import config from '../config/environment';
+import { discoverEmberDataModels } from "ember-cli-mirage";
+import { createServer } from 'miragejs';
+import EnvironmentConfig from '../config/environment';
```
calcaide, line 3: I changed the name here to avoid collisions with the `config` variable that Mirage provides (see line 16). Happy to discuss other names or different approach, I found this the more pragmatic to follow.

addons/api/mirage/config.js
```diff
@@ -1,20 +1,35 @@
...
-export default function () {
-  initializeMockIPC(this, config);
+const isTesting = EnvironmentConfig.environment === 'test';
+
+// Main function
+// More info about seerver configuration https://www.ember-cli-mirage.com/docs/advanced/server-configuration
+export default function(config) {
+  let finalConfig = {
+    ...config,
+    models: { ...discoverEmberDataModels(), ...config.models },
```
ZedLi, line 19: Do we need to also use the `applyEmberDataSerializers` function or no?

ZedLi (approved):
LGTM, just had a few questions

addons/api/mirage/config.js
calcaide: That is a very good point. I do not like variables without context like `config`. In this case was totally fine, because was obvius. But since now we have 2 config variables I love your idea.
- I will rename the mirage config varible to `mirageConfig`.
- I will rename the environment config variable to `environmentConfig`. The Upper camel case was a mistake

addons/api/mirage/config.js
```diff
@@ -1,20 +1,36 @@
-import config from '../config/environment';
+import { discoverEmberDataModels, applyEmberDataSerializers } from "ember-cli-mirage";
+import { createServer } from 'miragejs';
+import environmentConfig from '../config/environment';
```
randallmorey, line 3: Nice.  I prefer the lowercase naming here.

addons/api/mirage/config.js
```diff
@@ -1,20 +1,36 @@
-import config from '../config/environment';
+import { discoverEmberDataModels, applyEmberDataSerializers } from "ember-cli-mirage";
```
randallmorey, line 1:
```suggestion
import { discoverEmberDataModels, applyEmberDataSerializers } from 'ember-cli-mirage';
```

addons/api/mirage/config.js
```diff
@@ -1,20 +1,36 @@
...
+import { createServer } from 'miragejs';
+import environmentConfig from '../config/environment';
 import { authHandler, deauthHandler } from './route-handlers/auth';
 import { targetHandler } from './route-handlers/target';
 import { pickRandomStatusString } from './factories/session';
 import { Response } from 'miragejs';
```
randallmorey, line 7: Can this be reconciled with the `createServer` import from mirage above?

randallmorey (changes requested):

addons/api/mirage/config.js
calcaide: Goot catch!! my bad and also I realize the linter should have tell me that 🤔 I will review the linter config, perhaps we missing something.

cameronperera (approved):

PR #1507, opened 2022-11-30
Icu 6657 auth method model fragments refactor

ZedLi:
:tickets: [Jira ticket](https://hashicorp.atlassian.net/browse/ICU-6657)

## Description
This PR removes the last usage of fragments from auth methods and gets us off data model fragments.
We use the same trick as with host-sets and recreate the arrays and their contents to mark them as dirty in the model when editing.

:technologist: [Admin preview](https://boundary-ui-git-icu-6657-auth-method-model-fra-6f0d03-hashicorp.vercel.app/scopes/s_id6ozx3wue/auth-methods/am_wtbbeizy0q)


addons/api/package.json
```diff
@@ -31,8 +31,8 @@
     "ember-cli-babel": "^7.26.10",
     "ember-cli-htmlbars": "^6.1.0",
     "ember-cli-mirage": "^2.2.0",
+    "ember-copy": "^2.0.1",
```
calcaide, line 34: Asking without the context of using `ember-copy` within this PR: why we moved from devDeps to dep?

addons/api/mirage/config.js
```diff
@@ -81,7 +81,17 @@ export default function () {
     return authMethods.create(attrs);
   });
   this.get('/auth-methods/:id');
-  this.patch('/auth-methods/:id');
+  this.patch(
```
calcaide, line 84: Is this something we discovered during this work, or part of this refactor?

addons/api/package.json
ZedLi: see this thread: https://hashicorp.slack.com/archives/C04BFSLA2L9/p1669830260871929

addons/api/mirage/config.js
ZedLi: It's the same issue as with the `POST` in #1491. I noticed when updating the auth method we would get a translation error because there's no `state`

ui/admin/app/routes/scopes/scope/auth-methods/auth-method.js
```diff
@@ -22,6 +23,35 @@ export default class ScopesScopeAuthMethodsAuthMethodRoute extends Route {
...
+   * into a dirty state.  This ensures that `model.rollbackAttributes()` reverts
+   * to the original expected array.
+   *
+   * The deep copy implemented here is required to ensure that both the
+   * array itself and its members are all new.
+   *
+   * @param {authMethodModel} authMethod
+   */
+  @action
+  edit(authMethod) {
```
calcaide, line 37: I think we could simplify this part and also delete the need for `ember-copy`.

We can use [structuredClone()](https://developer.mozilla.org/en-US/docs/Web/API/structuredClone) for this purpose.

I tested this and to me ir worked but maybe I am missing context on `ember-data` dirty state or something.

```javascript
edit(authMethod) {
  // Deep copy authMethod model
  authMethod = structuredClone(authMethod);
}
```

ZedLi: That's really cool, I wasn't aware of that API. I would totally be down for using that and replacing `ember-copy`. It would be outside the scope of changing it in this PR but I think it's worth discussing; I'm not aware of any advantage of using `ember-copy`.

calcaide: Yeah, after last change isn't working as expected and I completly agree, this is out of the scope for this PR.

I will create a ticket on tech debt to explore deprecating `ember-copy` in favour of `structuredClone` or similar API's. I will also bring this topic to FE tech time to mention it 😃

addons/api/package.json
calcaide: Aha! that makes sense 👍

addons/api/mirage/config.js
calcaide: Ahh totally right!! Good job!

calcaide (approved):
LGTM!! Awesome work!!! Can't wait to merge this and... clean model-fragments!!!

DhariniJeeva (approved):

ui/admin/app/routes/scopes/scope/auth-methods.js
```diff
@@ -179,21 +179,20 @@ export default class ScopesScopeAuthMethodsRoute extends Route {
    */
   @action
   async addStringItem(authMethod, property, value) {
-    const array = authMethod.get(property);
-    array.addObject({ value });
+    const array = [...(authMethod.get(property) ?? []), { value }];
```
randallmorey, line 182: For readability, could this be an improvement...
```suggestion
    const existing = ...authMethod?.[property]  || [];
    const array = [...existing, { value }];
```

randallmorey (approved):

ui/admin/app/routes/scopes/scope/auth-methods.js
ZedLi: Yup I agree, I'll update that in a few places!

PR #1511, opened 2022-12-05
Icu 6896 upgrade ember data

ZedLi:
:tickets: [Jira ticket](https://hashicorp.atlassian.net/browse/ICU-6896)

## Description
This upgrades ember data to the latest `3.x` version. There was an error occurring after upgrading to `3.28.0` and any subsequent version (including the `4.x` versions).

It appears that ember-data had changed some of their internal state management and introduced a bug. Setting the `stateModel` (which is an object) as part of the `createRecord` call causes ember to error out (possibly because of their own ember run loop? seems like it assumes the model is ready before it's actually set, couldn't figure this out).

However setting the `stateModel` after creating the record works fine. As a workaround, that is what I do in this PR.

I also had to cherry-pick a fix from our ember 4 api upgrade branch.

### DIFF
--- a/addons/api/addon/models/base.js
+++ b/addons/api/addon/models/base.js
@@ -92,7 +92,6 @@ export default class BaseModel extends Model {
   /**
    * @type {boolean}
    */
-  @computed('hasDirtyAttributes', 'isSaving')
   get canSave() {
     return this.hasDirtyAttributes && !this.isSaving;
   }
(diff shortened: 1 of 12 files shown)


calcaide (approved):
LGTM!! Thanks for the fix!!!

cameronperera (approved):

PR #1515, opened 2022-12-07
feat: 🎸 Give more instructions for how to connect to tunnel

ZedLi:
✅ Closes: ICU-6527

:tickets: [Jira ticket](https://hashicorp.atlassian.net/browse/ICU-6527)

## Description

Updated the desktop client connect to give more instructions on what to do with the localhost tunnel.

light mode:
image

dark mode:
image

### DIFF
--- a/ui/desktop/app/templates/scopes/scope/projects/targets.hbs
+++ b/ui/desktop/app/templates/scopes/scope/projects/targets.hbs
@@ -22,7 +22,7 @@
               {{t 'resources.target.messages.proxy_url.title'}}
               ({{t (concat 'resources.target.types.' target.type)}})
             </h3>
-            {{#if (eq target.type 'ssh')}}
+            {{#if target.isSSH}}
               <Rose::CodeEditor
                 @codeValue={{concat
                   'ssh '
--- a/ui/desktop/app/templates/scopes/scope/projects/targets.hbs
+++ b/ui/desktop/app/templates/scopes/scope/projects/targets.hbs
@@ -22,32 +22,45 @@
               {{t 'resources.target.messages.proxy_url.title'}}
               ({{t (concat 'resources.target.types.' target.type)}})
             </h3>
-            <Rose::CodeEditor
-              @codeValue={{concat
-                'ssh '
-                session.proxy_address
-                ' -p '
-                session.proxy_port
-              }}
-              as |c|
-            >
-              <c.toolbar>
-                <:action>
-                  <span>{{t
-                      'resources.target.messages.connection-instructions.title'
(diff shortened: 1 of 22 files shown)


dvonanderson (approved):

ui/desktop/app/templates/scopes/scope/projects/targets.hbs
```diff
@@ -22,15 +22,44 @@
               {{t 'resources.target.messages.proxy_url.title'}}
               ({{t (concat 'resources.target.types.' target.type)}})
             </h3>
-            <p>
-              <Copyable
-                @text={{session.proxy}}
-                @buttonText={{t 'actions.copy-to-clipboard'}}
-                @acknowledgeText={{t 'states.copied'}}
+            {{#if (eq target.type 'ssh')}}
```
randallmorey, line 25:
```suggestion
            {{#if target.isSSH}}
```

randallmorey (changes requested):

calcaide (approved):

randallmorey (approved):

cameronperera (approved):

ZedLi:
There's currently a bug where the copy function doesn't work in desktop, will investigate later why it doesn't work

randallmorey (changes requested):
> There's currently a bug where the copy function doesn't work in desktop, will investigate later why it doesn't work

Rose code editor has a reference to a `CopyButton` component.  This component does not appear to be extant.  In any case, the code editor may be too peculiar for Rose itself.  I wonder if code editor can be moved into admin and desktop, and changed to use `Copyable`.

Currently, admin and desktop have their own copies of a `Copyable` component, which is not available in Rose because it does not originate from Structure.  It may make sense to promote Copyable to addons/core, provided sufficient tests can be added.  However, Rose Code Editor is relatively new and differs enough from existing patterns that I recommend it remain within the UIs for now.

https://github.com/hashicorp/boundary-ui/blame/main/addons/rose/addon/components/rose/code-editor/toolbar/index.hbs#L19

ZedLi:
> This component does not appear to be extant

What do you mean by it's not extant? It comes from this [package](https://github.com/jkusa/ember-cli-clipboard). I had [looked at it before](https://github.com/hashicorp/boundary-ui/pull/1207#discussion_r932768848) and it was similar to our `Copyable` and both are using ClipboardJS under the hood, but this one had the benefit of having test coverage so it seemed like a good thing to replace what we were already using.

Is this something we want to discuss more in tech time?

randallmorey (approved):

PR #1516, opened 2022-12-07
Refactor upgrade ember 4 addon core

calcaide:
:tickets: [Jira ticket](https://hashicorp.atlassian.net/browse/ICU-6813)

## Description

Upgrade to Ember 4.4 within `addon/core`

### DIFF
--- a/addons/core/.ember-cli
+++ b/addons/core/.ember-cli
@@ -5,5 +5,11 @@
 
     Setting `disableAnalytics` to true will prevent any data from being sent.
   */
-  "disableAnalytics": false
+  "disableAnalytics": false,
+
+  /**
+  Setting `isTypeScriptProject` to true will force the blueprint generators to generate TypeScript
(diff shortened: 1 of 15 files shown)


ZedLi (approved):
LGTM, one question I have is do we need that `ember-try` config we had in our other upgrades?

calcaide:
> 
@ZedLi 
Initially we had that same discussion with @randallmorey . Since the blueprints add it and they are addons we will leave it for now. 

We can discuss further within FE tech time. I do personally will like to not have it, but seems to be quite a "standard" thing within Ember.

calcaide:
> > Since the blueprints add it and they are addons we will leave it for now.
> 
> Sorry I meant I didn't see it in this PR but it was present in the other ones, should we add it in?

@ZedLi Yeah, in all Ember 4 related PR's we should add it.

DhariniJeeva (approved):

cameronperera (approved):

PR #1517, opened 2022-12-09
Refactor upgrade ember 4 addon rose fix linting issues

calcaide:
## Description

- Fix linter issues.
- Disable some linter rules. These components require to be refactored to glimmer component to be compliance with linter. We will come back to this later, now the work is focus on Ember 4.

**Be aware:**
- This branch is merging with `refactor-upgrade-ember-4-addon-rose`.
- This branch contains a 3 failing tests we will take care of those within `refactor-upgrade-ember-4-addon-rose`.

### DIFF
--- a/addons/rose/addon/components/rose/table/row/index.js
+++ b/addons/rose/addon/components/rose/table/row/index.js
@@ -1,3 +1,4 @@
+/* eslint-disable ember/no-classic-components, ember/no-classic-classes  */
 import Component from '@ember/component';
 import layout from './index';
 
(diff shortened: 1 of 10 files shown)


addons/rose/addon/components/rose/table/row/index.js
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable ember/no-classic-components, ember/no-classic-classes  */
```
ZedLi, line 1: Should we have some of these be handled in `.template-lintrc.js`? That way we could turn off some of these rules for multiple files in one place and they're centralized so we can easily revisit if the rules can be turned on again in the future.

Thoughts?

ZedLi (approved):
LGTM, just left one non blocking comment

PR #1518, opened 2022-12-09
fix: 🐛 infofield on host catalog plugin forms

DhariniJeeva:
the type and the helper text were flipped in the host catalog plugin forms

[azure](https://boundary-f4wwcezmk-hashicorp.vercel.app/scopes/s_1sheljmjgm/host-catalogs/hc_r9jfw7e4kg)

[aws](https://boundary-f4wwcezmk-hashicorp.vercel.app/scopes/s_1sheljmjgm/host-catalogs/hc_d035pwe9uu)

### DIFF
--- a/ui/admin/app/components/form/host-catalog/aws/index.hbs
+++ b/ui/admin/app/components/form/host-catalog/aws/index.hbs
@@ -101,14 +101,9 @@
       </:body>
     </form.fieldset>
   {{else}}
-    <InfoField
-      @value={{t 'resources.host-catalog.types.aws'}}
-      @icon='aws-color'
-      disabled={{form.disabled}}
-      as |F|
-    >
+    <InfoField @value={{t 'descriptions.provider'}} @icon='aws-color' as |F|>
       <F.Label>{{t 'titles.provider'}}</F.Label>
-      <F.HelperText>{{t 'descriptions.provider'}}
+      <F.HelperText>{{t 'resources.host-catalog.types.aws'}}
       </F.HelperText>
     </InfoField>
   {{/if}}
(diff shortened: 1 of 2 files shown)


ui/admin/app/components/form/host-catalog/aws/index.hbs
```diff
@@ -101,14 +101,9 @@
       </:body>
     </form.fieldset>
   {{else}}
-    <InfoField
-      @value={{t 'resources.host-catalog.types.aws'}}
-      @icon='aws-color'
-      disabled={{form.disabled}}
-      as |F|
-    >
+    <InfoField @value={{t 'descriptions.provider'}} @icon='aws-color' as |F|>
```
ZedLi, line 104: I'm a bit confused, wasn't the `descriptions.provider` previously the helper text before we migrated to `InfoField`? Why is it now the value instead of `Azure` or `AWS`?

dvonanderson: @DhariniJeeva yes that is correct.  We use a `hds::form::textinput::field`  that has better accessibility that what we previously had it.

calcaide (approved):
LGTM! Nice work!

As a suggestion for coming similar PR's, would you mind adding screenshots on how it look before and after? that helps on the review 😃  thank you

ui/admin/app/components/form/host-catalog/aws/index.hbs
randallmorey: The component itself has the elements reversed, which caused the confusion.  That will be fixed in a separate PR.  The value should be the value.  The helper text should be the description/help/explanation.

randallmorey (changes requested):

ui/admin/app/components/form/host-catalog/aws/index.hbs
DhariniJeeva: I'll close this pr as this is fixed in https://github.com/hashicorp/boundary-ui/pull/1525

PR #1519, opened 2022-12-12
Icu 7083 add stylelint to admin

dvonanderson:
:tickets: [Jira ticket](https://hashicorp.atlassian.net/browse/ICU-7083)

## Description
Add `stylelint` to admin in favor of `sasslint`. Used some disable and rules to suspend the warnings for `keyframes` syntax for `-webkit` and kebab-case on a few selectors that come from other addons.


ui/admin/app/styles/app.scss
```diff
@@ -509,6 +510,7 @@
   height: var(--secret-editor-height);
   margin-bottom: sizing.rems(l);
 
+  /* stylelint-disable selector-class-pattern */
   .secret-editor-json {
     .CodeMirror {
```
randallmorey, line 515: Is there a "next line" way to disable the rule?  That way developers can still add to the outer selector and be subject to linting.

randallmorey (approved):

randallmorey (changes requested):
Should this also remove the sasslint rc file from admin?

ui/admin/app/styles/app.scss
ZedLi: Alternatively maybe it's better to actually just add a custom regex for [this rule](https://stylelint.io/user-guide/rules/selector-class-pattern/) since it might come up more often and just add `CodeMirror` as an exception since it's not our class. 

Stylelint provides their regex for kebab case and we could just add to it for `CodeMirror` as well

calcaide: Indeed, I like @ZedLi idea!

ui/admin/.stylelintrc.js
```diff
@@ -0,0 +1,31 @@
...
+
+    // -Webkit animation CSS for older browsers.  Need a decision to remove `-webkit` syntax in
+    // favor of normal CSS
+    'property-no-vendor-prefix': null,
+    'at-rule-no-vendor-prefix': null,
+
+    // The few overrides for CodeMirror and HDS CSS aren't worth creating full regex expressions to match those
+    // style characteristics.
+    'selector-class-pattern': [
+      '(^([a-z][a-z0-9]*)(-[a-z0-9]*)*(--[a-z0-9]+)*$)|(^CodeMirror)',
```
ZedLi, line 24: What's this `(--[a-z0-9]+)*` group for?

dvonanderson: @ZedLi for the `HDS` overrides that we do in `app.scss`. HDS uses a double hyphen in some of their css.

calcaide (approved):
LGTM!!

ui/admin/.stylelintrc.js
ZedLi: OK if that's the intention, I think this could be simplified to:

```suggestion
      '(^([a-z][a-z0-9]*)(-{1,2}[a-z0-9]+)*$)|(^CodeMirror)',
```

randallmorey: I recommend providing an explanation and some matching examples in the comments.  Regexes can be hard to parse (pardon the weak pun).

randallmorey (changes requested):
Where there is a regex, I recommend providing an explanation and some matching examples in the comments. Regexes can be hard to parse (pardon the weak pun).

ui/admin/.stylelintrc.js
dvonanderson: @ZedLi @randallmorey got it!

ui/admin/.stylelintrc.js
```diff
@@ -0,0 +1,33 @@
...
+    'no-descending-specificity': null,
+
+    // -Webkit animation CSS for older browsers.  Need a decision to remove `-webkit` syntax in
+    // favor of normal CSS
+    'property-no-vendor-prefix': null,
+    'at-rule-no-vendor-prefix': null,
+
+    // As a team we want to stick to using kebab-case as our selector class pattern, but there are a couple
+    // of exceptions in which we overrride CSS from addons like CodeMirror and HDS, in which they use
+    // pascale-case and a double-kebabe-case. This regex checks for all of those 3 patterns and won't fail
```
ZedLi, line 23:
```suggestion
    // pascal case and a double-kebab-case. This regex checks for all of those 3 patterns and won't fail
```

ZedLi (approved):
LGTM!

randallmorey (approved):

cameronperera (approved):

PR #1520, opened 2022-12-12
OSS->ENT merge trigger

mdeggies:
## Description

Adds a github actions workflow to perform oss->ent merges from boundary-ui to boundary-ui-enterprise. Merges will initially only target the `main` branches in both repos, but this can be changed later on. This uses https://github.com/hashicorp/upstream-assistant under the hood.

NOTE: The team will need to create a GitHub PAT from a team service account that has write access to both the boundary-ui and boundary-ui-enterprise repos. The token should be added here https://github.com/hashicorp/boundary-ui/settings/secrets/actions and named `ELEVATED_GITHUB_TOKEN`.


cameronperera (approved):

mdeggies:
Awesome, thanks for the review! Just a reminder that the actions secret `ELEVATED_GITHUB_TOKEN` has to be set before merging. I don't see that in https://github.com/hashicorp/boundary-ui/settings/secrets/actions yet, so I'll hold off. cc @randallmorey

calcaide (approved):
LGTM! Thank you for the workflow

randallmorey (approved):
Added `ELEVATED_GITHUB_TOKEN`.  cc @mdeggies

DhariniJeeva (approved):

mdeggies:
OK, this should be ready to go. When the next PR is merged to main, check the github actions tab to see the workflow run, and check if the PR was auto-opened in the ENT repo.

PR #1522, opened 2022-12-12
feat: 🎸 Update byow install boundary command for hcp edition

cameronperera:
:tickets: [Jira ticket](https://hashicorp.atlassian.net/browse/ICU-7144)

## Description

### DIFF
--- a/ui/admin/app/components/form/worker/create-worker-led/index.js
+++ b/ui/admin/app/components/form/worker/create-worker-led/index.js
@@ -108,6 +108,12 @@ ${listenerText}
    * @type {string}
    */
   get installBoundaryText() {
+    if (this.features.isEnabled('byow-pki-hcp-cluster-id')) {
+      return `wget -q https://releases.hashicorp.com/boundary-worker/0.11.3+hcp/boundary-worker_0.11.3+hcp_linux_amd64.zip ;\\
+sudo apt-get update && sudo apt-get install unzip ;\\
+unzip *.zip ;\\
+boundary server -config="${this.configFilePath || '<path>'}/pki-worker.hcl"`;
+    }
     return `curl -fsSL https://apt.releases.hashicorp.com/gpg | sudo apt-key add - ;\\
 sudo apt-add-repository "deb [arch=amd64] https://apt.releases.hashicorp.com $(lsb_release -cs) main" ;\\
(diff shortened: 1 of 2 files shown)


ui/admin/app/components/form/worker/create-worker-led/index.js
```diff
@@ -108,6 +108,12 @@ ${listenerText}
    * @type {string}
    */
   get installBoundaryText() {
+    if (this.features.isEnabled('byow-pki-hcp-cluster-id')) {
+      return `wget -q https://releases.hashicorp.com/boundary-worker/0.11.3+hcp/boundary-worker_0.11.3+hcp_linux_amd64.zip ;\\
```
ZedLi, line 112: Are we hard coding the version for now? I saw there was some discussion of having a `latest` link or maybe even an APT release?

ui/admin/app/components/form/worker/create-worker-led/index.js
```diff
@@ -108,6 +108,12 @@ ${listenerText}
...
+sudo apt-get update && sudo apt-get install unzip ;\\
+unzip *.zip ;\\
```
ZedLi, line 114: Why do we unzip everything instead of just the file we need?

ui/admin/app/components/form/worker/create-worker-led/index.js
```diff
@@ -108,6 +108,12 @@ ${listenerText}
...
+boundary server -config="${this.configFilePath || '<path>'}/pki-worker.hcl"`;
```
ZedLi, line 115: Does this work? Don't we need to install it to their PATH, or at the minimum call it from the directory?

ui/admin/app/components/form/worker/create-worker-led/index.js
cameronperera: Yes. This is a temporary solution.

ui/admin/app/components/form/worker/create-worker-led/index.js
randallmorey: This is more or less just a suggestion and not intended to be the one true way to run a worker.  In production, administrators will almost certainly modify this script, or invoke in an entirely different way.

randallmorey (approved):

PR #1526, opened 2022-12-13
test: 💍 Add delete tests for JSON Credentials

cameronperera:
✅ Closes: ICU-6806

:tickets: [Jira ticket](https://hashicorp.atlassian.net/browse/ICU-6806)


DhariniJeeva (approved):

ui/admin/tests/acceptance/credential-store/credentials/delete-test.js
```diff
@@ -71,6 +79,10 @@ module(
         return this.server.schema.credentials.where({ type: 'ssh_private_key' })
           .length;
       };
+      getJSONCredentialCount = () => {
```
DhariniJeeva, line 82: Should this also be lowercase?

cameronperera: I didn't lowercase it because JSON is an acronym and it isn't the beginning of the variable name. Happy to change it to `getJsonCredentialCount` if that is better.

PR #1528, opened 2022-12-13
test: 💍 update list credential tests

cameronperera:
✅ Closes: ICU-6807

:tickets: [Jira ticket](https://hashicorp.atlassian.net/browse/ICU-6807)


randallmorey (commented):
Ideally a variable always begins with lowercase unless it's a class.

dvonanderson (approved):
Awesome

cameronperera:
> Ideally a variable always begins with lowercase unless it's a class.

Good point. I have been writing "JSON" long enough that is the only way I write it. I will update these.

PR #1529, opened 2022-12-13
Chore tweak target type content

randallmorey:
:tickets: [ICU-7314](https://hashicorp.atlassian.net/browse/ICU-7314)

## Description

:technologist: [Admin preview](https://boundary-icju6qtgy-hashicorp.vercel.app/scopes/s_1sheljmjgm/targets)

:desktop_computer: [Desktop preview](https://boundary-ui-desktop-7ri54k78x-hashicorp.vercel.app/scopes/global/projects/targets)

- Makes "SSH" the default target type (when available).
- Updates "TCP" target name to "Generic TCP".
- Improves clarity and brevity of target type descriptions.

#### Admin

Screenshot 2022-12-16 at 15 23 05

Screenshot 2022-12-13 at 16 20 18
Screenshot 2022-12-13 at 16 20 24

#### Desktop

Screenshot 2022-12-13 at 16 25 46


ui/admin/app/abilities/credential.js
```diff
@@ -6,7 +6,8 @@ export default class OverrideCredentialAbility extends CredentialAbility {
   @service features;
 
   get canRead() {
-    return (this.features.isEnabled('json-credentials') || this.model.type !== 'json')
+    return this.features.isEnabled('json-credentials') ||
+      this.model.type !== 'json'
```
cameronperera, line 10: Not blocking but I wanted to give context here. This was one lined for better readability, along with the explicit `()`. Moving the `this.model.type ...` portion to the next line had caused confusion.

randallmorey: It might exceed the line length of 80?  This happened after running `yarn format`.

ui/admin/app/components/form/worker/create-worker-led/index.js
```diff
@@ -112,7 +112,9 @@ ${listenerText}
       return `wget -q https://releases.hashicorp.com/boundary-worker/0.11.2+hcp/boundary-worker_0.11.2+hcp_linux_amd64.zip ;\\
 sudo apt-get update && sudo apt-get install unzip ;\\
 unzip *.zip ;\\
-./boundary-worker server -config="${this.configFilePath || '<path>'}/pki-worker.hcl"`;
+./boundary-worker server -config="${
+        this.configFilePath || '<path>'
+      }/pki-worker.hcl"`;
```
randallmorey, line 117: At first I was concerned that this changed the meaning, but it broke the lines within the JS interpolation.  I think I can factor out the value so that this remains on one line.

cameronperera: Good point. I did go a head and confirm that it does not cause issues.
Screen Shot 2022-12-13 at 4 16 53 PM

ui/admin/app/abilities/credential.js
cameronperera: Yep it sure does exceed 80. Do you think the readability of this is crucial enough that it deserves a rule exception? I assume that would block the format from changing it as well.

randallmorey: Hmm Actually there may be a more concise rewrite!  Let me propose something.

ui/admin/tests/unit/abilities/credential-test.js
```diff
@@ -4,87 +4,74 @@ import { setupTest } from 'ember-qunit';
 module('Unit | Abilities | credential', function (hooks) {
   setupTest(hooks);
 
-  test('canRead', function (assert) {
+  let features;
+
+  hooks.beforeEach(function () {
+    // Enable json-credentials feature by default
+    features = this.owner.lookup('service:features');
+    features.enable('json-credentials');
```
cameronperera, line 12: Do we need to specify this here? `json-credentials` are on by default in the testing environment? Although I am all for being explicit so this is non-blocking.

AdamBouhmad (approved):

ui/admin/tests/unit/abilities/credential-test.js
randallmorey: Ideally each test is self-contained and explicitly manages its own feature flags.  My reasons for this:

- If we explicitly set feature flags per test file, we don't need to rely on a global default for tests.  Global defaults become brittle since any changes will almost certainly break existing tests.
- For readability, it's nice to see up front which flags the tests rely on.

Open to discussion.

DhariniJeeva:
Thoughts on adding icons to the badges on the list view? We display icons for AWS, Azure, and managed groups

ui/admin/app/components/form/target/details/index.js
```diff
@@ -2,16 +2,23 @@ import Component from '@glimmer/component';
...
 export default class FormTargetComponent extends Component {
   // =properties
   /**
    * maps resource type with icon
    * @type {object}
    */
-  get mapResourceTypeWithIcon() {
-    return types.reduce((obj, type, i) => ({ ...obj, [type]: icons[i] }), {});
+  get typeMetas() {
+    const reversedTypes = [...types].reverse();
```
DhariniJeeva, line 17: Any particular reason for reversing this? Can we do 
```
const icons = {
  tcp: 'network',
  ssh: 'terminal-screen',
};
```

ui/admin/tests/unit/abilities/credential-test.js
cameronperera: Discussed this with the team. All in favor of explicitly managing feature flags per test file.

ui/admin/app/components/form/target/details/index.js
randallmorey: Because TCP is listed first in the model types (which makes sense), but on display we want SSH first.  The order of keys in an object is not necessarily the same as insertion order or code order, so going forward an array is a better option.

This is temporary anyway and will eventually be replaced by a helper.

randallmorey:
> Thoughts on adding icons to the badges on the list view? We display icons for AWS, Azure, and managed groups

Great idea!  I'll post a separate PR, since this requires a bit of effort to make work.

ui/admin/app/components/form/target/details/index.js
DhariniJeeva: Ahh! that is correct regarding the order of keys in an object. 🙈

cameronperera (approved):

ZedLi (approved):

ui/admin/app/components/form/target/details/index.hbs
```diff
@@ -43,33 +43,34 @@
...
+              (concat 'resources.target.help.' type.type)
             }}</R.Description>
-          <R.Icon @name={{icon}} />
+          <R.Icon @name={{type.icon}} />
         </G.RadioCard>
-      {{/each-in}}
+      {{/each}}
     </Hds::Form::RadioCard::Group>
-
+    <br />
```
calcaide, line 64: Are we using `<br />` to get some space? if so, can we achieve that with CSS?

calcaide (approved):
LGTM!! thanks for the work, I left a none blocking comment.

ui/admin/app/components/form/target/details/index.hbs
DhariniJeeva: Its been added to the hds override CSS, and we can get rid of the <br/>

DhariniJeeva (approved):

xingluw (approved):
LGTM

PR #1531, opened 2022-12-14
test: 💍 add tests for updating JSON credentials

cameronperera:
✅ Closes: ICU-6805

:tickets: [Jira ticket](https://hashicorp.atlassian.net/browse/ICU-6805)


ui/admin/tests/acceptance/credential-store/credentials/update-test.js
```diff
@@ -270,6 +362,29 @@ module(
...
+      assert.notEqual(instances.jsonCredential.name, mockInput);
+      await visit(urls.jsonCredential);
+      await click('form [type="button"]', 'Activate edit mode');
+      await fillIn('[name="name"]', mockInput);
+      assert.strictEqual(currentURL(), urls.jsonCredential);
+      try {
+        await visit(urls.credentials);
+      } catch (e) {
+        assert.dom('.rose-dialog').isVisible();
+        await click('.rose-dialog-footer button:first-child', 'Click Discard');
```
ZedLi, line 379: This works and there's no issues here but I just wanted to bring up I try and not use `element:first-child` or `element:last-child` because it's not intuitive to what I normally am trying to do. In this case I would probably do  `.rose-dialog-footer :first-child` or `.rose-dialog-footer button:first-of-type`. 

The difference is `.rose-dialog-footer button:first-child` is saying I want the first element under `.rose-dialog-footer` _and_ if it is a button, select it. In this case if we added a div or anything else before the button this would not find anything. Removing the `button` qualifier (`.rose-dialog-footer :first-child`) would just mean get the first child under the footer, whatever it is.

If we're trying to get the _first button_ under the footer (whether there's some siblings among them), most of the time really we're asking for is `.rose-dialog-footer button:first-of-type`. I personally find that usually when using `:first-child`, having the element qualifier isn't what was meant the majority of the time unless we were making it brittle on purpose.

ZedLi (approved):
LGTM, just left a side comment

ui/admin/tests/acceptance/credential-store/credentials/update-test.js
cameronperera: Very good point. After reading through this, `first-of-type` is definitely what we want in this situation. Thanks for pointing this out! I will make the updates.

PR #1532, opened 2022-12-14
Chore/upgrade ember data to 4

ZedLi:
## Description
Upgrade `ember-data` to 4.4.1.

~~Still getting deprecation messages from mirage but should have been fixed by #1504, will look a bit more into it.~~
The issue was when running tests, mirage was also looking at the dummy config

### DIFF
--- a/addons/api/package.json
+++ b/addons/api/package.json
@@ -28,7 +28,7 @@
     "ember-cli-htmlbars": "^6.1.0",
     "ember-cli-mirage": "^2.4.0",
     "ember-copy": "^2.0.1",
-    "ember-data": "^3.28.12",
+    "ember-data": "~4.4.0",
     "ember-fetch": "^8.1.1"
   },
   "devDependencies": {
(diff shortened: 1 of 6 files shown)


randallmorey (approved):

addons/api/package.json
```diff
@@ -28,7 +28,7 @@
     "ember-cli-htmlbars": "^6.1.0",
     "ember-cli-mirage": "^2.4.0",
     "ember-copy": "^2.0.1",
-    "ember-data": "^3.28.12",
+    "ember-data": "~4.4.0",
```
ZedLi, line 31: Changed our semver usage to `~` to match `ember-source` and we stay in step with the minor versions

addons/api/tests/dummy/mirage/config.js
```diff
@@ -1,22 +1,5 @@
...
-    this.get('/posts');
-    this.post('/posts');
-    this.get('/posts/:id');
-    this.put('/posts/:id'); // or this.patch
-    this.del('/posts/:id');
-
-    https://www.ember-cli-mirage.com/docs/route-handlers/shorthands
-  */
+export default function (config) {
+  return createServer({ ...config });
```
ZedLi, line 4: This was to fix the deprecation message thrown by mirage when running tests

calcaide: Nice! yeah I think here we do not need to use `models` neither `serializers` KISS!

calcaide (approved):
Awesome!! Let'z gooo!!!

cameronperera (approved):

PR #1536, opened 2022-12-16
chore: 🤖 add role principal read ability

ghost:
Adds an ability to check read permission on a role principal.  This is a convenience since principals are polymorphic and would otherwise require a type lookup before the read ability check.

Usage examples:
- `can 'readPrincipal role' user`
- `can 'readPrincipal role' group`
- `can 'readPrincipal role' managedGroup`


addons/api/addon/abilities/role.js
```diff
@@ -26,4 +35,22 @@ export default class RoleAbility extends ModelAbility {
...
+   * Returns the ability check for a principal model instance.
+   * Valid role principal models are: User, Group, and ManagedGroup.
+   * Throws an error if the principal type is invalid.
+   * @type {boolean}
+   * @throws {InvalidRolePrincipalTypeError}
+   */
+  get canReadPrincipal() {
+    const type = this.model.constructor.modelName;
+    const typeIsValid =
+      type === 'user' || type === 'group' || type === 'managed-group';
```
DhariniJeeva, line 49:
```suggestion
      type === 'user' || type === 'group' || type === 'managed group';
```

addons/api/addon/abilities/role.js
```diff
@@ -26,4 +35,22 @@ export default class RoleAbility extends ModelAbility {
...
+    if (!typeIsValid)
+      throw new InvalidRolePrincipalTypeError(
+        `Expected a role principal of type 'user', 'group', or 'managed-group'.  Got '${type}'.`
```
DhariniJeeva, line 52:
```suggestion
        `Expected a role principal of type 'user', 'group', or 'managed group'.  Got '${type}'.`
```

addons/api/addon/abilities/role.js
randallmorey: The input for this is an actual Ember Data model instance, in which the model name is `managed-group`.

addons/api/addon/abilities/role.js
randallmorey: The input for this is an actual Ember Data model instance, in which the model name is `managed-group`.

.github/workflows/oss-ent-merge-trigger.yml
```diff
@@ -0,0 +1,25 @@
+name: Trigger OSS to Enterprise Merge
```
calcaide, line 1: Is this entire file part of this PR? To me seems out of context. Could we have a separate PR just for the GH workflow?

addons/api/addon/models/role.js
```diff
@@ -75,7 +85,7 @@ export default class RoleModel extends GeneratedRoleModel {
         .queryBy('user', { id: ids }, { scope_id: 'global', recursive: true })
         .then((models) => models.map((model) => model));
     }
-
+    //this.store.query('user', { scope_id: 'global', recursive: true });
```
calcaide, line 88: Is this left in purpose?

addons/api/addon/models/role.js
```diff
@@ -103,6 +113,43 @@ export default class RoleModel extends GeneratedRoleModel {
     return resolve([]);
   }
 
+  /**
+   * A promise that resolves to an array of Group instances.
+   * When calling this getter, be sure to await resolution
+   * before interacting with the results.
+   * @type {Promise[GroupModel]}
+   */
+  get managedGroups() {
```
calcaide, line 122: If we are aiming for an async function (based on the coment above) shouldn't we define this function as `async`?

PR #1538, opened 2022-12-19
refactor: 💡 use constants for target types

randallmorey:
This pr introduces constants to represent available target types, as opposed to strings.  This is an improvement over duplicated strings for several reasons:

- Ties all instances of a resource type to a single source of truth:  its constant.
- Eliminates duplication of identical string values.
- Strings are error prone, since many tests may silently accept strings which are not valid types.
- This is a step toward the formalization of resource types, resulting in greater convenience and reliability of resource typing in the future (beginning with targets).
- Simplifies future refactors of resource types.  Where previously a search for an ambiguous string was required `'tcp'`, an unambiguous search may now be used `TYPE_TARGET_TCP`.

For example, in cases where target type previously was represented as a string:

```js
{ type: 'tcp' }
```

the type may now be represented as a constant, which is reusable everywhere the type is needed:

```js
import { TYPE_TARGET_TCP } from 'api/models/target';

{ type: TYPE_TARGET_TCP }
```

This is slightly more code.  But it reduces the proliferation of duplicate strings throughout the codebase and ties every use of a type back to a single source of truth.

### DIFF
--- a/addons/api/addon/models/target.js
+++ b/addons/api/addon/models/target.js
@@ -4,7 +4,7 @@ import { inject as service } from '@ember/service';
 
 export const TYPE_TARGET_TCP = 'tcp';
 export const TYPE_TARGET_SSH = 'ssh';
-export const types = [TYPE_TARGET_TCP, TYPE_TARGET_SSH];
+export const TYPES_TARGET = Object.freeze([TYPE_TARGET_TCP, TYPE_TARGET_SSH]);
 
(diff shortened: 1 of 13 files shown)


calcaide (approved):
Nice refactor!! simplifies and improves Dev Experience 🥳

cameronperera (approved):
I love every bit of this. Thanks for doing this refactor and the explanation for it.

ZedLi (approved):

PR #1542, opened 2022-12-21
Add mirage support principal managed groups

DhariniJeeva:
This PR adds mocks for managed groups in `role/principals list` and `add to principals` view for both global and org-level scope. 

This is part of https://hashicorp.atlassian.net/browse/ICU-7102 

global scope: 
Screen Shot 2022-12-20 at 1 37 24 PM

org Scope: 
Screen Shot 2022-12-20 at 1 36 50 PM

### DIFF
--- a/addons/api/mirage/factories/role.js
+++ b/addons/api/mirage/factories/role.js
@@ -28,9 +28,13 @@ export default factory.extend({
   withPrincipals: trait({
     afterCreate(role, server) {
       const { scope } = role;
+      const { id: scopeId } = scope;
       const users = server.createList('user', 2, { scope });
       const groups = server.createList('group', 2, { scope });
-      role.update({ users, groups });
+      const managedGroups = server.schema.managedGroups
+        .where({ scopeId })
+        .models.map((model) => model);
+      role.update({ users, groups, managedGroups });
     },
   }),
 
--- a/addons/api/mirage/factories/role.js
+++ b/addons/api/mirage/factories/role.js
@@ -31,9 +31,9 @@ export default factory.extend({
       const { id: scopeId } = scope;
       const users = server.createList('user', 2, { scope });
       const groups = server.createList('group', 2, { scope });
-      const managedGroups = server.schema.managedGroups
-        .where({ scopeId })
-        .models.map((model) => model);
+      const managedGroups = server.schema.managedGroups.where({
+        scopeId,
+      }).models;
       role.update({ users, groups, managedGroups });
     },
   }),
--- a/addons/api/mirage/scenarios/default.js
+++ b/addons/api/mirage/scenarios/default.js
@@ -65,7 +65,7 @@ export default function (server) {
     )[0];
     const managedGroups = server.createList('managed-group', 2, {
       scope,
-      authMethodId: oidcAuthMethod[0].id,
+      authMethodId: oidcAuthMethod.id,
     });
     role.update({ managedGroups });
   });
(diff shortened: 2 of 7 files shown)


addons/api/mirage/factories/role.js
```diff
@@ -28,9 +28,13 @@ export default factory.extend({
   withPrincipals: trait({
     afterCreate(role, server) {
       const { scope } = role;
+      const { id: scopeId } = scope;
       const users = server.createList('user', 2, { scope });
       const groups = server.createList('group', 2, { scope });
-      role.update({ users, groups });
+      const managedGroups = server.schema.managedGroups
+        .where({ scopeId })
+        .models.map((model) => model);
```
randallmorey, line 36:
```suggestion
        .models;
```

addons/api/mirage/scenarios/default.js
```diff
@@ -51,8 +51,43 @@ export default function (server) {
...
+      (auth) => auth.type === 'oidc'
+    );
+    const managedGroups = server.createList('managed-group', 2, {
+      scope,
+      authMethodId: oidcAuthMethod[0].id,
+    });
+    role.update({ managedGroups });
+  });
+
+  const OrgScopeRoles = server.createList(
```
randallmorey, line 73:
```suggestion
  const orgScopeRoles = server.createList(
```

addons/api/mirage/scenarios/default.js
```diff
@@ -51,8 +51,43 @@ export default function (server) {
...
+    'role',
+    5,
+    { scope: orgScope },
+    'withPrincipals'
+  );
+  //create managed groups for role/principals in orgScope
+  OrgScopeRoles.forEach((role) => {
```
randallmorey, line 80:
```suggestion
  orgScopeRoles.forEach((role) => {
```

addons/api/mirage/scenarios/default.js
```diff
@@ -51,8 +51,43 @@ export default function (server) {
...
+    1,
+    { scope: globalScope },
+    'withPrincipals'
+  );
+  //create managed groups for role/principals in globalScope
+  globalScopeRoles.forEach((role) => {
+    const { scope } = role;
+    const oidcAuthMethod = globalAuthMethods.filter(
+      (auth) => auth.type === 'oidc'
+    );
```
randallmorey, line 65:
```suggestion
    )[0];
```

addons/api/mirage/scenarios/default.js
```diff
@@ -51,8 +51,43 @@ export default function (server) {
...
+    const managedGroups = server.createList('managed-group', 2, {
+      scope,
+      authMethodId: oidcAuthMethod[0].id,
```
randallmorey, line 68:
```suggestion
      authMethodId: oidcAuthMethod.id,
```

randallmorey (changes requested):
Made some minor suggestions.  One question I had is why model managed groups with a `scope`?  IIRC this doesn't appear in the API.

DhariniJeeva:
> Made some minor suggestions. One question I had is why model managed groups with a `scope`? IIRC this doesn't appear in the API.

ah, you are right. we don't need scope id for the managed group, updated!

DhariniJeeva:
> Made some minor suggestions. One question I had is why model managed groups with a `scope`? IIRC this doesn't appear in the API.

actually that's a valid point, I'm double checking if everything works without the scope Id since this was already there in the code

DhariniJeeva:
> > Made some minor suggestions. One question I had is why model managed groups with a `scope`? IIRC this doesn't appear in the API.
> 
> actually that's a valid point, I'm double checking if everything works without the scope Id since this was already there in the code

After an offline discussion, it is understood that we add scopes to managed groups.

ZedLi (approved):

PR #1544, opened 2022-12-21
Add model support for managed groups

DhariniJeeva:
This PR refactors the role routes, moves the querying to the model, and adds support for managed groups.

Screen Shot 2022-12-20 at 5 12 22 PM

### DIFF
--- a/addons/api/addon/models/role.js
+++ b/addons/api/addon/models/role.js
@@ -1,7 +1,7 @@
 import GeneratedRoleModel from '../generated/models/role';
 import { attr } from '@ember-data/model';
 import { inject as service } from '@ember/service';
-import { resolve } from 'rsvp';
+import { all, resolve } from 'rsvp';
 
 export default class RoleModel extends GeneratedRoleModel {
   // =services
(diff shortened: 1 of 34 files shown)


randallmorey (approved):

cameronperera (approved):

addons/api/addon/models/role.js
```diff
@@ -37,6 +39,124 @@ export default class RoleModel extends GeneratedRoleModel {
...
+    // return a promise which resolves model instances for those IDs
+    if (ids?.length) {
+      return this.resourceFilterStore
+        .queryBy('group', { id: ids }, { scope_id: 'global', recursive: true })
+        .then((models) => models.map((model) => model));
+    }
+
+    // No principal IDs,
+    // return a promise resolving to an empty array
+    return resolve([]);
```
ZedLi, line 113: When should we still be using `rsvp` vs just normal promises? The [docs](https://api.emberjs.com/ember/4.4/functions/rsvp/resolve) says it just aliases to promises but are we doing some thing with the ember runloop or could this just be a normal promise?

randallmorey: It looks like [Ember is indeed integrating the runloop with RSVP](https://github.com/emberjs/ember.js/blob/master/packages/%40ember/-internals/runtime/lib/ext/rsvp.ts).  I haven't experimented with mixing native promises with RSVP.  It should technically be possible as long as RSVP implements the interface.  To be safe and idiomatically Ember, we tend to stick with RSVP though.

ZedLi: It does seem like `RSVP` is a product of its times considering it was before ES6 and the widespread adoption of native promises, this [thread](https://discuss.emberjs.com/t/should-we-still-use-rsvp-instead-of-native-promises/19243) and some of the linked threads seem to indicate it should be fine to use native promises now, it just has some effects in testing as the `RSVP` promises automatically wait for the app to "settle".

DhariniJeeva: maybe this could be a great tech time topic, given that we use RSVP in many places

ZedLi (approved):

PR #1546, opened 2022-12-21
chore: 🤖 add target multi-hop flags to environment config

cameronperera:
✅ Closes: ICU-7299

:tickets: [Jira ticket](https://hashicorp.atlassian.net/browse/ICU-7299)

## Description

This adds the needed feature flags for the target multi-hop worker filter changes.
`target-dual-worker-filters` and `target-dual-worker-filters-hcp` are edition flags that will be set to true in their respective build editions once work is complete.


ui/admin/config/environment.js
```diff
@@ -13,6 +13,9 @@ const featureEditions = {
     'byow-pki-hcp-cluster-id': false,
     'byow-pki-upstream': true,
     'vault-worker-filter': false,
+    'target-multi-hop': false,
```
randallmorey, line 16: Can this be changed to align with the other two flag names?

randallmorey (changes requested):

ui/admin/config/environment.js
cameronperera: Happy to change it. Are you thinking `target-dual-worker`? I originally thought to keep the naming slightly different since that flag is temporary versus the other two flags that are used for building the app.

ZedLi (approved):

DhariniJeeva (approved):

ui/admin/config/environment.js
DhariniJeeva: Why did we change from multi to dual?

cameronperera: Giving this a little thought I think this is a good change because we could potentially have other "multi-hop" work that is not directly related to the target worker filter changes. Using `target-dual-worker` will help if that is the case and we can easily split the feature flags. I hope that makes sense.

dvonanderson (approved):
LGTM

ZedLi (approved):

PR #1547, opened 2022-12-21
chore: 🤖 Update electron deps

ZedLi:
## Description

Update some of our electron dependencies. This was mainly to fix this [dependabot alert](https://github.com/hashicorp/boundary-ui/security/dependabot/52) as the more recent versions remove a dependency that pins `minimatch` to an old version which lets us upgrade it without pinning it with `resolutions`.

Went through the [breaking changes](https://github.com/electron/forge/releases) and nothing should affect us. 

Built everything locally and ran it to confirm everything still functions correctly. Built on Circle CI and works correctly.
image


cameronperera (approved):
LGTM

calcaide (approved):
Amazing work @ZedLi this small changes have a huge impact :hugs: :partying_face:

PR #1553, opened 2023-01-04
chore: 🤖 Update dependencies to fix minimatch vulnerability

ZedLi:
This is to resolve [this dependabot](https://github.com/hashicorp/boundary-ui/security/dependabot/51) alert. 
Updated `electronegativity` which upgrades their mocha version which was using the old minimatch version.

Pinned the minimatch version scoped under a package so we limit what the packages the yarn `resolutions` affects and so we know what package we need to update in the future so it can be unpinned.

### DIFF
--- a/package.json
+++ b/package.json
@@ -36,7 +36,7 @@
     "test:addons-rose": "cd addons/rose && yarn test",
     "test:ui-admin": "cd ui/admin && yarn test",
     "test:ui-desktop": "cd ui/desktop && yarn test",
-    "compliance:licenses": "license-checker --onlyAllow 'Apache*;Apache License, Version 2.0;Apache-2.0;Apache 2.0;Artistic-2.0;BSD;BSD-3-Clause;CC-BY-3.0;CC-BY-4.0;CC0-1.0;ISC;MIT;MPL-2.0;Public Domain;Python-2.0;Unicode-TOU;Unlicense;WTFPL' --excludePackages 'boundary-ui;admin@0.0.0;desktop@0.0.0;base64-js@0.0.2;cycle@1.0.3;cldr-core@41.0.0'",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -64,7 +64,6 @@
     "highlight.js": "^10.4.1",
     "hoek": "~4.2.0",
     "elliptic": "~6.5.4",
-    "react-dev-utils": "~11.0.4",
```
ZedLi, line 67: This was un-needed as it already was correctly resolving to `11.0.4`in the lock file and it was interfering with resolutions further downstream for `minimatch`.

package.json
```diff
@@ -105,7 +104,9 @@
     "terser": "^5.14.2",
     "jpeg-js": "^0.4.4",
     "got": "^11.8.5",
-    "loader-utils": "^2.0.4"
+    "loader-utils": "^2.0.4",
+    "**/@storybook/**/minimatch": "^3.0.5",
```
ZedLi, line 108: I'm thinking we should start pinning resolutions that are for security concerns in this way rather than a global pin which makes it hard to resolve dependencies down the line and makes it unclear what were the upstream packages that were using it and why it needed to be pinned.

This helps limit the scope of these global resolution overrides, though the syntax is a bit confusing. Their docmentation doesn't explain it very well but the [RFC](https://github.com/yarnpkg/rfcs/blob/f1d11a23cc87c5e89d909521197e6a5028e05e05/implemented/0000-selective-versions-resolutions.md#package-designation) gives more info, our normal resolutions are actually aliases for `**/package` but if you want to further scope the packages the alias doesn't seem to work.

calcaide: I LOVE IT!! defining resolutions this way is veeeeery descriptive of why we are doing it, which will facilitate A LOT when we have to audit or mantain them, awesome!

calcaide (approved):
LGTM! Nice job keeping security up to date!

PR #1554, opened 2023-01-04
Add egress and ingress worker filters to target model

cameronperera:
:tickets: [Jira ticket](https://hashicorp.atlassian.net/browse/ICU-7300)

## Description
```
    instances.target = this.server.create('target', {
      scope: instances.scopes.project,
    });
    // add hostSets to target after target has been created
    // due to the target Id being generated after creation
    instances.target.update({ hostSets: instances.hostCatalog.hostSets });
```
I made the above change to 2 test groups. After updating target ID to generate by type, the id is not actually created until target resource is created. By adding the host sets to the target after it has been created, it stopped giving me the error: `Mirage: Error: You're instantiating a host-set that has a targetId of 1, but that record doesn't exist in the database.`

I am not entirely sure where the `targetId of 1` is being generated. If there is a better solution please let me know.

[ICU-7300]: https://hashicorp.atlassian.net/browse/ICU-7300?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


addons/api/mirage/factories/target.js
```diff
@@ -23,22 +23,28 @@ export default factory.extend({
...
-  id: () => generateId('t_'),
-
   /**
    * -1 means "unlimited" and we want to generate these on occasion.
    */
   session_connection_limit: () =>
     faker.helpers.arrayElement([-1, faker.datatype.number()]),
   worker_filter: (i) => (i % 2 === 0 ? faker.random.words() : null),
+  egress_worker_filter: (i) => (i % 2 === 0 ? faker.random.words() : null),
+  ingress_worker_filter: (i) => (i % 3 === 0 ? faker.random.words() : null),
```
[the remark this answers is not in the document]
cameronperera: Correct. I will update it so they do not exist together. Good catch.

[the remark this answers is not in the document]
cameronperera: Fixed.

addons/api/mirage/factories/target.js
```diff
@@ -23,22 +23,28 @@ export default factory.extend({
...
     faker.helpers.arrayElement([-1, faker.datatype.number()]),
   worker_filter: (i) => (i % 2 === 0 ? faker.random.words() : null),
+  egress_worker_filter: (i) => (i % 2 !== 0 ? faker.random.words() : null),
+  ingress_worker_filter: (i) => (i % 2 !== 0 ? faker.random.words() : null),
   type: (i) => types[i % types.length],
   /**
    * Generates attributes fields by type.
    */
   afterCreate(target) {
+    const id =
```
ZedLi, line 39: Is there a reason the id is being created in this `afterCreate` instead of as part of this `target` class as it was previously? 

Regarding the error you're seeing, I would guess it's because of this since it doesn't have an id until after being created and the `host-set` has an association with `targets` in the models and isn't getting the right id because it doesn't have one yet.

cameronperera: I moved it to the `afterCreate` to give a more accurate id based on the target type. However, I can see that not being a good trade off for the test setup changes I had to make to accommodate this change. Happy to change it back if you feel I should.

cameronperera: You are right. I am going to try what I am thinking and push it up shortly. Thanks for the direction!

ZedLi (approved):

addons/api/mirage/factories/target.js
```diff
@@ -22,25 +22,23 @@ export default factory.extend({
       'add-credential-sources',
       'remove-credential-sources',
     ],
-
-  id: () => generateId('t_'),
+  id() {
+    return this.type === TYPE_TARGET_TCP
+      ? generateId('ttcp_')
+      : generateId('tssh_');
```
randallmorey, line 28: Unless we're using the ID prefixes in some way, it's probably overkill to type the mock IDs.

addons/api/mirage/factories/target.js
```diff
@@ -22,25 +22,23 @@ export default factory.extend({
...
+  },
 
   /**
    * -1 means "unlimited" and we want to generate these on occasion.
    */
   session_connection_limit: () =>
     faker.helpers.arrayElement([-1, faker.datatype.number()]),
   worker_filter: (i) => (i % 2 === 0 ? faker.random.words() : null),
+  egress_worker_filter: (i) => (i % 2 !== 0 ? faker.random.words() : null),
+  ingress_worker_filter: (i) => (i % 2 !== 0 ? faker.random.words() : null),
```
randallmorey, line 38: In the API, `worker_filter` is mutually exclusive with the new fields.  If it is set, the other two _may not_ be set, and vice versa.  And this mutual exclusivity is one of the assumptions used to drive the new migration UI.  I recommend reflecting this constraint in the mock data.

addons/api/mirage/factories/target.js
```diff
@@ -22,25 +22,23 @@ export default factory.extend({
...
-   * Generates attributes fields by type.
-   */
-  afterCreate(target) {
-    target.update({
-      attributes: {
-        default_port: faker.datatype.number(),
-      },
-    });
+  default_port() {
+    return this.type === TYPE_TARGET_TCP ? 443 : 22;
```
randallmorey, line 41: TCP is the default type, but this reads as if, implicitly, SSH is the default.  While not critical right now, this could also be a `switch` or use a data structure to make it easier to add additional type->port mappings.
```suggestion
    return this.type === TYPE_TARGET_SSH ? 22 : 443;
```

addons/api/mirage/generated/factories/target.js
```diff
@@ -1,16 +1,12 @@
 import { Factory } from 'ember-cli-mirage';
 import { faker } from '@faker-js/faker';
-import { TYPE_TARGET_TCP } from 'api/models/target';
 
 /**
  * GeneratedTargetModelFactory
  */
 export default Factory.extend({
-  type: TYPE_TARGET_TCP,
```
randallmorey, line 9: The default type is TCP.

addons/api/tests/unit/serializers/target-test.js
```diff
@@ -143,6 +146,76 @@ module('Unit | Serializer | target', function (hooks) {
...
+    const serializedRecord = serializer.serialize(snapshot);
+    assert.deepEqual(serializedRecord, {
+      name: 'User',
+      description: 'Description',
+      version: 1,
+      type: TYPE_TARGET_SSH,
+      scope_id: 'org_1',
+      session_max_seconds: 28800,
+      session_connection_limit: null,
+      worker_filter: null,
```
randallmorey, line 182: If worker wasn't set to begin with, should it be serialized?  May want to follow up with BE.

randallmorey (changes requested):

addons/api/mirage/factories/target.js
cameronperera: Screen Shot 2023-01-09 at 10 22 16 AM
This is how the data is being generated with these mirage changes. This should satisfy the constraint between `worker_filter` and the new properties.

addons/api/mirage/factories/target.js
cameronperera: 👍🏼 I can revert this change.

addons/api/mirage/factories/target.js
cameronperera: Since I am rolling back the `id` changes based on type, would you like me to roll this back as well? We aren't explicitly using it anywhere.

addons/api/mirage/generated/factories/target.js
cameronperera: This can actually be removed since we are setting the type dynamically in the target factory. I will remove this line.

addons/api/tests/unit/serializers/target-test.js
cameronperera: Followed up with the BE team. `worker_filter` being present and set to null should not be an issue.

addons/api/mirage/factories/target.js
cameronperera: Updated.

addons/api/mirage/factories/target.js
cameronperera: Rolled back.

dvonanderson (approved):

addons/api/tests/unit/serializers/target-test.js
```diff
@@ -145,6 +148,78 @@ module('Unit | Serializer | target', function (hooks) {
...
+      egress_worker_filter: 'egress worker',
+      ingress_worker_filter: null,
+      attributes: {
+        default_port: 1234,
+      },
+      address: null,
```
DhariniJeeva, line 191: do we have an address field?

cameronperera: Yes. @ZedLi just added that very recently for work that he is doing.

DhariniJeeva (approved):
LGTM! left a minor comment

PR #1556, opened 2023-01-06
Chore/parallelize tests

ZedLi:
## Description

Run some of our tests in parallel. I didn't include `core` or `auth` as they had such few tests that it didn't end up being faster in those addons. Also included running the tests/linting in all repos in parallel.

Running the tests in each repo in parallel brings the time to run tests down to ~8mins on our CI and running all the addon and ui tests in parallel brings it down to ~6 mins. For reference normally it takes around ~9 to 10 minutes.

image


addons/api/package.json
```diff
@@ -19,7 +19,7 @@
     "format": "yarn run format:js",
     "format:js": "prettier --write '{addon,app,config,stories,tests,mirage}/**/*.js' *.js",
     "start": "ember serve",
-    "test": "ember test",
+    "test": "ember exam --split=4 --parallel=1 --random",
```
ZedLi, line 22: To keep `--random` and have it actually use a random seed, it has to be the last flag otherwise it will interpret anything after as a seed. 

Interestingly though, `--parallel` also takes in an argument so it wants to be last as well so I have to set the argument to `--parallel` as `1`. The confusing part is changing the value for `--parallel` doesn't actually do anything without using `--load-balance` since it respects the `parallel` value from testem, so `--parallel=1` doesn't actually do anything and is the same as just `--parallel`. See [this explanation](https://github.com/ember-cli/ember-exam/issues/270#issuecomment-489134727).

I chose to split it by 4 as it seems like [github runners](https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners#supported-runners-and-hardware-resources) are on 2 core cpus? Not sure if it has 4 threads but I thought `4` was a reasonable number to start with to see how it runs in our CI. Let me know if you think it should be different, for context cloud ui uses 8.

cameronperera:
How did the tests take over a minute longer than previous runs? 🤯

ZedLi:
> How did the tests take over a minute longer than previous runs? 🤯

Probably have to play around with it since CI is much more resource constrained than our local machines. It's considerably faster locally but I might need to try a lower split number.

I also didn't think it would save much time since it's always seemed the build portion of testing takes up the lion's share of the time in CI, I imagine most of the time would be saved by running the repos concurrently instead of the tests.

I'll be playing around with some of these settings to test on the CI though.

addons/api/package.json
```diff
@@ -19,7 +19,7 @@
     "format": "yarn run format:js",
     "format:js": "prettier --write '{addon,app,config,stories,tests,mirage}/**/*.js' *.js",
     "start": "ember serve",
-    "test": "ember test",
+    "test": "ember exam --test-port 0 --split=4 --parallel=1 --random",
```
randallmorey, line 22: @ZedLi when parallel is set to 1, is this equivalent to not being run in parallel?  Should it be 2 or more?  Perhaps I'm misunderstanding?

randallmorey (changes requested):

addons/api/package.json
ZedLi: Take a look at this [comment](https://github.com/hashicorp/boundary-ui/pull/1556#discussion_r1063669761), it unfortunately is confusing but no this parallel setting has no affect on the actual number that is running.

randallmorey: Is there a reason to leave it when it has no effect?

DhariniJeeva (approved):
LGTM

addons/auth/package.json
```diff
@@ -15,7 +15,7 @@
     "format": "yarn run format:js",
     "format:js": "prettier --write '{addon,app,config,stories,tests}/**/*.js' *.js",
     "start": "ember serve",
-    "test": "ember test",
+    "test": "ember test --test-port 0",
```
calcaide, line 18: NOT blocking comment: As I saw in other package.json, defining `--test-port` to 0 means the framework will choose a port for each suite? just wondering, because if we assign same port to let's say auth and core, if they run at the same time it will fail?

ZedLi: Yes, it's so we don't run them all on the same port at the same time which will conflict.

package.json
```diff
@@ -14,28 +14,28 @@
   },
   "scripts": {
     "build": "yarn build:ui:admin",
-    "build:ui:admin": "cd ui/admin && yarn build",
-    "build:ui:admin:oss": "cd ui/admin && yarn build:oss",
-    "build:ui:admin:enterprise": "cd ui/admin && yarn build:enterprise",
-    "build:ui:admin:hcp": "cd ui/admin && yarn build:hcp",
-    "build:ui:desktop": "cd ui/desktop && yarn build",
+    "build:ui:admin": "yarn --cwd ui/admin build",
```
calcaide, line 17: That is C-L-E-A-N did not know about `--cwd`

calcaide (approved):
AMAZING work!!! I am sure the gains are more than we think. Definetly we need to persist some metrics from our CI on how long it takes (avg) to run tests, so when good stuff like this comes we can really know how good it is 😃  Thank you for the work

PR #1559, opened 2023-01-09
Target form UI: show deprecation message for worker_filter property when present

cameronperera:
:tickets: [Jira ticket](https://hashicorp.atlassian.net/browse/ICU-7381)

## Description
When a user views and/or updates a target that has a worker filter, a deprecation message will be shown.
Target form was updated to use HDS form components as well.

:technologist: [Admin preview](https://boundary-7sonfcfs0-hashicorp.vercel.app/scopes/s_1sheljmjgm/targets/t_nfp0up1fqj)


ui/admin/app/components/form/target/details/index.hbs
```diff
@@ -5,41 +5,47 @@
   @showEditToggle={{if @model.isNew false true}}
   as |form|
 >
-  <form.input
-    @name='name'
-    @type='text'
+  <Hds::Form::TextInput::Field
+    @isRequired={{true}}
+    @isOptional={{false}}
```
ZedLi, line 10: Are we marking all non-required fields as `optional` as well from here on out?

ui/admin/app/components/form/target/details/index.js
```diff
@@ -29,4 +30,13 @@ export default class FormTargetComponent extends Component {
   get icon() {
     return icons[this.args.model.type];
   }
+
+  /**
+   * determines when the worker filter deprecation message should be shown
+   * @type {boolean}
+   */
+  @computed('args.model.{isNew,worker_filter}')
```
ZedLi, line 38: Is this `computed` needed?

ui/admin/app/components/form/target/details/index.hbs
```diff
@@ -75,86 +81,141 @@
...
+      </F.Error>
     {{/if}}
-  </form.input>
+  </Hds::Form::TextInput::Field>
 
-  <form.input
-    @name='default_port'
-    @type='number'
+  <Hds::Form::TextInput::Field
+    @isRequired={{false}}
```
ZedLi, line 129: Isn't the default port required?

ui/admin/app/components/form/target/details/index.hbs
```diff
@@ -75,86 +81,141 @@
...
+        >
+          <A.Title>{{t
+              'resources.target.workers.deprecation_message.title'
+            }}</A.Title>
+          <A.Description>{{t
+              'resources.target.workers.deprecation_message.description'
+            }}</A.Description>
+        </Hds::Alert>
+      {{/if}}
+      <Hds::Form::TextInput::Field
```
ZedLi, line 173: Should this field be shown at all if we're not showing the deprecation message (i.e. the field shouldn't show if there is no worker filter)? Wouldn't we then normally be showing the ingress/egress filters so this worker filter should be part of the `showDeprecationMessage` if check?

ZedLi (changes requested):
Not sure what anyone else thinks but isn't the size of the message/font just too big compared to the rest of the content? It feels overwhelming to me

image

ui/admin/app/components/form/target/details/index.hbs
cameronperera: I thought that was part of the HDS migration. We would start marking all fields with required or optional. @randallmorey and @DhariniJeeva would you provide some input on this topic please?

ui/admin/app/components/form/target/details/index.js
cameronperera: Nope. I will remember that one day. Removing.

ui/admin/app/components/form/target/details/index.hbs
cameronperera: Funny enough, yes and no. For SSH Targets it is not required but it is for TCP. I marked it as required again.

ui/admin/app/components/form/target/details/index.hbs
cameronperera: Yes absolutely. I left it for now as a kind of place holder until I add the `egress` filter in the next story. I felt leaving a disabled input field looked better than just a "Workers" section with a description with no field at all. I am happy to move it within the check now but I will be updating it soon.

cameronperera:
> Not sure what anyone else thinks but isn't the size of the message/font just too big compared to the rest of the content? It feels overwhelming to me
> 
> image

I agree that it is large. I believe it is supposed to be pretty "in your face" to encourage people to migrate sooner. I will talk with design about it.

cameronperera:
Closing this PR to split the HDS migration and filter work.

PR #1551, opened 2022-12-23
refactor: 💡 desktop display of complex nested secrets

dvonanderson:
✅ Closes: ICU-7409

:tickets: [Jira ticket](https://hashicorp.atlassian.net/browse/ICU-7409)

## Description
A quick solution for displaying JSON secret objects in connection modal.  

### Screenshots (if appropriate):
Screen Shot 2022-12-23 at 12 10 02

PR #1561, opened 2023-01-10
chore: 🤖 add json credential to desktop mirage scenario

cameronperera:
Now when connecting to a target a JSON credential will be included as one of the credential sources via mirage.

It will be the second credential source in the list. Keep in mind that nested secrets still are not rendered correctly here. That is fixed in this [PR](https://github.com/hashicorp/boundary-ui/pull/1551).

## Description

:desktop_computer: [Desktop preview](https://boundary-ui-desktop-git-add-json-credential-to-62bfa9-hashicorp.vercel.app/scopes/global/projects/targets)

### DIFF
--- a/addons/api/mirage/scenarios/ipc.js
+++ b/addons/api/mirage/scenarios/ipc.js
@@ -79,6 +79,37 @@ export default function initializeMockIPC(server, config) {
           },
           {
             credential_source: {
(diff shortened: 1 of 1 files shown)


dvonanderson (approved):
LGTM

ZedLi (approved):

PR #1562, opened 2023-01-10
refactor: 💡 Migrate Target form to use HDS components

cameronperera:
:tickets: [Jira ticket](https://hashicorp.atlassian.net/browse/ICU-7512)

## Description

Migrate Target form to use HDS components

Added this [addon](https://github.com/mainmatter/ember-test-selectors) for testing purposes. This allows us to use `data-test-*` attributes for testing and will remove these attributes in production environments.

:technologist: [Admin preview](https://boundary-ui-git-icu-7512-target-form-hds-migration-hashicorp.vercel.app/scopes/s_1sheljmjgm/targets/new)

Screen Shot 2023-01-10 at 11 56 07 AM

### DIFF
--- a/addons/rose/package.json
+++ b/addons/rose/package.json
@@ -45,6 +45,7 @@
     "ember-focus-trap": "^1.0.1",
     "ember-named-blocks-polyfill": "^0.2.5",
     "ember-radio-button": "github:yapplabs/ember-radio-button.git#b2ab65f",
+    "ember-test-selectors": "^6.0.0",
     "ember-uuid": "^2.1.0",
     "jsonlint": "^1.6.3",
     "sass": "^1.55.0"
--- a/ui/admin/app/components/form/target/details/index.hbs
+++ b/ui/admin/app/components/form/target/details/index.hbs
@@ -148,8 +148,8 @@
   </Hds::Form::TextInput::Field>
 
   <Hds::Form::TextInput::Field
-    @isRequired={{true}}
-    @isOptional={{false}}
+    @isRequired={{@model.isTCP}}
+    @isOptional={{not @model.isTCP}}
     @value={{@model.default_port}}
     @isInvalid={{@model.errors.default_port}}
     @type='text'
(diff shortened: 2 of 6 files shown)


randallmorey (changes requested):
Is name required?  Are we sure default port is required?

cameronperera:
> Name is optional too. Are we sure default port is required?

According to the [protobuf](https://github.com/hashicorp/boundary/blob/main/internal/proto/controller/api/resources/targets/v1/target.proto#L223-L248) for Targets, SSH does not require the default_port but TCP does. I have been able to confirm this against the binary. 

Name is also required for Targets according to the [protobuf](https://github.com/hashicorp/boundary/blob/main/internal/proto/controller/api/resources/targets/v1/target.proto#L80-L87) and testing with the binary. Below is the error message received with leaving name empty.
```
{
   "kind":"InvalidArgument",
   "message":"Error in provided request.",
   "details":{
      "request_fields":[
         {
            "name":"name",
            "description":"This field is required."
         }
      ]
   }
}
```

ui/admin/app/components/form/target/details/index.hbs
```diff
@@ -75,86 +81,95 @@
...
+      </F.Error>
     {{/if}}
-  </form.input>
+  </Hds::Form::TextInput::Field>
 
-  <form.input
-    @name='default_port'
-    @type='number'
+  <Hds::Form::TextInput::Field
+    @isRequired={{true}}
```
ZedLi, line 151: If default port is only required for `tcp` can we check what type of target we have to make it required only for `tcp`?

ZedLi (approved):
Just left a small comment

ui/admin/app/components/form/target/details/index.hbs
cameronperera: Good idea. I can get that pushed up shortly.

DhariniJeeva:
Nice work! LGTM

DhariniJeeva (approved):

addons/rose/package.json
```diff
@@ -45,6 +45,7 @@
     "ember-focus-trap": "^1.0.1",
     "ember-named-blocks-polyfill": "^0.2.5",
     "ember-radio-button": "github:yapplabs/ember-radio-button.git#b2ab65f",
+    "ember-test-selectors": "^6.0.0",
```
calcaide, line 48: I do not see usage of `ember-test-selectors` within the PR, Why we are adding it as dependency?

PR #1563, opened 2023-01-10
refactor: 💡 Deprecate htmlbars-inline-precompile within addons

calcaide:
:tickets: [Jira ticket](https://hashicorp.atlassian.net/browse/ICU-7232)

## Description

Deprecate `htmlbars-inline-precompile` within addons.

**Why this change?** 

- `ember-cli` documentation recommend this change starting at `ember-cli` `4.0.0` and we are using `4.4.0`.
- We had some test using `import hbs from 'htmlbars-inline-precompile';` and others usign `import { hbs } from 'ember-cli-htmlbars';` so we want to consolidate for consistency. More info about the deprecation [here](https://github.com/ember-cli/ember-cli-htmlbars#tagged-template-usage--migrating-from-htmlbars-inline-precompile)


ZedLi (approved):
Thanks for cleaning this up!

cameronperera (approved):

PR #1575, opened 2023-01-11
Icu 7490 UI add new address field to target

ZedLi:
:tickets: [Jira ticket](https://hashicorp.atlassian.net/browse/ICU-7490)

## Description
Added an address field to support targets having a static address instead of a host. See the [figma](https://www.figma.com/file/gVJWQN2IWbsCBrS3ggknox/Assign-address-to-targets?node-id=1844%3A5779&t=i3ruNtWHaY4qWJyD-0).

The toggle is `Use Host Sources` instead of `Use Static Addresses`, but either one could be used and the default when creating could be checked or not either way, let me know which wording you prefer. 

There also has been a suggestion to move the address and port fields up, let me know if that would be better.

### Toggle Off:
image

### Toggle On:
image

### Error States:
image

image

### DIFF
--- a/ui/admin/app/routes/scopes/scope/targets.js
+++ b/ui/admin/app/routes/scopes/scope/targets.js
@@ -51,11 +51,11 @@ export default class ScopesScopeTargetsRoute extends Route {
   /**
    * Handle save.
    * @param {TargetModel} target
-   * @param {Event} e
+   * @param {object} options
    */
   @action
   @loading
-  @notifyError(({ message }) => message)
+  @notifyError(({ message }) => message, { catch: true })
   @notifySuccess(({ isNew }) =>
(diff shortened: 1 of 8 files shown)


ui/admin/app/routes/scopes/scope/targets.js
```diff
@@ -51,16 +51,21 @@ export default class ScopesScopeTargetsRoute extends Route {
   /**
    * Handle save.
    * @param {TargetModel} target
-   * @param {Event} e
+   * @param {object} options
    */
   @action
   @loading
-  @notifyError(({ message }) => message)
+  @notifyError(({ message }) => message, { catch: true })
```
ZedLi, line 58: When should we be squelching errors? After I curry this save function in the form, I started to get errors in tests from the adapter which I couldn't catch, is there a better way of solving this?

PR #1576, opened 2023-01-11
fix: 🐛 Remove json_object from payload for non-json credentials

cameronperera:
:tickets: [Jira ticket](https://hashicorp.atlassian.net/browse/ICU-7550)

## Description

Remove json_object from payload for non-json type credentials

### DIFF
--- a/ui/admin/app/routes/scopes/scope/credential-stores/credential-store/credentials/new.js
+++ b/ui/admin/app/routes/scopes/scope/credential-stores/credential-store/credentials/new.js
@@ -28,12 +28,16 @@ export default class ScopesScopeCredentialStoresCredentialStoreCredentialsNewRou
     const { id: credential_store_id } = this.modelFor(
       'scopes.scope.credential-stores.credential-store'
     );
-    let name, description;
+    let name, description, json_object;
     /**
      * this sets the credential attribute json_object to a string that
      * replicates an empty json blob as CodeMirror expects a string
      */
-    const json_object = '{}';
+    if (type === 'json') {
+      json_object = '{}';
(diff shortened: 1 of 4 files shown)


ZedLi (approved):

DhariniJeeva (approved):

ui/admin/app/routes/scopes/scope/credential-stores/credential-store/credentials/new.js
```diff
@@ -28,12 +28,16 @@ export default class ScopesScopeCredentialStoresCredentialStoreCredentialsNewRou
...
     /**
      * this sets the credential attribute json_object to a string that
      * replicates an empty json blob as CodeMirror expects a string
      */
-    const json_object = '{}';
+    if (type === 'json') {
+      json_object = '{}';
+    } else {
+      json_object = null;
+    }
```
randallmorey, line 40: The `json_object` is still being passed even though it's null.  Ideally, the key would be omitted entirely unless it is a json cred.

randallmorey (changes requested):

ui/admin/app/routes/scopes/scope/credential-stores/credential-store/credentials/new.js
cameronperera: Would omitting the key actually do anything? `json_object` is still an attribute of Credential, therefore it would still know about the key but it would just be null or undefined.

cameronperera: Fixed.

PR #1579, opened 2023-01-13
Icu 7490 UI add new address field to target

ZedLi:
:tickets: [Jira ticket](https://hashicorp.atlassian.net/browse/ICU-7490)

## Description
Added the new address field to target. The user will be prompted to delete host sources if they exist when adding an address and vice versa.

image

image

image


ui/admin/app/routes/scopes/scope/targets.js
```diff
@@ -55,7 +55,7 @@ export default class ScopesScopeTargetsRoute extends Route {
    */
   @action
   @loading
-  @notifyError(({ message }) => message)
+  @notifyError(({ message }) => message, { catch: true })
```
ZedLi, line 58: I catch here so as to not display any errors twice as the form is now also calling this save and also using `notifyError`.

cameronperera (approved):
LGTM!

calcaide (approved):
LGTM! Awesome job!

While testing this I found what I think is a bug for an edge case, I will DM you steps I follow to reproduce and a screen capture of the process.

Not sure if is a bug of the new implementation or we are missing something in Mirage that is making it look like a bug.

ZedLi:
> While testing this I found what I think is a bug for an edge case, I will DM you steps I follow to reproduce and a screen capture of the process.

Great catch! The issue was after removing the hosts, the model gets updated from the response which blows away the address field so I had to set it again right before saving.

calcaide (approved):
AMAZING WORK!!! Tested again and I could not reproduce the previous bug 😄

PR #1580, opened 2023-01-13
refactor: 💡 Simplify template for confirmation dialog

ZedLi:
## Description
While working on the network address feature in targets, I noticed our `application` template has unnecessary duplication of the dialog after adding the ability to change the confirmation button wording.
The goal of this PR is to simplify the template.


calcaide (approved):
LGTM!! Awesome simplification 👏

PR #1584, opened 2023-01-16
Create Worker Diagram for egress and ingress filters

cameronperera:

## Description

Create Worker Diagram for egress and ingress filters

"Front line workers" is not final yet. Will change before merge or once confirmed in another PR.

**Note: Not being rendered on any screen yet.**

Screen Shot 2023-01-15 at 6 54 53 PM

### DIFF
--- a/ui/admin/app/components/worker-diagram/dual-filter/index.hbs
+++ b/ui/admin/app/components/worker-diagram/dual-filter/index.hbs
@@ -1,0 +1,67 @@
+<div data-test-dual-filters>
+  {{#if (and @egressFilter @ingressFilter)}}
+    <OrderedSeriesDiagram data-test-egress-ingress-on as |D|>
+      <D.Item @icon='user'>
+        {{t 'resources.target.workers.diagram.client'}}
+      </D.Item>
+      <D.Item @icon='settings'>
+        {{t 'resources.target.workers.diagram.ingress-worker'}}
+      </D.Item>
+      <D.Group
+        @title={{t 'resources.target.workers.diagram.network'}}
+        @color='highlight'
+      >
+        <D.Item @icon='settings'>
+          {{t 'resources.target.workers.diagram.egress-worker'}}
+        </D.Item>
+        <D.Item @icon='server'>
+          {{t 'resources.target.workers.diagram.host'}}
+        </D.Item>
+      </D.Group>
+    </OrderedSeriesDiagram>
+  {{else if @ingressFilter}}
+    <OrderedSeriesDiagram data-test-ingress-on as |D|>
+      <D.Item @icon='user'>
+        {{t 'resources.target.workers.diagram.client'}}
+      </D.Item>
+      <D.Item @icon='settings'>
+        {{t 'resources.target.workers.diagram.front-line-worker'}}
+      </D.Item>
+      <D.Item @icon='server'>
+        {{t 'resources.target.workers.diagram.host'}}
+      </D.Item>
+    </OrderedSeriesDiagram>
+  {{else if @egressFilter}}
+    <OrderedSeriesDiagram data-test-egress-on as |D|>
+      <D.Item @icon='user'>
+        {{t 'resources.target.workers.diagram.client'}}
+      </D.Item>
+      <D.Item @icon='settings'>
+        {{t 'resources.target.workers.diagram.front-line-worker'}}
+      </D.Item>
+      <D.Group
+        @title={{t 'resources.target.workers.diagram.network'}}
+        @color='highlight'
+      >
+        <D.Item @icon='settings'>
+          {{t 'resources.target.workers.diagram.egress-worker'}}
+        </D.Item>
+        <D.Item @icon='server'>
+          {{t 'resources.target.workers.diagram.host'}}
+        </D.Item>
+      </D.Group>
+    </OrderedSeriesDiagram>
+  {{else}}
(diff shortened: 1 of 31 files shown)


ui/admin/config/environment.js
```diff
@@ -148,6 +148,7 @@ module.exports = function (environment) {
     ENV.featureFlags['ssh-target'] = true;
     ENV.featureFlags['vault-worker-filter'] = true;
     ENV.featureFlags['target-worker-filters-v2'] = true;
+    ENV.featureFlags['target-worker-filters-v2-ingress'] = true;
```
calcaide, line 151: NOT BLOCKING Question: Don't we need the flag turned on for test as well? I see the CI passed the test, I also see the test are passing in local but this flag, as per my understanding is turned off (`false`) within the test environment.

calcaide (approved):
Terrific job Cameron!!! Thanks!!

ui/admin/config/environment.js
cameronperera: We discussed controlling feature flags as close to the test cases as possible. You can see examples of it in this [PR](https://github.com/hashicorp/boundary-ui/pull/1585/files#diff-7c68e3f23829508db0431c1373bb7905a28b9d8bb60a42a245e4cd88b412743e). I use `hooks.beforeEach()` to set the feature flags for those test cases. I actually wasn't sure if I should leave this flag on for `development` or not. Happy to let the team decide which features we have when using vercel deployments.

ui/admin/app/components/worker-diagram/index.hbs
```diff
@@ -1,3 +1,10 @@
 <div data-test-worker-diagram>
-  <WorkerDiagram::SingleFilter @egressFilter={{@egressFilter}} />
+  {{#if (feature-flag 'target-worker-filters-v2-ingress')}}
+    <WorkerDiagram::DualFilter
+      @egressFilter={{@egressFilter}}
+      @ingressFilter={{@ingressFilter}}
+    />
+  {{else}}
+    <WorkerDiagram::SingleFilter @egressFilter={{@egressFilter}} />
+  {{/if}}
```
cameronperera, line 9: My initial plan was to have 3 logical paths but I went a different path for `DualFilter::HCP` so we only have 2 paths here. This now makes this logic overkill. We could easily move this fairly simple logic to the `target/details/index.hbs` template. I would like to hear other's thoughts on this.

ui/admin/config/environment.js
calcaide: @cameronperera Amazing!! I even like more this approach. I remember we discuss about it, but it skip to me you add it here.

Would you mind to add a task on Jira or somewhere we will not forget to document this pattern? I am sure others we will need 😄

ui/admin/app/components/worker-diagram/index.hbs
dvonanderson: @cameronperera. So what would be the outcome of the logic inside of the `target/detail` are you calling `<DualFilter />` and  making decisions about `ent` and `hcp`  in there?

ui/admin/app/components/worker-diagram/dual-filter/index.hbs
```diff
@@ -0,0 +1,67 @@
+<div data-test-dual-filters>
```
randallmorey, line 1: If the only reason this exists is to provide a data attribute, I think we should find another way.  Otherwise, we're implying that all components will always need wrappers.

randallmorey (changes requested):

ui/admin/app/styles/app.scss
```diff
@@ -613,6 +613,10 @@
       flex-direction: column;
       align-items: center;
       text-align: center;
+
+      .hds-icon-tile--color-hcp {
+        background-color: var(--subtle);
+      }
```
randallmorey, line 619: I would prefer not to provide this override to the default HDS styles, if possible.  Over time these accumulate.

randallmorey (changes requested):

ui/admin/config/environment.js
cameronperera: I think an epic or story has been made but if not I will do so.

ui/admin/app/components/worker-diagram/index.hbs
cameronperera: In this [PR](https://github.com/hashicorp/boundary-ui/pull/1585), you will see the use of this feature flag: `target-worker-filters-v2-hcp`. That will control the differences between `ent` and `hcp`. This branch is only for the `ent` version of Boundary.

ui/admin/config/environment.js
cameronperera: https://hashicorp.atlassian.net/browse/ICU-7605

ui/admin/app/components/worker-diagram/index.hbs
cameronperera: Updated this logic. Moved to `yielding` SingleFilter or DualFilter.

ui/admin/app/components/worker-diagram/dual-filter/index.hbs
cameronperera: Removed.

ui/admin/app/styles/app.scss
cameronperera: Removed.

dvonanderson (approved):
Awesome

DhariniJeeva (approved):

PR #1585, opened 2023-01-16
feat: 🎸 Add HCP logic to Worker Diagram component

cameronperera:
✅ Closes: ICU-7594

:tickets: [Jira ticket](https://hashicorp.atlassian.net/browse/ICU-7594)

## Description

Add HCP logic to Worker Diagram component

**Note: Not being rendered on any screen yet.**

Screen Shot 2023-01-15 at 9 50 22 PM


ui/admin/app/components/worker-diagram/dual-filter/index.hbs
```diff
@@ -25,7 +25,11 @@
         {{t 'resources.target.workers.diagram.client'}}
       </D.Item>
       <D.Item @icon='settings'>
-        {{t 'resources.target.workers.diagram.front-line-worker'}}
+        {{#if (feature-flag 'target-worker-filters-v2-hcp')}}
+          {{t 'resources.target.workers.diagram.ingress-worker'}}
+        {{else}}
+          {{t 'resources.target.workers.diagram.front-line-worker'}}
+        {{/if}}
```
cameronperera, line 32: I decided to go this route vs WorkerDiagram::DualFilter::HCP as it reduces duplicate code. Happy to create an HCP specific component if the team feels it is a better route.

ui/admin/tests/integration/components/worker-diagram/dual-filter/index-test.js
```diff
@@ -5,41 +5,153 @@ import { hbs } from 'ember-cli-htmlbars';
...
       await render(
         hbs`<WorkerDiagram::DualFilter @egressFilter={{false}} @ingressFilter={{false}} />`
       );
 
+      assert
+        .dom('.ordered-series-diagram-item:nth-child(2)')
+        .hasText('Front line worker');
+      assert
+        .dom('.ordered-series-diagram-item:nth-child(2)')
+        .doesNotIncludeText('HCP worker');
```
cameronperera, line 34: Added `.hasText()` and `.doesNotIncludeText()` asserts due to the translations being controlled by the feature flags. Happy to discuss alternatives if needed.

dvonanderson (approved):
LGTM

randallmorey (changes requested):
Let's stick to one module per test file, as is conventional in Ember.

ui/admin/tests/integration/components/worker-diagram/dual-filter/index-test.js
randallmorey: Let's use data test attributes for this?  We should avoid testing on hard coded strings in this manner.  If you really must, lookup the intl service in the test and fetch the value of the string via it's translation key.

DhariniJeeva (approved):

ui/admin/app/components/worker-diagram/dual-filter/index.hbs
```diff
@@ -37,7 +37,11 @@
         {{t 'resources.target.workers.diagram.client'}}
       </D.Item>
       <D.Item @icon='settings'>
-        {{t 'resources.target.workers.diagram.front-line-worker'}}
+        {{#if (feature-flag 'target-worker-filters-v2-hcp')}}
+          {{t 'resources.target.workers.diagram.hcp-worker'}}
+        {{else}}
+          {{t 'resources.target.workers.diagram.front-line-worker'}}
+        {{/if}}
```
cameronperera, line 44: It was suggested that we split this component into 2 new ones. `WorkerDiagram::DualFilter` and `WorkerDiagram::DualFilter::Hcp`. The only difference would be the logic here and on lines 64-68. Then we could remove the tests that check for the correct language rendered. We would just need to verify the correct component was rendered as the language will be unique to that component.

ui/admin/tests/integration/components/worker-diagram/dual-filter/enterprise/index-test.js
```diff
@@ -5,17 +5,33 @@ import { hbs } from 'ember-cli-htmlbars';
...
       await render(
         hbs`<WorkerDiagram::DualFilter @egressFilter={{false}} @ingressFilter={{false}} />`
       );
 
+      assert
+        .dom('.ordered-series-diagram-item:nth-child(2)')
+        .hasText('Any worker');
+      assert
+        .dom('.ordered-series-diagram-item:nth-child(2)')
+        .doesNotIncludeText('HCP worker');
```
cameronperera, line 34: This is one of the asserts that we could remove once the `WorkerDiagram::DualFilter` has been split into 2 new components.

ui/admin/tests/integration/components/worker-diagram/dual-filter/index-test.js
```diff
@@ -5,75 +5,72 @@ import { hbs } from 'ember-cli-htmlbars';
...
       await render(
         hbs`<WorkerDiagram::DualFilter @egressFilter={{false}} @ingressFilter={{false}} />`
       );
 
-      assert.dom('[data-test-dual-filter-egress-off-ingress-off]').isVisible();
       assert
-        .dom('[data-test-dual-filter-egress-on-ingress-off]')
-        .doesNotExist();
+        .dom('.ordered-series-diagram-item:nth-child(2)')
+        .hasText('Any worker');
```
cameronperera, line 31: Apologies, my previous comment about these tests might not have been worded correctly. We want to avoid testing against raw strings. We should be using the `data-test-*` attributes just to confirm the correct diagram is rendered.

cameronperera (commented):
LGTM

DhariniJeeva (approved):

PR #1586, opened 2023-01-16
chore: 🤖 Refactor onboarding

calcaide:
✅ Closes: ICU-7558

:tickets: [Jira ticket](https://hashicorp.atlassian.net/browse/ICU-7558)

## Description

- Rename `hostAddress` to `targetAddress`.
- Simplified onboarding routes:
  - The current routes:
    - Onboard landing: `/onboarding/quick-setup/choose-path`
    - Quick setup: `/onboarding/quick-setup/create-resources`
    - Quick setup success: `/onboarding/quick-setup/create-resources/success`
  - New routes:
    - Onboard landing (now is the form right away): `/onboarding`
    - Onboarding success: `/onboarding/success`.
- Simplified Styles.
- Refactor routes and templates to accomodate new routes mentioned above.
- Refactor all the test regarding onboarding.
- We implemented ad-hoc form validation.

**Vercel deployments:**
- Before applying changes: [Link](https://boundary-ui.vercel.app/onboarding/quick-setup/create-resources)
- After applying changes: [Link](https://boundary-qvooml83u-hashicorp.vercel.app/onboarding/quick-setup/create-resources)

### Screenshots:

Before changes: Quick setup / create resources
Screen Shot 2023-01-16 at 2 57 00 PM

After changes: Onboarding
Screen Shot 2023-01-25 at 11 54 30 AM

After changes: Onboarding validating form and save form button ALWAYS active:
Screen Shot 2023-01-31 at 10 38 34 AM

After changes: success screen (with new `icon` component):
Screen Shot 2023-01-31 at 10 40 10 AM

### Videos

**Onboard showing fields are required, after they are fill allows to save**

https://user-images.githubusercontent.com/9775006/215853097-efa52ac8-030f-4db9-9c09-c085122997c5.mov

**Onboard clicks DO this later**

https://user-images.githubusercontent.com/9775006/215555009-248075b0-8cc9-4f9d-af60-5a84f40f568a.mov


calcaide:
> The after image doesn't show host catalog or host set in the graphic but the text above it still explains that host catalog will be created. @calcaide, can you check the text? Ty!

@gsusmi I copied the same text than in the designs. I will ask @joshuaogle which will be the proper text to use here. Thank you for the catch

joshuaogle:
@calcaide Thanks for catching the helper text update, I changed it to : `Targets live in projects, and projects live in orgs. We will create these scopes for you but they can be edited or deleted later.`

It looks like we just renamed the host address label but just to make sure, this isn't just a name change, it should be using the new address field for targets and not using the old host source address.

ui/admin/app/routes/onboarding/quick-setup/create-resources.js
```diff
@@ -79,8 +82,8 @@ export default class OnboardingQuickSetupCreateResourcesRoute extends Route {
     hostSet.host_catalog_id = hostCatalog.id;
     await hostSet.save();
     host.host_catalog_id = hostCatalog.id;
-    if (hostAddress) {
-      host.address = hostAddress;
+    if (targetAddress) {
+      host.address = targetAddress;
```
ZedLi, line 86: I think we're trying to actually save the address to the target rather than using a host. The goal is to simplify the initial setup so we shouldn't be creating host catalogs, host sets, hosts, etc to add to the target

ZedLi (changes requested):

ui/admin/app/routes/onboarding/quick-setup/create-resources.js
calcaide: Yes, missed that part and working on it ATM  😄 I will provide changes

ui/admin/app/routes/onboarding/quick-setup/create-resources.js
```diff
@@ -58,46 +58,53 @@ export default class OnboardingQuickSetupCreateResourcesRoute extends Route {
...
       await target.save();
-      await target.addHostSources([hostSet.id]);
+      // Format role and role grants objects and persist them
       role.scopeID = org.id;
       role.grant_scope_id = project.id;
       await role.save();
       await role.saveGrantStrings([
         `type=target;actions=list`,
-        `id=${target.id};actions=authorize-session`,
+          `id=${target.id};actions=authorize-session`,
```
ZedLi, line 94: I think formatting got messed up here

calcaide: You are right. I updated the PR with this corrected thanks for the catch

randallmorey (approved):

cameronperera (approved):

addons/core/tests/integration/components/image-test.js
```diff
@@ -8,7 +8,7 @@ module('Integration | Component | Image', function (hooks) {
 
   test('it renders', async function (assert) {
     assert.expect(1);
-    await render(hbs`<Image @name="resource-creation" />`);
+    await render(hbs`<Image @name="logo-app" />`);
```
calcaide, line 11: This was unrelated to the topic of this PR.
But deleting the `resource-creation` imaged from core, reveal that we were using such image to test the `<image>` component which is not related to `resource-creation`. SInce we just need to test it renders an image I changed to the app logo which most likely will never be deleted.

ui/admin/app/components/form/onboarding/index.js
```diff
@@ -0,0 +1,4 @@
+/* eslint-disable ember/no-empty-glimmer-component-classes */
+import Component from '@glimmer/component';
+
+export default class FormOnboardingComponent extends Component {}
```
cameronperera, line 4: Why do we need an empty component class?

ui/admin/app/templates/onboarding/success.hbs
```diff
@@ -0,0 +1,90 @@
...
+      {{! We have created }}
+      <div class='create-resources-item'>
+        <span class='create-resources-item-header'>
+          <Rose::Icon @name='flight-icons/svg/check-circle-16' />
+          <h3>{{t 'onboarding.success.titles.creation-success'}}</h3>
+        </span>
+        <p>{{t 'onboarding.success.descriptions.role'}}</p>
+      </div>
+
+      {{! Donwload Desktop client }}
```
cameronperera, line 17:
```suggestion
      {{! Download Desktop client }}
```

ui/admin/app/templates/onboarding/success.hbs
```diff
@@ -0,0 +1,90 @@
...
+      <div class='create-resources-item'>
+        <span class='create-resources-item-header'>
+          <Rose::Icon @name='logo-app-background' @size='extra-large' />
```
cameronperera, line 20: Can we use the `<Hds::IconTile />` here?

ui/admin/tests/acceptance/onboarding/onboarding-test.js
```diff
@@ -0,0 +1,43 @@
...
+
+  hooks.beforeEach(() => {
+    authenticateSession({});
+  });
+
+  test('show targetAddress and targetPort fields', async function (assert) {
+    assert.expect(2);
+    await visit(urls.onboarding);
+    assert.ok('[name="targetAddress"]');
+    assert.ok(find('[name="targetPort"]'));
```
cameronperera, line 25: Can we update these tests to use the `assert.dom('selector').isVisible()` pattern?

cameronperera (changes requested):
Looks really great and simplified for our users! I left a couple of small change requests and a question about the empty class.

ui/admin/app/components/form/onboarding/index.hbs
```diff
@@ -0,0 +1,60 @@
...
+      as |F|
+    >
+      <F.Label>{{t 'onboarding.add-target.titles.target-port'}}</F.Label>
+      <F.HelperText>{{t
+          'onboarding.add-target.descriptions.target-port'
+        }}</F.HelperText>
+    </Hds::Form::TextInput::Field>
+  </div>
+
+  <form.fieldset>
```
DhariniJeeva, line 40: can we replace this with `<Hds::Form::Fieldset>`

calcaide: @DhariniJeeva Good comment. Initially I though to do so but didn't find any usage of `<Hds::Form::Fieldset>` with our codebase so I though to keep it simple. I changed the inputs to be HDS so I will try to get the `<Hds::Form::Fieldset>` that will be nice.
Thanks for the suggestion 😉

ui/admin/app/components/form/onboarding/index.js
calcaide: Ouch! I forgot to commit last change in here, you are right @cameronperera we shouldn't have empty classes. I will work on it and provide an update.

Thanks for the suggestion 😉

ui/admin/app/routes/onboarding/success.js
```diff
@@ -0,0 +1,29 @@
...
+    };
+  }
+
+  @action
+  showTargetList(scope) {
+    const {
+      project: { id: scopeID },
+      target: { id: targetID },
+    } = scope;
+    this.router.transitionTo('scopes.scope.targets.target', scopeID, targetID);
```
DhariniJeeva, line 27: What does the `targetId` do here? I understand we pass scopeID as the model; do we need to pass in targetID as well?

ui/admin/app/routes/onboarding.js
```diff
@@ -1,36 +1,122 @@
...
+    } catch (error) {
+      // Redirect to orgs
+      this.router.replaceWith('scopes.scope', 'global');
+      throw new Error();
+    }
   }
 
   /**
-   * If any orgs already exist, redirect to index to skip onboarding.
+   * Persist org and project previosly created in model()
```
DhariniJeeva, line 83:
```suggestion
   * Persist org and project previously created in the model
```

ui/admin/app/templates/onboarding/success.hbs
```diff
@@ -0,0 +1,90 @@
...
+          <h3>{{t 'onboarding.success.titles.desktop'}}</h3>
+        </span>
+        <p>{{t 'onboarding.success.descriptions.download'}}</p>
+
+        <Hds::Link::Standalone
+          @icon='download'
+          @text='{{t "onboarding.success.links.install"}}'
```
DhariniJeeva, line 27: I don't think the quotes are necessary with text, href

ui/admin/app/templates/onboarding/success.hbs
```diff
@@ -0,0 +1,90 @@
...
+        <Hds::Link::Standalone
+          @icon='docs-link'
+          @text='{{t "onboarding.success.links.cli"}}'
+          @href='{{doc-url "api-client.cli"}}'
+          isHrefExternal={{true}}
+        />
+
+        <Hds::Link::Standalone
+          @icon='docs-link'
+          @text='{{t "onboarding.success.links.api"}}'
```
DhariniJeeva, line 77: I don't think the quotes are necessary with text, href

ui/admin/app/templates/onboarding/success.hbs
```diff
@@ -0,0 +1,90 @@
...
+
+        <Hds::Link::Standalone
+          @icon='download'
+          @text='{{t "onboarding.success.links.install"}}'
+          @href='{{doc-url "downloads"}}'
+          isHrefExternal={{true}}
+        />
+
+        <Hds::Link::Standalone
+          @icon='docs-link'
```
DhariniJeeva, line 33: I don't think the quotes are necessary with text, href

DhariniJeeva (changes requested):

ui/admin/app/components/form/onboarding/index.hbs
cameronperera: @calcaide Originally we did not have an example of the `<HDS::Form::Fieldset>` but we do now! We added it during our multi-hop work. See the example [here](https://github.com/hashicorp/boundary-ui/blob/main/ui/admin/app/components/form/target/details/index.hbs#L180).

ui/admin/app/components/form/onboarding/index.hbs
```diff
@@ -0,0 +1,60 @@
...
+  @onSubmit={{fn @submit this.targetAddress this.targetPort}}
+  as |form|
+>
+  <div class='create-resources-form'>
+    <Hds::Form::TextInput::Field
+      name='targetAddress'
+      @type='text'
+      @value={{this.targetAddress}}
+      @isRequired={{true}}
+      @isInvalid={{this.targetAddress}}
```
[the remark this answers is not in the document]
randallmorey: Doesn't this essentially say "if a target address is set, then the field is invalid"?

ui/admin/app/routes/onboarding/success.js
randallmorey: Yes because both scope and target have dynamic parts.

ui/admin/app/components/form/onboarding/index.hbs
DhariniJeeva: We recently replaced HDS fieldset for [target forms](https://github.com/hashicorp/boundary-ui/blob/9b998ece1b04fb35091b1991aa8048828a850cfb/ui/admin/app/components/form/target/details/index.hbs#L180)

randallmorey (changes requested):
Can the PR be renamed to something more meaningful regarding quickstart?

ui/admin/app/components/form/onboarding/index.hbs
calcaide: Thanks for the references I was not aware from your work with `<HDS>>Form>>Fieldset>`. I also found interesting [its documentation](https://hds-website-hashicorp.vercel.app/components/form/primitives?tab=code) I was not obvious but found it 😉 
Changes applied.

ui/admin/app/components/form/onboarding/index.hbs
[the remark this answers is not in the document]
calcaide: Sorry, this was leftovers when I was trying some custom client validation. Forgot to commit the delition of it. Now is clean

calcaide:
> Can the PR be renamed to something more meaningful regarding quickstart?

As explained in the description of this PR, we refactored routes, the word `quickstart` dissapeared from it. I think it will be confusing using such word on the PR name.
I will rename it to refactor onboarding.

ui/admin/app/templates/onboarding/success.hbs
calcaide: Thanks for the catch!! 
I realize I messed up this file was not consistent, I was mixing single and double quotes. Now unnecessary quotes are deleted and just using single quotes.

ui/admin/app/templates/onboarding/success.hbs
calcaide: Thanks for the catch!! 
I realize I messed up this file was not consistent, I was mixing single and double quotes. Now unnecessary quotes are deleted and just using single quotes.

ui/admin/app/templates/onboarding/success.hbs
calcaide: Thanks for the catch!! 
I realize I messed up this file was not consistent, I was mixing single and double quotes. Now unnecessary quotes are deleted and just using single quotes.

ui/admin/app/routes/onboarding.js
```diff
@@ -1,36 +1,122 @@
...
+  @action
+  @loading
+  @notifyError(function () {
+    return this.intl.t('errors.onboarding-failed.description');
+  })
+  async createTarget(targetAddress, targetPort) {
+    // Create org and Project
+    try {
+      await this.createOrgAndProject();
+    } catch (error) {
```
ZedLi, line 66: Why catch and rethrow a generic error? Would it make sense to separate org and project creation even further and if one fails, drop them off at either the org or project page depending on what failed?

ui/admin/app/components/form/onboarding/index.hbs
```diff
@@ -0,0 +1,60 @@
+<Rose::Form
+  class='full-width'
+  @cancel={{@cancel}}
+  @disabled={{@model.target.isSaving}}
```
ZedLi, line 4: Do we have a model? I didn't see one being passed into the form

ui/admin/app/routes/onboarding.js
```diff
@@ -1,36 +1,122 @@
...
+    }
+
+    // Create target
+    try {
+      await this.createSampleTarget(targetAddress, targetPort);
+      // Redirect to success
+      this.router.transitionTo('onboarding.success');
+    } catch (error) {
+      // Redirect to orgs
+      this.router.replaceWith('scopes.scope', 'global');
```
ZedLi, line 77: Would it make more sense to drop them off at the targets page instead?

ui/admin/app/routes/onboarding.js
```diff
@@ -1,36 +1,122 @@
...
+    role.grant_scope_id = project.id;
+    await role.save();
+    await role.saveGrantStrings([
+      `type=target;actions=list`,
+      `id=${target.id};actions=authorize-session`,
+    ]);
+  }
+
+  @action
+  @loading
```
ZedLi, line 118: Is this `loading` decorator needed for a transition only?

ui/admin/app/templates/onboarding/success.hbs
calcaide: Awesome catch and even better suggestion! you got it 😉

ui/admin/app/routes/onboarding.js
calcaide: You are 100% right, that is not need at all here, deleted.

ui/admin/app/components/form/onboarding/index.hbs
calcaide: Good catch, my bad.

Changed to:
- Passing model.
- Not using `@disabled` in this form since it does not apply. Or at least not with the current implementation, correct me if I am wrong.

ui/admin/app/routes/onboarding.js
calcaide: Changed, now it throws the error passed by param

ui/admin/app/routes/onboarding.js
calcaide: You are right, because if the target failed, org already exhist.

ui/admin/app/components/form/onboarding/index.js
```diff
@@ -0,0 +1,48 @@
+import Component from '@glimmer/component';
+import { action } from '@ember/object';
+import { tracked } from '@glimmer/tracking';
+
+export default class FormOnboardingComponent extends Component {
+  // =properties
+  @tracked targetAddress;
+  @tracked targetAddressInValid = false;
```
ZedLi, line 8: I think these could have been getters that check the `targetAddress` or `targetPort` values to determine if they're valid

ui/admin/app/components/form/onboarding/index.hbs
```diff
@@ -0,0 +1,73 @@
...
+>
+  <div class='create-resources-form'>
+    <Hds::Form::TextInput::Field
+      name='targetAddress'
+      @type='text'
+      @value={{this.targetAddress}}
+      @isRequired={{true}}
+      @isInvalid={{this.targetAddressInValid}}
+      {{on 'input' (set-from-event this 'targetAddress')}}
+      {{on 'focusout' this.isTargetAddressInValid}}
```
ZedLi, line 14: Not sure what other's thoughts are but it feels a little weird to have it on the `focusout` event, I wonder if it might be better to validate when trying to save to mimic how our validation works on forms normally?

calcaide: @ZedLi 
Initially I wanted to do it at the same method we submit but encounter some scope limitations. Maybe I didn't though about it properly, let's catch up about this one during pod standup zoom call.

ui/admin/app/components/form/onboarding/index.js
calcaide: Since this comment could be related to the one above, let's discuss in our pod stand up zoom call.

calcaide: @ZedLi Refactored to a simplified way 😃 Thanks for the help!

ui/admin/app/components/form/onboarding/index.hbs
```diff
@@ -0,0 +1,63 @@
+<Rose::Form class='full-width' @cancel={{@cancel}} @onSubmit={{this.atSubmit}}>
+  <div class='create-resources-form'>
+    <Hds::Form::TextInput::Field
+      name='targetAddress'
+      @type='text'
+      @value={{this.targetAddress}}
+      @isRequired={{true}}
+      @isInvalid={{this.targetAddressInValid}}
+      {{on 'input' (set-from-event this 'targetAddress')}}
```
calcaide, line 9: @ZedLi Delete `onfocusout` event handlers. Now form validation happens within `atSubmit` method.

ui/admin/app/components/form/onboarding/index.hbs
```diff
@@ -0,0 +1,63 @@
...
+    <F.Legend>{{t 'onboarding.add-target.titles.target-setup-info'}}</F.Legend>
+    <F.Control>{{t
+        'onboarding.add-target.descriptions.target-setup-info'
+      }}</F.Control>
+    <F.Control><Image @name='ssh-target-creation' /></F.Control>
+  </Hds::Form::Fieldset>
+
+  <Hds::ButtonSet>
+    <Hds::Button @text='{{t "actions.save"}}' type='submit' />
+    <Hds::Button
```
calcaide, line 55: @ZedLi Since we refactor the form validation to happen at submit time rather than on input events, I delete the submit button disable functionality. Now submit button is always ON.

ui/admin/app/components/form/onboarding/index.hbs
DhariniJeeva: could we call this action`submit` for consistency?

cameronperera (approved):
LGTM

ui/admin/app/routes/onboarding.js
```diff
@@ -1,36 +1,146 @@
...
+  async createTarget(project, target, targetAddress, targetPort) {
+    target.scopeID = project.id;
+    target.address = targetAddress;
+    target.default_port = targetPort;
+
+    try {
+      await target.save();
+    } catch (error) {
+      // Redirect to targets global
+      this.router.replaceWith('scopes.scope.targets', 'global');
```
ZedLi, line 121: Should this be global, wouldn't it be the project scope?

ui/admin/app/components/form/onboarding/index.hbs
DhariniJeeva: yeah rename the action name from `atSubmit` to submit, to be consistent with other submit actions in forms

calcaide: If that was the change you mean already done 😃

ui/admin/app/components/form/onboarding/index.js
```diff
@@ -0,0 +1,32 @@
...
+  // =properties
+  @tracked targetAddress;
+  @tracked targetAddressInvalid = false; // change to method
+  @tracked targetPort;
+  @tracked targetPortInvalid = false; // change to method name
+
+  // =actions
+  @action
+  // change name to formSubmit
+  submit() {
```
DhariniJeeva, line 15: curious, what are these comments for? Are we changing these in the future?

ui/admin/app/components/form/onboarding/index.js
```diff
@@ -0,0 +1,32 @@
...
+    // Check targetAddress and targetPort are valid before submit
+    this.targetAddressInvalid = this.requiredFieldInvalid(this.targetAddress);
+    this.targetPortInvalid = this.requiredFieldInvalid(this.targetPort);
```
ZedLi, line 18: It might make more sense to just forgo the function and just do `!this.targetAddress` and `!this.targetPort` since it's not doing much

And a comment that this is necessary so that for the initial render it doesn't think the fields are invalid

ui/admin/app/routes/onboarding.js
calcaide: Yes, you are right. Changed 😃

ui/admin/app/components/form/onboarding/index.js
calcaide: @DhariniJeeva sorry... rushing it = I always forget something 😅 Deleted all unnecessary comments

ui/admin/app/components/form/onboarding/index.js
calcaide: You are right. I will quickly change it, run manual testing and push this change. Thanks for the catch!

calcaide: @ZedLi Actually this is neccessary, this is the only moment where the set state for `targetAddressInvalid` and `targetPortInvalid`

ui/admin/app/templates/onboarding/success.hbs
```diff
@@ -0,0 +1,94 @@
...
+    <:header>
+      <h2>{{t 'onboarding.success.titles.connect'}}</h2>
+    </:header>
+    <:body>
+      {{! We have created }}
+      <div class='create-resources-item'>
+        <span class='create-resources-item-header'>
+          <FlightIcon
+            @name='check-circle'
+            @color='var(--token-color-foreground-success)'
```
DhariniJeeva, line 13: this should live in scss, we could use the class`create-resources-item-header` above to target the icon

calcaide: @DhariniJeeva 
Not that I have a strong opinion, but what is the reason this should live in scss? I think this is the advantatge of tokens and the <FlightIcon> component spect [here link](https://hds-website-hashicorp.vercel.app/icons/usage-guidelines?tab=code) shows this as a way of using it.

ui/admin/tests/acceptance/onboarding/onboarding-test.js
calcaide: @cameronperera sorry I forgot about this comment. Changed! thanks for the catch

ui/admin/app/components/form/onboarding/index.js
calcaide: Sorry I missundertood the comment. Yes `requiredFieldInvalid` is not need at all, refactored to not use it.

ui/admin/app/templates/onboarding/success.hbs
randallmorey: Separation of concerns, the beauty and readability of this expression notwithstanding.  I agree that we should not mix styles and templates in this manner, even though there is an example of this usage.  I'm actually surprised icons don't use `currentColor` so that we can thus use [HDS color classes](https://hds-components-hashicorp.vercel.app/foundations/colors) or overrides in our stylesheets.

calcaide: @DhariniJeeva @randallmorey 
I understand separation of concerns, but this is a pattern we (almost all devs in Boundary) are already following, just look on your code editor and search for `@color` in our codebase. From those 22 matches follow this pattern, using `@color` in the template.
HDS give us the `@color` attribute and tokens for a reason. 

Since this looks to me as an opinionated "rule". I will suggest to bring it to tech time for discussion and whatever decission we take about it, we should documented. Because opinionated patterns should be documented, otherwise... will not be consistency with it.

calcaide: @randallmorey
Regarding the `currentColor` comment. That was my first approach but HDS did not support it, I opened [an slack thread](https://hashicorp.slack.com/archives/C7KTUHNUS/p1674860484678539) that mentions such thing.

randallmorey: The existing usage of `@color` is specifically with semantic color names, not style tokens.  Using style tokens is specifically the usage I am commenting on here.

calcaide: @randallmorey 
I understand you were commenting on the usage of a token. And as per my understanding is opinionated.
As per understanding of HDS seems like is opinionated too, this is extracted from [here](https://hds-components-hashicorp.vercel.app/foundations/colors):

"There are two different ways to apply a color to a UI element, via design tokens or via CSS helper classes.

Which one to use will depend on the context: since both refer to the same design token values, from the design system perspective there's no difference between the two methods (it's more of a preference for the "consumer" codebase)."

If you don't mind, I would like to bring this to FE tech time, discuss it and document it somewhere, so we are clear on what patterns to follow.

PR #1589, opened 2023-01-18
Icu 7301 UI target oss workflow

DhariniJeeva:
jira- part of [this](https://hashicorp.atlassian.net/browse/ICU-7301). Tests will be in a separate PR.

Screen Shot 2023-01-19 at 2 24 09 PM

 
when the update filter is clicked,
 
Screen Shot 2023-01-19 at 2 24 15 PM


ui/admin/app/components/form/target/details/index.js
```diff
@@ -37,4 +42,37 @@ export default class FormTargetComponent extends Component {
...
+  /**
+   * returns egress_worker_filter value from the model,
+   * but in case of updating the deprecated worker filter, value from worker_filter field is returned
+   * @type {boolean}
+   */
+  get egressWorkerFilterValue() {
+    return this.args.model.egress_worker_filter?.length
+      ? this.args.model.egress_worker_filter
+      : this.args.model.worker_filter;
+  }
```
randallmorey, line 66: When the user clicks update, the value is literally _copied_ into the egress and ingress field.  It will always come from the model.

ui/admin/app/components/form/target/details/index.js
```diff
@@ -37,4 +42,37 @@ export default class FormTargetComponent extends Component {
...
+  // =actions
+  @action
+  toggleEgressWorkerFilter() {
+    this.egressWorkerFilterEnabled = !this.egressWorkerFilterEnabled;
+  }
+  // =actions
+  @action
+  updateDeprecatedFilter() {
+    this.updateDeprecatedWorkerFilter = true;
+    this.egressWorkerFilterEnabled = !this.egressWorkerFilterEnabled;
```
randallmorey, line 76: The toggle's value is derived, not set.  It is enabled whenever `this.args.model.egress_worker_filter` has a value.

randallmorey (changes requested):

ui/admin/app/routes/scopes/scope/targets.js
```diff
@@ -51,16 +51,30 @@ export default class ScopesScopeTargetsRoute extends Route {
...
    */
   @action
   @loading
   @notifyError(({ message }) => message)
   @notifySuccess(({ isNew }) =>
     isNew ? 'notifications.create-success' : 'notifications.save-success'
   )
-  async save(target) {
-    await target.save();
+  async save(target, egressEnabled, updateDeprecatedWorkerFilter) {
```
randallmorey, line 62: The existing route save action should remain the same as always.  One way to achieve this is if the form component accepts an additional argument and switches the save action that it calls.

addons/api/addon/models/target.js
```diff
@@ -318,6 +318,34 @@ export default class TargetModel extends GeneratedTargetModel {
       options
     );
   }
+  /**
+   * save target with filters based on egress/ingress toggle
+   * @param {object} target
+   * @param {object} egress_worker_filter
+   * @return {Promise}
+   */
+  async saveWithToggles(target, egressEnabled = true) {
```
randallmorey, line 327: After reviewing, this should live in the route.  "egressEnabled" is a mostly UI concept and doesn't make sense in the API layer.

ui/admin/app/routes/scopes/scope/targets.js
```diff
@@ -51,16 +51,30 @@ export default class ScopesScopeTargetsRoute extends Route {
...
+    //by default, when the deprecated worker_filter field is available,
+    //copy the filter value into egress filter field if egress filter is enabled
+    if (
+      updateDeprecatedWorkerFilter &&
```
randallmorey, line 66: This is not a route level concern.  According to the RFC, saveWithToggles should accept the value of the toggles only.

randallmorey (changes requested):
Let's separate out the _save_ functionality from the component itself, for easier review.

addons/rose/addon/styles/hds/overrides.scss
```diff
@@ -3,3 +3,11 @@
 [class^='hds-form-group'] {
   margin-bottom: 1.5rem;
 }
+/* stylelint-disable selector-class-pattern */
```
DhariniJeeva, line 6: [selector from hds](https://github.com/hashicorp/design-system/blob/6cf5bff1f50ab2caaaa7f3ba346d353e6c353ff3/packages/components/app/styles/components/form/group.scss#L21) to override the font-weight for boundary. 

Screen Shot 2023-01-19 at 5 40 54 PM

ui/admin/app/routes/scopes/scope/targets.js
```diff
@@ -69,6 +69,33 @@ export default class ScopesScopeTargetsRoute extends Route {
...
 
+  /**
+   * save target with filters based on egress/ingress toggle
+   * @param {object} target
+   * @param {boolean} egressEnabled
+   * @return {Promise}
+   */
+  @action
+  @loading
+  async saveWithToggles(target, egressEnabled = true) {
```
ZedLi, line 80: Is there an advantage to having a separate save function that calls into the original save instead of just using the same save?

Beyond that, how come this logic doesn't live in the the save function in the form rather than here in the route?

ui/admin/app/components/form/target/details/index.js
```diff
@@ -57,31 +54,30 @@ export default class FormTargetComponent extends Component {
...
     this.egressWorkerFilterEnabled = true;
-    // When update is clicked, copy worker filter value into egress filter.
+    // When update is clicked, copy worker filter value into egress filter and clear the worker_filter
     this.args.model.egress_worker_filter = this.args.model.worker_filter;
     this.args.model.worker_filter = '';
   }
 
   @action
   @loading
-  @notifyError(({ message }) => message)
```
ZedLi, line 84: Won't we stop notifying errors now if removingHostSources fail? I think this is still needed.

ZedLi (changes requested):
Had a few questions

cameronperera (approved):

PR #1591, opened 2023-01-18
feat: 🎸 Update models for vault ssh cert feature

ZedLi:
✅ Closes: ICU-7546

:tickets: [Jira ticket](https://hashicorp.atlassian.net/browse/ICU-7546)

## Description
Updated models for the Vault SSH Cred feature.
This is a unique case where we have a breaking change where the backend is renaming `vault` to `vault-generic` but there isn't much functionality change, except for using types for translations.

I kept the original vault translation type as well as part of the model check for the original vault so we can still be accepting either of them for now. It didn't make sense to use a feature flag anywhere for this.

image

### DIFF
--- a/addons/api/addon/generated/models/credential-library.js
+++ b/addons/api/addon/generated/models/credential-library.js
@@ -100,15 +100,14 @@ export default class GeneratedCredentialLibraryModel extends BaseModel {
   })
   key_id;
 
-  @attr('object-array', {
+  @attr('object-as-array', {
     isNestedAttribute: true,
-    defaultValue: () => {},
     description:
       'Specifies a map of the critical options that the certificate should be signed for.',
   })
   critical_options;
 
-  @attr('object-array', {
+  @attr('object-as-array', {
     isNestedAttribute: true,
     description:
       'Specifies a map of the extensions that the certificate should be signed for.',
--- a/addons/api/addon/generated/models/credential-library.js
+++ b/addons/api/addon/generated/models/credential-library.js
@@ -67,4 +67,51 @@ export default class GeneratedCredentialLibraryModel extends BaseModel {
     description: 'The path in Vault to request credentials from.',
   })
   path;
+
+  @attr('string', {
+    isNestedAttribute: true,
+    description: 'The username to use when making an SSH connection.',
+  })
+  user_name;
+
+  @attr('string', {
+    isNestedAttribute: true,
+    description: 'The desired key type to use when generating a private key.',
+  })
+  key_type;
+
+  @attr('number', {
+    isNestedAttribute: true,
+    description:
+      'Specifies the number of bits used when generating the private key. Not used if key type is ed25519',
+  })
+  key_bits;
+
+  @attr('string', {
+    isNestedAttribute: true,
+    description: 'Specifies the requested time to live for the certificate.',
+  })
+  ttl;
+
+  @attr('string', {
+    isNestedAttribute: true,
+    description:
+      'Specifies the key id that the created certificate should have.',
+  })
+  key_id;
+
+  @attr('object-array', {
+    isNestedAttribute: true,
+    defaultValue: () => {},
+    description:
+      'Specifies a map of the critical options that the certificate should be signed for.',
+  })
+  critical_options;
+
+  @attr('object-array', {
+    isNestedAttribute: true,
+    description:
+      'Specifies a map of the extensions that the certificate should be signed for.',
+  })
+  extensions;
 }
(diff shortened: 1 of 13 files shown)


randallmorey (changes requested):
The types we use in our models exactly reflect the API.  The only change is an _addition_ of one type.  The existing type should remain unchanged (it is `vault`).  `vault-generic` is not a real type.  Can this be reworked with that in mind?

addons/core/translations/resources/en-us.yaml
```diff
@@ -450,7 +450,8 @@ credential-library:
       help: The body of the HTTP request the library sends to Vault when requesting credentials. Only allowed when HTTP method is set to POST.
   types:
     vault: Vault
-    static: Static
```
randallmorey, line 453: Is this removed?

randallmorey (changes requested):

addons/api/addon/models/credential-library.js
```diff
@@ -28,6 +32,14 @@ export default class CredentialLibraryModel extends GeneratedCredentialLibraryMo
    * @type {boolean}
    */
   get isVault() {
-    return this.type === 'vault';
+    return this.type === VAULT_GENERIC || this.type === VAULT;
```
randallmorey, line 35:
```suggestion
    return this.isVaultGeneric || this.isVaultSSHCert;
```

addons/api/addon/models/credential-library.js
```diff
@@ -28,6 +32,14 @@ export default class CredentialLibraryModel extends GeneratedCredentialLibraryMo
    * @type {boolean}
    */
   get isVault() {
```
randallmorey, line 34: We should add another getter too, `isVaultGeneric`.

ui/admin/app/components/cred-source-type-badge/index.hbs
```diff
@@ -1,6 +1,6 @@
 {{#if @model.isUnknown}}
   <Hds::Badge @text={{t 'resources.credential.types.unknown'}} />
-{{else if @model.isVault}}
+{{else if (or @model.isVault @model.isVaultSSHCert)}}
```
randallmorey, line 3:
```suggestion
{{else if @model.isVault}}
```

addons/core/translations/resources/en-us.yaml
```diff
@@ -450,7 +450,8 @@ credential-library:
       help: The body of the HTTP request the library sends to Vault when requesting credentials. Only allowed when HTTP method is set to POST.
   types:
     vault: Vault
```
randallmorey, line 452: This would go away too I believe.
```suggestion
```

addons/api/addon/transforms/object-array.js
```diff
@@ -0,0 +1,18 @@
+import Transform from '@ember-data/serializer/transform';
+
+export default class ObjectArrayTransform extends Transform {
```
randallmorey, line 3: This name may be a bit confusing.  It implies an array of objects is the type, e.g. `[{}, {}]`.  But it's more of an object _to_ array transform.

randallmorey (changes requested):

addons/api/addon/generated/models/credential-library.js
```diff
@@ -67,4 +67,51 @@ export default class GeneratedCredentialLibraryModel extends BaseModel {
...
+  @attr('string', {
+    isNestedAttribute: true,
+    description:
+      'Specifies the key id that the created certificate should have.',
+  })
+  key_id;
+
+  @attr('object-array', {
+    isNestedAttribute: true,
+    defaultValue: () => {},
```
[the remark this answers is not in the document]
ZedLi: Woops good catch, that shouldn't be there

cameronperera (approved):

calcaide (approved):
LGTM!! Thank you for this

addons/api/addon/models/credential-library.js
```diff
@@ -7,10 +7,19 @@ export const options = {
   http_method: ['GET', 'POST'],
 };
 
+// TODO: Remove `vault` type once we enable the feature
+export const TYPE_CREDENTIAL_LIBRARY_VAULT = 'vault';
```
ZedLi, line 11: Do you have any opinions on this @randallmorey? Do you think it's worth keeping the old type? It's not easily "feature-flaggable" so it's not the best

randallmorey: Will the three types coincide?  If not, this should be removed.  Generally this should reflect the state of the API.  However, since it's not yet merged on the backend, we should wait on merge just in case.

PR #1597, opened 2023-01-20
feat: 🎸 Setup types in cred library forms and use HDS component

ZedLi:
:tickets: [Jira ticket](https://hashicorp.atlassian.net/browse/ICU-7673)

## Description
Initial change to allow credential libraries to have types. Also converted normal `vault` 

image

image


ui/admin/app/routes/scopes/scope/credential-stores/credential-store/credential-libraries/new.js
```diff
@@ -13,12 +16,22 @@ export default class ScopesScopeCredentialStoresCredentialStoreCredentialLibrari
...
    */
   model() {
-    const { id: credential_store_id, type } = this.modelFor(
+    const { id: credential_store_id } = this.modelFor(
       'scopes.scope.credential-stores.credential-store'
     );
+
     return this.store.createRecord('credential-library', {
-      type,
+      type: TYPE_CREDENTIAL_LIBRARY_VAULT_GENERIC,
```
randallmorey, line 24: We still need to get the user-selected type from the query param, don't we?  Generic is the default if type is not provided.

randallmorey (changes requested):

ui/admin/app/routes/scopes/scope/credential-stores/credential-store/credential-libraries/new.js
ZedLi: I was going to add in that functionality in the next PR but I can add it here

calcaide (approved):
LGTM!! Awesome work!!

addons/core/translations/resources/en-us.yaml
```diff
@@ -459,6 +459,9 @@ credential-library:
   types:
     vault-generic: Generic Secrets
     vault-ssh-cert: SSH Certificate Credentials
+  help:
+    vault-generic: Authorize sessions using credentials brokered by Vault
```
randallmorey, line 463: Can we describe about the type without talking about authorization or sessions?

randallmorey (changes requested):

addons/core/translations/resources/en-us.yaml
ZedLi: Do you have any suggestions? I'm not sure what else it would say

DhariniJeeva (approved):

addons/core/translations/resources/en-us.yaml
ZedLi: I'm going to defer any rewording for later once I can get some more input from others as well and merge it for now

PR #1598, opened 2023-01-20
refactor: 💡 Move save back to route to unblock other work

ZedLi:
Move save functionality back to route

### DIFF
--- a/ui/admin/app/templates/scopes/scope/targets/new.hbs
+++ b/ui/admin/app/templates/scopes/scope/targets/new.hbs
@@ -17,7 +17,7 @@
   <page.body>
     <Form::Target
       @model={{@model}}
-      @submit={{route-action 'save' @model}}
+      @submit={{route-action 'saveWithAddress' @model}}
       @cancel={{route-action 'cancel' @model}}
       @changeType={{route-action 'changeType'}}
     />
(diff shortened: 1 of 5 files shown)


ui/admin/app/templates/scopes/scope/targets/new.hbs
```diff
@@ -17,7 +17,7 @@
   <page.body>
     <Form::Target
       @model={{@model}}
-      @submit={{route-action 'save' @model}}
+      @submit={{route-action 'saveWithAddress' @model}}
```
randallmorey, line 20: A new target cannot have host sources, since it must first be saved before they can be added.
```suggestion
      @submit={{route-action 'save' @model}}
```

randallmorey (changes requested):

cameronperera (approved):
LGTM

DhariniJeeva:
so I'm seeing two error notification since we have notifyError in both save and saveAddress

DhariniJeeva:
I see two error notifications when I integrate this with my filters work. I tried removing   `@notifyError(({ message }) => message)` from `saveWithAddress` and was still getting the error notification for host source function failure. So should we remove the error decorator from saveWithAddress?

DhariniJeeva (approved):

PR #1606, opened 2023-01-25
refactor: 💡 fix styling for ordered series diagram

dvonanderson:
## Description
Added wrapper and min-height to `ordered-series-diagram` to ensure when toggles are clicked we don't get weird height behavior.

Screen Shot 2023-01-25 at 13 30 59

### DIFF
--- a/ui/admin/app/styles/app.scss
+++ b/ui/admin/app/styles/app.scss
@@ -619,7 +619,7 @@
 
       &-title {
         font-weight: var(--token-typography-font-weight-semibold);
-        margin-top: 0.5rem;
+        margin-top: sizing.rems(xs);
         width: 4rem;
       }
 
@@ -634,8 +634,8 @@
       display: flex;
 
       &-content {
-        border-radius: 0.5rem;
-        padding: 0 0.75rem 0.75rem;
+        border-radius: sizing.rems(xs);
+        padding: 0 sizing.rems(s) sizing.rems(s);
       }
 
       &-items {
@@ -661,7 +661,7 @@
 
     &-separator-icon {
       color: var(--token-color-foreground-faint);
-      margin: 0.75rem 0.5rem;
+      margin: sizing.rems(s) sizing.rems(xs);
     }
   }
 }
@@ -698,5 +698,5 @@
 
 .target-worker-filter-explainer {
   min-height: 1rem;
-  margin: 1.125rem 0 1.5rem;
+  margin: 1.125rem 0 sizing.rems(l);
 }
--- a/ui/admin/app/styles/app.scss
+++ b/ui/admin/app/styles/app.scss
@@ -598,72 +598,71 @@
   }
 }
 
-.ordered-series-diagram-wrapper {
+.worker-diagram-wrapper {
   min-height: 9.375rem;
-  margin-top: sizing.rems(s);
+  margin-top: sizing.rems(l);
+}
+.ordered-series-diagram {
+  --group-title-line-height: 2.25rem;
 
-  .ordered-series-diagram {
-    --group-title-line-height: 2.25rem;
+  display: flex;
 
+  &-item {
+    display: flex;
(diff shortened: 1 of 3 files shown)


ui/admin/app/styles/app.scss
```diff
@@ -598,69 +598,71 @@
...
+        text-align: center;
       }
-    }
-  }
 
-  &-group {
-    display: flex;
+      &-title {
+        font-weight: var(--token-typography-font-weight-semibold);
+        margin-top: 0.5rem;
```
cameronperera, line 622: Should we be using `sizing.rems(xs)` in place of rem directly for these styles?
`_sizing.scss` should have some default sizes already defined.

cameronperera (changes requested):

ui/admin/app/styles/app.scss
dvonanderson: @cameronperera good catch

ui/admin/app/styles/app.scss
```diff
@@ -598,69 +598,71 @@
...
+.ordered-series-diagram-wrapper {
+  min-height: 150px;
 
-  width: 100%;
-  display: flex;
-  margin-bottom: 1rem;
-
-  &-item {
+  .ordered-series-diagram {
+    --group-title-line-height: 2.25rem;
```
cameronperera, line 605:
```suggestion
    --group-title-line-height: sizing.rems(xl);
```

ui/admin/app/styles/app.scss
```diff
@@ -598,69 +598,71 @@
...
       }
-    }
-  }
 
-  &-group {
-    display: flex;
+      &-title {
+        font-weight: var(--token-typography-font-weight-semibold);
+        margin-top: sizing.rems(xs);
+        width: 4rem;
```
cameronperera, line 623:
```suggestion
        width: sizing.rems(xxl) + sizing.rems(m);
```

ui/admin/app/styles/app.scss
```diff
@@ -696,5 +698,5 @@
 
 .target-worker-filter-explainer {
   min-height: 1rem;
```
cameronperera, line 700:
```suggestion
  min-height: sizing.rems(m);
```

ui/admin/app/styles/app.scss
```diff
@@ -696,5 +698,5 @@
...
-  margin: 1.125rem 0 1.5rem;
+  margin: 1.125rem 0 sizing.rems(l);
```
cameronperera, line 701:
```suggestion
  margin: (sizing.rems(m) + sizing.rems(xxxs)) 0 sizing.rems(l); // 18 0 24
```

ui/admin/app/styles/app.scss
```diff
@@ -598,69 +598,71 @@
   }
 }
 
-.ordered-series-diagram {
-  --group-title-line-height: 2.25rem;
+.ordered-series-diagram-wrapper {
+  min-height: 150px;
```
cameronperera, line 602: For these larges ones I think we make exceptions? @randallmorey is that the general rule?

cameronperera (changes requested):

ui/admin/app/styles/app.scss
DhariniJeeva: I think we can keep all units the same and convert this to rem, 9 or 10 rem could work here

dvonanderson: @DhariniJeeva totally agree with you.  There are a few instances in which we use container height and width as px but each fix.

randallmorey (changes requested):
For the min-height fix, I was imagining this going into the target form itself for now.  It needs to contain the migration alert and the diagram description too.  Not sure we know enough about how the order series diagram will be used or evolve to make this kind of change in the component.

PR #1608, opened 2023-01-25
feat: 🎸 Add vault ssh cert type for credential library

ZedLi:
:tickets: [Jira ticket](https://hashicorp.atlassian.net/browse/ICU-7678)

## Description
Add support for the new vault ssh cert form in credential libraries.

:technologist: [Admin preview](https://boundary-ui-git-icu-7678-add-new-cred-library-type-hashicorp.vercel.app/scopes/s_1sheljmjgm/credential-stores/cs_vdcb6n0nu8/credential-libraries/new?type=vault-ssh-cert)

image


ui/admin/app/components/mapping-list/index.hbs
```diff
@@ -0,0 +1,64 @@
...
+            disabled={{@disabled}}
+            {{on 'input' (set-from-event option 'key')}}
+          />
+          <Hds::Form::TextInput::Field
+            @value={{option.value}}
+            @type='text'
+            disabled={{@disabled}}
+            {{on 'input' (set-from-event option 'value')}}
+          />
+          <div>
```
ZedLi, line 24: I needed these wrapper divs around the buttons to provide a container for whitespace so I could left align the different sized buttons.

Happy to hear of a better alternative to accomplish this.

ui/admin/app/styles/app.scss
```diff
@@ -698,3 +698,23 @@
...
+  width: 100%;
+
+  li {
+    padding: sizing.rems(xs) 0;
+    display: grid;
+    grid-template-columns: 1fr 1fr 5%;
+    align-items: center;
+    gap: sizing.rems(l);
+
+    > div {
```
randallmorey, line 712: Can this strictly use class names rather than elements?  It's a bit more flexible.

randallmorey (changes requested):

randallmorey (changes requested):
Looks like there are borders immediately below the text fields, and the "add" button juts out beyond the border.

Screenshot 2023-01-25 at 16 23 13

ui/admin/app/components/form/credential-library/vault-ssh-cert/index.js
```diff
@@ -1,4 +1,69 @@
...
+  @action
+  shouldDisplayKeyBits({ target: { value } }) {
+    if (value === 'ed25519') {
+      this.showKeyBits = false;
+      this.args.model.key_type = null;
+    } else {
+      this.showKeyBits = true;
+      this.args.mode.key_type = value;
+    }
+  }
```
randallmorey, line 42: There are a couple things going on here that could be separated:

- Set `key_type` on the model independently of any display logic.
- Convert `showKeyBits` to a getter that is based on the value of `key_type`.

Then `shouldDisplayKeyBits` can be dropped.

randallmorey (changes requested):

ui/admin/app/components/form/credential-library/vault-ssh-cert/index.js
```diff
@@ -1,4 +1,69 @@
...
+
+  /**
+   * Curried function to add the option. We recreate a new array after adding
```
randallmorey, line 45:
```suggestion
   * Adds an key/value option object. We recreate a new array after adding
```

ui/admin/app/components/form/credential-library/vault-ssh-cert/index.js
```diff
@@ -1,4 +1,69 @@
...
+   */
+  @action
+  addOption(field, { key, value }) {
+    const existingArray = this.args.model[field] ?? [];
+    const newArray = [...existingArray, { key, value }];
+    set(this.args.model, field, newArray);
+  }
+
+  /**
+   * Curried function to remove the option by index. We recreate a new array after
```
randallmorey, line 59:
```suggestion
   * Removes an option by index. We recreate a new array after
```

randallmorey (changes requested):
Edited some language in the comments.  The functions don't need to know how they're being used (and currying implies a single argument).

ui/admin/app/components/form/credential-library/vault-ssh-cert/index.js
ZedLi: Good call, updated it

ui/admin/app/styles/app.scss
ZedLi: Updated to use class names, the div one I changed to `*` to make it clear it's to clear any existing bottom margin (mainly from HDS)

ZedLi:
> Looks like there are borders immediately below the text fields, and the "add" button juts out beyond the border.

Good catch, I totally didn't see that. I updated the styles for it

ui/admin/app/components/mapping-list/index.hbs
randallmorey: Can this have a class name?

ui/admin/app/styles/app.scss
```diff
@@ -698,3 +698,23 @@
...
+    gap: sizing.rems(l);
+
+    > * {
+      margin: 0;
+    }
+  }
+
+  > :not(:last-child) {
+    border-bottom: 1px solid var(--token-color-border-primary);
+  }
```
randallmorey, line 719: Simplification of the border rule / positive framing:

```suggestion
  &-item {
    border-bottom: 1px solid var(--token-color-border-primary);
    padding: sizing.rems(xs) 0;
    display: grid;
    grid-template-columns: 1fr 1fr 5%;
    align-items: center;
    gap: sizing.rems(l);

    > * {
      margin: 0;
    }
  }

  &:last-child {
    border-bottom: 0;
  }
```

randallmorey (changes requested):

ui/admin/app/components/mapping-list/index.hbs
ZedLi: Sure, but even if there's no associated styles with the class?

ui/admin/app/styles/app.scss
ZedLi: I updated it but I personally found the previous style a bit cleaner/easier to read for me since it doesn't declare the border to set it back to 0 immediately and didn't have the border declaration in two places.

Curious what makes this style easier for you to understand

addons/rose/package.json
```diff
@@ -34,7 +34,7 @@
   },
   "dependencies": {
     "@embroider/test-setup": "^0.48.1",
-    "@hashicorp/design-system-components": "^1.3.2",
+    "@hashicorp/design-system-components": "^1.5.1",
```
calcaide, line 37: Thanks for this!!

ui/admin/app/components/form/credential-library/vault-ssh-cert/index.js
```diff
@@ -1,4 +1,55 @@
...
+
+  /**
+   * Removes an option by index. We recreate a new array after
+   * splicing out the item so that ember is aware that the array has been modified.
+   * @param field {string}
+   * @param index {number}
+   */
+  @action
+  removeOptionByIndex(field, index) {
+    const newArray = this.args.model[field].filter((_, i) => i !== index);
```
calcaide, line 52: Clean!

ui/admin/app/components/mapping-list/index.js
```diff
@@ -0,0 +1,33 @@
...
+import { action } from '@ember/object';
+
+export default class MappingListComponent extends Component {
+  @tracked
+  newOptionKey;
+  @tracked
+  newOptionValue;
+
+  @action
+  updateArray() {}
```
DhariniJeeva, line 12: do we need this?

ui/admin/app/components/mapping-list/index.hbs
```diff
@@ -0,0 +1,90 @@
...
+        </H.Tr>
+      </:head>
+      <:body as |B|>
+        {{#each @options as |option index|}}
+          <B.Tr>
+            <B.Td>
+              <Hds::Form::TextInput::Field
+                @value={{option.key}}
+                @type='text'
+                disabled={{@disabled}}
```
calcaide, line 24: Non blocking comment, asking from curiosity, wher this `@disabled` come from? tracked prop?

ui/admin/app/components/mapping-list/index.js
ZedLi: Woops good catch, that was a leftover action that I forgot to remove.

ui/admin/app/styles/app.scss
```diff
@@ -698,3 +698,35 @@
   min-height: 1rem;
   margin: 1.125rem 0 1.5rem;
 }
+
+.mapping-table {
+  /* stylelint-disable selector-class-pattern */
+  td > div,
```
calcaide, line 704: Non blocking comment: I am fine disabling this rule, but asking for curiosity, it has the `td > div, .hds-form-field__control:not(:first-child)` selector that cause the linter to complain?

ui/admin/app/components/mapping-list/index.hbs
ZedLi: It's getting passed in by the consumer, usually from the form's determination of `disabled`

ui/admin/app/styles/app.scss
ZedLi: It's the usage of `__` by HDS classes. I had to use some of the HDS classes as otherwise I couldn't quite get the right CSS specificity to override their styles correctly.

Open to other suggestions.

ui/admin/app/routes/scopes/scope/credential-stores/credential-store/credential-libraries/credential-library.js
```diff
@@ -19,4 +21,27 @@ export default class ScopesScopeCredentialStoresCredentialStoreCredentialLibrari
...
+   *
+   * The deep copy implemented here is required to ensure that both the
+   * array itself and its members are all new.
+   *
+   * @param {credentialLibraryModel} credentialLibrary
+   */
+  @action
+  edit(credentialLibrary) {
+    if (credentialLibrary.critical_options) {
+      credentialLibrary.critical_options = copy(
```
calcaide, line 38: Non-blocking comment and 100% personal opinion, not a requirement, just a discussion for the sake of learning.

I wonder if the usage of `ember-copy` brings some magic behind the scenes or if in the future we can rely on `array.from()` [docs here](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/from) (or similar) to shallow copy arrays and ember-data mark them as dirty state.

ui/admin/app/styles/app.scss
calcaide: Ahh I see, I was wondering what was the cause, now we know is HDS classes and eventually we might need to do this again, just to be aware team wise 😉

calcaide (approved):
LGTM!! Awesome work 💪

ui/admin/app/routes/scopes/scope/credential-stores/credential-store/credential-libraries/credential-library.js
ZedLi: Unfortunately that only makes a shallow copy, we specifically use the `copy` with the deep copy parameter as `true` for deep copies.

addons/rose/package.json
ZedLi: Yup I had to update this for the new HDS table component!

ui/admin/app/components/form/credential-library/vault-ssh-cert/index.hbs
```diff
@@ -1,2 +1,287 @@
...
+    disabled={{form.disabled}}
+    {{on 'input' (set-from-event @model 'key_type')}}
+    as |F|
+  >
+    <F.Label>{{t 'resources.credential-library.form.key_type.label'}}</F.Label>
+    <F.HelperText>{{t
+        'resources.credential-library.form.key_type.help'
+      }}</F.HelperText>
+    <F.Options>
+      <option value=''></option>
```
randallmorey, line 142: What's the purpose of this item?  Should this at least have visible text?

ui/admin/app/styles/app.scss
randallmorey: Can we use class names rather than anonymous html elements for clarity?

randallmorey (changes requested):
Just a couple points of feedback.  For the styles, can we use class names rather than HTML elements?  It's more flexible in case we want to change the elements we use.

ui/admin/app/styles/app.scss
DhariniJeeva: if these table style changes are not just specific to credential library resource and applicable to boundary UI overall, we could move these to `override.scss` where we override HDS styles. We did a similar thing for fieldsets recently.

ui/admin/app/components/form/credential-library/vault-ssh-cert/index.hbs
ZedLi: It's to show an empty option so a user actually has to select it. 

There's a somewhat problematic UX where without the option, the model itself doesn't have a value but the select just displays the first option and it doesn't get saved but it _looks_ like it did even if you viewed the form again. 

I would have to set a default option on the model to get around this since we don't set anything until the user triggers the `onInput` which I'm open to, but I thought this might be better since a user has to make a conscious decision. 

I'm not sure if it's a common pattern/UX but the code examples in [HDS](https://helios.hashicorp.design/components/form/select?tab=code) show the usage of them.

ui/admin/app/styles/app.scss
ZedLi: Hmm they're not for boundary overall, just the usage of their table in this instance, I'll try using class names though; I think I had some difficulty with CSS specificity but that wasn't with my own classes.

addons/core/translations/resources/en-us.yaml
```diff
@@ -464,6 +464,27 @@ credential-library:
     http_request_body:
       label: HTTP Method POST Request Body
       help: The body of the HTTP request the library sends to Vault when requesting credentials. Only allowed when HTTP method is set to POST.
+    username:
+      label: Username
+      help: This is the username that will be used to connect to Vault.
```
AdamBouhmad, line 469: I would change this to be inline with the rfc: `The username to use when making an SSH connection. This will be included in the valid_principles when making the request to Vault.`

ui/admin/app/styles/app.scss
ZedLi: I wasn't able to override correctly the HDS styles using my own classes still (without using !important), not sure if you have suggestions for this.

ui/admin/app/components/form/credential-library/vault-ssh-cert/index.hbs
randallmorey: Should it say "Choose a type" or something?

randallmorey (changes requested):

ui/admin/app/components/form/credential-library/vault-ssh-cert/index.hbs
ZedLi: Yup that makes sense, I'll update it

ui/admin/app/components/form/field/key-value-list/index.hbs
```diff
@@ -0,0 +1,98 @@
+<Hds::Form::Fieldset
+  @layout={{@layout}}
+  @isRequired={{@isRequired}}
+  @isOptional={{@isOptional}}
+  as |F|
+>
+  {{yield F}}
+  <F.Control>
+    <Hds::Table class='mapping-table' name={{@name}}>
```
randallmorey, line 9:
```suggestion
    <Hds::Table class='key-value-list' name={{@name}}>
```

ui/admin/app/routes/scopes/scope/credential-stores/credential-store/credential-libraries/credential-library.js
```diff
@@ -19,4 +20,28 @@ export default class ScopesScopeCredentialStoresCredentialStoreCredentialLibrari
...
+    if (credentialLibrary.critical_options) {
+      credentialLibrary.critical_options = structuredClone(
+        credentialLibrary.critical_options
+      );
+    }
+    if (credentialLibrary.extensions) {
+      credentialLibrary.extensions = structuredClone(
+        credentialLibrary.extensions
+      );
+    }
```
randallmorey, line 45: `structuredClone` works with `undefined` and `null`, so it should be safe to do unguarded.
```suggestion
    const { critical_options, extensions } = credentialLibrary;
    credentialLibrary.critical_options = structuredClone(critical_options);
    credentialLibrary.extensions = structuredClone(extensions);
```

randallmorey (changes requested):

addons/core/translations/resources/en-us.yaml
```diff
@@ -474,6 +474,7 @@ credential-library:
     key_type:
       label: Key Type
       help: Specifies the desired key type to use when generating a private key.
+      initial_option: Choose a type
```
randallmorey, line 477: This seems like it would have wide applicability.

randallmorey (changes requested):

addons/core/translations/resources/en-us.yaml
ZedLi: Moved to the more generic forms folder instead

DhariniJeeva (approved):

randallmorey (approved):

PR #1611, opened 2023-01-30
Refactor mapping list

ghost:
Simplification of mapping list CSS and component code, with added comments for clarity.

### DIFF
--- a/ui/admin/app/components/mapping-list/index.hbs
+++ b/ui/admin/app/components/mapping-list/index.hbs
@@ -41,7 +41,7 @@
                 {{on 'input' (set-from-event option 'value')}}
                 as |F|
               >
-                <F.Label classes='sr-only'>{{t 'form.value.label'}}</F.Label>
+                <F.Label class='sr-only'>{{t 'form.value.label'}}</F.Label>
               </Hds::Form::TextInput::Field>
             </B.Td>
             <B.Td>


ui/admin/app/components/mapping-list/index.hbs
```diff
@@ -36,20 +38,21 @@
                 {{on 'input' (set-from-event option 'value')}}
                 as |F|
               >
-                <F.Label>{{t 'form.value.label'}}</F.Label>
+                <F.Label classes='sr-only'>{{t 'form.value.label'}}</F.Label>
```
randallmorey, line 41:
```suggestion
                <F.Label class='sr-only'>{{t 'form.value.label'}}</F.Label>
```

ui/admin/app/components/mapping-list/index.hbs
```diff
@@ -36,20 +38,21 @@
...
-                @text={{t 'actions.remove'}}
-                @color='critical'
-                @icon='trash'
-                @isIconOnly={{true}}
-                disabled={{@disabled}}
-                {{on 'click' (fn this.removeOptionByIndex index)}}
-              />
-
-            </B.Td>
+            {{#if @removeOptionByIndex}}
```
ZedLi, line 44: Should the display of these cells be controlled by whether we have the function? The reason why I had the indirect function call was to guard against not including the function but I think it should still be required.

Wouldn't this also not work with the `last-child` css that shrinks the last cell?

ZedLi (changes requested):

ui/admin/app/components/mapping-list/index.js
```diff
@@ -1,16 +1,35 @@
...
+   */
   @action
   addOption() {
-    if (this.args.addOption) {
+    assert(
+      '[boundary-admin-key-value-list-field] `@addOption` is required.',
+      this.args.addOption
+    );
+
+    if (this.newOptionKey) {
```
ZedLi, line 32: Do we want to always require a key to be able to add? If so, shouldn't we be disabling the add button instead?

ui/admin/app/styles/app.scss
```diff
@@ -710,33 +710,61 @@
...
-    display: none;
-  }
+      &:last-child {
+        // ...except in the last cell, which doesn't need right padding
+        padding-right: 0;
 
-  border: 0;
+        // Shrink last cell to minimum possible width
+        width: 1%;
+        white-space: nowrap;
```
ZedLi, line 743: What's the purpose of adding `nowrap` here, shouldn't we always be expecting a button of some sort?

ui/admin/app/components/mapping-list/index.hbs
randallmorey: Good point.

ui/admin/app/styles/app.scss
randallmorey: In case the button has multiple words in it.

ui/admin/app/components/mapping-list/index.hbs
ZedLi: Good to know, wasn't aware of assert usage

ZedLi (approved):
LGTM

PR #1614, opened 2023-02-01
feat: 🎸 Enable feature flag for vault ssh cert work

ZedLi:
## Description
Enable the feature for enterprise and HCP only.


cameronperera (approved):
Should line 18 be deleted from the `oss` edition flags. If it is not present, it will default to `false`. I do see the benefit of leaving it to clarify that it should be `false` for OSS though.

PR #1622, opened 2023-02-03
fix: 🐛 Fix undefined values from resource filter

ZedLi:
## Description
There can be a bug in workers when you have an active filter on a worker that is deleted and it is the only worker with the active tag being filtered. When we return to the workers route, the `resource-filter` tries to find that tag from the list of tags from the workers but it doesn't find it anymore. 

This aims to fix it by filtering out the `undefined` tags that result from not finding the resources.

### DIFF
--- a/addons/api/addon/services/resource-filter-store.js
+++ b/addons/api/addon/services/resource-filter-store.js
@@ -61,9 +61,7 @@ export class ResourceFilter {
    * @return {string}
    */
   equals(key, value) {
-    return value !== null && value !== undefined
-      ? `"/item/${key}" == "${value}"`
-      : null;
+    return value !== null ? `"/item/${key}" == "${value}"` : null;
   }
 
   /**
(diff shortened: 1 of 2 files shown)


cameronperera (approved):

addons/api/addon/services/resource-filter-store.js
```diff
@@ -61,7 +61,9 @@ export class ResourceFilter {
    * @return {string}
    */
   equals(key, value) {
-    return value !== null ? `"/item/${key}" == "${value}"` : null;
+    return value !== null && value !== undefined
```
ZedLi, line 64: This had to be updated because the change to the resource filter can return `undefined` if `deserializedValue` is null due to the optional chaining... 🤯 

Side note: how do we feel about doing `value != null` to check for both null and undefined, I assumed we don't like that so was explicit.

DhariniJeeva (approved):

PR #1623, opened 2023-02-03
fix: 🐛 Fix null date time erroring out

ZedLi:
## Description
This fixes when a worker doesn't properly connect and you try to view the details page for the worker and it errors out due to not having a date time.

### DIFF
--- a/addons/core/addon/helpers/relative-datetime-live.js
+++ b/addons/core/addon/helpers/relative-datetime-live.js
@@ -45,6 +45,10 @@ export default class extends Helper {
   // =methods
 
   compute([date]) {
+    if (!date) {
+      return;
+    }
+
     const delta = date.valueOf() - this.clockTick.now;
     const greatestMeaningfulUnit = unitMappings.reduce(
       (currentValue, previousValue) => {
--- a/addons/core/addon/helpers/relative-datetime-live.js
+++ b/addons/core/addon/helpers/relative-datetime-live.js
@@ -45,9 +45,7 @@ export default class extends Helper {
   // =methods
 
   compute([date]) {
-    if (!date) {
-      return;
-    }
+    if (!date) return;
 
     const delta = date.valueOf() - this.clockTick.now;
     const greatestMeaningfulUnit = unitMappings.reduce(
--- a/ui/admin/app/components/form/worker/index.hbs
+++ b/ui/admin/app/components/form/worker/index.hbs
@@ -99,7 +99,10 @@
     <form.input
       @name='last seen'
       @type='datetime'
-      @value={{concat relativeDate ', ' customFormat}}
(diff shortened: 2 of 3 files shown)


addons/core/addon/helpers/relative-datetime-live.js
```diff
@@ -45,6 +45,10 @@ export default class extends Helper {
   // =methods
 
   compute([date]) {
+    if (!date) {
+      return;
+    }
```
randallmorey, line 50:
```suggestion
    if (!date) return;
```

ui/admin/app/components/form/worker/index.hbs
```diff
@@ -99,7 +99,10 @@
     <form.input
       @name='last seen'
       @type='datetime'
-      @value={{concat relativeDate ', ' customFormat}}
+      @value={{if
+        @model.last_status_time
+        (concat relativeDate ', ' customFormat)
+      }}
```
randallmorey, line 105: This display value seems like a good candidate for a getter in the form js.

randallmorey (changes requested):

addons/core/addon/helpers/relative-datetime-live.js
calcaide: @randallmorey 
I have no strong opinion about 1 liner conditionals (I think is opinonated). 
But if this is what we want, I think could be benefitial for us as a team discuss it and enforced via prettier?

ui/admin/app/components/form/worker/index.hbs
ZedLi: We use a few template helpers for this, should I be calling them directly in the js file?

addons/core/addon/helpers/relative-datetime-live.js
ZedLi: I personally dislike them but I see that we've used them elsewhere so I've updated it for now. 

Totally agree on discussing it and enforcing it though

ui/admin/app/components/form/worker/index.hbs
randallmorey: It would simplify the template a bit.  But probably not necessary right now.

randallmorey (approved):

cameronperera (approved):

addons/core/addon/helpers/relative-datetime-live.js
calcaide: I personally don't like them either, but since it's a personal opinion I think is fair to bring to FE tech time, discuss it as team, take a decision (which I will be totally fin with) and let the tooling enforcing it for us 😃

calcaide (approved):
LGTM!! Good work!

PR #1630, opened 2023-02-10
chore: 🤖 create enable and disable feature test helpers

cameronperera:
## Description

With shifting to having the test ENV features controlled by each test module, I created the `enableFeature` and `disableFeature` test helpers. 

For now I only updated one module of each `acceptance`, `integration`, and `unit` tests. Happy to update more in this PR or a later one.

### DIFF
--- a/ui/admin/tests/acceptance/targets/create-test.js
+++ b/ui/admin/tests/acceptance/targets/create-test.js
@@ -1,12 +1,5 @@
 import { module, test } from 'qunit';
-import {
-  visit,
-  currentURL,
-  find,
-  click,
-  fillIn,
-  getContext,
-} from '@ember/test-helpers';
+import { visit, currentURL, find, click, fillIn } from '@ember/test-helpers';
 import { setupApplicationTest } from 'ember-qunit';
 import setupMirage from 'ember-cli-mirage/test-support/setup-mirage';
 import { Response } from 'miragejs';
@@ -17,6 +10,7 @@ import {
   //invalidateSession,
 } from 'ember-simple-auth/test-support';
 import { TYPE_TARGET_TCP, TYPE_TARGET_SSH } from 'api/models/target';
+import { disableFeature, enableFeature } from '../../helpers/features-service';
 
 module('Acceptance | targets | create', function (hooks) {
   setupApplicationTest(hooks);
@@ -25,7 +19,6 @@ module('Acceptance | targets | create', function (hooks) {
   let getTargetCount;
   let getTCPTargetCount;
   let getSSHTargetCount;
-  let featuresService;
 
   const instances = {
     scopes: {
@@ -46,8 +39,6 @@ module('Acceptance | targets | create', function (hooks) {
   };
 
   hooks.beforeEach(function () {
-    const { owner } = getContext();
-    featuresService = owner.lookup('service:features');
     instances.scopes.global = this.server.create('scope', { id: 'global' });
     instances.scopes.org = this.server.create('scope', {
       type: 'org',
@@ -85,7 +76,7 @@ module('Acceptance | targets | create', function (hooks) {
   });
 
   test('can create type `ssh` when `target-worker-filters-v2` is disabled', async function (assert) {
-    featuresService.disable('target-worker-filters-v2');
+    disableFeature('target-worker-filters-v2');
     assert.expect(4);
     const targetCount = getTargetCount();
     const sshTargetCount = getSSHTargetCount();
@@ -111,7 +102,7 @@ module('Acceptance | targets | create', function (hooks) {
   });
 
   test('can create type `ssh` when `target-worker-filters-v2` is enabled', async function (assert) {
-    featuresService.enable('target-worker-filters-v2');
+    enableFeature('target-worker-filters-v2');
     assert.expect(4);
     const targetCount = getTargetCount();
     const sshTargetCount = getSSHTargetCount();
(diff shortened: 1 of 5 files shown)


dvonanderson (approved):
very nice

ui/admin/tests/helpers/features-service.js
```diff
@@ -0,0 +1,13 @@
+import { getContext } from '@ember/test-helpers';
+
+export function enableFeature(flag) {
```
ZedLi, line 3: Do we need to re-create this function? It technically exists currently in the [addon](https://github.com/kategengler/ember-feature-flags/blob/v6.0.0/addon-test-support/index.js)

calcaide: @ZedLi I assume, @cameronperera correct me if I am wrong, adding this test helper will deprecate the `ember-feature-flags`? So we can stop using such package?

ui/admin/tests/integration/components/worker-diagram/index-test.js
```diff
@@ -3,39 +3,49 @@ import { setupRenderingTest } from 'admin/tests/helpers';
...
     await render(hbs`<WorkerDiagram />`);
 
     assert.dom('[data-test-dual-filter-egress-off-ingress-off]').isVisible();
   });
 
   test('it renders a HCP dual filter diagram when `target-worker-filters-v2-hcp` is enabled', async function (assert) {
-    featuresService = this.owner.lookup('service:features');
-    featuresService.enable(targetWorkerFilterIngress);
-    featuresService.enable(targetWorkerFilterHCP);
+    // featuresService = this.owner.lookup('service:features');
```
calcaide, line 44: Could we delete the commented old code?
My understanding is that we are leaving this comments as a backup if something fails? We can always look up git history 😃

ui/admin/tests/acceptance/targets/create-test.js
```diff
@@ -111,7 +102,7 @@ module('Acceptance | targets | create', function (hooks) {
   });
 
   test('can create type `ssh` when `target-worker-filters-v2` is enabled', async function (assert) {
-    featuresService.enable('target-worker-filters-v2');
+    enableFeature('target-worker-filters-v2');
```
calcaide, line 105: None blocking comment: I am wondering about the context of enableFeature.

Enabling a specific feature in a test , will be persisted for the next test case? or each test case starts with a default fresh state of feature flags?

In other words, If we enable `target-worker-filters-v2` in this test, in the next test within the same describe block, will `target-worker-filters-v2` be enabled or disable?

calcaide (approved):
C-L-E-A-N!! Nice work @cameronperera, I left a couple of none blocking comments 😉

ui/admin/tests/helpers/features-service.js
ZedLi: No, we'd still need the addon to handle the feature flag service functionality, this is just for a test helper that the addon hasn't released yet

ui/admin/tests/integration/components/worker-diagram/index-test.js
cameronperera: Ooops. I thought I got them all. I will remove these before merging.

calcaide: lol being there done that :sweat_smile:

ui/admin/tests/acceptance/targets/create-test.js
cameronperera: I have not run into issues with enabling a feature in one test case and it still being enabled in another test case, even within the same test module. I think if this was an issue, `yarn test` would likely never be successful as it randomizes our tests to ensure stuff like this doesn't happen.

ui/admin/tests/helpers/features-service.js
cameronperera: @ZedLi is correct, this does not remove the need for the addon. This just helps with toggling flags in our tests.

To your original question, my understanding is we had moved to just enabling and disabling feature flags ourselves instead of having a mix of the `enableFeature('flag')` from the addon and `featuresService.disable('flag')`. I don't think we are even using the `enableFeature` helper from the addon. From what I can tell we only use our current method. Happy to remove this `enableFeature` test-helper if the team prefers that route.

calcaide: Gotcha! Understood, thanks for clarification folks 😉

calcaide: As per @ZedLi comment and seeing Cameron answer I think Cameron is right, we aren't using `enableFeature` from the addon. I do not have a strong opinion but I would leave it becasue seems consistent to me our service has `enableFleature` and `disableFeature`. But as I said, no strong opinion 😉

ui/admin/tests/acceptance/targets/create-test.js
calcaide: @cameronperera you are right, if sharing scope out of the `test()` function our test suite would fail. But I was wondering if we explicitly did the sanity check with this 👍

ui/admin/tests/helpers/features-service.js
ZedLi: It's not a big deal since it's such a simple function, but I lean towards using the existing functionality since it's the exact same function and should have been provided by the addon for this service.

I think we didn't use the `enableFeature` initially because we didn't re-create the helper for the `disableFeature` so it wasn't symmetrical, but I think it makes sense for us to include the helper for `disableFeature` and maybe add a comment why we made this helper and not the `enableFeature` for context.

cameronperera:
No longer needed as the team is moving in a different direction for feature flags and tests.

PR #1634, opened 2023-02-15
chore: use structuredClone instead of ember-copy addon

DhariniJeeva:
:tickets: [Jira ticket](https://hashicorp.atlassian.net/browse/ICU-7153)

this pr removes ember-copy and uses `structuredClone` instead.


cameronperera (approved):
LGTM

calcaide (approved):
LGTM! Nice work!!

ZedLi (approved):
I think this is great but I just want to make sure we're aware that I don't think this is an exact replacement for `ember-copy` as [structuredClone](https://developer.mozilla.org/en-US/docs/Web/API/Web_Workers_API/Structured_clone_algorithm) doesn't work on some things (namely functions) which I believe some ember specific objects have (and I encountered errors before when trying to copy) e.g. ember models, ember arrays, etc.

DhariniJeeva:
after our tech time discussion, we will move forward with removing this addon for now, in the future if we come across cases as Zed mentioned we will revisit adding this addon.

PR #1635, opened 2023-02-17
test: 💍 Swap playwright to @playwright/test package

calcaide:

## Description

Playwright documentation recommends to use @playwright/test now as the only playwright package.

**Why this change?**

@moduli is working on adding e2e to Admin UI. While working on the PoC we realize desktop client was using `playwright` yarn package and Admin UI `@playwright/test`, which caused name collisions within the monorepo.

After digging in Palywright documentation and changelog, seems like the recommended way to go is to use `@playwright/test`. Change that on the Desktop app and now both e2e test can be run without name collision.

Tested on @moduli  branch `moduli-e2e-tests-ICU-7418` succesfully 😉

### DIFF
--- a/ui/desktop/package.json
+++ b/ui/desktop/package.json
@@ -62,7 +62,7 @@
     "@faker-js/faker": "^7.3.0",
     "@glimmer/component": "^1.1.2",
     "@glimmer/tracking": "^1.1.2",
-    "@playwright/test": "^1.15.0",
+    "@playwright/test": "^1.30.0",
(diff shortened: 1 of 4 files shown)


moduli (approved):
Started a new branch against this one and verified that it works.

ZedLi (approved):

lisbet-alvarez (approved):

DhariniJeeva (approved):

PR #1637, opened 2023-02-21
test(admin/e2e): Add first playwright test

moduli:

## Description

This PR introduces [Playwright](https://playwright.dev/) tests to the Boundary Admin UI project. There will be a series of PRs created to make reviewing simpler. This is the first in the series.

This PR contains the following...
- Adds Playwright to `ui/admin`
- Adds a test that uses the Admin UI to create the necessary resources to connect to a target. The test then uses the CLI to start a session and checks in the Admin UI that a session is created

Note: There will most likely be some refactoring/reorganizing as we learn more about how this test suite will get used.

https://user-images.githubusercontent.com/2474253/220419143-825549b0-f795-49a2-9bbf-b71ccbc865d5.mov

### DIFF
--- a/ui/admin/tests/e2e/README.md
+++ b/ui/admin/tests/e2e/README.md
@@ -3,8 +3,7 @@
 This test suite tests the Boundary Admin UI in an end-to-end setting. It was designed to be run in a
 variety of environments as long as the appropriate environment variables are set. You can choose to
 spin up your own infrastructure or use [Enos](https://github.com/hashicorp/boundary/tree/main/enos)
-to generate it for you. There is an enos scenario in the `boundary` repository that runs these
-tests.
+to generate it for you.
 
 The test suite uses the [Playwright](https://playwright.dev/) framework.
 
@@ -13,7 +12,13 @@ Setup...
 ```bash
 cd boundary-ui/ui/admin
 yarn install
-npx playwright install
+npx playwright install # this installs the browsers used by Playwright
+```
+
+Ensure the `boundary` CLI is available on the path. You can validate this by making sure this
+command returns something.
+```bash
+which boundary
 ```
 
 Run tests...
--- a/ui/admin/tests/e2e/README.md
+++ b/ui/admin/tests/e2e/README.md
@@ -25,7 +25,7 @@ Run tests...
 ```bash
 export BOUNDARY_ADDR=
 export E2E_PASSWORD_ADMIN_LOGIN_NAME=
-export E2E_PASSWORD_ADMIN_LOGIN_PASSWORD=
+export E2E_PASSWORD_ADMIN_PASSWORD=
 export E2E_PASSWORD_AUTH_METHOD_ID=
 ... # Tests may have additional variables that it needs
 cd boundary-ui/ui/admin
--- a/ui/admin/tests/e2e/README.md
+++ b/ui/admin/tests/e2e/README.md
@@ -1,0 +1,44 @@
+# Admin UI e2e tests
+
+This test suite tests the Boundary Admin UI in an end-to-end setting. It was designed to be run in a
+variety of environments as long as the appropriate environment variables are set. You can choose to
+spin up your own infrastructure or use [Enos](https://github.com/hashicorp/boundary/tree/main/enos)
+to generate it for you. There is an enos scenario in the `boundary` repository that runs these
+tests.
+
+The test suite uses the [Playwright](https://playwright.dev/) framework.
+
+## Getting Started
+Setup...
+```bash
+cd boundary-ui/ui/admin
+yarn install
+npx playwright install
+```
+
+Run tests...
+```bash
+export BOUNDARY_ADDR=
+export E2E_PASSWORD_ADMIN_LOGIN_NAME=
+export E2E_PASSWORD_ADMIN_LOGIN_PASSWORD=
+export E2E_PASSWORD_AUTH_METHOD_ID=
+... # Tests may have additional variables that it needs
+cd boundary-ui/ui/admin
+yarn run e2e
+```
+
+Here are some additional commands to assist with debugging.
+```bash
+PWDEBUG=console yarn playwright test --headed --config ./tests/e2e/playwright.config.js login.spec.js
+PWDEBUG=console yarn playwright test --headed --config ./tests/e2e/playwright.config.js login.spec.js:13 --debug
+PWDEBUG=console yarn playwright test --headed --config ./tests/e2e/playwright.config.js login.spec.js --debug
+```
+
+## Developing Tests
+
+It is recommended to use locators that resemble how users interact with the application. `getByRole`
+should be the most prioritized locator as this is the closest way to how users and
+accessibility features perceive the page. `getByLabel` should be used for form fields.
+
+Additional information can be found [here](https://playwright.dev/docs/locators#locating-elements)
+and [here](https://testing-library.com/docs/queries/about/#priority).
--- a/ui/admin/tests/e2e/global-setup.js
+++ b/ui/admin/tests/e2e/global-setup.js
@@ -1,6 +1,6 @@
 /* eslint-disable no-undef */
 const { chromium } = require('@playwright/test');
-const { checkEnv } = require('./helpers/general');
+const { checkEnv, authenticatedState } = require('./helpers/general');
 
 module.exports = async () => {
   await checkEnv([
@@ -23,8 +23,6 @@ module.exports = async () => {
   await page.getByRole('button', { name: 'Sign In' }).click();
   await page.getByRole('navigation', { name: 'General' }).waitFor();
   await page.getByText(process.env.E2E_PASSWORD_ADMIN_LOGIN_NAME).waitFor();
-  await page
-    .context()
-    .storageState({ path: './tests/e2e/artifacts/authenticated-state.json' });
+  await page.context().storageState({ path: authenticatedState });
   await browser.close();
 };
--- a/ui/admin/tests/e2e/global-setup.js
+++ b/ui/admin/tests/e2e/global-setup.js
@@ -1,0 +1,30 @@
+/* eslint-disable no-undef */
+const { chromium } = require('@playwright/test');
+const { checkEnv } = require('./helpers/general');
+
+module.exports = async () => {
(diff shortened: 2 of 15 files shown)


yarn.lock
```diff
@@ -141,27 +141,6 @@
     semver "^6.3.0"
     source-map "^0.5.0"
 
-"@babel/core@^7.14.8":
```
moduli, line 144: These changes to `yarn.lock` weren't originally here when I created the PR...

calcaide: @moduli Do not worry, these `yarn.lock` changes happend because we got a `.lock` file out of date, we will merge an specific fix for it once is done I let you know, then we update this branch and the change will be gone 😉

ui/admin/tests/e2e/global-setup.js
```diff
@@ -0,0 +1,30 @@
+/* eslint-disable no-undef */
```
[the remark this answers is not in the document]
calcaide: As per my understanding, eventually we will need, but I will say do not worry just now because in the UI repository we did not applied them to any files.

ui/admin/tests/e2e/global-setup.js
```diff
@@ -0,0 +1,30 @@
...
+const { chromium } = require('@playwright/test');
+const { checkEnv } = require('./helpers/general');
+
+module.exports = async () => {
+  await checkEnv([
```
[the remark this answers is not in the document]
calcaide: I think this is a very good place to start with, the file name is very descriptive and we are just starting with this. I suggest to leave them here and let's see when we encounter the first case with an specific ENV var how this works for us?

ui/admin/tests/e2e/tests/target.spec.js
```diff
@@ -0,0 +1,106 @@
...
+  createNewOrg,
+  createNewProject,
+  createNewHostCatalog,
+  createNewHostSet,
+  createNewHostInHostSet,
+  createNewTarget,
+  addHostSourceToTarget,
+} = require('../helpers/boundary-ui');
+
+test.use({ storageState: './tests/e2e/artifacts/authenticated-state.json' });
```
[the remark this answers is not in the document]
cameronperera: As a chronic misspell-er, a constant would be nice.

ui/admin/tests/e2e/global-setup.js
calcaide, line 6: I love this approach and that solves a problem we have in Desktop client e2e, so I will copy this 😉

ui/admin/tests/e2e/tests/target.spec.js
```diff
@@ -0,0 +1,106 @@
...
+    'E2E_TARGET_IP',
+    'E2E_SSH_USER',
+    'E2E_SSH_KEY_PATH',
+    'E2E_SSH_PORT',
+  ]);
+
+  await checkBoundaryCli();
+});
+
+test('Cancel a session', async ({ page }) => {
```
cameronperera, line 32:
```suggestion
test('Verify session created then cancel the session', async ({ page }) => {
```
Small suggestion. It seems this is really wanting to confirm that a session was created & we could cancel it.

ui/admin/tests/e2e/tests/target.spec.js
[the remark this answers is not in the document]
calcaide: +1 to that! nice suggestion

cameronperera (dismissed):
LGTM. I left a small suggestion but not blocking.

calcaide (approved):
LGTM!! Terrific job @moduli !! We will copy over `desktop/e2e` some of the patterns you brought here.
Thank you a lot for this!!

ui/admin/tests/e2e/global-setup.js
calcaide, line 1: **For visibility:** We shouldn't be ignoring the linter here and that's NOTHING to do with this PR. This is a tech debt item we need to solve. For e2e that run using `playright` (so node js) we have to write CommonJS but for the rest of our code that runs in the browser we have to write modules.

So we need to discuss that as a team and if we keep both types we should change linter configuration to allow Common JS just for tests.

As I said earlier, NOT related to this PR and NOT to be fixed in this PR.

ui/admin/tests/e2e/README.md
```diff
@@ -0,0 +1,44 @@
+# Admin UI e2e tests
+
+This test suite tests the Boundary Admin UI in an end-to-end setting. It was designed to be run in a
+variety of environments as long as the appropriate environment variables are set. You can choose to
+spin up your own infrastructure or use [Enos](https://github.com/hashicorp/boundary/tree/main/enos)
+to generate it for you. There is an enos scenario in the `boundary` repository that runs these
```
randallmorey, line 6: Can we link to the scenario on the text "an enos scenario"?
```suggestion
to generate it for you. There is [an enos scenario](/) in the `boundary` repository that runs these
```

ui/admin/tests/e2e/README.md
```diff
@@ -0,0 +1,44 @@
...
+tests.
+
+The test suite uses the [Playwright](https://playwright.dev/) framework.
+
+## Getting Started
+Setup...
+```bash
+cd boundary-ui/ui/admin
+yarn install
+npx playwright install
```
randallmorey, line 16: What is the purpose of this command?  Can it be included in the package.json as a `pre:e2e` step?  One less thing for developers to worry about.

ui/admin/tests/e2e/README.md
```diff
@@ -0,0 +1,44 @@
...
+
+This test suite tests the Boundary Admin UI in an end-to-end setting. It was designed to be run in a
+variety of environments as long as the appropriate environment variables are set. You can choose to
+spin up your own infrastructure or use [Enos](https://github.com/hashicorp/boundary/tree/main/enos)
+to generate it for you. There is an enos scenario in the `boundary` repository that runs these
+tests.
+
+The test suite uses the [Playwright](https://playwright.dev/) framework.
+
+## Getting Started
```
randallmorey, line 11: It seems that the CLI is required to be in the path in order to run e2e tests.  Is that documented?

randallmorey (changes requested):

ui/admin/tests/e2e/README.md
moduli: It hasn't been created yet. I'll remove this line and update it once it's there.

ui/admin/tests/e2e/README.md
moduli: That's correct. I'll update the README with that info.

ui/admin/tests/e2e/README.md
moduli: It's a one-time thing to install the browsers used by Playwright. Does the `pre:` imply something/is that a special command?

calcaide: My understanding of `npx playwright install` as Michael mention is an optional command to pre-install all browsers (one time) within your local machine before running the test. If you do not use this command is totally fine, browsers will be install (just one time) when developer runs the test.

I will suggest (for now) to not add any yarn script to run `npx playwright install`. I will leave it on the documentation as optional to keep it simple until we anticipate we have this need.

@randallmorey Happy to have a char over the `npx playwright install` detailed script if you want

ui/admin/tests/e2e/README.md
```diff
@@ -0,0 +1,49 @@
...
+to generate it for you.
+
+The test suite uses the [Playwright](https://playwright.dev/) framework.
+
+## Getting Started
+Setup...
+```bash
+cd boundary-ui/ui/admin
+yarn install
+npx playwright install # this installs the browsers used by Playwright
```
calcaide, line 15: I will mark this as optional, since browsers will be install per demand anyways if they are not at running the test.

moduli: Is it optional? I *think* I was greeted with a message to run this command before I could run tests.

calcaide: Documentation changed quite a bit since I put together e2e for Desktop, but at that time was 100% optional, and [following documentation right now](https://playwright.dev/docs/cli#install-browsers) seems to me it is.

But thinking twice... I does not hurt running it 😉

ui/admin/tests/e2e/README.md
```diff
@@ -0,0 +1,49 @@
...
+command returns something.
+```bash
+which boundary
+```
+
+Run tests...
+```bash
+export BOUNDARY_ADDR=
+export E2E_PASSWORD_ADMIN_LOGIN_NAME=
+export E2E_PASSWORD_ADMIN_LOGIN_PASSWORD=
```
calcaide, line 28: I think I found a misspelling: I set the documented env variables, run the test and got a complain saying I do not have set `E2E_PASSWORD_ADMIN_PASSWORD` but the one we documented is `E2E_PASSWORD_ADMIN_LOGIN_PASSWORD`

calcaide: Also, as soon I run the test it complains I am missing: `E2E_TARGET_IP,E2E_SSH_USER,E2E_SSH_KEY_PATH,E2E_SSH_PORT` env variables.

Do you think we should document those as well?

moduli: Yup. There's a typo in README. We need to set`E2E_PASSWORD_ADMIN_PASSWORD`. I'll update the README. 

So those are the extra "test-specific" environment variables that I was referencing earlier. Those are defined in each test file. You can see it near the top of `target.spec.js`

ui/admin/tests/e2e/README.md
calcaide: @moduli @randallmorey 
You guys are right. Is best to have the `npx playwright install` and also a `pre:` script that takes care of it.

I checked and if you already have the browsers install, running `npx playwright install` does not try to resintall them again, so is better to guarantee test suite integrity

ui/admin/tests/e2e/README.md
calcaide: Gotcha!! now I see them on `target.spec.js` 😉

calcaide (approved):

cameronperera (approved):

PR #1639, opened 2023-02-21
fix: 🐛 Fix failure in GH Actions Workflows

calcaide:
## Description

Changes this PR introduces:
- Change` node-version` definition from`'16.x'` to `16`. Both do the same, provide the greatest node 16 version. **THIS CHANGE** does not fix any problem, neither causes it, is 100% to simplify it. Feel free to argue this change and ask for the previous definition to be back.
- Deletes a conditional that was making the workflow to skip run the jobs with the specific setup. This was introduce it by mistake when we implement cache for the build job in CI. Until now it worked because the GH action `actions/setup-node@v3` was using node `v16` as default. Now `actions/setup-node@v3` is using node `v18` by default, so is when issues start popping and we realize about this.
- Fixes an outdated `yarn.lock`.

More context on the reasons under this PR [this slack thread](https://hashicorp.slack.com/archives/C012GTH1C4E/p1677010254594309)

### DIFF
--- a/.github/workflows/build-admin-ui.yaml
+++ b/.github/workflows/build-admin-ui.yaml
@@ -19,7 +19,7 @@ jobs:
             node-modules-
       - uses: actions/setup-node@v3
         with:
-          node-version: '16.x'
+          node-version: 16
       - run: yarn install
   build-admin-ui-oss:
     name: Build Admin UI OSS
@@ -37,7 +37,7 @@ jobs:
       - name: Build
         uses: actions/setup-node@v3
         with:
-          node-version: '16.x'
+          node-version: 16
         if: steps.node-modules-cache.outputs.cache-hit != 'true'
       - run: yarn run build:ui:admin:oss
       - name: Upload artifact admin-ui
@@ -66,7 +66,7 @@ jobs:
       - name: Build
         uses: actions/setup-node@v3
         with:
-          node-version: '16.x'
+          node-version: 16
         if: steps.node-modules-cache.outputs.cache-hit != 'true'
       - run: yarn run build:ui:admin:enterprise
       - name: Upload artifact enterprise admin UI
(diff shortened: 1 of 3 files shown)


dvonanderson (dismissed):

ZedLi:
Is the way `node-version` is defined wrong? The [documentation](https://github.com/actions/setup-node#supported-version-syntax) says it uses normal semantic versioning which means it should still be valid

ZedLi:
Should we also skip the step on `build-admin-ui` to force it to use 16?

calcaide:
As @ZedLi mentioned on two comments:
- Wrong assumption on my side, the definition of node was still valid.
- For `monorepo-validate.yaml` we were skipping the step to setup `node v16`.

I would suggest to change the node version definition as `16`, doing so GH Actions will apply the greatest v16. But open to change it back to `16.x`.

.github/workflows/build-admin-ui.yaml
```diff
@@ -37,7 +37,7 @@ jobs:
       - name: Build
         uses: actions/setup-node@v3
         with:
-          node-version: '16.x'
+          node-version: 16
         if: steps.node-modules-cache.outputs.cache-hit != 'true'
```
ZedLi, line 41: I think these also need to be removed in this file or else the [build](https://github.com/hashicorp/boundary-ui/actions/runs/4236321296/jobs/7361104771) will also fail.

ZedLi (changes requested):

.github/workflows/build-admin-ui.yaml
calcaide: That is correct, changes applied.

ZedLi (approved):
LGTM, thanks for cleaning this up!

cameronperera (approved):
LGTM

lisbet-alvarez (approved):

DhariniJeeva (approved):

PR #1640, opened 2023-02-21
fix: 🐛 Fix generic-vault getting extra null fields when saving

ZedLi:
✅ Closes: ICU-8141

:tickets: [Jira ticket](https://hashicorp.atlassian.net/browse/ICU-8141)

## Description
There was a regression introduced in 0.12.0 where we can't save the generic-vault type for credential libraries. The issue is we were sending extra `null` fields from the `vault-ssh-certificate` type and it's erroring out.

### DIFF
--- a/addons/api/addon/generated/models/credential-library.js
+++ b/addons/api/addon/generated/models/credential-library.js
@@ -48,6 +48,7 @@ export default class GeneratedCredentialLibraryModel extends BaseModel {
   //= attributes (vault)
 
   @attr('string', {
+    for: 'vault-generic',
     isNestedAttribute: true,
     description:
       'The HTTP method the library uses when requesting credentials from Vault.',
@@ -56,6 +57,7 @@ export default class GeneratedCredentialLibraryModel extends BaseModel {
   http_method;
 
   @attr('string', {
+    for: 'vault-generic',
     isNestedAttribute: true,
     description:
       'The body of the HTTP request the library sends to Vault when requesting credentials. Only valid if `http_method` is set to `POST`.',
@@ -69,18 +71,21 @@ export default class GeneratedCredentialLibraryModel extends BaseModel {
   path;
 
   @attr('string', {
+    for: 'vault-ssh-certificate',
     isNestedAttribute: true,
     description: 'The username to use when making an SSH connection.',
   })
   username;
 
   @attr('string', {
+    for: 'vault-ssh-certificate',
     isNestedAttribute: true,
     description: 'The desired key type to use when generating a private key.',
   })
   key_type;
 
   @attr('number', {
+    for: 'vault-ssh-certificate',
     isNestedAttribute: true,
     description:
       'Specifies the number of bits used when generating the private key. Not used if key type is ed25519',
@@ -88,12 +93,14 @@ export default class GeneratedCredentialLibraryModel extends BaseModel {
   key_bits;
 
   @attr('string', {
+    for: 'vault-ssh-certificate',
     isNestedAttribute: true,
     description: 'Specifies the requested time to live for the certificate.',
   })
   ttl;
 
   @attr('string', {
+    for: 'vault-ssh-certificate',
     isNestedAttribute: true,
     description:
       'Specifies the key id that the created certificate should have.',
@@ -101,6 +108,7 @@ export default class GeneratedCredentialLibraryModel extends BaseModel {
   key_id;
 
   @attr('object-as-array', {
+    for: 'vault-ssh-certificate',
     isNestedAttribute: true,
     description:
       'Specifies a map of the critical options that the certificate should be signed for.',
@@ -108,6 +116,7 @@ export default class GeneratedCredentialLibraryModel extends BaseModel {
   critical_options;
 
   @attr('object-as-array', {
+    for: 'vault-ssh-certificate',
     isNestedAttribute: true,
     description:
       'Specifies a map of the extensions that the certificate should be signed for.',
--- a/addons/api/addon/serializers/credential-library.js
+++ b/addons/api/addon/serializers/credential-library.js
@@ -1,5 +1,8 @@
 import ApplicationSerializer from './application';
-import { TYPE_CREDENTIAL_LIBRARY_VAULT_GENERIC } from '../models/credential-library';
+import {
+  TYPE_CREDENTIAL_LIBRARY_VAULT_GENERIC,
+  TYPE_CREDENTIAL_LIBRARY_VAULT_SSH_CERTIFICATE,
+} from '../models/credential-library';
 
 export default class CredentialLibrarySerializer extends ApplicationSerializer {
   // =properties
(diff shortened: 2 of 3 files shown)


addons/api/addon/serializers/credential-library.js
```diff
@@ -14,16 +17,41 @@ export default class CredentialLibrarySerializer extends ApplicationSerializer {
...
     if (serialized.attributes) {
-      if (serialized.type !== TYPE_CREDENTIAL_LIBRARY_VAULT_GENERIC) {
-        delete serialized.attributes.http_method;
-      }
-
       // Serialize `http_request_body` only if `http_method` is POST
-      if (!serialized.attributes?.http_method?.match(/post/i))
-        delete serialized.attributes.http_request_body;
+      if (!serialized.attributes?.http_method?.match(/post/i)) {
+        serialized.attributes.http_request_body = null;
```
ZedLi, line 53: I discovered a bug while testing and am purposely setting it to null in the scenario where someone is `PATCH`ing to a GET when there was an existing body. The API will error out if we don't set it to `null`.

See this [thread](https://hashicorp.slack.com/archives/C014AHXUL3G/p1677021391347479) for more info.

addons/api/addon/generated/models/credential-library.js
```diff
@@ -48,6 +48,7 @@ export default class GeneratedCredentialLibraryModel extends BaseModel {
   //= attributes (vault)
 
   @attr('string', {
+    for: 'vault-generic',
```
ZedLi, line 51: I ended up using the `for` concept we've talked about before to control whether we serialize the attribute. For some reason however, I couldn't import the type constant and use it, I would always get undefined.

Does anyone know if there's a trick to importing or something I missed?

DhariniJeeva: Hmm, I tried this branch locally, removed `for`, and could see the imported type values, and I didn't get `undefined` when I edited a `vault-generic` cred store. Is there a different scenario where you get undefined?

addons/api/addon/serializers/credential-library.js
```diff
@@ -14,16 +17,41 @@ export default class CredentialLibrarySerializer extends ApplicationSerializer {
...
+      case TYPE_CREDENTIAL_LIBRARY_VAULT_GENERIC:
+        return this.serializeVaultGeneric(...arguments);
+      case TYPE_CREDENTIAL_LIBRARY_VAULT_SSH_CERTIFICATE:
+      default:
+        return super.serialize(...arguments);
+    }
+  }
+
+  serializeAttribute(snapshot, json, key, attribute) {
+    let value = super.serializeAttribute(...arguments);
```
DhariniJeeva, line 31:
```suggestion
    const value = super.serializeAttribute(...arguments);
```

addons/api/addon/generated/models/credential-library.js
ZedLi: Did you mean you replaced the strings with the constants and were able to save?

I was trying 
```
import {
  TYPE_CREDENTIAL_LIBRARY_VAULT_GENERIC,
  TYPE_CREDENTIAL_LIBRARY_VAULT_SSH_CERTIFICATE,
} from '../models/credential-library';
```

and

```
import {
  TYPE_CREDENTIAL_LIBRARY_VAULT_GENERIC,
  TYPE_CREDENTIAL_LIBRARY_VAULT_SSH_CERTIFICATE,
} from 'api/models/credential-library';
```

And I wasn't getting a value in the `for` option when serializing while using these constants

DhariniJeeva: Ahh, I misinterpreted the original comment. I get `undefined` as well; however, if I move the constants to `baseModel`, I can access them just fine by importing from there, and I can also access constants from other models, such as targets.  🤔

DhariniJeeva (approved):

addons/api/addon/serializers/credential-library.js
```diff
@@ -14,16 +17,41 @@ export default class CredentialLibrarySerializer extends ApplicationSerializer {
...
+    // or isn't undefined, we delete it from the json
+    if (
+      options.isNestedAttribute &&
+      options.for !== type &&
+      options.for !== undefined
+    ) {
+      delete json.attributes[key];
+    }
+
+    return value;
```
lisbet-alvarez, line 45: This is just a general question about the `serializeAttribute` method. I don't understand why there is a `return value` since [the docs](https://api.emberjs.com/ember-data/4.9/classes/JSONAPISerializer/methods/serializeAttribute?anchor=serializeAttribute) don't specify the method returning anything.

ZedLi: Yes you're right, there is no return value so it doesn't do anything. I admittedly was following a previous example we had but I can clean this up

lisbet-alvarez (approved):

cameronperera (approved):

addons/api/addon/generated/models/credential-library.js
randallmorey: Fine for now.  In the future I think types will live in a separate place from models.

calcaide (approved):
LGTM!

addons/api/addon/serializers/credential-library.js
```diff
@@ -14,16 +17,39 @@ export default class CredentialLibrarySerializer extends ApplicationSerializer {
...
+
+    // For any attribute that doesn't match its `for`
+    // or isn't undefined, we delete it from the json
+    if (
+      options.isNestedAttribute &&
+      options.for !== type &&
+      options.for !== undefined
+    ) {
+      delete json.attributes[key];
+    }
```
randallmorey, line 43: Tried to reframe this in the positive.  IMHO conditionals stated
in the positive are easier to understand.  I'm also wondering
if it matters if an attribute is nested, shouldn't it match the
record type as long as `for` is specified?  And does it matter
if options.for is strictly not undefined or simply that it has
a value?  We can always get strict about typing later on when
we introduce formal resource types.
```suggestion
    // If an attribute has a `for` option, it must match the
    // record's `type`, else the attribute is excluded
    // from serialization.
    if (options?.for && options.for !== type) {
      delete json.attributes[key];
    }
```

addons/api/addon/serializers/credential-library.js
randallmorey: Should the associated code comment be updated?

randallmorey (dismissed):

addons/api/addon/serializers/credential-library.js
```diff
@@ -14,16 +17,39 @@ export default class CredentialLibrarySerializer extends ApplicationSerializer {
...
    * @param {Snapshot} snapshot
    */
-  serialize() {
+  serialize(snapshot) {
+    switch (snapshot.record.type) {
+      case TYPE_CREDENTIAL_LIBRARY_VAULT_GENERIC:
+        return this.serializeVaultGeneric(...arguments);
+      case TYPE_CREDENTIAL_LIBRARY_VAULT_SSH_CERTIFICATE:
+      default:
+        return super.serialize(...arguments);
```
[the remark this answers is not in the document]
randallmorey: Once we introduce formal resource types, an attempt to serialise an unknown type is probably an error.

addons/api/addon/serializers/credential-library.js
ZedLi: Yeah it shouldn't matter whether it's a nested attribute, I didn't have it originally but don't recall why I decided to add it, I think I might have actually just been testing, I'll change this. 

I do agree I think this is easier to read, I'll update it

addons/api/addon/serializers/credential-library.js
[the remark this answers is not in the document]
ZedLi: 🤔 I kept it like this since the default (which shouldn't be hit) would just do the least amount of extra serialization. I actually feel like maybe if we get to the default case we should be erroring out since that would be unexpected

addons/api/addon/serializers/credential-library.js
ZedLi: Yup I'll update that

addons/api/addon/serializers/credential-library.js
```diff
@@ -14,16 +17,40 @@ export default class CredentialLibrarySerializer extends ApplicationSerializer {
...
+    }
+  }
+
+  serializeAttribute(snapshot, json, key, attribute) {
+    super.serializeAttribute(...arguments);
+    const { options } = attribute;
+    const { type } = snapshot.record;
+
+    // For any attribute that doesn't match its `for`
+    // or isn't undefined, we delete it from the json
```
randallmorey, line 36:
```suggestion
    // If an attribute has a `for` option, it must match the
    // record's `type`, else the attribute is excluded
    // from serialization.
```

randallmorey (changes requested):

randallmorey (approved):

calcaide (approved):

PR #1643, opened 2023-02-24
test(admin/e2e): Add tests for Credential Stores

moduli:

## Description

This PR adds more tests to the Admin UI e2e test suite (https://github.com/hashicorp/boundary-ui/pull/1637). There are two tests added in this PR.

One test does the following...
- Creates a Static Credential Store
- Stores a Username/Key Pair credential in the store and attaches the credential to a target
- Uses the CLI to authorize a session and retrieve the credentials
- Confirms that the credentials match

The other test does something similar, but with Vault...
- Uses the Vault CLI to set up a Vault server
- Creates a Username/Key Pair credential in Vault
- Creates a Vault Credential Store
- Creates a Vault Credential Library pointing to the Vault credential and attaches the library to a target
- Uses the CLI to authorize a session and retrieve the credentials
- Confirms that the credentials match

### DIFF
--- a/ui/admin/tests/e2e/helpers/vault-cli.js
+++ b/ui/admin/tests/e2e/helpers/vault-cli.js
@@ -1,0 +1,13 @@
+/* eslint-disable no-undef */
+const { execSync } = require('child_process');
+
+/**
+ * Checks that the vault cli is available
+ */
+exports.checkVaultCli = async () => {
+  try {
+    execSync('which vault');
+  } catch (e) {
+    throw new Error('vault does not exist on the path');
+  }
+};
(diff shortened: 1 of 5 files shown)


ui/admin/tests/e2e/helpers/vault-cli.js
```diff
@@ -0,0 +1,13 @@
+/* eslint-disable no-undef */
+const { execSync } = require('child_process');
+
+/**
+ * Checks that the vault cli is available
+ */
+exports.checkVaultCli = async () => {
```
calcaide, line 7: Does this means we need vault install to run the test? or just the CLI presence?

moduli: As currently set up, this test needs both the Vault CLI and a Vault server instance running.

ZedLi (approved):
Looks great, thanks for adding these!

cameronperera (approved):
LGTM

PR #1645, opened 2023-02-24
feat: 🎸 initial LDAP authentication support

randallmorey:
This pr lists LDAP on the signup/login form and allows authentication via LDAP on both admin and desktop UI

[jira](https://hashicorp.atlassian.net/browse/ICU-9213)

admin ui:
Screenshot 2023-05-10 at 11 43 15 AM

Screenshot 2023-05-10 at 11 43 29 AM

Screenshot 2023-05-10 at 11 43 42 AM

desktop: 

Screenshot 2023-05-10 at 11 45 41 AM

Screenshot 2023-05-10 at 11 55 14 AM

Screenshot 2023-05-10 at 11 55 37 AM


addons/core/addon/components/form/authenticate/index.hbs
```diff
@@ -9,7 +9,6 @@
   {{component
     (concat 'form/authenticate/' @model.type)
     submit=@submit
-    cancel=false
```
DhariniJeeva, line 12: I removed this as we don't have a cancel button on the authenticate forms

addons/core/tests/integration/components/form/authenticate-test.js
```diff
@@ -0,0 +1,31 @@
...
+ * SPDX-License-Identifier: MPL-2.0
+ */
+
+import { module, test } from 'qunit';
+import { setupRenderingTest } from 'ember-qunit';
+import { render, find } from '@ember/test-helpers';
+import { hbs } from 'ember-cli-htmlbars';
+import { setupIntl } from 'ember-intl/test-support';
+
+module('Integration | Component | form | authenticate', function (hooks) {
```
cameronperera, line 12: Could we add a some more tests/asserts for this form? My thoughts are with the current test, we could delete the 2 input fields and this tests would still pass. My suggestion would be to assert the inputs and form buttons are present.

cameronperera (changes requested):

lisbet-alvarez (approved):
LGTM!

cameronperera (approved):
Looks great!

PR #1646, opened 2023-02-27
chore: 🤖 remove window object access patterns

dvonanderson:
✅ Closes: ICU-7152

:tickets: [Jira ticket](https://hashicorp.atlassian.net/browse/ICU-7152)

## Description
Deprecate window object service pattern in favor of `ember-browser-services`. Remove `browser-object` service and test.


ui/admin/app/components/form/worker/create-worker-led/index.js
```diff
@@ -17,7 +17,7 @@ class Tag {
 export default class FormWorkerCreateWorkerLedComponent extends Component {
   // =services
   @service features;
-  @service browserObject;
+  @service('browser/window') window;
```
ZedLi, line 20: Is this also supposed to update how we make references to these globals in desktop as well?

We still have references to the other way we access the globals such as in [this example](https://github.com/hashicorp/boundary-ui/blob/1f3a8ef135d1eadcfe6cc9428e80d32e1dbd223a/ui/desktop/app/routes/cluster-url.js#L31)

dvonanderson: @ZedLi yes good catch!

ui/admin/app/components/form/worker/create-worker-led/index.js
```diff
@@ -30,7 +30,7 @@ export default class FormWorkerCreateWorkerLedComponent extends Component {
   @tracked newWorkerValue;
 
   get clusterIDFromURL() {
-    const hostname = this.browserObject.hostname;
```
calcaide, line 33: Are the `?` used here for [optional chaining](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining)? if so, due to the importance of `window` and `location` we can assure it will exist.

calcaide (changes requested):
Nice work Devon!!! looking good!! As already mentioned, on a comment by Zed, the only thing I am missing in the PR is the methods within `ui/desktop`. Great work :wink:

ui/admin/app/components/form/worker/create-worker-led/index.js
dvonanderson: @calcaide yea not sure on this one, I took expression from the hostname getter inside of the `browserObject` service.  My guess is they didn't want it to throw an error if it didn't exist, but I can see how that would never be a problem.

calcaide: Yeah, as Zed mention is fine eitherway and this is a not blocking comment (sorry I didn't make that clear at first).

Yeah, this could be a problem on server side rendering but I don't think we are going to reproduce it since we do not perform that.

How my mind see it: if we do not have `window` on the current context of Boundary Admin UI, the browser, then we are in trouble :sweat_smile:

But you guys are right, better to be safe than sorry.

ui/desktop/app/routes/cluster-url.js
```diff
@@ -29,10 +29,8 @@ export default class ClusterUrlRoute extends Route {
    * @type {Window}
    */
   get window() {
-    // The Ember way of accessing globals...
-    const document = getOwner(this).lookup('service:-document').documentElement;
     // defaultView === window, but without using globals directly
-    return document.parentNode.defaultView;
+    return this.document.documentElement.parentNode.defaultView;
```
ZedLi, line 33: I think this can be just returning the window rather than going through all this indirection

dvonanderson:
Got some whacky local stuff going on in which I can't rebase this.   Gonna redo it.

PR #1647, opened 2023-02-27
test(admin/e2e): Add login tests

moduli:

## Description

This PR adds more tests to the Admin UI e2e test suite (https://github.com/hashicorp/boundary-ui/pull/1637). These tests revolve around logging in.

The tests include the following...
- Log in, log out, and then log back in
- Attempt to log in with an invalid password
- Attempt to log in with only a username
- Attempt to log in with only a password

### DIFF
--- a/ui/admin/tests/e2e/tests/login.spec.js
+++ b/ui/admin/tests/e2e/tests/login.spec.js
@@ -1,0 +1,74 @@
+/* eslint-disable no-undef */
+const { test, expect } = require('@playwright/test');
+
+test('Log in, log out, and then log back in', async ({ page }) => {
+  await page.goto('/');
+
+  await page
+    .getByLabel('Login Name')
+    .fill(process.env.E2E_PASSWORD_ADMIN_LOGIN_NAME);
+  await page
+    .getByLabel('Password')
+    .fill(process.env.E2E_PASSWORD_ADMIN_PASSWORD);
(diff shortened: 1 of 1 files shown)


ui/admin/tests/e2e/tests/login.spec.js
```diff
@@ -0,0 +1,74 @@
...
+  await page.getByRole('button', { name: 'Sign In' }).click();
+  await expect(page.getByRole('navigation', { name: 'General' })).toBeVisible();
+  await expect(
+    page.getByText(process.env.E2E_PASSWORD_ADMIN_LOGIN_NAME)
+  ).toBeEnabled();
+
+  await page.getByText(process.env.E2E_PASSWORD_ADMIN_LOGIN_NAME).click();
+  await page.getByRole('button', { name: 'Sign Out' }).click();
+  await expect(page.getByRole('button', { name: 'Sign In' })).toBeVisible();
+
```
moduli, line 22: Is there anything else we can check to make sure we're logged out?

cameronperera: I am not sure. The only other thing I can think of (which is something I can see as a real user interaction) is to try to directly change the URL, wait for it to resolve, and confirm that we still see the "Sign In" button. Not sure if that is possible or not and is likely overkill for this test.

cameronperera (approved):

PR #1652, opened 2023-03-02
chore: 🤖 deprecate window object service

dvonanderson:
✅ Closes: ICU-7152

:tickets: [Jira ticket](https://hashicorp.atlassian.net/browse/ICU-7152)

## Description
Deprecate `browser-object` service in favor of `ember-browser-services`.  Re pushed PR to have cleaner commits.


ZedLi:
Are we still going to cleanup the desktop usages?

calcaide (changes requested):
Looking good @dvonanderson !! I am missing the clean up on the `ui/desktop` are we going to put that within this PR or in a separate one?

dvonanderson:
> Are we still going to cleanup the desktop usages?

@zed @calcaide yes working on desktop now.  We only reference the window object once in the `cluster-url` route.  Seems to be working correctly, but test are failing.

dvonanderson:
@zed or @calcaide  having some challenges with the last piece to this PR.  `cluster-url.js` is the only file in the desktop application that interacts with the window object.  After using the `browser/window` service to grab the `window.location.origin` all of the acceptance test for `cluster-url` pass with the exception for the last two.  Those test are referencing the `autoOrigin` value in the desktop environment file.  Any suggestions?
```
Error while processing route: cluster-url Cannot read properties of undefined (reading 'origin') TypeError: Cannot read properties of undefined (reading 'origin')
```

ZedLi (approved):

ui/desktop/tests/acceptance/cluster-url-test.js
```diff
@@ -115,6 +111,8 @@ module('Acceptance | clusterUrl', function (hooks) {
 
   test('visiting index without a clusterUrl specified redirects to clusterUrl route', async function (assert) {
     assert.expect(2);
+    this.owner.register('service:browser/window', WindowMockIPC);
```
calcaide, line 114: What is the reason why we delete this from the `beforeEach` statement and move it to each test? I assume it is because not all test in the test file use it?

If the reason is not all test use it, I think we could extract this code to a function before the `describe` block, then, each test that needs the mock of the `service:browser/window` just call that function.

Let me know your thoughts :wink:

dvonanderson: @calcaide absolutely

calcaide (approved):
Awesome @dvonanderson Terrific work!!!

PR #1657, opened 2023-03-16
Refactor edition enablement

randallmorey:
:tickets: [Jira ticket](https://hashicorp.atlassian.net/browse/ICU-8382)

## Description

This PR introduces dynamic runtime enabling of feature editions (OSS, HCP, Enterprise).  This allows for dynamic runtime edition toggling during development for a more rapid and realistic experience.

Screenshot 2023-03-16 at 15 26 20

### DIFF
--- a/addons/core/translations/en-us.yaml
+++ b/addons/core/translations/en-us.yaml
@@ -28,7 +28,6 @@ titles:
   actions: Actions
   documentation: Documentation
   abandon-confirm: Discard unsaved changes?
-  toggle-color-theme: Toggle Color Theme
   welcome-to-boundary: Welcome to Boundary
   label: Label
   value: Value
@@ -38,6 +37,12 @@ titles:
   success: Success
   create-resources: Create Resources
   choose-an-option: Choose an option
+  feature-edition: 'Edition: {edition}'
+editions:
+  oss: OSS
+  ent: Enterprise
+  enterprise: Enterprise
+  hcp: HCP
 descriptions:
   empty-set: There are no items to display yet.  You may be able to add items or try back later.
   cluster-url-initialization: To get started, please enter your cluster URL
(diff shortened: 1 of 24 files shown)


ui/admin/config/environment.js
```diff
@@ -9,28 +9,36 @@ const APP_NAME = process.env.APP_NAME || 'Boundary';
...
+  ...baseEdition,
+  byow: true,
+  'byow-pki-upstream': true,
+  'json-credentials': true,
+  'static-credentials': true,
+  'target-worker-filters-v2': true,
+  'target-network-address': true,
 };
 featureEditions.enterprise = {
   ...featureEditions.oss,
```
ZedLi, line 38: Should we be inheriting from `oss`? I thought the idea of having a base was so we would only be turning features on, rather than turning any off if they're mutually exclusive. 

Currently we have one flag `byow-pki-upstream` that is mutually exclusive and would still need to set to false for  hcp

randallmorey: `byow-pki-upstream` and `byow-pki-hcp-cluster-id` are currently the only mutually exclusive flags.  This could be refactored.

There isn't really a "base" edition.  A more apt name might be _available features_.  It just declares which flags exist.  Logically, enterprise does indeed "inherit" its features from OSS.  Ent and HCP are supersets (mostly).

dvonanderson (approved):
Nice

addons/core/translations/en-us.yaml
```diff
@@ -38,6 +37,12 @@ titles:
   success: Success
   create-resources: Create Resources
   choose-an-option: Choose an option
+  feature-edition: 'Edition: {edition}'
+editions:
+  oss: OSS
+  ent: Enterprise
+  enterprise: Enterprise
```
cameronperera, line 44: Do we need both of these translations?

ui/admin/config/environment.js
randallmorey: Address in #1658

PR #1658, opened 2023-03-17
refactor: 💡 remove explicit byow-pki-upstream, as it is default

randallmorey:
:tickets: [✅ Closes: ICU-8551](https://hashicorp.atlassian.net/browse/ICU-8551)

## Description

This PR removes the feature flag`byow-pki-upstream` for being mutually exclusive with `byow-pki-hcp-cluster-id`.  The removed flag becomes the default if `byow-pki-hcp-cluster-id` is disabled.


lisbet-alvarez (approved):
LGTM!

PR #1661, opened 2023-03-28
Feature metadata

randallmorey:
This PR introduces a controller metadata mock endpoint for discovering the edition and enabled features.  It also adds a feature select in development for more rapid experimentation.

Screenshot 2023-03-28 at 09 29 19


ui/admin/app/instance-initializers/feature-edition.js
```diff
@@ -1,6 +1,24 @@
...
+  let features;
+
+  // Attempt to load edition for the metadata JSON
+  try {
+    const response = await fetch(url);
+    const json = await response.json();
+    edition = edition || json?.license?.edition;
+    features = json?.license?.features;
+  } catch (e) {
+    // if the request fails, the service will initialize the default edition
```
calcaide, line 16: None blocking comment: In case of failure, do we want the user know?

ui/admin/app/services/feature-edition.js
```diff
@@ -3,7 +3,7 @@ import { inject as service } from '@ember/service';
 import { tracked } from '@glimmer/tracking';
 import config from 'admin/config/environment';
 
-const { selectedEdition, featureEditions } = config;
+console.log(config.features.licensedFeatures);
```
calcaide, line 6: None blocking comment: this console, is for development usage? I understand it as it will console log the list of licensed features?

DhariniJeeva, line 6: do we need this?

ui/admin/app/templates/application.hbs
```diff
@@ -89,14 +89,28 @@
...
-                          @label={{t (concat 'editions.' edition)}}
+                          @label={{edition}}
                           @value={{edition}}
                         />
                       </dropdown.item>
                     {{/each}}
                   </form.radioGroup>
                 </Rose::Form>
               </dropdown.section>
+              <dropdown.separator />
```
calcaide, line 100: The current dropdown is `Rose`, Can we use an HDS dropdwon here?

ui/admin/app/routes/application.js
```diff
@@ -20,6 +20,7 @@ export default class ApplicationRoute extends Route {
   @service confirm;
   @service router;
   @service intl;
+  @service features;
```
calcaide, line 23: None blocking comment: where `features` service lives? I see `feature-editions` within `app/services/feature-edition` but not `features`.

ui/admin/app/templates/application.hbs
DhariniJeeva: Currently, HDS doesn't support radio button within a dropdown

ui/admin/app/instance-initializers/feature-edition.js
randallmorey: The metadata endpoint isn't required and should be allowed to error.  When that happens we default to the OSS edition.

ui/admin/app/services/feature-edition.js
randallmorey: Good catch this should be removed.

ui/admin/app/templates/application.hbs
randallmorey: Not really, because this is in main app header, which doesn't support HDS components.  The overall app chrome can be migrated in the future when HDS supports all of the components that appear within it.

ui/admin/app/routes/application.js
randallmorey: https://www.npmjs.com/package/ember-feature-flags

ui/admin/app/services/feature-edition.js
randallmorey, line 6:
```suggestion
```

ui/admin/app/instance-initializers/feature-edition.js
calcaide: I understood that part. My question is: Would the user be informed? or because the endpoint isn't required (as you mention) the user does not need to know?

addons/core/translations/en-us.yaml
```diff
@@ -40,7 +40,6 @@ titles:
   feature-edition: 'Edition: {edition}'
 editions:
```
DhariniJeeva, line 41: Do we still need this translation? Seems like we don't use this anywhere in the code

calcaide (approved):

ui/admin/app/instance-initializers/feature-edition.js
```diff
@@ -1,6 +1,24 @@
...
-  service.initialize();
+  const url = '/metadata.json';
+  let edition;
+  let features;
+
+  // Attempt to load edition for the metadata JSON
+  try {
+    const response = await fetch(url);
+    const json = await response.json();
+    edition = edition || json?.license?.edition;
```
ZedLi, line 13: Couldn't this have been:
```suggestion
    const edition = json?.license?.edition;
```

It doesn't look like `edition` can get initialized before this line?

ui/admin/app/services/feature-edition.js
```diff
@@ -12,31 +11,43 @@ export default class FeatureEditionService extends Service {
 
   // =attributes
 
+  licensedFeatures = Object.keys(config.features.licensedFeatures);
```
ZedLi, line 14: What is this `licensedFeatures` used for?

addons/core/translations/en-us.yaml
randallmorey: Removed.

ui/admin/app/instance-initializers/feature-edition.js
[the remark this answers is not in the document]
randallmorey: The service is just responsible for setting an edition and enabling extra features.  It doesn't know or care where the information comes from.  Controller metadata is broader in conception that just edition or features and thus its handling is separated into an initializer.

ui/admin/app/services/feature-edition.js
randallmorey: https://github.com/hashicorp/boundary-ui/pull/1661/files?diff=split&w=0#diff-d2befbbcfc2fd4071e76ff3126f8e2cec284877fdf26740bddfa7b8b1f5727a9R103

ZedLi (approved):

DhariniJeeva (approved):

cameronperera (approved):

ui/admin/tests/acceptance/workers/list-test.js
```diff
@@ -68,6 +73,7 @@ module('Acceptance | workers | list', function (hooks) {
 
   test('Users can navigate to workers with only list permission', async function (assert) {
     assert.expect(1);
+    featuresService.enable('byow');
```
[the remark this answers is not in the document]
randallmorey: When a flag is enabled via the service, it is safe to assume that it works since the library should already be unit tested.  The enable call also serves as self-documenting code, clearly indicating to developers that the test relies on the flag.  

There are some cases where we specifically test the negative case:  a flag is _disabled_.  Explicit calls to _disable_ flags are no longer necessary, since they are all disabled by default.  That results in a loss of self-documentation.  However, we can reclaim that lost readability by testing the state of a flag when a test explicitly relies on it being disabled.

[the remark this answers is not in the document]
randallmorey: Upon reflection, a better approach would be to encapsulate feature flagging into test helpers.  In a test a developer might call `withFeatures('byow', 'ssh-target')`.  Internally that would result in two calls:  one to enable the features and one to assert they are enabled (or vice versa).

PR #1663, opened 2023-03-31
Resolutions audit: Delete 3 oldest resolutions

calcaide:
:tickets: [Jira ticket](https://hashicorp.atlassian.net/browse/ICU-8761)

## Description

Delete 3 oldest resolutions

### DIFF
--- a/package.json
+++ b/package.json
@@ -49,7 +49,6 @@
     "lint-staged": "^13.0.2"
   },
   "resolutions": {
-    "bl": "~1.2.3",
     "tough-cookie": "~2.3.3",
     "npm": "~6.14.6",
     "npm-registry-fetch": "~8.1.1",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -49,9 +49,6 @@
     "lint-staged": "^13.0.2"
   },
   "resolutions": {
-    "bl": "~1.2.3",
-    "tough-cookie": "~2.3.3",
-    "npm": "~6.14.6",
```
ZedLi, line 54: I wonder why we had to pin npm in the first place, it looks like without this resolution we don't even need it in our yarn lock file since it has no dependencies 🤔

ZedLi (approved):
LGTM, thanks a ton for doing this auditing, it's very much appreciated!!

cameronperera (approved):
Thanks for this work Carlos!

package.json
calcaide: I have no idea why it was there. It was add more than 2 years ago. I tested all the build procedures and works np.

After we finish the `resolutions` audit (it will take long) doing a dependencies audit and review what we have, IMHO is a good idea.

PR #1664, opened 2023-04-05
feat: 🎸 Add support to display external name for dynamic hosts

ZedLi:
✅ Closes: ICU-8795

:tickets: [Jira ticket](https://hashicorp.atlassian.net/browse/ICU-8795)

## Description
Added support for dynamic host catalogs to display the external name for hosts. Take a look at the mockup [here](https://hashicorp.atlassian.net/browse/ICU-7472), and the [PRD](https://docs.google.com/document/d/181RpEqvKAg-BWNEOEgPCRuu64_GWHBYZmT5Cjbf1uTA/edit#).

I kept the same behavior as we currently have but instead display the external name if there's no boundary name (~~I think there can only be a boundary name if a user manually updates the host's name from the CLI? Not sure since it looks like name isn't populated by default with dynamic hosts~~ Dynamic hosts won't have a name).

The name column for dynamic hosts will now be `External Name`

image

image

image

Desktop:
image

### DIFF
--- a/addons/api/mirage/factories/host.js
+++ b/addons/api/mirage/factories/host.js
@@ -10,7 +10,9 @@ import generateId from '../helpers/id';
 
 export default factory.extend({
   id: () => generateId('h_'),
-
+  name() {
+    if (this.type === 'static') return faker.random.words();
+  },
   type() {
     return this.hostCatalog?.type || factory.attrs.type;
   },
@@ -34,8 +36,16 @@ export default factory.extend({
     }
     return result;
   },
+  // Return external fields only for plugins
   external_id() {
-    return faker.datatype.uuid();
+    if (this.type === 'plugin') {
+      return faker.datatype.uuid();
+    }
+  },
+  external_name(i) {
+    if (this.type === 'plugin') {
+      return `${this.plugin.name} provided name ${i}`;
+    }
   },
   // Only aws plugins have dns_names
   dns_names() {
--- a/ui/admin/app/components/form/host/aws/index.hbs
+++ b/ui/admin/app/components/form/host/aws/index.hbs
@@ -11,51 +11,70 @@
   as |form|
 >
 
-  <form.input
+  <Hds::Form::TextInput::Field
     @value={{@model.type}}
-    @label={{t 'form.type.label'}}
-    readonly={{true}}
-    @disabled={{true}}
-  />
-
-  <form.input
-    @name='name'
-    @type='name'
-    @value={{@model.name}}
-    @label={{t 'form.name.label'}}
-    @error={{@model.errors.name}}
-    @helperText={{t 'form.name.help'}}
-    readonly={{false}}
-    as |field|
+    name='type'
+    disabled={{true}}
+    as |F|
   >
-    {{#if @model.errors.name}}
-      <field.errors as |errors|>
-        {{#each @model.errors.name as |error|}}
-          <errors.message>{{error.message}}</errors.message>
-        {{/each}}
-      </field.errors>
-    {{/if}}
-  </form.input>
+    <F.Label>{{t 'form.type.label'}}</F.Label>
+  </Hds::Form::TextInput::Field>
 
-  <form.textarea
-    @name='description'
-    @type='description'
+  {{#if @model.external_name}}
+    <Hds::Form::TextInput::Field
+      @value={{@model.external_name}}
+      @isInvalid={{@model.errors.external_name}}
+      name='external_name'
+      disabled={{true}}
+      as |F|
+    >
+      <F.Label>{{t 'form.external_name.label'}}</F.Label>
+      <F.HelperText>{{t 'form.external_name.help'}}</F.HelperText>
+      {{#if @model.errors.external_name}}
+        <F.Error as |E|>
+          {{#each @model.errors.external_name as |error|}}
+            <E.Message>{{error.message}}</E.Message>
+          {{/each}}
+        </F.Error>
+      {{/if}}
+    </Hds::Form::TextInput::Field>
+  {{/if}}
+
+  <Hds::Form::TextInput::Field
+    @value={{@model.external_id}}
+    @isInvalid={{@model.errors.external_id}}
+    name='external_id'
+    disabled={{true}}
+    as |F|
+  >
+    <F.Label>{{t 'form.external_id.label'}}</F.Label>
+    <F.HelperText>{{t 'form.external_id.help'}}</F.HelperText>
+    {{#if @model.errors.external_id}}
+      <F.Error as |E|>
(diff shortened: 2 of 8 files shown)


ui/admin/app/components/form/host/aws/index.hbs
```diff
@@ -69,27 +88,33 @@
...
-          @disabled={{true}}
+          disabled={{true}}
+          @width='unset'
         />
-      {{/each}}
-    </:body>
-  </form.fieldset>
+      </F.Control>
+    {{/each}}
+  </Hds::Form::Fieldset>
```
ZedLi, line 103: This displays a bit differently now: it shows the border for the input fields for each of these without overriding HDS styles. I thought it actually looked fine and the behavior was a bit better than before. 

It looks like if there were too many fields, it would force them all on one line regardless and become unable to display them (though this is unlikely)
image

Here's a normal scenario: 
image

Let me know your thoughts

ui/admin/app/components/form/host/aws/index.hbs
```diff
@@ -11,51 +11,70 @@
...
+    name='type'
+    disabled={{true}}
+    as |F|
+  >
+    <F.Label>{{t 'form.type.label'}}</F.Label>
+  </Hds::Form::TextInput::Field>
+
+  {{#if @model.external_name}}
+    <Hds::Form::TextInput::Field
+      @value={{@model.external_name}}
```
DhariniJeeva, line 25: should we add `isRequired/isOptional` arg here?

ui/admin/app/components/form/host/azure/index.hbs
```diff
@@ -68,19 +87,20 @@
...
+    <F.Legend>{{t 'form.ip_addresses.label'}}</F.Legend>
+    <F.HelperText>{{t 'form.ip_addresses.help'}} </F.HelperText>
+    {{#each @model.ip_addresses as |ip_address|}}
+      <F.Control>
+        <Hds::Form::TextInput::Base
           @value={{ip_address}}
-          readonly={{true}}
-          @disabled={{true}}
+          disabled={{true}}
+          @width='unset'
```
DhariniJeeva, line 99: Why are we unsetting this width here?

ui/admin/app/components/form/host/aws/index.hbs
```diff
@@ -11,51 +11,70 @@
...
     @value={{@model.description}}
-    @label={{t 'form.description.label'}}
-    @error={{@model.errors.description}}
-    @helperText={{t 'form.description.help'}}
-    disabled={{@model.isSaving}}
-    readonly={{false}}
-    as |field|
+    @isInvalid={{@model.errors.description}}
+    name='description'
+    disabled={{true}}
```
DhariniJeeva, line 65: Would `form.disabled` work here?

ui/admin/app/components/form/host/aws/index.hbs
DhariniJeeva: I like the bordered inputs.

ui/admin/app/components/form/host/aws/index.hbs
ZedLi: I didn't add them to these fields because these are "non-standard" forms since they are basically readonly and can't be modified as the dynamic hosts are being pulled externally.

It didn't make sense to me to include `isOptional` or `isRequired` for these fields, any thoughts to this?

ui/admin/app/components/form/host/aws/index.hbs
ZedLi: We could but it also will always be disabled, I think I preferred to be explicit and set all the fields to `disabled=true` since we were already doing that for some fields and since the form will never be updated

ui/admin/app/components/form/host/aws/index.hbs
DhariniJeeva: Yeah, in that case, we don't need to.

ui/admin/app/templates/scopes/scope/host-catalogs/host-catalog/host-sets/host-set/hosts/index.hbs
```diff
@@ -31,10 +31,14 @@
...
-            <row.headerCell>{{t 'form.name.label'}}</row.headerCell>
+            {{#if @model.hostSet.isStatic}}
+              <row.headerCell>{{t 'form.name.label'}}</row.headerCell>
+            {{else}}
+              <row.headerCell>{{t 'form.external_name.label'}}</row.headerCell>
+            {{/if}}
             <row.headerCell>{{t 'form.type.label'}}</row.headerCell>
             <row.headerCell>{{t 'form.id.label'}}</row.headerCell>
-            {{#if @model.isStatic}}
+            {{#if @model.hostSet.isStatic}}
```
ZedLi, line 41: I think this was a bug since it wasn't ever displaying the actions header

DhariniJeeva (approved):

ui/admin/app/templates/scopes/scope/host-catalogs/host-catalog/host-sets/host-set/hosts/index.hbs
lisbet-alvarez: Non-blocking question: What was the reason for having `@model.isStatic` in the first place since it seems like hosts belonging to a static host-catalog are listed in this [file ](https://github.com/hashicorp/boundary-ui/blob/main/ui/admin/app/templates/scopes/scope/host-catalogs/host-catalog/hosts/index.hbs) (in [vercel](https://boundary-ui-git-icu-8795-dynamic-host-catalog-0d4d11-hashicorp.vercel.app/scopes/s_1sheljmjgm/host-catalogs/hc_pkq4tp9lzg/hosts) )?

lisbet-alvarez (approved):
LGTM!

cameronperera (approved):

ui/admin/app/templates/scopes/scope/host-catalogs/host-catalog/host-sets/host-set/hosts/index.hbs
lisbet-alvarez: Ah I def. misunderstood! Thanks for clarifying!

addons/api/mirage/factories/host.js
```diff
@@ -10,7 +10,9 @@ import generateId from '../helpers/id';
 
 export default factory.extend({
   id: () => generateId('h_'),
-
+  name() {
+    if (this.type === 'static') return faker.random.words();
```
calcaide, line 14: None blocking question, If the host has no type, then it returns no name?

calcaide (approved):
LGTM! Awesome work!!

PR #1667, opened 2023-04-13
test(admin/e2e): Add create target with address test

stasryzhov:
:tickets: [Jira ticket](https://hashicorp.atlassian.net/browse/ICU-8420)

## Description
This PR creates a new UI test that does the following:
- Prepares test data: org, project
- Creates a target with an address directly added to it
- Connects to the target
- Cancels the connection
- Deletes test org

### DIFF
--- a/ui/admin/tests/e2e/helpers/boundary-cli.js
+++ b/ui/admin/tests/e2e/helpers/boundary-cli.js
@@ -40,6 +40,7 @@ exports.authenticateBoundaryCli = async () => {
 /**
  * Uses the boundary CLI to connect to the specified target
  * @param {string} targetId ID of the target to be connected to
+ * @returns ChildProcess representing the result of the command execution
  */
 exports.connectToTarget = async (targetId) => {
   let connect;
--- a/ui/admin/tests/e2e/helpers/boundary-cli.js
+++ b/ui/admin/tests/e2e/helpers/boundary-cli.js
@@ -4,7 +4,7 @@
  */
 
 /* eslint-disable no-undef */
-const { execSync } = require('child_process');
+const { execSync, exec } = require('child_process');
 
 /**
  * Checks that the boundary cli is available
@@ -36,3 +36,37 @@ exports.authenticateBoundaryCli = async () => {
     console.log(`${e.stderr}`);
   }
 };
+
+exports.connectToTarget = async (target) => {
+  let connect;
+  try {
+    connect = exec(
+      'boundary connect' +
+      ' -target-id=' +
+      target.id +
+      ' -exec /usr/bin/ssh --' +
+      ' -l ' +
+      process.env.E2E_SSH_USER +
+      ' -i ' +
+      process.env.E2E_SSH_KEY_PATH +
+      ' -o UserKnownHostsFile=/dev/null' +
+      ' -o StrictHostKeyChecking=no' +
+      ' -o IdentitiesOnly=yes' + // forces the use of the provided key
+      ' -p {{boundary.port}}' +
+      ' {{boundary.ip}}'
+    )
+  } catch (e) {
+    console.log(`${e.stderr}`);
+  }
+  return connect
+}
+
+exports.deleteOrg = async (orgId) => {
+  try {
+    exec(
+      'boundary scopes delete -id=' + orgId
+    )
+  } catch (e) {
+    console.log(`${e.stderr}`);
+  }
+}
--- a/ui/admin/tests/e2e/helpers/boundary-cli.js
+++ b/ui/admin/tests/e2e/helpers/boundary-cli.js
@@ -37,36 +37,42 @@ exports.authenticateBoundaryCli = async () => {
   }
 };
 
-exports.connectToTarget = async (target) => {
+/**
+ * Uses the boundary CLI to connect to the specified target
+ * @param {string} targetId ID of the target to be connected to
+ */
+exports.connectToTarget = async (targetId) => {
   let connect;
   try {
     connect = exec(
       'boundary connect' +
-      ' -target-id=' +
-      target.id +
-      ' -exec /usr/bin/ssh --' +
-      ' -l ' +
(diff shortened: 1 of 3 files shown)


ui/admin/tests/e2e/tests/target.spec.js
```diff
@@ -109,3 +119,38 @@ test('Verify session created then cancel the session', async ({ page }) => {
     }
   }
 });
+
+test('Verify session created to target with address then cancel the session', async ({page}) => {
```
moduli, line 122:
```suggestion
test('Verify session created to target with address, then cancel the session', async ({page}) => {
```
We may also want to update the name of the first test to something like
`Verify session created to target with host, then cancel the session`

ui/admin/tests/e2e/tests/target.spec.js
```diff
@@ -34,6 +39,10 @@ test.beforeAll(async () => {
   await checkBoundaryCli();
 });
 
+test.afterEach(async () => {
+  await deleteOrg(org.id);
+})
+
```
moduli, line 45: Instead of doing this approach, what if we added a `try/finally` in a test to do the cleanup (i.e. https://github.com/microsoft/playwright/discussions/10941#discussioncomment-1813578)? That way, the cleanup is more explicit when reading a test and we remove the need of global variables. What are your thoughts on that?

ui/admin/tests/e2e/tests/target.spec.js
```diff
@@ -18,8 +20,11 @@ const {
   createNewHostSet,
   createNewHostInHostSet,
   createNewTarget,
+  createNewTargetWithAddress,
+  waitForSessionToBeVisible,
```
moduli, line 24: I'm cool with refactoring the first test to make use of `connectToTarget` and `waitForSessionToBeVisible`

ui/admin/tests/e2e/tests/target.spec.js
stasryzhov: I personally like how the framework capabilities (e.g. `afterEach` method) allow to get rid of duplications in test methods and move test data creation and removal out of them. In our case, we already have a `try/finally` statement used in both tests, so reusing it for the clean up makes sense 👍

ui/admin/tests/e2e/tests/target.spec.js
```diff
@@ -4,12 +4,14 @@
  */
 
 /* eslint-disable no-undef */
-const { test, expect } = require('@playwright/test');
-const { exec, execSync } = require('child_process');
-const { checkEnv, authenticatedState } = require('../helpers/general');
+const {test} = require('@playwright/test');
+const {execSync} = require('child_process');
+const {checkEnv, authenticatedState} = require('../helpers/general');
```
cameronperera, line 9: Non-blocking but I am curious as to why the spaces before and after `test` and the other imports were removed? I thought this was a setting controlled by our linter.

cameronperera (approved):
Left a small comment that isn't blocking. Looks good.

ui/admin/tests/e2e/tests/target.spec.js
stasryzhov: Thank you for pointing this out, @cameronperera 👍 
It happened because I committed changes via an IDE Git plugin and it skipped running the hooks. I've ran the hooks and fixed the formatting now 👍

ui/admin/tests/e2e/tests/target.spec.js
```diff
@@ -59,53 +65,57 @@ test('Verify session created then cancel the session', async ({ page }) => {
...
       .getByRole('cell', { name: targetName })
       .locator('..')
       .getByRole('button', { name: 'Cancel' })
       .click();
   } finally {
+    await deleteOrg(org.id);
     // End `boundary connect` process
     if (connect) {
-      connect.kill('SIGTERM');
+      connect.kill();
```
moduli, line 118: Do we want the `SIGTERM` here?

ui/admin/tests/e2e/tests/target.spec.js
moduli, line 1: When running these tests locally, there was one pop-up in the UI I was seeing that wasn't a "Success" dialog. I was seeing it occasionally when running this command `yarn run e2e:dev:chrome target.spec.js --debug` and it was occurring somewhere during the second test when it was either waiting for the session or canceling it. I'll try to reproduce it and capture what it was.

moduli: Screenshot 2023-04-14 at 2 54 20 PM

Hm. Looks like this can happen whenever we're canceling the session?

ui/admin/tests/e2e/tests/target.spec.js
stasryzhov: It applies `SIGTERM` under the hood if no argument is provided: The doc:

> If no argument is given, the process will be sent the 'SIGTERM'

We may want to specify it to highlight our intention though

ui/admin/tests/e2e/tests/target.spec.js
moduli: After discussion with Stan, we will look to investigate this in a separate ticket.

calcaide (approved):
Nice done!! these test are looking good :smiley:

ui/admin/tests/e2e/helpers/boundary-cli.js
```diff
@@ -36,3 +36,43 @@ exports.authenticateBoundaryCli = async () => {
     console.log(`${e.stderr}`);
   }
 };
+
+/**
+ * Uses the boundary CLI to connect to the specified target
+ * @param {string} targetId ID of the target to be connected to
+ */
```
moduli, line 43: This comment is missing a `returns` line

Example:
```
/**
 * Uses the UI to create a new project. Assumes you have selected the desired org.
 * @returns Name of the project
 */
```

DhariniJeeva (approved):

PR #1672, opened 2023-04-19
chore: 🤖 add json5 to resolutions per dependabot alerts

lisbet-alvarez:
Addresses dependabot alerts [#70](https://github.com/hashicorp/boundary-ui/security/dependabot/70) & [#71](https://github.com/hashicorp/boundary-ui/security/dependabot/71)

## Description
`"**/babel-core/json5": "^2.2.2"` was added to force dependencies using json5 ^0.5.1 to use ^2.2.2

[json5](https://github.com/json5/json5/releases?page=1)

### DIFF
--- a/package.json
+++ b/package.json
@@ -104,9 +104,7 @@
     "loader-utils": "^2.0.4",
     "**/@storybook/**/minimatch": "^3.0.5",
     "**/ember-electron/**/minimatch": "^3.0.5",
-    "**/babel-core/json5": "^2.2.2",
-    "**/@babel/core/json5": "^2.2.2",
-    "**/@storybook/**/json5": "^2.2.2"
+    "**/babel-core/json5": "^2.2.2"
   },
   "config": {
     "commitizen": {
--- a/package.json
+++ b/package.json
@@ -103,7 +103,10 @@
     "got": "^11.8.5",
     "loader-utils": "^2.0.4",
     "**/@storybook/**/minimatch": "^3.0.5",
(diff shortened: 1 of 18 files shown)


package.json
```diff
@@ -103,7 +103,10 @@
     "got": "^11.8.5",
     "loader-utils": "^2.0.4",
     "**/@storybook/**/minimatch": "^3.0.5",
-    "**/ember-electron/**/minimatch": "^3.0.5"
+    "**/ember-electron/**/minimatch": "^3.0.5",
+    "**/babel-core/json5": "^2.2.2",
```
ZedLi, line 107: I think you only need this one to force the `^0.5.1` to the most recent version. According to your description, the other two actually have the correct semver needed to update to the most recent version. This can be done by manually removing the out of date entries in the `yarn.lock` and redoing an install to get the latest version that matches the semver. 

I suspect if you remove the two below and run `yarn install` again nothing will change since you have the correct `yarn.lock` file now.

lisbet-alvarez: Ahh that makes so much more sense, thank you!

ZedLi (approved):

package.json
calcaide: Nicely executed @lisbet-alvarez !! As @ZedLi already mention, very clean the scope of the specific dep that needs to be resolved. Definetly a pattern to follow in the future!

calcaide (approved):
NIcely done @lisbet-alvarez !! I love the pattern of scope down the dep that needs to be resolved, kudos to you and @ZedLi for it!!

PR #1673, opened 2023-04-24
refactor: 💡 removed whitespace from layout in add-members view

dvonanderson:
✅ Closes: ICU-5986

:tickets: [Jira ticket](https://hashicorp.atlassian.net/browse/ICU-5986)

## Description
Removed whitespace from layout in `add-members` view for groups by replacing `Rose::Table` with `Hds::Table`.  We were able to achieve the same look as the previous whitespace formatting hack by using  a `Hds::Table ` and a `<HelperText>` to wrap the badge for `account_name` instead of custom `css`.  

## Note
There will be a small bit of `override` css that will globally remove margin from form-fields when uses inside of an `Hds::Table`

image

### DIFF
--- a/addons/rose/addon/styles/hds/overrides.scss
+++ b/addons/rose/addon/styles/hds/overrides.scss
@@ -17,9 +17,10 @@
   }
 }
 
-// removes field bottom margin (for tables with checkbox and badge in a cell)
+// removes field bottom margin (for tables with input elements in cell)
 .hds-table {
-  [class^=hds-form-field], [class^=hds-form-group] {
+  [class^='hds-form-field'],
+  [class^='hds-form-group'] {
     margin-bottom: 0;
   }
 }
--- a/addons/rose/addon/styles/hds/overrides.scss
+++ b/addons/rose/addon/styles/hds/overrides.scss
@@ -19,7 +19,7 @@
 
 // removes field bottom margin (for tables with checkbox and badge in a cell)
 .hds-table__td {
-  .hds-form-field--layout-flag {
+  [class^=hds-form-field], [class^=hds-form-group] {
     margin-bottom: 0;
   }
 }
--- a/addons/rose/addon/styles/hds/overrides.scss
+++ b/addons/rose/addon/styles/hds/overrides.scss
@@ -16,3 +16,10 @@
     }
   }
 }
+
+// removes field bottom margin (for tables with checkbox and badge in a cell)
+.hds-table__td {
+  .hds-form-field--layout-flag {
+    margin-bottom: 0;
+  }
+}
--- a/addons/rose/addon/styles/hds/overrides.scss
+++ b/addons/rose/addon/styles/hds/overrides.scss
@@ -18,7 +18,7 @@
 }
 
 // removes field bottom margin (for tables with checkbox and badge in a cell)
-.hds-table__td {
+.hds-table {
   [class^=hds-form-field], [class^=hds-form-group] {
     margin-bottom: 0;
   }
--- a/addons/rose/addon/styles/hds/overrides.scss
+++ b/addons/rose/addon/styles/hds/overrides.scss
@@ -17,10 +17,17 @@
   }
 }
 
-// removes field bottom margin (for tables with input elements in cell)
 .hds-table {
+  border-collapse: revert;
+
   [class^='hds-form-field'],
   [class^='hds-form-group'] {
     margin-bottom: 0;
+  }
+}
+
+.hds-dropdown {
+  .hds-dropdown-toggle-icon {
+    color: inherit;
   }
 }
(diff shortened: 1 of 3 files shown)


randallmorey (changes requested):
Can the cell contents be centered vertically?  https://helios.hashicorp.design/components/table?tab=code#vertical-alignment

dvonanderson:
> Can the cell contents be centered vertically? https://helios.hashicorp.design/components/table?tab=code#vertical-alignment

ui/admin/app/components/form/group/add-members/index.hbs
```diff
@@ -17,45 +16,38 @@
...
-                <Hds::Badge @text={{user.accountName}} />
+    <Hds::Table
+      @model={{this.availableUsers}}
+      @columns={{array
+        (hash label=(t 'form.name.label'))
+        (hash label=(t 'form.id.label'))
+        (hash label=(t 'resources.scope.title'))
+        (hash label=(t 'form.description.label'))
+      }}
+      @valign='middle'
```
DhariniJeeva, line 27: Maybe it's just me, but I like the look of the row when it is not vertically aligned here, as it tries to align with the badge from the first cell. (especially with the badge on the third column doesn't sit well with that padding)

This is with @valign='middle'
Screen Shot 2023-04-25 at 1 25 35 PM


this is without @valign='middle'
Screen Shot 2023-04-25 at 1 25 56 PM

DhariniJeeva (approved):

addons/rose/addon/styles/hds/overrides.scss
```diff
@@ -16,3 +16,10 @@
     }
   }
 }
+
+// removes field bottom margin (for tables with checkbox and badge in a cell)
+.hds-table__td {
+  .hds-form-field--layout-flag {
```
randallmorey, line 22: This suggestion overrides the exact selector used to add margin in the first place, scoped by table.
```suggestion
  [class^=hds-form-field], [class^=hds-form-group] {
```

randallmorey (changes requested):

addons/rose/addon/styles/hds/overrides.scss
```diff
@@ -16,3 +16,10 @@
     }
   }
 }
+
+// removes field bottom margin (for tables with checkbox and badge in a cell)
+.hds-table__td {
```
randallmorey, line 21: Even simpler:
```suggestion
.hds-table {
```

randallmorey (changes requested):

joshuaogle:
I looked all over and couldn't find any guidance from the HDS team, which surprises me a bit but I also don't see any complex examples in Figma. I think aligning to the center makes sense for us :+1: You would only notice the difference with long descriptions and it makes aligning these mixed components much easier.

ui/admin/app/components/form/group/add-members/index.hbs
joshuaogle: I looked all over and couldn't find any guidance from the HDS team, which surprises me a bit but I also don't see any complex examples in Figma. I think aligning to the center makes sense for us 👍 You would only notice the difference with long descriptions and it makes aligning these mixed components much easier.

joshuaogle: I do normally like to top-align the tables if they are more text heavy, but when you have text with different line-heights the text-baselines get further apart and it gets funky. I have done all sorts of tricks to make it work but I'm not sure we want to do that here. Unless you can think of a way to make them closer

DhariniJeeva: I agree it doesn't make much difference to change that, and in this case, middle alignment seems like the right path. Appreciate the feedback!

ZedLi:
I notice the table doesn't have rounded corners, it seems like we never ported over the HDS override here:

```
.hds-table {
  border-collapse: revert;
}
```

lisbet-alvarez (approved):
Nice! LGTM!

DhariniJeeva (approved):

ui/admin/app/components/form/group/add-members/index.hbs
calcaide: This is a very nice and interesting topic. The classic problem of elements vertical alignment when elements have different baselines.

I agree with @joshuaogle  here, better to keep it as simple as possible since we can get into a rabbit hole trying with vertical alignments.

I just wanted to mention, I am sure `@valign` is an abstraction of a [cell vertical-align](https://developer.mozilla.org/en-US/docs/Web/CSS/vertical-align). As a suggestion, perhaps we can try to align with value baseline rather than middle? Does that look any better?

I am fine either way, but also I think whatever decision we take, we should be consistent with it, specially thinking on the HDS adoption when we want to replace all the rose table for the HDS ones.

cc: @joshuaogle @DhariniJeeva

PR #1679, opened 2023-05-03
docs: ✏️ Add i18n documentation

calcaide:
## Description

Add Boundary developer conventions document, start with i18n conventions.

### DIFF
--- a/documentation/README.md
+++ b/documentation/README.md
@@ -4,7 +4,7 @@
 
 We use [ember-intl](https://ember-intl.github.io/ember-intl/) to manage i18n. All the code related to i18n lives within [the core addon](https://github.com/hashicorp/boundary-ui/tree/main/addons/core).
 
-All the features provided by emebr-intl are available to use, on top of that, the Boundary team [provides some custom helpers](https://github.com/hashicorp/boundary-ui/tree/main/addons/core/addon/helpers).
+All the features provided by ember-intl are available to use, on top of that, the Boundary team [provides some custom helpers](https://github.com/hashicorp/boundary-ui/tree/main/addons/core/addon/helpers).
 
 ### Dates & Time
 
--- a/documentation/README.md
+++ b/documentation/README.md
@@ -14,7 +14,7 @@ The desired date format for the Boundary teams is: `YYYY-MM-DD HH-MM-SS TZ`
 
 ### Strings
 
-Plain text strings must be avoided, we use a rules [(`no-bare-strings`)](https://github.com/ember-template-lint/ember-template-lint/blob/master/docs/rule/no-bare-strings.md) within our linter to avoid committing plain text strings.
+Plain text strings must be avoided, we use a rule [(`no-bare-strings`)](https://github.com/ember-template-lint/ember-template-lint/blob/master/docs/rule/no-bare-strings.md) within our linter to avoid committing plain text strings.
 
 To learn how to add a translation, [check documentation](https://ember-intl.github.io/ember-intl/docs/quickstart#2-add-your-first-translation).
 
--- a/documentation/README.md
+++ b/documentation/README.md
@@ -1,0 +1,23 @@
+# Boundary Developer conventions
+
+## I18n:
+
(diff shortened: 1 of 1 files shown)


documentation/README.md
```diff
@@ -0,0 +1,23 @@
+# Boundary Developer conventions
+
+## I18n:
+
+We use [ember-intl](https://ember-intl.github.io/ember-intl/) to manage i18n. All the code related to i18n lives within [the core addon](https://github.com/hashicorp/boundary-ui/tree/main/addons/core).
+
+All the features provided by emebr-intl are available to use, on top of that, the Boundary team [provides some custom helpers](https://github.com/hashicorp/boundary-ui/tree/main/addons/core/addon/helpers).
```
ZedLi, line 7:
```suggestion
All the features provided by ember-intl are available to use, on top of that, the Boundary team [provides some custom helpers](https://github.com/hashicorp/boundary-ui/tree/main/addons/core/addon/helpers).
```

ZedLi, line 7: Are the custom helpers really related to internationalization? For the date helpers specifically, in retrospect we should've just used the [date formatter](https://ember-intl.github.io/ember-intl/docs/helpers/format-date) from the addon

ZedLi (approved):
Thanks for the hard work in writing documentation!

documentation/README.md
calcaide: @ZedLi 
You are right. In retrospect, I arrive to the same conclusion, we could rely within `date formatter`. I will create a low priority ticket, to audit the usage of dates within our UI's with next goals:

- Make sure all the times and dates in our UI are using `<time>` HTML element, [documentation here](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/time) and show timezone.
- Try to rely 100% on the `ember-intl` [date formatter](https://ember-intl.github.io/ember-intl/docs/helpers/format-date).
- Make sure the default or any other format we define within `format.js` shows the timezone.
- Clean up, if need, custom helpers.
- Clean up, if need, formats not used within `format.js` [code here](https://github.com/hashicorp/boundary-ui/blob/main/ui/admin/app/formats.js). i.e: I don't think at the moment we are using any `number` ([code here](https://github.com/hashicorp/boundary-ui/blob/main/ui/admin/app/formats.js#L38-L52)), or others.
- Update documentation (basicaslly what this PR brings) with the work done if need. I.e: delete from documentation the mention to custom helpers if we end up deprecating them.

Thoughs?

calcaide:
> Thanks for the hard work in writing documentation!

@ZedLi feedback much appreciate it!! thank you!!

documentation/README.md
```diff
@@ -0,0 +1,23 @@
...
+
+### Dates & Time
+
+Dates and time must be expressed when possible using the [`<time>` HTML element](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/time).
+
+The desired date format for the Boundary teams is: `YYYY-MM-DD HH-MM-SS TZ`
+
+### Strings
+
+Plain text strings must be avoided, we use a rules [(`no-bare-strings`)](https://github.com/ember-template-lint/ember-template-lint/blob/master/docs/rule/no-bare-strings.md) within our linter to avoid committing plain text strings.
```
lisbet-alvarez, line 17:
```suggestion
Plain text strings must be avoided, we use a rule [(`no-bare-strings`)](https://github.com/ember-template-lint/ember-template-lint/blob/master/docs/rule/no-bare-strings.md) within our linter to avoid committing plain text strings.
```

lisbet-alvarez (approved):
Awesome! This documentation is very helpful! LGTM!

documentation/README.md
calcaide: @lisbet-alvarez Nice catch!!

cameronperera (approved):
Wow! This is great Carlos. Thanks for writing this.

DhariniJeeva (approved):
Thank you for putting this together! LGTM!

PR #1680, opened 2023-05-03
feat: 🎸 Download latest HCP binary for worker config commands

ZedLi:
✅ Closes: ICU-7265

:tickets: [Jira ticket](https://hashicorp.atlassian.net/browse/ICU-7265)

## Description
Changed the worker config commands to just download the latest HCP boundary binary. 
Used the `/latest` endpoint from the hashicorp releases and grabbed the URL to be used in a command substitution.

Note that our edition feature switching doesn't properly update the worker page since ember isn't aware of any changes as we use the feature check in getters. You can update it just by filling out a field on the page. 

image

### DIFF
--- a/ui/admin/app/components/form/worker/create-worker-led/index.js
+++ b/ui/admin/app/components/form/worker/create-worker-led/index.js
@@ -118,8 +118,10 @@ ${listenerText}
 sudo apt-add-repository "deb [arch=amd64] https://apt.releases.hashicorp.com $(lsb_release -cs) main" ;\\
 sudo apt-get update && sudo apt-get install boundary ;\\
 boundary server -config="${configPath}/pki-worker.hcl"`;
-    const hcpContent = `wget -q https://releases.hashicorp.com/boundary-worker/0.12.0+hcp/boundary-worker_0.12.0+hcp_linux_amd64.zip ;\\
-sudo apt-get update && sudo apt-get install unzip ;\\
+
+    const hcpContent = `sudo apt-get update && sudo apt-get install jq unzip ;\\
+wget -q "$(curl -fsSL "https://api.releases.hashicorp.com/v1/releases/boundary-worker/latest?license_class=hcp" \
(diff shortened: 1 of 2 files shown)


dvonanderson (approved):

ui/admin/app/components/form/worker/create-worker-led/index.js
```diff
@@ -118,8 +118,10 @@ ${listenerText}
 sudo apt-add-repository "deb [arch=amd64] https://apt.releases.hashicorp.com $(lsb_release -cs) main" ;\\
 sudo apt-get update && sudo apt-get install boundary ;\\
 boundary server -config="${configPath}/pki-worker.hcl"`;
-    const hcpContent = `wget -q https://releases.hashicorp.com/boundary-worker/0.12.0+hcp/boundary-worker_0.12.0+hcp_linux_amd64.zip ;\\
-sudo apt-get update && sudo apt-get install unzip ;\\
+
+    const hcpContent = `sudo apt-get update && sudo apt-get install jq unzip ;\\
+wget -q "$(curl -fsSL "https://api.releases.hashicorp.com/v1/releases/boundary-worker/latest?license_class=hcp" \
+| jq -r '.builds[] | select(.arch == "amd64" and .os == "linux") | .url')" ;\\
```
lisbet-alvarez, line 124: Non-blocking question: is the reason for having `sudo apt-get install jq unzip` on line 122 because `jq` and `unzip` might not be already installed in the user's linux system? If so, should `wget` also be added to `sudo apt-get install jq unzip`, or is that not necessary ?

lisbet-alvarez (approved):
LGTM!

calcaide (approved):
LGTM!! Thank you for taking care of this!

ui/admin/app/components/form/worker/create-worker-led/index.js
ZedLi: We could, I thought about it but `wget` is normally installed with most linux distros IIRC. `jq` and `unzip` aren't so they definitely to be installed.

PR #1682, opened 2023-05-09
fix: 🐛 support metadata access on another API host

randallmorey:
This fix adds support for fetching `metadata.json` from an arbitrary API host as specified by the `API_HOST` env var.  This is expected to affect only development environments where the UI and controller are running on different hosts/ports.


cameronperera (approved):

lisbet-alvarez (approved):
LGTM!

calcaide (approved):
Thanks for the fix!!

Tested against boundary binary and works no problem!!

PR #1688, opened 2023-05-15
Icu 9226 disable all actions for ldap accounts within users

lisbet-alvarez:
✅ Closes: ICU-9226
:tickets: [Jira ticket](https://hashicorp.atlassian.net/browse/ICU-9226)

## Description

Shows that ldap accounts do not appear in the add-accounts route
:technologist: [Admin preview #1](https://boundary-ui-git-icu-9226-disable-all-actions-f-d5c52f-hashicorp.vercel.app/scopes/global/users/u_uab8iflhfu/add-accounts)

Shows that user with an ldap account does not have the "Remove" action.
:technologist: [Admin preview #2](https://boundary-ui-git-icu-9226-disable-all-actions-f-d5c52f-hashicorp.vercel.app/scopes/global/users/u_p3rrlsb0rh/accounts)


addons/api/addon/models/account.js
```diff
@@ -5,20 +5,35 @@
 
 import GeneratedAccountModel from '../generated/models/account';
 import { computed } from '@ember/object';
-import { equal } from '@ember/object/computed';
+import {
+  TYPE_AUTH_METHOD_PASSWORD,
+  TYPE_AUTH_METHOD_OIDC,
+  TYPE_AUTH_METHOD_LDAP,
+} from './auth-method';
```
lisbet-alvarez, line 12: Im currently using the same type values declared in the auth-method model since an account will always have the same type as its auth-method. Is this reasonable or should this account model have its own type values.

randallmorey (changes requested):
@lisbet-alvarez two thoughts:
- for ease of review, can the type constant refactor be a separate PR?
- where LDAP accounts or their actions are filtered out, there seem to be hard-coded type checks.  Can these use Ember Can abilities?  This could potentially be in combination with a feature flag to make enabling them easier later on.

randallmorey (changes requested):

ui/admin/app/components/form/user/add-accounts/index.js
```diff
@@ -35,7 +36,8 @@ export default class FormUserAddAccountsComponent extends Component {
     // Get IDs for accounts already added to the current user
     const alreadyAddedAccountIDs = this.args.model.account_ids;
     const notAddedAccounts = this.args.accounts.filter(
-      ({ id }) => !alreadyAddedAccountIDs.includes(id)
+      ({ id, type }) =>
+        !alreadyAddedAccountIDs.includes(id) && type !== TYPE_AUTH_METHOD_LDAP
```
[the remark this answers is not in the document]
lisbet-alvarez: Great idea! I'll make that change, thx!

ui/admin/tests/acceptance/users/accounts-test.js
```diff
@@ -50,28 +69,47 @@ module('Acceptance | users | accounts', function (hooks) {
...
+    instances.globalUser = this.server.schema.users.findBy({
+      scopeId: 'global',
+    });
     accountsCount = instances.user.accountIds.length;
+    urls.globalScope = `/scopes/global`;
     urls.users = `/scopes/${instances.scopes.org.id}/users`;
     urls.user = `${urls.users}/${instances.user.id}`;
+    urls.globalUser = `${urls.globalScope}/users/${instances.globalUser.id}`;
     urls.accounts = `${urls.user}/accounts`;
+    urls.ldapAccounts = `${urls.globalUser}/accounts`;
```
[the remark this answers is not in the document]
lisbet-alvarez: ahh yup the name is misleading I will refactor this. Thx for catching this!

addons/api/addon/models/account.js
```diff
@@ -5,20 +5,43 @@
...
 
 export default class AccountModel extends GeneratedAccountModel {
   // =attributes
 
   /**
    * @type {boolean}
    */
-  @equal('type', 'password') isPassword;
+  get isPassword() {
```
DhariniJeeva, line 21: would this not work? `@equal('type', TYPE_AUTH_METHOD_PASSWORD) isPassword`

addons/api/addon/models/account.js
DhariniJeeva: I think it is reasonable to import this since we will always have the same type as auth-methods

addons/api/addon/models/account.js
dvonanderson: @DhariniJeeva Is there an effort to reduce computed ember magic?

lisbet-alvarez: the existing computed property works, i refactored this to use a plain getter since computed properties are not necessary anymore to get a value.

ui/admin/app/components/form/user/add-accounts/index.js
```diff
@@ -30,12 +34,14 @@ export default class FormUserAddAccountsComponent extends Component {
    * Accounts not already added to the user.
    * @type {[AccountModel]}
    */
-  @computed('args.{accounts.[],model.account_ids.[]}')
+  @computed('args.{accounts.[],model.account_ids.[]}', 'can')
```
DhariniJeeva, line 37: Do we need this computed property here?

ui/admin/app/routes/scopes/scope/users/user/add-accounts.js
```diff
@@ -27,8 +27,8 @@ export default class ScopesScopeUsersUserAddAccountsRoute extends Route {
   }
 
   /**
-   * Returns the current target, all host catalogs, and all host sets.
-   * @return {{target: TargetModel, hostCatalogs: [HostCatalogModel], hostSets: [HostSetModel]}}
+   * Returns the current user, all auth methods, and all accounts.
+   * @return {{user: UserModel, authMethods: [AuthMethodModel], accounts: [AccountModel]}}
```
DhariniJeeva, line 31: good catch

ui/admin/app/components/form/user/add-accounts/index.js
```diff
@@ -30,12 +33,14 @@ export default class FormUserAddAccountsComponent extends Component {
...
    */
-  @computed('args.{accounts.[],model.account_ids.[]}')
   get filteredAccounts() {
     // Get IDs for accounts already added to the current user
     const alreadyAddedAccountIDs = this.args.model.account_ids;
+    // TODO: Remove the addAccount ability check during Phase 2
     const notAddedAccounts = this.args.accounts.filter(
-      ({ id }) => !alreadyAddedAccountIDs.includes(id)
+      (account) =>
+        this.can.can('addAccount account', account) &&
```
ZedLi, line 42: Did we want to actually filter out the ldap accounts or just disable them to prevent users from trying to add them? Should we still let them see it?

lisbet-alvarez: Great question! Had a discussion with my pod sync and we agreed to exclude ldap accounts from the [add-accounts](https://boundary-ui-git-icu-9226-disable-all-actions-f-d5c52f-hashicorp.vercel.app/scopes/global/users/u_1r8fkwwiqj/add-accounts) route since that feels like the better ux because a user expects to see a list of accounts that could be added.

lisbet-alvarez: @joshuaogle We would also appreciate your thoughts on this.

lisbet-alvarez: Alrighty, change was applied and this is the new [preview](https://boundary-ui-git-icu-9226-disable-all-actions-f-d5c52f-hashicorp.vercel.app/scopes/global/users/u_uab8iflhfu/add-accounts).  Users can now see disabled ldap accounts in the add-accounts route.

ui/admin/app/components/form/user/add-accounts/index.hbs
```diff
@@ -30,11 +30,15 @@
...
           <body.row as |row|>
             <row.cell>
-              <form.checkbox
-                @label={{account.id}}
-                @description={{account.description}}
-                @onChange={{fn this.toggleAccount account}}
-              />
+              {{!TODO: Remove the addAccount ability check during Phase 2}}
+              <Hds::Form::Checkbox::Field
+                disabled={{if (can 'addAccount account' account) false true}}
```
ZedLi, line 35: Not sure which one reads better or if we have a convention but we also could have used `not` instead of a ternary

ZedLi (approved):
LGTM!

DhariniJeeva (approved):

calcaide (approved):
LGTM!! Thanks for this work!

randallmorey (approved):

DhariniJeeva (approved):

calcaide (approved):
LGTM!

PR #1695, opened 2023-05-23
Refactor type to use const values

lisbet-alvarez:
## Description
Some of this work was originally in a previous pr but was taken out in order to increase readabiity and separate concerns.


cameronperera (approved):

dvonanderson (approved):
Awesome work Lis

PR #1698, opened 2023-05-26
fix: 🐛 Fix serializing client secrets in oidc auth methods

ZedLi:
## Description
We are currently serializing client secrets in the OIDC auth methods form even if they're null. Annotated it as a secret so it won't get serialized if it doesn't exist 

See this [thread](https://hashicorp.slack.com/archives/C012GTH1C4E/p1685109648980819) for more details.

### DIFF
--- a/addons/api/addon/generated/models/auth-method.js
+++ b/addons/api/addon/generated/models/auth-method.js
@@ -61,7 +61,7 @@ export default class GeneratedAuthMethodModel extends BaseModel {
 
   @attr('string', { isNestedAttribute: true }) client_id;
 
-  @attr('string', { isNestedAttribute: true }) client_secret;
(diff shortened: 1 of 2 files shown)


dvonanderson (approved):

cameronperera (approved):

jimlambrt:
ty!

randallmorey (changes requested):
Would the `isSecret` attribute option work for this?

ZedLi:
> Would the `isSecret` attribute option work for this?

I mentioned this part as part of the comment but no because the `isNestedSecret` assumes we are serializing it under a `secrets` object but I believe this route currently assumes it's just under `attributes` still.

DhariniJeeva (approved):

randallmorey (approved):

PR #1701, opened 2023-05-30
fix: 🐛 radio card group name

DhariniJeeva:
https://hashicorp.atlassian.net/browse/ICU-9510

This issue is due to using the same name for two radio groups within the same form

 Before:
Screenshot 2023-05-30 at 4 35 41 PM

After:
Screenshot 2023-05-30 at 4 36 13 PM

### DIFF
--- a/ui/admin/app/components/form/host-catalog/static/index.hbs
+++ b/ui/admin/app/components/form/host-catalog/static/index.hbs
@@ -71,32 +71,6 @@
         </G.RadioCard>
       {{/each}}
     </Hds::Form::RadioCard::Group>
-
-    {{#if @model.isPlugin}}
-
-      <Hds::Form::RadioCard::Group @name='type' as |G|>
-        <G.Legend>{{t 'titles.provider'}}</G.Legend>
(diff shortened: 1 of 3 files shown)


ui/admin/app/components/form/host-catalog/static/index.hbs
```diff
@@ -71,32 +71,6 @@
         </G.RadioCard>
       {{/each}}
     </Hds::Form::RadioCard::Group>
-
-    {{#if @model.isPlugin}}
```
DhariniJeeva, line 75: This block is not being used as we [default](https://github.com/hashicorp/boundary-ui/blob/main/ui/admin/app/routes/scopes/scope/host-catalogs/new.js#L58) to `aws` if the model type is `plugin` and the [root](https://github.com/hashicorp/boundary-ui/blob/main/ui/admin/app/components/form/host-catalog/index.hbs#L7) `index.hbs` takes you to the aws form.

lisbet-alvarez (approved):
Nice Work!! LGTM!

randallmorey (approved):

PR #1707, opened 2023-06-07
Optimise Desktop API queries

randallmorey:
:tickets: [icu-9554](https://hashicorp.atlassian.net/browse/icu-9554)

## Description

Improves performance of Desktop API queries:

- Removes unnecessary supplemental queries
- Removes bexpr filter expressions, filtering instead on the client
- Disables polling to reduce number of queries originating from Desktop


calcaide (approved):
Nice!! thanks for this!

Tested manually and validated all the points mark on the description.

cameronperera (approved):

PR #1713, opened 2023-06-13
Chore icu 9072 audit resolutions

calcaide:
:tickets: [Jira ticket](https://hashicorp.atlassian.net/browse/ICU-9072)

### caniuse-lite: 
- After running `yarn why caniuse-lite` is used by different deps.
- After deleting `caniuse-lite` from resolutions, resolves a higer version than the one we had in resolutions. We can delete safely from resolutions `caniuse-lite`.

How to test:
- Build API and Rose addons succesfully.

### immer
- After running `yarn why immer` is used by storybook deps.
- Unfortunetly, if we delete `immer` in resolutions, we start resolving a version `8.0.1` which has a security vulnerability.
- We will keep `immer` in resolutions but being very specific, just affects the `immer` within `@storybook`.

No need to test, because there are no changes within `yarn.lock` neither on the version resolved.

### node-notifier
- After running `yarn why node-notifier` is used by `ember-cli`.
- After deleting `node-notifier` from resolutions we resolve a higher version. Before `8.0.2` to `10.0.0`.

How to test:
- Build succesfully addons + UI's (to validate ember-cli).

### DIFF
--- a/package.json
+++ b/package.json
@@ -50,7 +50,6 @@
   },
   "resolutions": {
     "request": "^2.88.0",
-    "immer": "^9.0.6",
     "node-notifier": "^8.0.1",
     "highlight.js": "^10.4.1",
     "hoek": "~4.2.0",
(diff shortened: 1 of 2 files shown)


ZedLi (approved):
Thanks for this work!!

lisbet-alvarez (approved):
Awesome work!

PR #1716, opened 2023-06-13
chore: 🤖 remove deprecated set-output cmd from jira.yml

### LINKED ISSUE #1689: GitHub Actions - deprecated warnings found - action required!
Workflow Name: Jira Sync
Branch: main
Run URL: https://github.com/hashicorp/boundary-ui/actions/runs/4955222045

save-state deprecation warnings: 0
set-output deprecation warnings: 0
node12 deprecation warnings: 1

Please review these deprecation warnings as soon as possible and merge in the necessary updates.

GitHub will be removing support for these commands and plan to fully disable them on 31st May 2023. At this time, any workflow that still utilizes these commands will fail. See https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/.

GitHub have not finalized a date for deprecating node12 yet but have indicated that this will be summer 2023. So it is advised to switch to node16 asap. See https://github.blog/changelog/2022-09-22-github-actions-all-actions-will-begin-running-on-node16-instead-of-node12/.

If you need any help, please reach out to us in #team-rel-eng.

lisbet-alvarez:
✅ Closes: ICU-9783
closes #1689 
:tickets: [Jira ticket](https://hashicorp.atlassian.net/browse/ICU-9783)

## Description
remove deprecated set-output cmd from jira.yml workflow


cameronperera (approved):
LGTM.

.github/workflows/jira.yml
```diff
@@ -25,7 +25,7 @@ jobs:
         else
           echo "Actor ${{ github.actor }} is not a ${TEAM} team member"
         fi
-        echo "::set-output name=role::${ROLE}"
+        echo "role=${ROLE}" >> $GITHUB_OUTPUT
       env:
         GITHUB_TOKEN: ${{ secrets.JIRA_SYNC_GITHUB_TOKEN }}
```
ZedLi, line 30: Unrelated but I assume it's broken bc the token is incorrect. I wonder if we can just use the `${{ secrets.GITHUB_TOKEN }}` or if it actually needs extra permissions to get the team membership

ZedLi (approved):

calcaide (approved):
LGTM!

.github/workflows/jira.yml
lisbet-alvarez: I think you are correct about their being an issue with the token! According to [these docs](https://docs.github.com/en/rest/guides/getting-started-with-the-rest-api?apiVersion=2022-11-28#:~:text=Instead%20of%20using%20the%20gh%20auth%20login%20command%2C%20pass%20your%20token%20as%20an%20environment%20variable%20called%20GH_TOKEN) github does recommend on using `GITHUB_TOKEN` in order to authenticate to use the `hub api`. I also think the main issue could be with the env variable name because in the docs the name is `GH_TOKEN` not `GITHUB_TOKEN`.

PR #1718, opened 2023-06-14
chore: 🤖 add .copywrite.hcl config file

cameronperera:
## Description

I added this `.copywrite.hcl` config file via `copywrite init` and updated which files and directories to ignore. Please take a look at the ignore rules I have added. I can add or remove specific rules if needed.


dvonanderson (approved):
LGTM!

PR #1722, opened 2023-06-15
Chore icu 9075 audit resolutions

calcaide:
:tickets: [Jira ticket](https://hashicorp.atlassian.net/browse/ICU-9075)

### underscore
- After running `yarn why underscore` 3rd party deps uses it.
- After deleting `underscore` from resolutions, 1 of the deps (`nomnom`) still resolving `underscore@1.6.0` which has security vulnerabilities, as per [synk information](https://security.snyk.io/package/npm/underscore). We will add a more specific resolution.

### hosted-git-info
- After running `yarn why hosted-git-info` 3rd party deps uses it, also different versions are used.
- After deleting `hosted-git-info` from resolutions, just 1 dep still resolving `hosted-git-info@2.8.9` with has security vulnerabilities, as per [synk information](https://security.snyk.io/package/npm/hosted-git-info). We will add a more specific resolution.

### trim
- After running `yarn why trim` just one dep uses it.
- After deleting `trim` from resolutions we resolve `trim@0.0.1` wich has security vulnerabilities as per [synk information](https://security.snyk.io/package/npm/trim). We will add a more specific resolution.

CI test run with all these changes, [link here](https://github.com/hashicorp/boundary-ui-releases/actions/runs/5273631945).


ZedLi (approved):
Great wor

package.json
```diff
@@ -91,7 +88,10 @@
     "**/ember-electron/**/minimatch": "^3.0.5",
     "**/babel-core/json5": "^2.2.2",
     "**/react-syntax-highlighter/refractor/prismjs": "^1.27.0",
-    "**/@storybook/**/immer": "^9.0.6"
+    "**/@storybook/**/immer": "^9.0.6",
+    "**/nomnom/underscore": "^1.12.1",
+    "**/normalize-package-data/hosted-git-info": "^3.0.8",
+    "**/remark-parse/trim": "~1.0.1"
```
ZedLi, line 94: It doesn't matter too much but just curious does it actually need to be `~` vs `^`?

calcaide: @ZedLi 
I just copied what was used before to resolve `trim`. But you are right, in this case I would say `^` is better, because it includes for minor patches to be resolve, while `~` will just allow patches. I will update the PR.
Nice catch!! and appreciate the feedback :wink:

DhariniJeeva (approved):

lisbet-alvarez (approved):
🙌

cameronperera (approved):

PR #1727, opened 2023-06-16
Chore icu 9123 upgrade storybook

calcaide:
:tickets: [Jira ticket](https://hashicorp.atlassian.net/browse/ICU-9123).

## Description

**BE AWARE:** this PR will not succesfully pass the CI run until [ICU-9076](https://github.com/hashicorp/boundary-ui/pull/1724) & [ICU-9077](https://github.com/hashicorp/boundary-ui/pull/1725) are merged. The work done on those PR's will allow Storybook to upgrade.

- Deprecate `--static-dir` CLI flag, [here info](https://github.com/storybookjs/storybook/blob/next/MIGRATION.md#deprecated---static-dir-cli-flag).
- Add webpack5 as builder, [here info](https://github.com/storybookjs/storybook/blob/next/MIGRATION.md#webpack-5)
- Upgrade Storybook from `6.3.8` to `6.4.0`.
- Fix unmeet Peer deps warnings.

[ICU-9076]: https://hashicorp.atlassian.net/browse/ICU-9076?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[ICU-9077]: https://hashicorp.atlassian.net/browse/ICU-9077?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


addons/rose/package.json
```diff
@@ -66,11 +66,12 @@
     "@storybook/cli": "^6.3.10",
     "@storybook/components": "^6.3.12",
     "@storybook/core-events": "^6.3.0",
-    "@storybook/ember": "^6.3.8",
+    "@storybook/ember": "6.4.0",
     "@storybook/ember-cli-storybook": "^0.6.0",
     "@storybook/theming": "^6.3.0",
     "babel-loader": "^8.2.5",
     "babel-plugin-ember-modules-api-polyfill": "^2.12.0",
+    "babel-plugin-htmlbars-inline-precompile": "3",
```
ZedLi, line 74: Why is this needed and how come it's pinned to `3`? It looks like it's an archived package

addons/rose/package.json
```diff
@@ -66,11 +66,12 @@
     "@storybook/cli": "^6.3.10",
     "@storybook/components": "^6.3.12",
     "@storybook/core-events": "^6.3.0",
-    "@storybook/ember": "^6.3.8",
+    "@storybook/ember": "6.4.0",
```
ZedLi, line 69: Will we also continue upgrading to `6.5` and maybe `7.0`? IIRC the fix needed to get ember somewhat working with storybook was made in 7 but also eventually backported to `6.5`. 

I had tested a little while back after they had backported but still couldn't get it to work but with all these dependency fixes we might be there?

ZedLi (approved):
Super excited to get this upgraded, this has been a real thorn! Thanks for doing this work!!

addons/rose/package.json
calcaide: It is a peerDep dependency, I know is pin to 3 and not ideal, but is how storybook 6.4 has it defined. Hopefully, soon we will upgrade Storybook and this could be gone.

I left a screenshot I took during development that shows the peer dep, look at the ones pending on `rose`

Screenshot from 2023-06-15 15-37-06

addons/rose/package.json
calcaide: Absolutley!! We plan on keep the upgrade going on, but first, after my experience working with so many dependency jungle is that we should move forward auditing resolutions, because that was in first place what unlocked this upgrade :smile:
I know it sounds cumbersome but I think is the smart way to work wich such depedency hell.

lisbet-alvarez (approved):

PR #1734, opened 2023-06-20
chore: 🤖 add config for copywrite tool and compliance scan

cameronperera:
## Description

add config for copywrite tool and compliance scan

Please let me know if we should make any changes.

### DIFF
--- a/.copywrite.hcl
+++ b/.copywrite.hcl
@@ -1,0 +1,20 @@
+schema_version = 1
+
+project {
+  license        = "MPL-2.0"
+  copyright_year = 2020
+
+  # (OPTIONAL) A list of globs that should not have copyright/license headers.
+  # Supports doublestar glob patterns for more flexibility in defining which
+  # files or folders should be ignored
+  header_ignore = [
+    ".github/**",
+    ".husky/**",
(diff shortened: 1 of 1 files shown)


.copywrite.hcl
```diff
@@ -0,0 +1,20 @@
+schema_version = 1
+
+project {
+  license        = "MPL-2.0"
+  copyright_year = 2020
```
ZedLi, line 5: Shouldn't this be `2023`?

ZedLi (approved):

.copywrite.hcl
cameronperera: It defaults to the year the repo was created according to the [README](https://github.com/hashicorp/copywrite#config-structure) for the copywrite tool. This entry is optional though so we could exclude it if we wanted to.

> (OPTIONAL) Represents the year that the project initially began
  Default: <the year the repo was first created>
  copyright_year = 0

dvonanderson (approved):

DhariniJeeva (approved):

lisbet-alvarez (approved):
Looks great!

PR #1746, opened 2023-06-27
refactor: 💡 add support for named block content in form actions

lisbet-alvarez:
✅ Closes: ICU-9507

:tickets: [Jira ticket](https://hashicorp.atlassian.net/browse/ICU-9507)

## Description
Add support for named block content in rose form actions.
Checkout ticket for details.

:technologist: [Admin preview](https://boundary-ui-git-icu-9507-add-support-for-block-cef1e7-hashicorp.vercel.app/scopes/global/authenticate/am_3enprm7y2t)

:desktop_computer: [Desktop preview](https://boundary-ui-desktop-git-icu-9507-add-support-f-74fad1-hashicorp.vercel.app/cluster-url)

### DIFF
--- a/addons/rose/addon/components/rose/form/actions/index.hbs
+++ b/addons/rose/addon/components/rose/form/actions/index.hbs
@@ -4,17 +4,28 @@
 }}
 
 <Hds::ButtonSet class='rose-form-actions' ...attributes>
-  <Hds::Button
-    @text={{@submitText}}
-    disabled={{@submitDisabled}}
-    type='submit'
-  />
-  {{#if this.showCancel}}
+  {{#if (has-block 'actions')}}
+    {{yield
+      (hash
+        submitDisabled=@submitDisabled
+        cancelDisabled=@cancelDisabled
+        cancel=@cancel
+      )
+      to='actions'
+    }}
+  {{else}}
     <Hds::Button
-      @text={{@cancelText}}
-      @color='secondary'
-      disabled={{@cancelDisabled}}
-      {{on 'click' @cancel}}
+      @text={{@submitText}}
+      disabled={{@submitDisabled}}
+      type='submit'
     />
+    {{#if this.showCancel}}
+      <Hds::Button
+        @text={{@cancelText}}
+        @color='secondary'
+        disabled={{@cancelDisabled}}
+        {{on 'click' @cancel}}
+      />
+    {{/if}}
   {{/if}}
 </Hds::ButtonSet>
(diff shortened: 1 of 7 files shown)


addons/rose/addon/components/rose/form/actions/index.hbs
```diff
@@ -4,17 +4,28 @@
...
-    disabled={{@submitDisabled}}
-    type='submit'
-  />
-  {{#if this.showCancel}}
+  {{#if (has-block 'actions')}}
+    {{yield
+      (hash
+        submitDisabled=@submitDisabled
+        cancelDisabled=@cancelDisabled
+        cancel=@cancel
```
ZedLi, line 12: Do we need to pass these variables down as a convenience? Aren't they already available from the form?

My take on this is if we're yielding something custom that falls out of the normal use case, they can use whatever they need from the form component directly rather than us trying to pass down all the things they might need.

Any thoughts?

ZedLi (approved):
Looks great!

addons/rose/addon/components/rose/form/actions/index.hbs
lisbet-alvarez: That's a good point, `submitDisabled` and `cancelDisabled` are essentially the same thing as `form.disabled`. `cancelDisabled` does have a bit more of logic but we could leave that up to each use case. I think `cancel` might be the only one that is needed here since it uses `handleCancel` ?

ZedLi: I think you could still do `form.handleCancel` if you needed access to that?

DhariniJeeva: Yup, I also think we don't need to pass it down as they are all already available through the `form` and can be accessed through that and keep this component simple
    `{{yield to='actions'}}`

DhariniJeeva (approved):
Awesome job! LGTM! 🎉

addons/rose/addon/components/rose/form/actions/index.hbs
calcaide: I agree with @ZedLi and @DhariniJeeva . I like the idea of calling those as API `form.handleCancel`. That is more intuitive (IMHO).

calcaide (approved):
Nice one @lisbet-alvarez !! good call preping the field for future work :star_struck:

addons/rose/addon/components/rose/form/actions/index.hbs
lisbet-alvarez: Thank you @ZedLi , @DhariniJeeva , and @calcaide! Updated! 😎

PR #1749, opened 2023-06-29
fix: 🐛 Removing status badging from desktop client

dvonanderson:
✅ Closes: ICU-9790

:tickets: [Jira ticket](https://hashicorp.atlassian.net/browse/ICU-9790)

## Description
Removed status badges for `targets` and `sessions` in the desktop client app.

Screenshot 2023-06-29 at 11 08 08

Screenshot 2023-06-29 at 11 08 22

Screenshot 2023-06-29 at 11 08 35

### DIFF
--- a/ui/desktop/app/templates/scopes/scope/projects/sessions/index.hbs
+++ b/ui/desktop/app/templates/scopes/scope/projects/sessions/index.hbs
@@ -122,6 +122,7 @@
             <row.headerCell>{{t 'resources.target.title'}}</row.headerCell>
             <row.headerCell>{{t 'form.proxy.label'}}</row.headerCell>
             <row.headerCell>{{t 'form.started.label'}}</row.headerCell>
+            <row.headerCell>{{t 'form.status.label'}}</row.headerCell>
             <row.headerCell />
           </header.row>
         </table.header>
@@ -171,6 +172,9 @@
                 <time datetime={{format-date-iso model.created_time}}>
                   {{format-date-iso-human model.created_time}}
                 </time>
+              </row.cell>
+              <row.cell>
+                <SessionStatus @model={{model}} />
               </row.cell>
               <row.cell>
                 {{#if model.isCancelable}}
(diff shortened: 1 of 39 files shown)


ui/desktop/app/templates/scopes/scope/projects/sessions/index.hbs
```diff
@@ -173,9 +172,6 @@
                   {{format-date-iso-human model.created_time}}
                 </time>
               </row.cell>
-              <row.cell>
-                <SessionStatus @model={{model}} />
-              </row.cell>
```
ZedLi, line 178: Did we want to remove this? I thought we were just removing the active label for targets

dvonanderson: @ZedLi @randallmorey I think this was the main issue. The status badge doesn't update if the user terminates the session, via the user's terminal, only when the `cancel` action is pressed.

ZedLi: I believe that's expected since killing the SSH connection doesn't end the session. See this [thread](https://hashicorp.slack.com/archives/C012GTH1C4E/p1686693010861389). 

Theoretically we could fix the active badge on the target issue by refreshing both sessions and targets when manually refreshing so the sessions gets updated as well which would then correctly know if its active, but it probably makes more sense to remove it. 

I would assume we want to keep the session status though.

dvonanderson: @ZedLi thank you for the information let me work through a quick refactor

DhariniJeeva (approved):

ui/desktop/app/templates/scopes/scope/projects/targets/target/-header.hbs
ZedLi, line 1: Where is this header being used?

ZedLi (approved):
Looks good!

PR #1750, opened 2023-06-29
chore: 🤖 Update semver

ZedLi:
## Description
Just updating semver in the yarn.lock file since it looks like this dependabot [alert](https://github.com/hashicorp/boundary-ui/pull/1738) wasn't fully resolved as we still had some versions still resolving to `7.3.5`.


calcaide (approved):
oops!! forgot the `electron-app.package.json` when updated.
Awesome work!!! Thanks for this!!

PR #1753, opened 2023-07-04
Chore icu 9081 audit resolutions

calcaide:
:tickets: [Jira ticket](https://hashicorp.atlassian.net/browse/ICU-9081)

### jsonpointer
- After running `yarn why jsonpointer` no dependencies use it.
- After deleting `jsonpointer` from resolutions, yarn.lock does not resolve it anymore, so it is safe to delete from resolutions.

### json-schema
- After running `yarn why json-schema` ui/desktop uses it.
- After deleting `json-schema` from resolutions, we resolve an older version with security vulnerabilities, as per [snyk information](https://security.snyk.io/package/npm/json-schema). We will clean up `json-schema` resolution for `electron-app/package.json` project, and modify to a more specific resolutions within root `package.json`.

### follow-redirects
- After running `yarn why follow-redirects` one dep uses it.
- After deleting `follow-redirects` from resolutions, we keep resolving `1.15.2` which has no security vulnerabilities as per [synk information](https://security.snyk.io/package/npm/follow-redirects).

### DIFF
--- a/package.json
+++ b/package.json
@@ -53,7 +53,6 @@
   },
   "resolutions": {
     "request": "^2.88.0",
-    "ember-cli-babel": "^7.26.8",
     "jsonpointer": "5.0.0",
     "json-schema": "^0.4.0",
     "follow-redirects": "^1.14.8",
--- a/package.json
+++ b/package.json
@@ -53,7 +53,6 @@
   },
   "resolutions": {
     "request": "^2.88.0",
-    "ansi-regex": "^5.0.1",
     "ember-cli-babel": "^7.26.8",
     "jsonpointer": "5.0.0",
     "json-schema": "^0.4.0",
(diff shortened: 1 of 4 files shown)


dvonanderson (approved):

ZedLi (approved):
Looks good but i had question about this:

> We will clean up json-schema resolution for electron-app/package.json project, and modify to a more specific resolutions within root package.json.

What did you mean by this? It looks like it was removed from the yarn.lock

calcaide:
> 

@ZedLi 

You are right, I should have explained this better.

We had `json-schema` declared in resolutions within `ui/desktop/electron-app/package.json`. Which made that project resolves `json-schema` even tho if you do not use it within `ui/desktop/electron-app/package.json` but we do within `ui/desktop/package.json`. 
So deleting the resolution within `ui/desktop/electron-app/package.json` forces no more yarn to resolve it within this project, for that reason, `ui/desktop/electron-app/yarn.lock` deletes the `json-schema` entry.

Within `boundary-ui/package.json` we add an specific resolution for `json-schema`. The reason for it is that `ui/desktop/package.json` uses it through `@electron-forge`.

PR #1754, opened 2023-07-07
refactor: 💡 use hds checkbox field instead of rose checkbox

lisbet-alvarez:
✅ Closes: ICU-9423

:tickets: [Jira ticket](https://hashicorp.atlassian.net/browse/ICU-9423)

## Description
* Refactor uses of Rose checkboxes to HDS checkbox field.
* Any Rose table that used checkboxes was also refactored to Hds table because the Rose table interfered with the Hds checkbox alignment.

Before:
image
image

After:
image
image


addons/api/mirage/scenarios/ipc.js
```diff
@@ -172,10 +172,12 @@ export default function initializeMockIPC(server, config) {
     /**
      * Do nothing when attempting to minimize, toggle fullscreen,
      * and close a browser window
+     * * Do nothing when canceling a session invokes the stop method
      */
     minimizeWindow() {}
     closeWindow() {}
     toggleFullscreenWindow() {}
+    stop() {}
```
lisbet-alvarez, line 180: This was added because cancelSession actions in desktop ui using `this.ipc.invoke('stop, {...}` were causing an error due to `stop()` not being defined. This error prevented the success toast from displaying. This is only an issue in mirage. 
image

ui/admin/app/components/form/host-set/add-hosts/index.js
```diff
@@ -41,14 +41,14 @@ export default class FormHostSetAddHostsIndexComponent extends Component {
...
   @action
-  toggleHost(host) {
-    if (!this.selectedHostIDs.includes(host.id)) {
-      this.selectedHostIDs.addObject(host.id);
+  toggleHost(hostId) {
+    if (!this.selectedHostIDs.includes(hostId)) {
+      this.selectedHostIDs.addObject(hostId);
     } else {
-      this.selectedHostIDs.removeObject(host.id);
+      this.selectedHostIDs.removeObject(hostId);
```
lisbet-alvarez, line 51: I refactored this action (and other similar ones) to just use the resource id as the parameter instead of the model. Since only the id is really necessary. Thoughts ?

ui/desktop/app/controllers/scopes/scope/projects/sessions/index.js
```diff
@@ -15,7 +16,8 @@ export default class ScopesScopeProjectsSessionsIndexController extends Controll
    */
   get sorted() {
     const sessions = this.model;
-    const sortedSessions = sessions
+    if (sessions.length <= 1) return sessions;
+    const sortedSessions = A(sessions)
```
lisbet-alvarez, line 20: I encountered an error when filtering sessions. Seems like using an ember array fixes that, im assuming because `sortBy` is a method particular to ember ?
image

ui/admin/app/components/form/target/add-host-sets/index.hbs
```diff
@@ -17,37 +17,33 @@
...
+              {{on 'change' (fn this.toggleHostSet B.data.id)}}
+              as |F|
+            >
+              <F.Label>{{B.data.id}}</F.Label>
+              <F.HelperText>{{B.data.description}}</F.HelperText>
+            </Hds::Form::Checkbox::Field>
+          </B.Td>
+          <B.Td>{{B.data.name}}</B.Td>
+          <B.Td><HostCatalogTypeBadge @model={{B.data}} /></B.Td>
+          <B.Td><code>{{B.data.host_catalog_id}}</code></B.Td>
```
lisbet-alvarez, line 43: Should the `host_catalog_id` be embedded in a `<Copyable>` component ?

DhariniJeeva: This route is for adding host sources to a target. Not sure what the user would do with the copied host catalog id from this view, so I don't think it is necessary to wrap it with copyable. thoughts?

lisbet-alvarez: Good point, ill leave it as is then. Thank you!

ui/admin/app/components/form/auth-method/oidc/index.hbs
```diff
@@ -81,15 +81,24 @@
...
+      )
     }}
-  />
+    as |F|
+  >
+    <F.Label>{{t 'form.disable_discovered_config_validation.label'}}</F.Label>
+    <F.HelperText>{{t
+        'form.disable_discovered_config_validation.help'
+      }}</F.HelperText>
+  </Hds::Form::Checkbox::Field>
```
DhariniJeeva, line 101:
```suggestion
    <F.HelperText>
      {{t 'form.disable_discovered_config_validation.help'}}
    </F.HelperText>
  </Hds::Form::Checkbox::Field>
```

ui/desktop/app/controllers/scopes/scope/projects/sessions/index.js
```diff
@@ -15,7 +16,8 @@ export default class ScopesScopeProjectsSessionsIndexController extends Controll
    */
   get sorted() {
     const sessions = this.model;
-    const sortedSessions = sessions
+    if (sessions.length <= 1) return sessions;
```
DhariniJeeva, line 19:
```suggestion
    if (!sessions.length) return sessions;
```

DhariniJeeva (approved):
Nice work! LGTM!

ui/desktop/app/controllers/scopes/scope/projects/sessions/index.js
ZedLi: That's surprising, do you know at what point it's returning a normal array instead of ember array from the model? I'm curious how this was working before

ui/admin/app/components/form/credential-store/vault/index.hbs
```diff
@@ -199,25 +199,31 @@
...
 
-  <form.checkbox
-    @label={{t 'resources.credential-store.form.tls_skip_verify.label'}}
-    @helperText={{t 'resources.credential-store.form.tls_skip_verify.help'}}
-    @checked={{@model.tls_skip_verify}}
-    @onChange={{fn
-      (mut @model.tls_skip_verify)
-      (if @model.tls_skip_verify false true)
+  <Hds::Form::Checkbox::Field
+    @isInvalid={{@model.errors.tls_skip_verify}}
```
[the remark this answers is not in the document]
lisbet-alvarez: Good catch! Updated!

calcaide (approved):
@lisbet-alvarez terrific job!!! awesome work!!

ui/desktop/app/controllers/scopes/scope/projects/sessions/index.js
ZedLi: Does an ember array `.filter` not return it as another ember array? I'm just trying to understand where it becomes a normal array and whether we should actually be converting to an ember array upstream

ZedLi: I might do the actual converting in the `model` hook since the expectation might be that it should have already been an ember array for the consumers, any thoughts?

ZedLi (approved):
Looks good! Left some non blocking comments

ui/desktop/app/controllers/scopes/scope/projects/sessions/index.js
lisbet-alvarez: Yes, I agree, and thats a good point! Updated!

PR #1758, opened 2023-07-07
chore: 🤖 Patch @storybook and all its deps to ~6.4.0

calcaide:
:tickets: [Jira ticket](https://hashicorp.atlassian.net/browse/ICU-9958)

## Description

Patch `@storybook` and all the dependencies to `~6.4.0`.


addons/rose/package.json
```diff
@@ -57,18 +57,18 @@
...
-    "@storybook/addon-docs": "~6.3.8",
-    "@storybook/addons": "^6.3.9",
-    "@storybook/api": "^6.3.12",
-    "@storybook/cli": "^6.3.10",
-    "@storybook/components": "^6.3.12",
-    "@storybook/core-events": "^6.3.0",
-    "@storybook/ember": "6.4.0",
+    "@storybook/addon-a11y": "~6.4.0",
+    "@storybook/addon-actions": "~6.4.0",
+    "@storybook/addon-controls": "~6.4.0",
```
calcaide, line 62: This dependency is "lowered" to align with all the `@storybook` dependencies being at `~6.4.0`.

ZedLi: Yeah I think this is better, I've noticed there's been incompatibilities between different minor versions of storybook addons so we should just make sure they're consistent

DhariniJeeva (approved):

lisbet-alvarez (approved):

PR #1762, opened 2023-07-11
[ICU-10024] feat: Revise auth-method model

dvonanderson:
:tickets: [Jira ticket](https://hashicorp.atlassian.net/browse/ICU-10024)

## Description
Refactor `auth-method` model to include LDAP attributes.


addons/api/addon/transforms/account-attribute-map-array.js
```diff
@@ -0,0 +1,16 @@
+import Transform from '@ember-data/serializer/transform';
+
+export default class AccountAttributeMapArrayTransform extends Transform {
```
ZedLi, line 3: Is this literally the same as `AccountClaimMapArrayTransform`? I don't think we should duplicate the exact same transform if so

dvonanderson: Yea was going to pose that question.  Should we rename and refactor it for use with both?

addons/api/addon/generated/models/auth-method.js
```diff
@@ -101,9 +101,79 @@ export default class GeneratedAuthMethodModel extends BaseModel {
   })
   idp_ca_certs;
 
-  @attr('account-claim-map-array', {
+  @attr('account-value-map-array', {
     emptyArrayIfMissing: true,
     isNestedAttribute: true,
   })
   account_claim_maps;
+
```
DhariniJeeva, line 109: Can we add a comment here saying these attributes are for LDAP?

addons/api/addon/serializers/auth-method.js
```diff
@@ -4,7 +4,28 @@
...
+    'callback_url',
+    'disable_discovered_config_validation',
+    'dry_run',
+    'claims_scopes',
+    'signing_algorithms',
+    'allowed_audiences',
+    'idp_ca_certs',
+    'account_claim_maps',
+  ],
+  ldap: [],
```
DhariniJeeva, line 27: I assume this will be populated when we work on the serializer ticket?

addons/api/addon/serializers/auth-method.js
```diff
@@ -17,11 +38,27 @@ export default class AuthMethodSerializer extends ApplicationSerializer {
    */
   serialize(snapshot) {
     switch (snapshot.record.type) {
-      case TYPE_AUTH_METHOD_OIDC:
-        return this.serializeOIDC(...arguments);
-      default:
+      case TYPE_AUTH_METHOD_PASSWORD:
         return this.serializeDefault(...arguments);
+      default:
+        return this.serializeAuthMethod(...arguments);
```
DhariniJeeva, line 44: Naming this `serializeDefault` reads better. Thoughts?

DhariniJeeva, line 44:
```suggestion
      case TYPE_AUTH_METHOD_PASSWORD:
        return this.serializeAuthMethod(...arguments);
      default:
        return this. serializeDefault(...arguments);
```

addons/api/addon/generated/models/auth-method.js
dvonanderson: @DhariniJeeva yes good call

addons/api/addon/serializers/auth-method.js
dvonanderson: @DhariniJeeva yes!

calcaide (approved):

addons/api/addon/serializers/auth-method.js
```diff
@@ -4,7 +4,28 @@
  */
 
 import ApplicationSerializer from './application';
-import { TYPE_AUTH_METHOD_OIDC } from 'api/models/auth-method';
+import { TYPE_AUTH_METHOD_PASSWORD } from 'api/models/auth-method';
+
+const fieldsByType = {
```
ZedLi, line 9: I'm not the biggest fan of this since we can start utilizing the `for` field we've conceptualized already in the model and can generalize now. It also helps documentation for field types in the model rather than the serializer. 

Doing it this way requires us to keep adding similar logic to `serializeAttribute` for each serializer that needs this logic

ZedLi (changes requested):

addons/api/addon/serializers/auth-method.js
dvonanderson: @ZedLi good point,  I like the concept of using `for` and was now of the reason asked that question in an earlier thread.  I will put in a change, and also add a ticket to switch over the `host-catalogs` serializer as well

addons/api/addon/serializers/auth-method.js
```diff
@@ -17,18 +17,35 @@ export default class AuthMethodSerializer extends ApplicationSerializer {
...
     }
   }
 
+  serializeAttribute(snapshot, json, key, attribute) {
+    super.serializeAttribute(...arguments);
+    const { type } = snapshot.record;
+    const { options } = attribute;
+
+    // If an attribute has a `for` option, it must match the
+    // record's `type`, else the attribute is excludedOptional user-set description for identification purposes
```
ZedLi, line 33: I think this comment got messed up

addons/api/addon/generated/models/auth-method.js
```diff
@@ -55,55 +55,239 @@ export default class GeneratedAuthMethodModel extends BaseModel {
   })
   version;
 
-  @attr('string', { readOnly: true, isNestedAttribute: true }) state;
+  // =attributes(oidc, ldap)
 
-  @attr('string', { isNestedAttribute: true }) issuer;
+  @attr('string', {
```
ZedLi, line 60: Does this need a `for` that includes oidc and ldap?

addons/api/addon/serializers/auth-method.js
```diff
@@ -17,18 +17,35 @@ export default class AuthMethodSerializer extends ApplicationSerializer {
...
+    // from serialization.
+    if (options?.for && options.for !== type) {
```
ZedLi, line 35: I think you might need to be able to handle the `for` as an array as well? This might be worth pulling out into the application serializer at this point.

addons/api/addon/generated/models/auth-method.js
dvonanderson: @ZedLi It works as designed, but if I include the `for` option, I would have the refactor the `changeState` method for the `adapterOptions`?

ZedLi: Not sure I understand, why is that? It's only for OIDC and LDAP right?

dvonanderson: @ZedLi Yes, but we delete the nested attributes for `password`,  so my question would be, should it not need a `for` option because the state attribute is for both?

ZedLi: Are you saying because `password` also deletes attributes we don't need it? We should still annotate since if we add another auth method in the future that assumption might not hold

addons/api/addon/serializers/auth-method.js
```diff
@@ -17,19 +17,35 @@ export default class AuthMethodSerializer extends ApplicationSerializer {
...
+
   /**
-   * Default serialization omits `attributes`.
+   * Password serialization omits `attributes`.
    * @return {object}
    */
-  serializeDefault() {
+  serializePassword() {
     let serialized = super.serialize(...arguments);
+    console.log(serialized);
```
DhariniJeeva, line 48:
```suggestion
```

addons/api/addon/serializers/auth-method.js
```diff
@@ -40,11 +56,11 @@ export default class AuthMethodSerializer extends ApplicationSerializer {
    * @param {Snapshot} snapshot
    * @return {object}
    */
-  serializeOIDC(snapshot) {
+  serializeDefault(snapshot) {
     let serialized = super.serialize(...arguments);
     const state = snapshot?.adapterOptions?.state;
     if (state) {
-      serialized = this.serializeOIDCWithState(snapshot, state);
+      serialized = this.serializeDefaultWithState(snapshot, state);
```
DhariniJeeva, line 63:
```suggestion
      serialized = this.serializeWithState(snapshot, state);
```

addons/api/addon/serializers/auth-method.js
```diff
@@ -57,7 +73,7 @@ export default class AuthMethodSerializer extends ApplicationSerializer {
    * @param {string} state
    * @return {object}
    */
-  serializeOIDCWithState(snapshot, state) {
+  serializeDefaultWithState(snapshot, state) {
```
DhariniJeeva, line 76:
```suggestion
  serializeWithState(snapshot, state) {
```

addons/api/addon/serializers/auth-method.js
```diff
@@ -17,19 +17,35 @@ export default class AuthMethodSerializer extends ApplicationSerializer {
...
 
+  serializeAttribute(snapshot, json, key, attribute) {
+    super.serializeAttribute(...arguments);
+    const { type } = snapshot.record;
+    const { options } = attribute;
+
+    // If an attribute has a `for` option, it must match the
+    // record's `type`, else the attribute excluded
+    // from serialization.
+    if (options?.for && options.for !== type && !options.for.includes(type)) {
```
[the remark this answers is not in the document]
lisbet-alvarez: i think this part is missing in case its not a nested attribute
```
} else {
    delete json[key];
 }
```

addons/api/addon/serializers/auth-method.js
```diff
@@ -17,19 +17,35 @@ export default class AuthMethodSerializer extends ApplicationSerializer {
...
     }
   }
 
+  serializeAttribute(snapshot, json, key, attribute) {
+    super.serializeAttribute(...arguments);
+    const { type } = snapshot.record;
+    const { options } = attribute;
+
+    // If an attribute has a `for` option, it must match the
+    // record's `type`, else the attribute excluded
```
DhariniJeeva, line 33:
```suggestion
    // record's `type`, else the attribute is excluded
```

addons/api/addon/serializers/auth-method.js
DhariniJeeva, line 35: since type is a string, 
```suggestion
   if (options?.for && !options.for.includes(type)) {
```

DhariniJeeva (changes requested):

addons/api/addon/serializers/auth-method.js
DhariniJeeva, line 35:
```suggestion
    if (options?.for && !options.for.includes(type)) {
```

DhariniJeeva (approved):

addons/api/addon/serializers/auth-method.js
```diff
@@ -32,7 +32,7 @@ export default class AuthMethodSerializer extends ApplicationSerializer {
     // If an attribute has a `for` option, it must match the
     // record's `type`, else the attribute is excluded
     // from serialization.
-    if (options?.for && !options.for.includes(type)) {
+    if (type !== 'password' && options?.for && !options.for.includes(type)) {
```
DhariniJeeva, line 35: why do we need this hardcoded check? Can't we make use of `for`?

dvonanderson: @DhariniJeeva I think the serialize workflow is pushing the password type into the serializeAttribute with no `nestedAttributes`, so when it gets sent through `setPassword` serializer it was getting a null. Fixed

addons/api/addon/serializers/auth-method.js
```diff
@@ -17,18 +17,32 @@ export default class AuthMethodSerializer extends ApplicationSerializer {
...
+
+    // If an attribute has a `for` option, it must match the
+    // record's `type`, else the attribute is excluded
+    // from serialization.
+    if (options?.for && !options.for.includes('oidc', 'ldap')) {
+      if (options.isNestedAttribute) {
+        delete json.attributes[key];
+      }
+    }
+  }
```
DhariniJeeva, line 39:
```suggestion
  serializeAttribute(snapshot, json, key, attribute) {
    const { type } = snapshot.record;
    super.serializeAttribute(...arguments);
    const { options } = attribute;

    // If an attribute has a `for` option, it must match the
    // record's `type`, else the attribute is excluded
    // from serialization.
    if (options?.for && !options?.for?.includes(type)) {
      if (options?.isNestedAttribute) {
        delete json?.attributes?.[key];
      }
    }
  }
```

DhariniJeeva: can you check if this works? 

the test failure was due to an empty attributes object which gets deleted in `serializeDefault`.

PR #1772, opened 2023-07-13
refactor: 💡 application serializeAttribute

DhariniJeeva:
:tickets: [Jira ticket](https://hashicorp.atlassian.net/browse/ICU-10195)

Refactor application attribute serializer method to include a check for `for` option and delete attributes not matched to that specific record type.


addons/api/addon/serializers/application.js
```diff
@@ -72,6 +73,16 @@ export default class ApplicationSerializer extends RESTSerializer {
       }
       delete json[key];
     }
+    // If an attribute has a `for` option, it must match the
+    // record's `type`, else the attribute is excluded
+    // from serialization.
+    if (options?.for && !options.for.includes(recordType)) {
```
ZedLi, line 79: Can we actually be more specific than using `includes` for strings and do a real equality check? This works fine if `for` is an array but for strings I'm concerned that in the future we can have bugs where we have a type that is a substring of another type.

ZedLi (commented):
Can we also add some tests for this functionality?

addons/api/addon/serializers/application.js
DhariniJeeva: Good point, Done!

lisbet-alvarez (approved):

dvonanderson (approved):

addons/api/addon/serializers/application.js
```diff
@@ -72,6 +73,20 @@ export default class ApplicationSerializer extends RESTSerializer {
       }
       delete json[key];
     }
+    // If an attribute has a `for` option, it must match the
+    // record's `type`, else the attribute is excluded
+    // from serialization.
```
ZedLi, line 78: We should also probably be able to handle secrets but I don't think we need to worry about it yet, maybe just make a note in the comment that we don't handle it yet and can add support if/when we need it?

ZedLi (approved):
Looks good!

cameronperera (approved):

PR #1781, opened 2023-07-14
chore: 🎸 add support for ldap managed group in serializer

DhariniJeeva:
[jira](https://hashicorp.atlassian.net/browse/ICU-10058)

updated string to string-array for group_names in model

[domain model](https://developer.hashicorp.com/boundary/docs/concepts/domain-model/managed-groups#ldap-managed-group-information-and-attributes-beta)

### DIFF
--- a/addons/api/addon/generated/models/managed-group.js
+++ b/addons/api/addon/generated/models/managed-group.js
@@ -53,6 +53,7 @@ export default class GeneratedManagedGroupModel extends BaseModel {
   // =attributes (OIDC)
 
   @attr('string', {
+    for: 'oidc',
     isNestedAttribute: true,
     description:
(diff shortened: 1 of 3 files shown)


addons/api/addon/generated/models/managed-group.js
```diff
@@ -53,6 +53,7 @@ export default class GeneratedManagedGroupModel extends BaseModel {
   // =attributes (OIDC)
 
   @attr('string', {
+    for: 'oidc',
```
DhariniJeeva, line 56: filter_string is supported only in OIDC type

cameronperera (approved):

lisbet-alvarez (approved):

PR #1786, opened 2023-07-18
refactor: 💡 refactor test and serializer

dvonanderson:
✅ Closes: ICU-10025

:tickets: [Jira ticket](https://hashicorp.atlassian.net/browse/ICU-10025)

## Description
Refactor `auth-method` mirage serializer to delete secret attributes. Refactor `auth-method` mirage serializer unit test for LDAP

### DIFF
--- a/addons/api/mirage/serializers/auth-method.js
+++ b/addons/api/mirage/serializers/auth-method.js
@@ -18,6 +18,9 @@ export default ApplicationSerializer.extend({
     const isPrimary =
       model.scope && model.scope.primaryAuthMethodId === model.id;
     json.is_primary = isPrimary;
+    delete json.attributes?.client_secret;
+    delete json.attributes?.client_certificate_key;
+    delete json.attributes?.bind_password;
     return json;
   },
 });
--- a/addons/api/tests/unit/serializers/auth-method-test.js
+++ b/addons/api/tests/unit/serializers/auth-method-test.js
@@ -10,6 +10,7 @@ import setupMirage from 'ember-cli-mirage/test-support/setup-mirage';
 import {
   TYPE_AUTH_METHOD_PASSWORD,
   TYPE_AUTH_METHOD_OIDC,
+  TYPE_AUTH_METHOD_LDAP,
 } from 'api/models/auth-method';
 
 module('Unit | Serializer | auth method', function (hooks) {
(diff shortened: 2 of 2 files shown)


addons/api/mirage/serializers/auth-method.js
```diff
@@ -18,6 +18,9 @@ export default ApplicationSerializer.extend({
     const isPrimary =
       model.scope && model.scope.primaryAuthMethodId === model.id;
     json.is_primary = isPrimary;
+    delete json.attributes?.client_secret;
+    delete json.attributes?.client_certificate_key;
+    delete json.attributes?.bind_password;
```
dvonanderson, line 23: Do we need to delete the `secret` attributes from the serialized data?

ZedLi: Shouldn't we be deleting them if the actual API and our models don't return them so we actually mimic them? We already have a pattern in mirage where we delete secrets, perhaps it's just the OIDC serializer just now correctly handling it in mirage?

addons/api/tests/unit/serializers/auth-method-test.js
```diff
@@ -271,4 +429,68 @@ module('Unit | Serializer | auth method', function (hooks) {
...
+        user_dn: 'cn=users,dc=example,dc=com',
+        user_attr: 'uid',
+        user_filter: '({{.UserAttr}}={{.Username}})',
+        enable_groups: true,
+        group_dn: 'cn=groups,dc=example,dc=com',
+        group_attr: 'cn',
+        group_filter: '(member={{.UserDN}})',
+        certificates: certificates,
+        client_certificate: clientCerticate,
+      },
```
lisbet-alvarez, line 460: Non-blocking question: to test normalization works shouldn't the `client_certificate_key` be added here and then we test that it gets set to an empty string? Im probably misunderstanding so this is more of a question for understanding.

lisbet-alvarez (approved):

addons/api/mirage/serializers/auth-method.js
calcaide: I do not remember, but @ZedLi  comment, sounds familiar to me. I want to think, we delete the secrets because the API does not expect them back, neither we use them on the ember model?
The work I can remember about this was the `normalizeNestedSecrets` on the application serializer, [code here](https://github.com/hashicorp/boundary-ui/blob/main/addons/api/addon/serializers/application.js#L277-L299)

But @DhariniJeeva comment is also to consider.

Perhaps we can set a pattern for these cases and document it in [Boundary developer conventions](https://github.com/hashicorp/boundary-ui/tree/main/documentation)? not saying we have to, but asking thoughts on it 😃

addons/api/tests/unit/serializers/auth-method-test.js
ZedLi: Yup we should have a test for normalization that confirms it sets the secrets to an empty string correctly

addons/api/mirage/serializers/auth-method.js
DhariniJeeva: this was discussed offline- yup, it was the OIDC that didn't have a pattern, we delete it in other resources in Mirage. we should be following that pattern and adding it to the docs make sense and would be useful for future references

addons/api/tests/unit/serializers/auth-method-test.js
dvonanderson: @ZedLi @lisbet-alvarez changes incoming

DhariniJeeva (approved):
LGTM!

cameronperera (approved):

addons/api/tests/unit/serializers/auth-method-test.js
```diff
@@ -262,13 +422,81 @@ module('Unit | Serializer | auth method', function (hooks) {
...
+        user_dn: 'cn=users,dc=example,dc=com',
+        user_attr: 'uid',
+        user_filter: '({{.UserAttr}}={{.Username}})',
+        enable_groups: true,
+        group_dn: 'cn=groups,dc=example,dc=com',
+        group_attr: 'cn',
+        group_filter: '(member={{.UserDN}})',
+        certificates: certificates,
+        client_certificate: clientCerticate,
+        client_certificate_key: '',
```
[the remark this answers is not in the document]
dvonanderson: @zed great point

PR #1787, opened 2023-07-18
refactor: 💡 refactor mirage factory

dvonanderson:
✅ Closes: ICU-10284

:tickets: [Jira ticket](https://hashicorp.atlassian.net/browse/ICU-10284)

## Description
Refactor `LDAP/managed-groups` factory to only return specific `auth-method` type

### Before
Screenshot 2023-07-18 at 13 20 19

### After
Screenshot 2023-07-18 at 13 19 34

### DIFF
--- a/addons/api/mirage/scenarios/default.js
+++ b/addons/api/mirage/scenarios/default.js
@@ -68,9 +68,11 @@ export default function (server) {
     const oidcAuthMethod = globalAuthMethods.filter(
       (auth) => auth.type === 'oidc'
     )[0];
+    const { type } = oidcAuthMethod;
     const managedGroups = server.createList('managed-group', 2, {
       scope,
       authMethodId: oidcAuthMethod.id,
+      type,
     });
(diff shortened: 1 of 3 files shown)


DhariniJeeva (approved):

addons/api/mirage/scenarios/default.js
```diff
@@ -68,9 +68,11 @@ export default function (server) {
     const oidcAuthMethod = globalAuthMethods.filter(
       (auth) => auth.type === 'oidc'
```
DhariniJeeva, line 69: this can be replaced with the imported type constant

dvonanderson: @DhariniJeeva yes, good catch

dvonanderson: @DhariniJeeva added

lisbet-alvarez (approved):

PR #1791, opened 2023-07-19
Icu 10307 create a new helper

DhariniJeeva:
[](url):tickets: [Jira ticket](https://hashicorp.atlassian.net/browse/ICU-10307)

 
Screenshot 2023-07-18 at 8 33 02 PM

### DIFF
--- a/addons/core/addon/helpers/truncate-list.js
+++ b/addons/core/addon/helpers/truncate-list.js
@@ -11,14 +11,16 @@ import { helper } from '@ember/component/helper';
  */
 
 export default helper(function truncateList(params /*, hash*/) {
-  const list = params[0];
-  const numberOfItems = Object.keys(list).length;
-  const items = list.map(({ value }) => value);
+  const list = params?.[0];
+  if (list) {
+    const numberOfItems = Object.keys(list).length;
+    const items = list.map(({ value }) => value);
 
-  if (numberOfItems > 3) {
-    const remainingItems = items.splice(3).length;
-    return items.join(', ') + `, +${remainingItems} more`;
-  } else {
-    return items.join(', ');
+    if (numberOfItems > 3) {
+      const remainingItems = items.splice(3).length;
+      return items.join(', ') + `, +${remainingItems} more`;
+    } else {
+      return items.join(', ');
+    }
   }
 });
--- a/addons/core/addon/helpers/truncate-list.js
+++ b/addons/core/addon/helpers/truncate-list.js
@@ -27,16 +27,13 @@ export default class extends Helper {
    * @return string
    */
 
-  compute([list, limit = 3]) {
+  compute([translation, list, limit = 3]) {
     if (list) {
       const numberOfItems = Object.keys(list).length;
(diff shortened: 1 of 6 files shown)


randallmorey (changes requested):
Good start.  A couple of questions:

* Should this allow a customizable number of max items, defaulting to 3?
* Can this use the Intl t helper so as not to hard-code the content?

dvonanderson (approved):
LGTM!

addons/core/addon/helpers/truncate-list.js
```diff
@@ -0,0 +1,45 @@
...
+
+  compute([list, limit = 3]) {
+    if (list) {
+      const numberOfItems = Object.keys(list).length;
+      const items = list.map(({ value }) => value);
+      if (numberOfItems > limit) {
+        const remainingItems = items.splice(limit).length;
+        return (
+          items.join(', ') +
+          `, +${remainingItems} ${this.intl.t('actions.more')}`
```
ZedLi, line 38: We should make this phrase entirely in the translation and take in a variable since some languages might have a different order between the number and counter

DhariniJeeva: That's a good point! will update

calcaide: Awesome catch!! Indeed with @ZedLi comment

addons/core/tests/integration/helpers/truncate-list-test.js
```diff
@@ -0,0 +1,52 @@
...
+];
+
+const listWithNoItem = [];
+
+const listWithOneItem = [{ key: 1, value: 'boundary' }];
+module('Integration | Helper | truncate-list', function (hooks) {
+  setupRenderingTest(hooks);
+  setupIntl(hooks);
+
+  test('it renders correctly', async function (assert) {
```
calcaide, line 26: None blocking comment: Can we use a more descriptive title for this test case? the rest of test cases have a very descriptive name, something like: it renders correctly when there is a list with more than 3 items?

addons/core/addon/helpers/truncate-list.js
```diff
@@ -0,0 +1,45 @@
...
+  // =compute method
+
+  /**
+   * This helper truncates the list if there are more items than the limit specified
+   * and returns the original list appended with +N remaining items
+   * @param {array} list List of items
+   * @return string
+   */
+
+  compute([list, limit = 3]) {
```
calcaide, line 30: I like making the limit a parameter 😃

addons/core/tests/integration/helpers/truncate-list-test.js
DhariniJeeva: makes sense! will add

lisbet-alvarez (approved):
LGTM!

ZedLi (approved):
Looks good!

cameronperera (approved):

calcaide (approved):
Nicely done!! thanks for this!!

PR #1797, opened 2023-07-25
refactor: 💡 update to allow state change & update edit action

lisbet-alvarez:
✅ Closes: ICU-10078 & 10079

:tickets: [Jira ticket](https://hashicorp.atlassian.net/browse/ICU-10078)
:tickets: [Jira ticket](https://hashicorp.atlassian.net/browse/ICU-10079)

## Description
Allows state change for ldap auth-method.
Verifies that cancelling changes rolls back attributes.
Verifies that deleting ldap auth method is successful.
Verifies that editing an ldap auth method is successful.
Acceptance tests for delete and update.

:technologist: [Admin preview](https://boundary-ui-git-icu-10078-enable-edit-mode-hashicorp.vercel.app/scopes/global/auth-methods/am_ec8st2tp51)

### DIFF
--- a/addons/api/addon/models/auth-method.js
+++ b/addons/api/addon/models/auth-method.js
@@ -20,6 +20,7 @@ export const TYPES_AUTH_METHOD = Object.freeze([
  * Enum options per auth method type and field.
  */
 export const options = {
+  state: ['inactive', 'active-private', 'active-public'],
   oidc: {
     signing_algorithms: [
       'RS256',
@@ -33,7 +34,6 @@ export const options = {
       'PS512',
       'EdDSA',
     ],
-    state: ['inactive', 'active-private', 'active-public'],
     account_claim_maps: {
       to: ['sub', 'name', 'email'],
     },
@@ -42,7 +42,6 @@ export const options = {
     account_attribute_maps: {
       to: ['fullName', 'email'],
     },
-    state: ['inactive', 'active-private', 'active-public'],
   },
 };
 
@@ -118,9 +117,11 @@ export default class AuthMethodModel extends GeneratedAuthMethodModel {
    */
   changeState(state, options = { adapterOptions: {} }) {
     const defaultAdapterOptions = {
-      method: 'change-state',
       state,
     };
+    if (this.isOIDC) {
+      defaultAdapterOptions.method = 'change-state';
+    }
     return this.save({
       ...options,
       adapterOptions: {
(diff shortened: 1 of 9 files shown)


addons/api/addon/serializers/auth-method.js
```diff
@@ -51,6 +57,31 @@ export default class AuthMethodSerializer extends ApplicationSerializer {
...
 
+  /**
+   * If `adapterOptions.state` is set, the serialization should
+   * include **only state** and version.  Normally, this is not serialized.
+   * @param {Snapshot} snapshot
+   * @return {object}
+   */
+  serializeLDAP(snapshot) {
+    let serialized = super.serialize(...arguments);
+    // Deleting these attributes is a temporary fix.
```
lisbet-alvarez, line 68: Changes in this file are temporary. The reason being that sending these fields to the backend prevents a user from being able to edit the form. There is also a bug related to this, visit this [slack thread](https://hashicorp.slack.com/archives/C014AHXUL3G/p1690405498990289) for more info.

addons/api/addon/models/auth-method.js
```diff
@@ -118,9 +117,11 @@ export default class AuthMethodModel extends GeneratedAuthMethodModel {
    */
   changeState(state, options = { adapterOptions: {} }) {
     const defaultAdapterOptions = {
-      method: 'change-state',
       state,
     };
+    if (this.isOIDC) {
```
DhariniJeeva, line 122: why is this only for OIDC?

addons/api/addon/serializers/auth-method.js
DhariniJeeva: I'm ok with a temp fix for now so it does not break the UI, could we add a ticket to track this? The ticket can be linked to the BE ticket so we can clean this up once the fix from BE is in.

addons/api/addon/models/auth-method.js
lisbet-alvarez: great question, only oidc has a 'change-state' endpoint. For ldap auth-methods, specifying a method causes the API to return an error. 
image

joshuaogle (approved):
Looks nice!

addons/api/addon/models/auth-method.js
DhariniJeeva: So we can't change the LDAP state after creation?

lisbet-alvarez: Ldap state can be updated, its just that specifying the 'change-state' method for ldap will cause an error. The `serializeWithState` method in the serializer takes care of sending a payload with just the state.

dvonanderson (approved):
Awesome work Lis

PR #1798, opened 2023-07-25
feat: 🎸 Create LDAP CRUD:read

dvonanderson:
✅ Closes: ICU-10064

:tickets: [Jira ticket](https://hashicorp.atlassian.net/browse/ICU-10064)

## Description
Originally this ticket was supposed to change the details view to support HMAC, but we have decided not to show the HMAC field.  This PR will serve as a quick fix for some mirage data for `client_certificate_key`

### DIFF
--- a/ui/admin/app/components/form/auth-method/ldap/index.hbs
+++ b/ui/admin/app/components/form/auth-method/ldap/index.hbs
@@ -160,7 +160,7 @@
     {{/if}}
   </Hds::Form::Fieldset>
 
-  {{#if (or @model.isNew form.isEditable)}}
+  {{#if @model.isNew}}
     <Hds::Form::TextInput::Field
       @isOptional={{true}}
       @value={{@model.client_certificate}}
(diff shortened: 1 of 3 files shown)


ui/admin/app/components/form/auth-method/ldap/index.hbs
```diff
@@ -212,6 +212,19 @@
     </Hds::Form::TextInput::Field>
   {{/if}}
 
+  {{#unless @model.isNew}}
+    <Hds::Form::TextInput::Field
```
ZedLi, line 216: I don't have context for this but do we need to show the `HMAC`? Is it useful for the user in this case, especially considering how large the form already is? 

I know we show it in the OIDC form and I think vault cred store but we stopped showing it in the newer forms we've built since then.

If we do display it though, these fields should be [readonly](https://helios.hashicorp.design/components/form/text-input#readonly-disabled-and-hidden-fields) instead of just disabled.

dvonanderson: @ZedLi This is a question I presented to Randy, whose recommendation was to show it.  Fast forward, I wanted to display both the `client_certificate,` `client_certificate_key`, `bind_dn`, and `bind_password` and do them as we do `secret_access_key`.  @DhariniJeeva and @joshuaogle what do you think?

Screenshot 2023-07-25 at 16 46 48

DhariniJeeva: Good point. The user doesn't need to see the HMAC info. I'm ok with getting rid of those from the form. We should also do this in OIDC for consistency in auth method forms. @dvonanderson, can you please add a ticket in the backlog to remove the HMAC from OIDC if we decide not to show the field?

joshuaogle: That sounds great to me too, and I like the idea of adopting that in OIDC. Thanks!

DhariniJeeva (approved):

joshuaogle (approved):

PR #1799, opened 2023-07-25
chore: 🤖 add post connection workflow feature flag

cameronperera:
✅ Closes: ICU-10269

:tickets: [Jira ticket](https://hashicorp.atlassian.net/browse/ICU-10269)

## Description

add post connection workflow feature flag and merge into LLB for post connection workflow

### DIFF
--- a/ui/desktop/config/environment.js
+++ b/ui/desktop/config/environment.js
@@ -65,6 +65,7 @@ module.exports = function (environment) {
     featureFlags: {
       search: false,
       filter: true,
+      'post-connection-workflow': false,
     },
   };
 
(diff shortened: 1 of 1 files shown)


calcaide (approved):
The beginning of post connection workflow!! let's goo!!! 🥳

ui/desktop/config/environment.js
```diff
@@ -85,6 +86,7 @@ module.exports = function (environment) {
       : true;
 
     // Enable features in development
+    ENV.featureFlags['post-connection-workflow'] = true;
```
calcaide, line 89: @cameronperera Shouldn't we include the flag within test environment as well?

cameronperera: We did in the past but we decided to move to enabling a feature flag in the test that directly needs it to be enabled.

calcaide: @cameronperera that is totally correct and I completly forget about it. Thanks for the reminder!

PR #1808, opened 2023-07-27
chore: 🤖 upgrade faker js from v7 to v8

lisbet-alvarez:
✅ Closes: ICU-10373

:tickets: [Jira ticket](https://hashicorp.atlassian.net/browse/ICU-10373)

## Description
Check out this [guide for upgrading](https://fakerjs.dev/guide/upgrading.html#deprecations-and-other-changes) from v7 to v8.

Deprecated modules such as `random`, most of `datatype` (with exception of the `boolean` method) were replaced to use other modules such as `word`, `number`, `string`.

There was also a [change](https://fakerjs.dev/guide/upgrading.html#methods-will-throw-on-empty-data-set-inputs) to the `faker.helpers.arrayElement` method.

There should be no faker-related deprecation warnings in the dev tools console with this upgrade.


addons/api/mirage/factories/credential-library.js
```diff
@@ -30,14 +30,13 @@ export default factory.extend({
...
         return {
           http_method: 'GET',
-          http_request_body: faker.random.word(),
+          http_request_body: faker.word.words(1),
           path: faker.system.directoryPath(),
         };
       case TYPE_CREDENTIAL_LIBRARY_VAULT_SSH_CERTIFICATE:
         return {
-          username: faker.random.word(),
-          key_type: faker.random.word(),
```
lisbet-alvarez, line 39: removed key_type because its a select field that already options

cameronperera (approved):
Nice work on this upgrade!

calcaide (approved):
Amazing work on the upgrade!! The refactor is delightful!! left a couple of none blocking comments 😉

addons/api/mirage/factories/account.js
```diff
@@ -32,25 +32,28 @@ export default factory.extend({
...
-        let groups = [];
-        for (let i = 0; i < groupsAmount; i++) {
-          groups.push(faker.random.word());
+        const randomNum = faker.number.int({ min: 3, max: 5 });
+        let member_of_groups = [];
+        const login_name = faker.internet.userName();
+        let dn = `uid=${login_name}`;
+        for (let i = 0; i < randomNum; i++) {
+          member_of_groups.push(faker.word.noun());
+          dn += `,dc=${faker.internet.domainSuffix()}`;
```
calcaide, line 41: None blocking comment: Can you elaborate the change in `dn`? before we were returning a string. Now we set it with an `uid` string, and later on the loop we concat a string per iteration. What is the purpose of it?

ZedLi (approved):
Looks great, thanks for this work!

addons/api/mirage/factories/target.js
```diff
@@ -11,6 +11,8 @@ import generateId from '../helpers/id';
 import { TYPES_TARGET, TYPE_TARGET_SSH } from 'api/models/target';
 
 const randomBoolean = (chance = 0.5) => Math.random() < chance;
+const randomFilter = () =>
```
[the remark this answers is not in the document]
ZedLi: Yes, this was a nice change as the filters before didn't accurately reflect what the filter field would look like.

addons/api/mirage/factories/account.js
ZedLi: I assume it's to make a more accurate `dn` [field](https://developer.hashicorp.com/boundary/docs/concepts/domain-model/accounts#dn), take a look at what they can look like [here](https://learn.microsoft.com/en-us/previous-versions/windows/desktop/ldap/distinguished-names).

She's setting an initial UID and a random number of DC fields as part of the string

DhariniJeeva (approved):

PR #1809, opened 2023-07-27
fix missing account ids

DhariniJeeva:
During the recent HDS migration, the account IDs were incorrectly added.

before
Screenshot 2023-07-27 at 1 33 32 PM

after:
 
Screenshot 2023-07-27 at 1 37 11 PM

### DIFF
--- a/ui/admin/app/templates/scopes/scope/auth-methods/auth-method/accounts/index.hbs
+++ b/ui/admin/app/templates/scopes/scope/auth-methods/auth-method/accounts/index.hbs
@@ -155,11 +155,11 @@
               </B.Td>
               <B.Td>
                 <Copyable
-                  @text={{B.data.account.id}}
+                  @text={{B.data.id}}
                   @buttonText={{t 'actions.copy-to-clipboard'}}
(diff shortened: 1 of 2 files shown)


calcaide (approved):
Nice one!! thanks for catching this!! :dancer:

calcaide:
@DhariniJeeva 
**None blocking comment:** I notice this PR brings changes within `yarn.lock` but your changes does not seem to provoke that. Perhaps your branch is out of sync or we have an outdated `yarn.lock` in main, just mentioning in case you want to check that 😉

DhariniJeeva:
> @DhariniJeeva **None blocking comment:** I notice this PR brings changes within `yarn.lock` but your changes does not seem to provoke that. Perhaps your branch is out of sync or we have an outdated `yarn.lock` in main, just mentioning in case you want to check that 😉

yeah, I noticed that locally, I believe the main is out of sync?

calcaide:
@DhariniJeeva Gotcha!! then thsi PR will fix it!! thanks for it! 🥳

PR #1811, opened 2023-07-31
refactor: 💡 Remove targets/target-id/sessions route and tests

cameronperera:
✅ Closes: ICU-10317

:tickets: [Jira ticket](https://hashicorp.atlassian.net/browse/ICU-10317)

## Description

Remove `targets/target-id/sessions route` and tests

:desktop_computer: [Desktop preview](https://boundary-ui-desktop-7yu928wrj-hashicorp.vercel.app/scopes/global/projects/targets/t_l5fhet6083)

Screen Shot 2023-07-31 at 2 24 43 PM


calcaide (commented):
Sessions has a controller, [here code](https://github.com/hashicorp/boundary-ui/blob/main/ui/desktop/app/controllers/scopes/scope/projects/targets/target/sessions.js), do we want to delete the controller? or perhaps other parts of the app are using it?

calcaide (approved):
Left a comment regarding a controller 😉 good job!! Let's goo!! 🧹

cameronperera:
> Sessions has a controller, [here code](https://github.com/hashicorp/boundary-ui/blob/main/ui/desktop/app/controllers/scopes/scope/projects/targets/target/sessions.js), do we want to delete the controller? or perhaps other parts of the app are using it?

Good catch, I did not check for controllers. Thanks! I will remove that now.

PR #1815, opened 2023-08-01
update electron version on desktop client

DhariniJeeva:
:tickets: [Jira ticket](https://hashicorp.atlassian.net/browse/ICU-10455)

Upgraded electron to **25.3.2**
here are the [ release notes](https://releases.electronjs.org/release/v25.3.2), and checked[ breaking changes](https://www.electronjs.org/docs/latest/breaking-changes#planned-breaking-api-changes-250). We only use one of the deprecated methods which has been replaced by the recommended method. 

Test:

- tested these changes by building locally on Intel mac and everything works as expected. 

[ci release pipeline](https://github.com/hashicorp/boundary-ui-releases/actions/runs/5732175104)

Screenshot 2023-08-01 at 3 16 31 PM


calcaide (approved):
Awesome upgrade!! Now we are up to date with security!

cameronperera (approved):

ui/desktop/electron-app/src/index.js
```diff
@@ -173,8 +174,7 @@ app.on('ready', async () => {
       : path.normalize(`${emberAppDir}${absolutePath}`);
 
     if (isDev) console.log('[serving]', request.url);
-
-    callback({ path: normalizedPath });
+    return net.fetch(`file:///${normalizedPath}`);
```
ZedLi, line 177: I haven't tested this but should it be 
```suggestion
    return net.fetch(`file://${normalizedPath}`);
```

Doesn't the `normalizedPath` include a leading slash?

ZedLi (commented):
Is the pipeline broken for mac signing?

ui/desktop/electron-app/src/index.js
DhariniJeeva: normalizedPath has a leading slash, interestingly, it didn't throw any error with three vs four slashes(incl the normalizedPath), could it be because the file protocol ignores the extra slash?

ZedLi: Honestly it's a little confusing, there's a literal [section](https://en.wikipedia.org/wiki/File_URI_scheme#How_many_slashes?) on wikipedia on how many slashes 🤷

ZedLi (approved):
Thanks for looking into this!!

PR #1821, opened 2023-08-03
refactor: 💡 remove unneeded div for layout of target details

cameronperera:
✅ Closes: ICU-10513

:tickets: [Jira ticket](https://hashicorp.atlassian.net/browse/ICU-10513)

## Description

 remove unneeded div for layout of target details

### DIFF
--- a/ui/desktop/app/templates/scopes/scope/projects/targets/target.hbs
+++ b/ui/desktop/app/templates/scopes/scope/projects/targets/target.hbs
@@ -24,7 +24,6 @@
   </page.actions>
 
   <page.body>
-    <div class='copy-url-container'>{{t 'actions.copy-to-clipboard'}}</div>
     <Rose::Layout::BodyContent as |bc|>
       <bc.Body class='target-details-body'>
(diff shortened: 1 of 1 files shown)


ui/desktop/app/templates/scopes/scope/projects/targets/target.hbs
```diff
@@ -24,7 +24,6 @@
   </page.actions>
 
   <page.body>
-    <div class='copy-url-container'>{{t 'actions.copy-to-clipboard'}}</div>
```
DhariniJeeva, line 27: Are we moving this copy button somewhere else?

DhariniJeeva (approved):

ui/desktop/app/templates/scopes/scope/projects/targets/target.hbs
cameronperera: Yes sorry I should have added that context. The copy button will only exist on the Session Detail screen, so the Target details screen no longer needs this.

PR #1822, opened 2023-08-03
fix: 🐛 accounts update refactor

dvonanderson:
✅ Closes: ICU-10503

:tickets: [Jira ticket](https://hashicorp.atlassian.net/browse/ICU-10503)

## Description
Changed `type` and `auth_method_id` on the `accounts` model to `readOnly` so they are not sent in API payload.

### Error when trying to send `type` and `auth_method_id` as part of payload
Screenshot 2023-08-03 at 11 08 12

### After making attributes read-only
image

### DIFF
--- a/addons/api/addon/generated/models/account.js
+++ b/addons/api/addon/generated/models/account.js
@@ -13,11 +13,13 @@ export default class GeneratedAccountModel extends BaseModel {
   // =attributes
 
   @attr('string', {
+    readOnly: true,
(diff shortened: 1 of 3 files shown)


DhariniJeeva (approved):

DhariniJeeva:
Before you merge, can you check if change works with other auth methods too?

dvonanderson:
> Before you merge, can you check if change works with other auth methods too?

@DhariniJeeva, we are good to go on all auth methods.
image

PR #1825, opened 2023-08-04
Icu 10491 create route for session/session-id

cameronperera:
:tickets: [Jira ticket](https://hashicorp.atlassian.net/browse/ICU-10491) Route creation
:tickets: [Jira ticket](https://hashicorp.atlassian.net/browse/ICU-10492) Header
:tickets: [Jira ticket](https://hashicorp.atlassian.net/browse/ICU-10493) Layout

## Description

Create route for session/session-id
Add base header and layout for screen

**_No logic routing logic for ending the session or redirecting away from detail screen_**
_**Also no direct route for user, requires direct url manipulation, add `/session-id` to end of sessions route**_

:desktop_computer: [Desktop preview](https://boundary-ui-desktop-git-icu-10491-create-route-4609f4-hashicorp.vercel.app/scopes/global/projects/sessions/ss_69vzrnh490)

Screen Shot 2023-08-04 at 7 34 47 AM

### DIFF
--- a/ui/desktop/app/styles/app.scss
+++ b/ui/desktop/app/styles/app.scss
@@ -506,6 +506,7 @@ $chevron-down: url("data:image/svg+xml;utf-8,<svg viewBox='0 0 24 24' fill='%231
   }
 }
 
+.session-details-header,
 .target-details-header {
   display: flex;
   align-items: center;
--- a/addons/core/translations/resources/en-us.yaml
+++ b/addons/core/translations/resources/en-us.yaml
@@ -280,6 +280,7 @@ session:
       description: No active sessions available to display yet.
   actions:
     connect: Connect
+    end: End session
   credential:
     title: Credential
     title_plural: Credentials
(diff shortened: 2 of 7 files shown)


ui/desktop/app/styles/app.scss
```diff
@@ -506,6 +506,7 @@ $chevron-down: url("data:image/svg+xml;utf-8,<svg viewBox='0 0 24 24' fill='%231
   }
 }
 
+.session-details-header,
```
calcaide, line 509: None blocking comment:
session-detail-header and target-details-header are basically the same, we just add the domain (`session-` or `target-`) but the code is the  same, the design is the same, isn't?
Do you think it is doable to refactor this something more reusable? perhaps defining this such as:  `.rose-layout-page-header > h1` or something that does not add the domain name to the class?

calcaide (approved):
Left a none blocking comment 😉  nice work!!

ZedLi (approved):

ui/desktop/app/styles/app.scss
cameronperera: That is exactly what I had planned. I don't believe these page headers will require different styles but I was keeping it split on purpose for now. But the plan is to look at the styling for both pages and see what I can refactor into a single style block that can be used by both screens.

PR #1826, opened 2023-08-04
refactor: 💡 Refactor ldap account view and serializer

dvonanderson:
✅ Closes: ICU-10514

:tickets: [Jira ticket](https://hashicorp.atlassian.net/browse/ICU-10514)

## Description
Refactor LDAP account view to only show `InfoField` fields if values are not `undefined.`  I also fixed a bug for updating accounts.

### Before
image

### After if the value is undefined.
Screenshot 2023-08-04 at 11 44 36


addons/api/addon/serializers/account.js
```diff
@@ -52,9 +52,8 @@ export default class AccountSerializer extends ApplicationSerializer {
       serialized.attributes = {};
       const { login_name } = snapshot.record;
       if (login_name) serialized.attributes.login_name = login_name;
-    }
-    // LDAP does not support updating the auth method ID or type.
-    if (!snapshot.record.isNew) {
+    } else {
+      // OIDC does not support updating the readonly auth method ID or type attributes.
```
calcaide, line 56: None blocking comment: Why we are mentioning here OIDC and deleting fields if this is `serializeLDAP` ?

calcaide (approved):
LGTM!! Thanks for the work

addons/api/addon/generated/models/account.js
```diff
@@ -13,13 +13,11 @@ export default class GeneratedAccountModel extends BaseModel {
   // =attributes
 
   @attr('string', {
-    readOnly: true,
```
ZedLi, line 16: Are these readonly fields or not? It's not clear to me in this PR since it seems like they're not supposed to be able to be updated but now we're manually deleting them in the serializer? What's the actual bug?

ui/admin/app/components/form/account/ldap/index.hbs
```diff
@@ -86,27 +86,40 @@
...
+          <InfoField @value={{@model.email}} as |F|>
+            <F.Label>{{t 'resources.account.form.email.label'}}</F.Label>
+          </InfoField>
+        {{/if}}
+        {{#if @model.full_name}}
+          <InfoField @value={{@model.full_name}} as |F|>
+            <F.Label>{{t 'resources.account.form.full_name.label'}}</F.Label>
+          </InfoField>
+        {{/if}}
+        {{#if @model.first_name}}
```
ZedLi, line 105: Should this be `@model.dn`?

addons/api/addon/serializers/account.js
DhariniJeeva: good catch, this should be LDAP

DhariniJeeva (approved):

ui/admin/app/components/form/account/ldap/index.hbs
```diff
@@ -86,27 +86,40 @@
...
+          <InfoField @value={{@model.dn}} as |F|>
+            <F.Label>{{t 'resources.account.form.dn.label'}}</F.Label>
+          </InfoField>
+        {{/if}}
+
+        {{#if @model.member_of_groups}}
```
ZedLi, line 111: Is this an array? Should we be checking the length instead?

ZedLi (dismissed):

addons/api/addon/generated/models/account.js
dvonanderson: @ZedLi the original bug was not allowing the name and the description of an account to be update.  We set the fields to readOnly and although this works for updating the name and description of the account.  It was causing a he creation of a new account to fail because it wanted the auth_method_id and the type.  This is why we are deleting the fields in the serializer if it's not new

ui/admin/app/components/form/account/ldap/index.hbs
dvonanderson: yes good catch @ZedLi

PR #1830, opened 2023-08-07
test(admin/e2e): Updates to the dynamic host catalog e2e test

moduli:
## Description

This PR modifies a dynamic host catalog end-to-end test, which creates an AWS dynamic host catalog and a host set and verifies it can navigate to each host in the host set. Updates include the following:
- Fixes the test so that it navigates to the host list page after navigating to a host in the host set
- Removes the need to create a second host set. It doesn't provide any additional testing coverage here. This also reduces a resource that we would need to create in AWS, reducing setup time.

### DIFF
--- a/ui/admin/tests/e2e/tests/dynamic-host-catalog.spec.js
+++ b/ui/admin/tests/e2e/tests/dynamic-host-catalog.spec.js
@@ -220,6 +220,7 @@ test.describe('AWS', async () => {
           ', ACTUAL: ' +
           rowCount
       );
+      await page.getByRole('link', { name: 'Hosts' }).click();
     }
   });
 });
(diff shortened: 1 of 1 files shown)


ui/admin/tests/e2e/tests/dynamic-host-catalog.spec.js
```diff
@@ -156,70 +154,7 @@ test.describe('AWS', async () => {
...
-    } while (i < 5);
-
-    if (!hostsAreVisible) {
-      throw new Error(
-        'Hosts are not visible. EXPECTED: ' +
-          expectedHosts.length +
-          ', ACTUAL: ' +
-          rowCount
-      );
+      await page.getByRole('link', { name: 'Hosts' }).click();
```
stasryzhov, line 157: Do we want to explicitly assert/expect anything after/instead of the click?

stasryzhov (approved):

calcaide (approved):
Nice one!!! awesome simplification #lessismore

PR #1831, opened 2023-08-08
chore: 🤖 update hds to v2.9.0

cameronperera:
## Description

update hds to v2.9.0

### DIFF
--- a/addons/core/package.json
+++ b/addons/core/package.json
@@ -31,7 +31,6 @@
     "doc:toc": "doctoc README.md"
   },
   "dependencies": {
-    "@hashicorp/design-system-components": "^2.9.0",
     "api": "*",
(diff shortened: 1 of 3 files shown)


addons/core/package.json
```diff
@@ -31,6 +31,7 @@
     "doc:toc": "doctoc README.md"
   },
   "dependencies": {
+    "@hashicorp/design-system-components": "^2.9.0",
```
ZedLi, line 34: Don't we import this in `rose` instead of `core`?

cameronperera: Yep. I must have tab completed to the wrong directory. Thanks for catching that. Fixed.

cameronperera:
Closing due to yarn.lock conflicts

PR #1833, opened 2023-08-09
Upgrade api to ember 4.12

ZedLi:
## Description
Initial upgrade of ember and ember-data to 4.12 for API addon.


package.json
```diff
@@ -82,7 +82,8 @@
     "**/@storybook/**/ansi-regex": "^5.0.1",
     "**/jsprim/json-schema": "^0.4.0",
     "**/@storybook/**/yuidocjs/markdown-it": "^12.3.2",
-    "**/ember-cli/markdown-it-terminal/markdown-it": "^12.3.2"
+    "**/ember-cli/markdown-it-terminal/markdown-it": "^12.3.2",
+    "**/ember-cli-mirage/@embroider/macros": "^1.0.0"
```
ZedLi, line 86: This resolution is needed because there's a current issue when building with multiple versions of `@embroider/macros`. See this [issue](https://github.com/embroider-build/embroider/issues/592) and this [one](https://github.com/miragejs/ember-cli-mirage/issues/2343) for more info.

addons/api/.stylelintignore
```diff
@@ -0,0 +1,8 @@
+# unconventional files
+/blueprints/*/files/
+
+# compiled output
+/dist/
+
+# addons
+/.node_modules.ember-try/
```
DhariniJeeva, line 8: is this part of the upgrade?

DhariniJeeva (approved):
LGTM! 

left a non-blocking question

addons/api/.stylelintignore
ZedLi: Yeah it was automatically added as part of the upgrade, I can actually remove stylelint since we shouldn't ever need it in the API addon

addons/api/package.json
```diff
@@ -10,67 +10,70 @@
   "scripts": {
     "build:development": "ember build",
     "build": "ember build --environment=production",
-    "lint": "npm-run-all --aggregate-output --continue-on-error --parallel \"lint:!(fix)\"",
-    "lint:fix": "npm-run-all --aggregate-output --continue-on-error --parallel lint:*:fix",
+    "lint": "concurrently \"npm:lint:*(!fix)\" --names \"lint:\"",
```
calcaide, line 13: None blocking comment: I guess `ember-cli` deprecated `npm-run-all` and installed `concurrently`? or we did so?

calcaide (approved):
LGTM!! Great upgrade!! first addon to pass the 4.0 and to mebrace newer node support!

addons/api/package.json
ZedLi: Yeah this was all `ember-cli` 😆

cameronperera (approved):
Nice work!

addons/api/addon/adapters/host-catalog.js
```diff
@@ -20,8 +20,12 @@ export default class HostCatalogAdapter extends ApplicationAdapter {
   buildURL(modelName, id, snapshot, requestType) {
     let url = super.buildURL(...arguments);
     if (requestType === 'createRecord') {
-      const pluginName = snapshot.attr('plugin')?.name;
-      if (pluginName) url = `${url}?plugin_name=${pluginName}`;
+      try {
+        const pluginName = snapshot.attr('plugin')?.name;
+        if (pluginName) url = `${url}?plugin_name=${pluginName}`;
+      } catch (e) {
+        // Ignore any adapter errors here
```
ZedLi, line 27: These try catches are needed now because for some reason, if there's a 404 you get a cache error by just trying to access the snapshot attribute

ui/admin/tests/acceptance/authentication-test.js
```diff
@@ -124,10 +123,7 @@ module('Acceptance | authentication', function (hooks) {
 
   test('visiting auth method when the scope cannot be loaded is still allowed', async function (assert) {
     assert.expect(1);
-    setupOnerror(() => {
```
ZedLi, line 127: Seems like this doesn't work anymore, possibly because ember changed what uses ember runloop in the most recent changes and this only catches things on the runloop.

cameronperera (approved):

PR #1835, opened 2023-08-09
chore: fix to get rid of deprecate-non-strict-relationships warning

DhariniJeeva:
[jira](https://hashicorp.atlassian.net/browse/ICU-10574)

This pr fixes the [non-strict relationship deprecation warning](https://rfcs.emberjs.com/id/0739-ember-data-deprecate-non-strict-relationships/)

Below is the screenshot of the said warning
Screenshot 2023-08-14 at 2 00 01 PM


ZedLi (approved):
Looks great! Thanks for fixing this deprecation. I gotta say when they're explicitly laid out like that it gets confusing thinking about it 🤯

cameronperera (approved):

PR #1840, opened 2023-08-14
refactor: 💡 initial change to target connect workflow

cameronperera:
✅ Closes: ICU-10605

:tickets: [Jira ticket](https://hashicorp.atlassian.net/browse/ICU-10605)

## Description

initial change to target connect workflow
this is the first change to this refactor work. Our plan is to update how we connect to a target without hosts (target address) and targets with host individually. This is also focused on happy path only. How we display errors to the user will change a good bit and will get its own PR.

I decided to skip the failing tests for now as I am still refactoring this workflow.

:desktop_computer: [Desktop preview](https://boundary-ui-desktop-git-icu-10605-refactor-con-d52ca6-hashicorp.vercel.app/scopes/global/projects/targets)
Click Connect from either the targets list or the target detail screen to see new workflow

connect_refactor


lisbet-alvarez (approved):
LGTM!!

ui/desktop/app/routes/scopes/scope/projects/sessions/session.js
```diff
@@ -17,25 +17,11 @@ export default class ScopesScopeProjectsSessionsSessionRoute extends Route {
...
    * @param {object} params
    * @param {string} params.session_id
-   * @return {{session: SessionModel, target: TargetModel}}
+   * @return {SessionModel}
    */
-  async model({ session_id }) {
-    let target = null;
-    const session = await this.store.findRecord('session', session_id, {
+  model({ session_id }) {
+    return this.store.peekRecord('session', session_id, {
```
ZedLi, line 23: Was this meant to be a `peekRecord` now? It doesn't support `reload: true` but I'm also curious as to why the switch

cameronperera: Good call out on the `reload: true`. I updated this to `peedRecord` because we will never be able to access the session until after it has been created and loaded into the data store. So we don't need to use `findRecord`. I even tested with closing the desktop client with active sessions and opening the client back up. However, when you close the desktop client, it shuts down all active or pending sessions.

ZedLi: I don't think that's always true as you pointed out with the CLI. But also especially as we move to paginated list responses where not all resources are guaranteed to be loaded into the datastore from the route hierarchy so we can't assume they're already loaded.

I would save the trouble and just always load singular resources

cameronperera: Fixed.

ZedLi (approved):

PR #1842, opened 2023-08-14
chore: fix sortby deprecation warning

DhariniJeeva:
jira-  https://hashicorp.atlassian.net/browse/ICU-10575

This pr installs `lodash` to use the functions that were previously supported by ember, such as `sortby` 

**screenshot of the deprecation warning**
Screenshot 2023-08-14 at 3 35 50 PM

### DIFF
--- a/ui/admin/app/routes/scopes/scope/sessions.js
+++ b/ui/admin/app/routes/scopes/scope/sessions.js
@@ -12,7 +12,7 @@ import runEvery from 'ember-pollster/decorators/route/run-every';
 import { notifySuccess, notifyError } from 'core/decorators/notify';
 import config from '../../../config/environment';
 import { resourceFilter } from 'core/decorators/resource-filter';
-
+import { sortBy } from 'lodash';
 const POLL_TIMEOUT_SECONDS = config.sessionPollingTimeoutSeconds;
 
 export default class ScopesScopeSessionsRoute extends Route {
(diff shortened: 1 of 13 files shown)


lisbet-alvarez (approved):
Nice Work!!

ui/admin/app/routes/scopes/scope/sessions.js
```diff
@@ -12,7 +12,7 @@ import runEvery from 'ember-pollster/decorators/route/run-every';
 import { notifySuccess, notifyError } from 'core/decorators/notify';
 import config from '../../../config/environment';
 import { resourceFilter } from 'core/decorators/resource-filter';
-
+import { sortBy } from 'lodash';
```
ZedLi, line 15:
```suggestion
import sortBy from 'lodash/sortBy';
```

We really shouldn't use these type of imports since it will pull in all of lodash. The normal lodash package can't be tree shaked unless you use `lodash-es` so we'll have the entire package in our builds. I haven't looked into how our app is built and whether it would get tree shaked correctly either so we should just stick with individual imports. 

We _could_ use the [lodash babel plugin](https://github.com/lodash/babel-plugin-lodash) but honestly we're probably not using too many lodash functions so it would be just as easy to do the individual imports.

ZedLi (changes requested):
Looks great but just had a comment regarding how we import lodash

ui/admin/app/routes/scopes/scope/sessions.js
```diff
@@ -80,9 +80,10 @@ export default class ScopesScopeSessionsRoute extends Route {
       )
     );
     // Sort sessions by time created...
-    let sortedSessionAggregates = A(sessionAggregates)
-      .sortBy('session.created_time')
-      .reverse();
+    let sortedSessionAggregates = sortBy(
+      A(sessionAggregates),
```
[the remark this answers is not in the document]
DhariniJeeva: good catch, it is not needed.

ui/admin/app/routes/scopes/scope/sessions.js
DhariniJeeva: ahh, I didn't know that lodash isn't tree shakable by default.

cameronperera (approved):

ZedLi (approved):

PR #1843, opened 2023-08-15
test: 💍 Add create password auth method test

stasryzhov:
✅ Closes: ICU-10556

:tickets: [ICU-10556](https://hashicorp.atlassian.net/browse/ICU-10556)

## Description
The PR adds new UI test for the create password auth-method workflow and does the following:
1. Creates new Org
2. Creates new Auth Method
3. Adds new Account to the Auth Method
4. Sets new password to the Account
5. Makes the Auth Method primary
6. Creates new User
7. Adds the Account to the User
8. Cleans up by deleting the Org

[ICU-10556]: https://hashicorp.atlassian.net/browse/ICU-10556?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.github/workflows/build-admin-ui.yaml
+++ b/.github/workflows/build-admin-ui.yaml
@@ -17,7 +17,7 @@ jobs:
           key: node-modules-${{ hashFiles('**/yarn.lock') }}
           restore-keys: |
             node-modules-
-      - uses: actions/setup-node@e33196f7422957bea03ed53f6fbb155025ffc7b8 # v3.7.0
+      - uses: actions/setup-node@5e21ff4d9bc1a8cf6de233a3057d20ec6b3fb69d # v3.8.1
         with:
           node-version: 16
       - run: yarn install
@@ -35,7 +35,7 @@ jobs:
           restore-keys: |
             node-modules-
       - name: Build
-        uses: actions/setup-node@e33196f7422957bea03ed53f6fbb155025ffc7b8 # v3.7.0
+        uses: actions/setup-node@5e21ff4d9bc1a8cf6de233a3057d20ec6b3fb69d # v3.8.1
         with:
           node-version: 16
       - run: yarn run build:ui:admin:oss
@@ -63,7 +63,7 @@ jobs:
           restore-keys: |
             node-modules-
       - name: Build
-        uses: actions/setup-node@e33196f7422957bea03ed53f6fbb155025ffc7b8 # v3.7.0
+        uses: actions/setup-node@5e21ff4d9bc1a8cf6de233a3057d20ec6b3fb69d # v3.8.1
         with:
           node-version: 16
       - run: yarn run build:ui:admin:enterprise
@@ -86,7 +86,7 @@ jobs:
           restore-keys: |
             node-modules-
       - name: Build
-        uses: actions/setup-node@e33196f7422957bea03ed53f6fbb155025ffc7b8 # v3.7.0
+        uses: actions/setup-node@5e21ff4d9bc1a8cf6de233a3057d20ec6b3fb69d # v3.8.1
         with:
           node-version: 16
       - run: yarn run build:ui:admin:hcp
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -34,7 +34,7 @@ jobs:
 
     # Initializes the CodeQL tools for scanning.
     - name: Initialize CodeQL
-      uses: github/codeql-action/init@85c77f1dfc42a47cc98299e8779c151d2159b120 # codeql-bundle-v2.14.0
+      uses: github/codeql-action/init@f9a7c6738f28efb36e31d49c53a201a9c5d6a476 # codeql-bundle-v2.14.2
       with:
         languages: ${{ matrix.language }}
         # If you wish to specify custom queries, you can do so here or in a config file.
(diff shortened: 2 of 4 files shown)


ui/admin/tests/e2e/tests/auth-method.spec.js
```diff
@@ -0,0 +1,54 @@
...
+    await addAccountToAuthMethod(
+      page,
+      'UI Test Account',
+      'test-user',
+      'password'
+    );
+    await setPasswordToAccount(page, 'password2');
+    await createNewUser(page, 'UI Test User');
+    await addAccountToUser(page);
+  } finally {
```
moduli, line 51: What if we also set the auth method to be the primary auth method for this scope (there are buttons either on the auth methods list page or in the auth method details page)? 

We should also create a similar test on the backend and actually log in as that user. I'll create that ticket.

moduli, line 51: Not sure how you want to handle checks for deletion. Should we do deletes for user, account, auth methods, in this test or create a separate a test that will create and delete everything?

ui/admin/tests/e2e/helpers/boundary-ui.js
```diff
@@ -287,3 +287,121 @@ exports.addBrokeredCredentialsToTarget = async (
   await page.getByRole('button', { name: 'Dismiss' }).click();
   await expect(page.getByRole('link', { name: credentialName })).toBeVisible();
 };
+
+/**
+ * Uses the UI to create new Auth Method. Assumes you have selected the desired scope.
+ * @param {Page} page Playwright page object
+ * @param {string} authMethodName Name of new auth method
+ */
+exports.createNewAuthMethod = async (page, authMethodName) => {
```
calcaide, line 296: None blocking question: I am curious how we specify the type of auth method we are creating? I can't see were we are specifyng auth method type. I can see on line 301 & 302 of this same file we are selecting the type password. Would it be possible type is a parameter of this method?

ui/admin/tests/e2e/tests/auth-method.spec.js
stasryzhov: I'd rather have a separate test or set of tests for resources deletion scenarios.
In such scenarios we would not need to create the Boundary resources via UI; CLI would be more efficient for the test data preparation stage of the test. Once the test data is prepared, we'd delete the Boundary resources one by one using UI. 

This approach would help achieve two things:

1. The delete test would concentrate on verifying deletion of the resources via UI only
2. The auth-methods test would concentrate on creating auth-methods related resources via UI only and remain relatively small for this test level.

ui/admin/tests/e2e/helpers/boundary-ui.js
stasryzhov: >  I am curious how we specify the type of auth method we are creating?

It's hardcoded:
`await page.getByText('Password', { exact: true }).click();`

It's possible to add the parameter. However in the method we'd have two big branches: one for password method and the other for the OIDC as they have different set of required fields. 
Now this method is used for creating password auth method only. So, I'd propose sticking to the current implementation and rename a method to make it clear what auth method type it creates

```suggestion
exports.createNewPasswordAuthMethod = async (page, authMethodName) => {
```

What do you think @calcaide?

calcaide: @stasryzhov I think you explained very clear and I agree with you 😃 Thanks for clarification and modification, appreciate it!!

calcaide (approved):
Amazing job!! thanks for it!

moduli:
FYI: We still want to update this test to use the UI to set the auth method to be the primary auth method.

ui/admin/tests/e2e/helpers/boundary-ui.js
```diff
@@ -287,3 +287,140 @@ exports.addBrokeredCredentialsToTarget = async (
...
+ * Uses the UI to make the first available auth method primary.
+ * Assumes you have created new auth method.
+ * @param {Page} page Playwright page object
+ */
+exports.makeAuthMethodPrimary = async (page) => {
+  await page
+    .getByRole('navigation', { name: 'IAM' })
+    .getByRole('link', { name: 'Auth Methods' })
+    .click();
+  await page.click('[aria-label="Manage"]');
```
moduli, line 326: Would a simpler locator work better here?
```
await page.getByText('Manage').click();
```
I want to try to avoid using attributes when possible.

stasryzhov: Agreed. Interesting behaviour:
This one works:
`page.getByRole("button", { name: 'Manage' }) `

This one does not:
`page.getByText('Manage')`

moduli: That's fine. Were there other instances of the word `Manage` on the page?

stasryzhov: There is one if there is one auth method created as it is in the test.
When using `page.getByText('Manage')`, it keeps waiting for the element but could not find it as 'Manage' is in the`area-label` property only.

moduli (approved):

ui/admin/tests/e2e/tests/auth-method.spec.js
```diff
@@ -0,0 +1,55 @@
+/**
+ * Copyright (c) HashiCorp, Inc.
+ * SPDX-License-Identifier: MPL-2.0
```
moduli, line 3: Can you check if we need to make these `BUSL` now?

stasryzhov: This repo is under MPL-2.0 at the moment

moduli (approved):

PR #1845, opened 2023-08-15
chore: 🤖 silence `deprecate_early_static` warnings

DhariniJeeva:
:tickets: [Jira ticket](https://hashicorp.atlassian.net/browse/ICU-10577)

this pr will silence the `deprecate_early_static` [warning](https://rfcs.emberjs.com/id/0741-ember-data-deprecate-model-static-field-access-without-lookup/), which affects only when mocks are used. This is due to the discoverEmberDataModels method that accesses the static fields of the model directly. 

Screenshot 2023-08-15 at 11 02 49 AM

### DIFF
--- a/addons/core/addon/initializers/deprecations.js
+++ b/addons/core/addon/initializers/deprecations.js
@@ -7,6 +7,11 @@ import { registerDeprecationHandler } from '@ember/debug';
 
 export function initialize() {
   registerDeprecationHandler((message, options, next) => {
+    //silence deprecate-early-static warnings till we find a way to
+    //fix ember-cli-mirage discoverEmberDataModels
+    if (options?.id === 'ember-data:deprecate-early-static') {
+      return;
+    }
     if (options?.url) {
       message += options.url;
     }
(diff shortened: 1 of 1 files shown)


addons/core/addon/initializers/deprecations.js
```diff
@@ -7,6 +7,11 @@ import { registerDeprecationHandler } from '@ember/debug';
 
 export function initialize() {
   registerDeprecationHandler((message, options, next) => {
+    //silence deprecate-early-static warnings till we find a way to
```
ZedLi, line 10: Maybe it would also help to add a TODO as part of the comment?

ZedLi (approved):

addons/core/addon/initializers/deprecations.js
DhariniJeeva: done

lisbet-alvarez (approved):

calcaide (approved):
Nice!! thanks for looking into this!!

PR #1848, opened 2023-08-15
chore: 🤖 update hds to 2.10.0

cameronperera:
## Description

update hds to 2.10.0

HDS v2.9.0 introduced Copy components that use `ember-cli-clipboard@1.0.0`. This introduced a dependency issue with `ember-arg-types` that uses `prop-types@15.7.1` but it needs to be at v15.8.1. Added a resolution for that now and will remove if HDS fixes this.

### DIFF
--- a/package.json
+++ b/package.json
@@ -84,7 +84,7 @@
     "**/@storybook/**/yuidocjs/markdown-it": "^12.3.2",
     "**/ember-cli/markdown-it-terminal/markdown-it": "^12.3.2",
     "**/ember-cli-mirage/@embroider/macros": "^1.0.0",
-    "**/ember-arg-types/prop-types": "^15.8.1"
+    "**/rose/**/ember-arg-types/prop-types": "^15.8.1"
   },
   "config": {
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -83,7 +83,8 @@
     "**/jsprim/json-schema": "^0.4.0",
     "**/@storybook/**/yuidocjs/markdown-it": "^12.3.2",
     "**/ember-cli/markdown-it-terminal/markdown-it": "^12.3.2",
-    "**/ember-cli-mirage/@embroider/macros": "^1.0.0"
+    "**/ember-cli-mirage/@embroider/macros": "^1.0.0",
+    "**/ember-arg-types/prop-types": "^15.8.1"
```
calcaide, line 87: @cameronperera 
None blocking comment: if `ember-arg-types` is used only within `addons/rose` (needs to be check), I would suggest to add the `rose` within the resolution, just to narrow it down there and avoid possible issues. I.E: `**/rose/**/ember-arg-types/prop-types`.

PR #1851, opened 2023-08-18
fix: 🐛 proxy promise deprecation

DhariniJeeva:
:tickets: [Jira ticket](https://hashicorp.atlassian.net/browse/ICU-10635)

 
This pr gets rid of the [promiseArray deprecation warning](https://api.emberjs.com/ember-data/4.12/classes/CurrentDeprecations/properties?anchor=DEPRECATE_PROMISE_PROXIES)

Since the listed methods in the link above return a promise now, we need to resolve this object before we access them in the templates.

[preview](https://boundary-ui-git-icu-10635-fix-proxy-promises-warning-hashicorp.vercel.app/scopes/global/roles/r_22a3qt99e6)

 
Screenshot 2023-08-18 at 2 13 25 PM

### DIFF
--- a/ui/admin/app/routes/scopes/scope/roles/role/index.js
+++ b/ui/admin/app/routes/scopes/scope/roles/role/index.js
@@ -22,7 +22,9 @@ export default class ScopesScopeRolesRoleIndexRoute extends Route {
       : (await this.store.query('scope', { scope_id: currentScope.id })).map(
           (scope) => ({
             model: scope,
-            subScopes: this.store.query('scope', { scope_id: scope.id }),
+            subScopes: this.store
+              .query('scope', { scope_id: scope.id })
+              .then((models) => models.map((model) => model)),
           })
         );
     const grantScopes = [
--- a/ui/admin/app/routes/scopes/scope/roles/role/index.js
+++ b/ui/admin/app/routes/scopes/scope/roles/role/index.js
@@ -5,7 +5,6 @@
 
 import Route from '@ember/routing/route';
 import { inject as service } from '@ember/service';
-import { all } from 'rsvp';
 
 export default class ScopesScopeRolesRoleIndexRoute extends Route {
   // =services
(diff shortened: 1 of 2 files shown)


ui/admin/app/routes/scopes/scope/roles/role/index.js
```diff
@@ -22,7 +22,9 @@ export default class ScopesScopeRolesRoleIndexRoute extends Route {
       : (await this.store.query('scope', { scope_id: currentScope.id })).map(
           (scope) => ({
             model: scope,
-            subScopes: this.store.query('scope', { scope_id: scope.id }),
+            subScopes: this.store
+              .query('scope', { scope_id: scope.id })
+              .then((models) => models.map((model) => model)),
```
ZedLi, line 27: What's the purpose of this map?

Do we need to `.then` this, can't we await the query?

DhariniJeeva: good point. Actually, `.map` isn't needed here! I'll remove that. 

since the query returns an array, using await like the below code block, was just returning an unresolved array of promises. I could try to use `promise.all`, but using `.then` seemed simpler. 

>    const subScopes = currentScope.isProject
>       ? []
>       : (await this.store.query('scope', { scope_id: currentScope.id })).map(
>           async (scope) => ({
>             model: scope,
>             subScopes: await this.store.query('scope', { scope_id: scope.id }),
>           })
>         );

ZedLi: Oh interesting 🤔 , it doesn't return a promise that resolves an array, it's an array of promises? That seems surprising since it seems [docs](https://api.emberjs.com/ember-data/4.12/classes/Store/methods/query?anchor=query) say `store.query` just returns a promise

ZedLi (approved):
Looks good! Just left a comment regarding the usage of `rsvp`

ui/admin/app/routes/scopes/scope/roles/role/index.js
```diff
@@ -5,6 +5,7 @@
 
 import Route from '@ember/routing/route';
 import { inject as service } from '@ember/service';
+import { all } from 'rsvp';
```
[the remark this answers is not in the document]
DhariniJeeva: We don't need to use rsvp. I changed it to `Promise.all`

cameronperera (approved):

PR #1855, opened 2023-08-21
Icu 10315 add credentials panel

cameronperera:
:tickets: [Jira ticket](https://hashicorp.atlassian.net/browse/ICU-10315)

## Description

Adds a credentials panel to the sessions detail screen if credentials exist after a connection has been established

:desktop_computer: [Desktop preview](https://boundary-ui-desktop-git-icu-10315-add-credenti-5d988f-hashicorp.vercel.app/scopes/global/projects/sessions/1)

Credentials:
Screen Shot 2023-08-21 at 5 14 05 PM

View API JSON:
Screen Shot 2023-08-21 at 5 14 11 PM

No Credentials:
Screen Shot 2023-08-21 at 5 14 32 PM


ui/desktop/app/routes/scopes/scope/projects/sessions/session.js
```diff
@@ -12,6 +12,11 @@ export default class ScopesScopeProjectsSessionsSessionRoute extends Route {
   @service store;
 
   // =methods
+  resetController(controller, isExiting, transition) {
```
calcaide, line 15: None blocking comment: would you mind adding a comment to describe what resetController does?

ui/desktop/app/styles/app.scss
```diff
@@ -530,3 +530,110 @@ $chevron-down: url("data:image/svg+xml;utf-8,<svg viewBox='0 0 24 24' fill='%231
     background-color: var(--token-color-palette-neutral-50);
   }
 }
+
+.session-details {
+  .rose-layout-body-content {
+    margin-top: 1.5rem;
```
calcaide, line 536: Is this `margin-top` something speicifc to session-details? or could be applied to targets too?

ui/desktop/app/styles/app.scss
```diff
@@ -530,3 +530,110 @@ $chevron-down: url("data:image/svg+xml;utf-8,<svg viewBox='0 0 24 24' fill='%231
...
+  }
+}
+
+.credential-section-header {
```
calcaide, line 540: The markup classes and semantic elements are IMHO a bit confusing. I will start a thread on slack to discuss it 😉

ui/desktop/app/routes/scopes/scope/projects/sessions/session.js
```diff
@@ -12,6 +12,11 @@ export default class ScopesScopeProjectsSessionsSessionRoute extends Route {
...
+    if (isExiting && transition.targetName !== 'error') {
```
DhariniJeeva, line 16: why do we need to reset the controller explicitly here? Wouldn't this automatically happen when we change routes? 🤔

cameronperera: As Carlos mentioned above, I am going to add a comment for this but I am going to give more details here. Controllers are [singletons](https://guides.emberjs.com/v4.4.0/routing/controllers/#toc_where-and-when-to-use-controllers). So `isRawApiVisible` will be shared among all instances of that route. For example, when I toggle the view to see the API JSON for the credential, then transition away from that screen, the `isRawApiVisible` property is still `true`. So if I were to go back to another session details screen, the API JSON view would automatically be visible even though it shouldn't. So Ember has the [resetController](https://api.emberjs.com/ember/4.4/classes/Route/methods/resetController?anchor=resetController) hook that allows you to reset the value to whatever you want when the user transitions away from that route.

ZedLi: Can you explain why this needs to be in the controller in the first place? It's not obvious to me why this is the case

ui/desktop/app/styles/app.scss
```diff
@@ -530,3 +530,111 @@ $chevron-down: url("data:image/svg+xml;utf-8,<svg viewBox='0 0 24 24' fill='%231
...
+
+  /* stylelint-disable selector-class-pattern */
+  .hds-form-field--layout-flag {
+    margin-bottom: 0;
+  }
+}
+
+.credential-panel-body {
+  border: 0.063rem solid var(--token-color-border-primary);
+  border-radius: 0.313rem;
```
ZedLi, line 559: These are incredibly specific rems, should these be pixels?

ui/desktop/app/styles/app.scss
```diff
@@ -530,3 +530,111 @@ $chevron-down: url("data:image/svg+xml;utf-8,<svg viewBox='0 0 24 24' fill='%231
...
+    right: $border-width;
+    top: $border-width;
+    width: calc(100% + (2 * $border-width));
+  }
+
+  .secret-raw .CodeMirror {
+    border-bottom-left-radius: 0.313rem;
+    border-bottom-right-radius: 0.313rem;
+  }
+}
```
ZedLi, line 640: What are these needed for?

ui/desktop/app/styles/app.scss
cameronperera: Both screens need it. I updated the class name to not be specific to just sessions.

ui/desktop/app/styles/app.scss
cameronperera: Discussed and refactored.

ui/desktop/app/styles/app.scss
```diff
@@ -530,3 +530,111 @@ $chevron-down: url("data:image/svg+xml;utf-8,<svg viewBox='0 0 24 24' fill='%231
...
+  .raw-credential-title {
+    display: flex;
+    justify-content: space-between;
+
+    .hds-copy-button {
+      padding: 0;
+      background: unset;
+      border: unset;
+      box-shadow: unset;
+      min-height: unset;
```
ZedLi, line 621: Should we be overriding these styles or just adhering to the standard HDS styling for this?

ui/desktop/app/templates/scopes/scope/projects/sessions/session.hbs
```diff
@@ -15,18 +15,119 @@
...
+                      @isIconOnly={{true}}
+                      @textToCopy={{raw-json credential.rawCredential}}
+                    />
+                  </div>
+                  <div class='hds-typography-body-100'>
+                    {{credential.source.description}}
+                  </div>
+                </div>
+
+                <div class='secret-raw'>
```
ZedLi, line 71: Is this extra div needed?

ui/desktop/app/styles/app.scss
```diff
@@ -530,3 +530,111 @@ $chevron-down: url("data:image/svg+xml;utf-8,<svg viewBox='0 0 24 24' fill='%231
...
+  display: flex;
+  margin-top: 0.75rem;
+
+  .credential-header {
+    padding: 1.5rem;
+    width: 33%;
+  }
+}
+
+.credential-secret {
```
calcaide, line 569: None blocking comment: If `.credential-secret` just lives inside of a `.credential-panael-body` I will suggest to nested inside `.credential-panel-body` as we did with `.credential-header`. Thoughts?

ui/desktop/app/routes/scopes/scope/projects/sessions/session.js
cameronperera: I decided to put this in the controller because this is not part of the model data and only used to toggle how the user views the credential UI. Is there a better place this should live?

ui/desktop/app/styles/app.scss
cameronperera: I am fine with using pixels. My understanding is we are using rem as our standard. Apologies if that changed and I missed it.

ui/desktop/app/routes/scopes/scope/projects/sessions/session.js
ZedLi: Is there a reason this couldn't just live in the backing class of the template?

ui/desktop/app/styles/app.scss
cameronperera: The container styles are to make the editor the width of the parent plus the border width to match the wireframe. Happy to update this with a better solution.

The `border-bottom-[left|right]-radius` is to add radius to just the bottom corners. 
Screen Shot 2023-08-23 at 8 49 46 AM

ui/desktop/app/styles/app.scss
ZedLi: I think that's true for typography and layout but I don't think that makes sense for something like `border-radius`, would it make sense for `border-radius` to change depending on the root font size?

ui/desktop/app/styles/app.scss
ZedLi: If the issue is the size is not correct, should we be instead passing the correct sizing info to code mirror itself rather than trying to override it ourselves?

ui/desktop/app/styles/app.scss
cameronperera: I originally left it with the standard HDS styling. I decided to change it so that the copy buttons matched with the `<HiddenSecret/>` component. I was going to remove this if we have time to update all of the copy buttons to the HDS version. I can address this in another PR asap if you'd like but I didn't want the scope to grow too much for this PR.

ui/desktop/app/templates/scopes/scope/projects/sessions/session.hbs
cameronperera: Nope. Removed.

ui/desktop/app/styles/app.scss
cameronperera: Good point. I will update this.

ui/desktop/app/styles/app.scss
cameronperera: I originally had it nested but I got a lint error because it was too many items nested.
`app/styles/app.scss
 587:9  ✖  Expected ".credential-panel-body .credential-secret .secret-container .hidden-secret .rose-button-ghost" to have no more than 4 compound selectors  selector-max-compound-selectors`

ui/desktop/app/routes/scopes/scope/projects/sessions/session.js
cameronperera: Going to move the credential panel to a component after chatting offline.

cameronperera: Done.

ui/desktop/app/styles/app.scss
cameronperera: It isn't that the size isn't correct. Just that sometimes we need the border-radius and in some cases (workers) we don't use a border-radius. As for the width, this is the only place I have found that the width defined by code mirror was not sufficient. We could add border-radius to code mirror by default but I would still need to remove the top left and right for my work. I would prefer that change happen outside of this PR though.

ui/desktop/tests/integration/components/credentials-panel-test.js
```diff
@@ -0,0 +1,65 @@
...
+    this.set('credentials', [credential]);
+
+    await render(hbs`
+      <CredentialsPanel @credentials={{this.credentials}} />
+    `);
+
+    assert.dom('.credential-name').hasText('Credential Library 1');
+  });
+
+  test('it code editor when toggle clicked', async function (assert) {
```
DhariniJeeva, line 54:
```suggestion
  test('it shows code editor when toggle clicked', async function (assert) {
```

DhariniJeeva (approved):

ui/desktop/tests/integration/components/credentials-panel-test.js
cameronperera: Thanks for catching that.

ui/desktop/app/styles/app.scss
ZedLi: It seems like to me you're trying to compensate for the extra border that's on the container that includes the header and the editor.

If you moved the border to just the header instead, do you need to fiddle with the code editor position and calculate width at all?

ui/desktop/app/styles/app.scss
DhariniJeeva: It is okay if the HDS copy buttons aren't consistent with Boundary's old copy buttons. Ideally, we should use the hds style and try not to override the styles. We can create a ticket in the backlog to update the rest of the copy buttons when we get time, this is the approach we followed when migrating components.

cameronperera: Overrides removed.

ui/desktop/app/styles/app.scss
cameronperera: Absolutely doable. Fixed now. Thanks for that suggestion.

ui/desktop/app/styles/app.scss
calcaide: I wouldn't disable the rule. I never had this issue so I can't talk 100% with confidence. But the rule you mention, as [per my understanding](https://stylelint.io/user-guide/rules/selector-max-compound-selectors/#:~:text=Limit%20the%20number%20of%20compound%20selectors%20in%20a%20selector.&text=A%20compound%20selector%20is%20a,%60%2C%20%2B%20%2C%20%3E%20).) is about how many selectors you use to chain, now about how many selectors are nested.

Perhaps we can do something like:
```
.credential-secret {
& > .credential-secret {}
& > .credential-heder {}
}

cameronperera: Styles updated to please stylelint rule

PR #1857, opened 2023-08-21
feat: 🎸 Initial setup of indexedDb

ZedLi:
## Description
Initial setup of indexedDB with dexie to sit in front of our normal API.

### DIFF
--- a/addons/api/addon-test-support/helpers/indexed-db.js
+++ b/addons/api/addon-test-support/helpers/indexed-db.js
@@ -1,0 +1,30 @@
+/**
+ * This test helper can help set up and cleanup the database in your tests.
+ *
+ * ```js
+ * import { setupIndexedDb } from 'api/test-support/helpers/indexed-db';
+ *
+ * module('Acceptance | my test', function(hooks) {
+ *  setupApplicationTest(hooks);
+ *  setupIndexedDb(hooks);
+ *
+ *   // add your actual tests here
+ * });
+ * ```
+ *
+ * @param hooks
+ */
+export function setupIndexedDb(hooks) {
+  hooks.beforeEach(function () {
+    const indexedDb = this.owner.lookup('service:indexed-db');
+    indexedDb.setup(`test-indexed-db-${Date.now()}`);
+  });
+
+  hooks.afterEach(async function () {
+    const indexedDb = this.owner.lookup('service:indexed-db');
+
+    await indexedDb.db.delete();
+  });
+}
+
+export default setupIndexedDb;
--- a/addons/api/addon/handlers/indexed-db-handler.js
+++ b/addons/api/addon/handlers/indexed-db-handler.js
@@ -16,7 +16,8 @@ const IndexedDbHandler = {
         //  requests and cache population requests
 
         // Go through normal flow if we don't yet support the model
-        if (!supportedModels.includes(type)) {
+        // or if we don't have an indexedDb instance
+        if (!supportedModels.includes(type) || !this.indexedDb.db) {
           break;
         }
 
--- a/addons/api/addon/handlers/indexed-db-handler.js
+++ b/addons/api/addon/handlers/indexed-db-handler.js
@@ -1,0 +1,57 @@
+import { modelIndexes } from 'api/services/indexed-db';
+
+/**
+ * Handler to sit in front of the API request layer
+ * so we can handle any caching with indexedDb first
+ */
+const IndexedDbHandler = {
+  async request(context, next) {
+    switch (context.request.op) {
+      case 'query': {
+        const { store, data } = context.request;
+        const { type, query } = data;
+        const supportedModels = Object.keys(modelIndexes);
+
+        // TODO: Differentiate between user initiated paginated
+        //  requests and cache population requests
+
+        // Go through normal flow if we don't yet support the model
+        if (!supportedModels.includes(type)) {
+          break;
+        }
+
+        const adapter = store.adapterFor(type);
+        const schema = store.modelFor(type);
+
+        const payload = await adapter.query(store, schema, query);
+        const serializer = store.serializerFor(type);
+        const normalizedPayload = serializer.normalizeResponse(
+          store,
+          schema,
+          payload,
+          null,
+          'query'
+        );
+
+        const { data: payloadData } = normalizedPayload;
+
+        // TODO: Just to show how to add data to indexedDb
+        await this.indexedDb.db[type].bulkPut(payloadData);
+        await this.indexedDb.db[type]
+          .where('attributes.type')
+          .equals('ssh')
+          .toArray();
+
+        // TODO: Manually insert the normalized data into ember data store and don't call next()
+        //  Also investigate to see if we can return cache first and make API request in background
+        break;
+      }
+      default:
+        return next(context.request);
+    }
+
+    return next(context.request);
+  },
+};
+
+export default IndexedDbHandler;
--- a/addons/api/addon/handlers/indexed-db-handler.js
+++ b/addons/api/addon/handlers/indexed-db-handler.js
@@ -11,13 +11,15 @@ const IndexedDbHandler = {
         const { store, data } = context.request;
         const { type, query } = data;
         const supportedModels = Object.keys(modelIndexes);
+        const { db: indexedDb } = this.indexedDb ?? {};
 
         // TODO: Differentiate between user initiated paginated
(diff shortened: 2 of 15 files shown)


addons/api/addon/handlers/indexed-db-handler.js
```diff
@@ -0,0 +1,57 @@
...
+          schema,
+          payload,
+          null,
+          'query'
+        );
+
+        const { data: payloadData } = normalizedPayload;
+
+        // TODO: Just to show how to add data to indexedDb
+        await this.indexedDb.db[type].bulkPut(payloadData);
```
ZedLi, line 39: I've made the choice to actually store the normalized data instead of the raw JSON in indexed DB, any concerns with this?

addons/api/addon/services/store.js
```diff
@@ -0,0 +1,21 @@
...
+  // =services
+  @service indexedDb;
+
+  requestManager = new RequestManager();
+
+  constructor(args) {
+    super(args);
+
+    // Bind `this` so the handler can have access to the indexedDb service
+    IndexedDbHandler.request = IndexedDbHandler.request.bind(this);
```
ZedLi, line 17: Not a fan of needing to do this but couldn't think of a better way to handle this. Open to suggestions

addons/api/addon/handlers/indexed-db-handler.js
cameronperera: Was there something that caused you to change your approach? Are there any cons to storing the normalized data?

ZedLi: The raw JSON isn't an accurate representation of what data the user interacts with, we'd still need to handle serialization/normalization each time we query for data which would complicate searching and filtering. 

I also want to be able to use the data as is and move it into the ember data store.

cameronperera: That's right. You just didn't want the hydrated models. I am down with normalizing it first.

addons/api/addon/services/indexed-db.js
```diff
@@ -0,0 +1,40 @@
+import Service from '@ember/service';
+import Dexie from 'dexie';
+import { inject as service } from '@ember/service';
+
+// List of resources we support to be cached and searched in
+// indexedDb. Any field that can be searched upon should be added to
+// this index. Increment the version number of the database whenever
+// indexes change.
+export const modelIndexes = {
+  target: '&id, attributes.type',
```
DhariniJeeva, line 10: what does `&id` mean?

ZedLi: Ah good question, this just indicates as part of the index that the field is unique, take a look at the docs [here](https://dexie.org/docs/Version/Version.stores()#schema-syntax) for some examples

cameronperera (approved):

ui/desktop/app/routes/application.js
```diff
@@ -35,7 +37,16 @@ export default class ApplicationRoute extends Route {
...
     this.toggleTheme(theme);
-    return this.clusterUrl.updateClusterUrl();
+    await this.clusterUrl.updateClusterUrl();
+
+    // Setup the DB from a successful authentication restoration
+    if (this.session.isAuthenticated) {
+      const userId = this.session.data?.authenticated?.user_id;
+      const clusterUrl = this.clusterUrl.rendererClusterUrl;
+      if (userId && clusterUrl) {
+        this.indexedDb.setup(formatDbName(userId, clusterUrl));
```
lisbet-alvarez, line 47: non-blocking question for understanding: why is the `indexedDb` setup method being called here when the session service `handleAuthentication` method already calls the `setup` method? Another question, why is the session service put in the `ui/desktop` folder and not the `addons/api` folder ?

addons/api/addon/handlers/indexed-db-handler.js
```diff
@@ -0,0 +1,61 @@
...
+        const { data: payloadData } = normalizedPayload;
+
+        // TODO: Just to show how to add data to indexedDb
+        // await this.indexedDb.db[type].bulkPut(payloadData);
+        // await this.indexedDb.db[type]
+        //   .where('attributes.type')
+        //   .equals('ssh')
+        //   .toArray();
+
+        // TODO: Manually insert the normalized data into ember data store and don't call next()
```
lisbet-alvarez, line 49: Non-blocking question for understanding: once this TODO is completed does that mean  that the `LegacyNetworkHandler` use will be removed ?

ui/desktop/app/routes/application.js
ZedLi: Great questions.

The session service is in the `ui/desktop` because it has to live in the application portion of the code. Depending if it's in admin UI or desktop, we will have slightly different logic (namely the URL to use). Otherwise the `api` addon would have to be aware of its clients. I'm actually still even debating whether some of these services should be living in `core` or `api`.

I'm requiring the database name to need the user's ID (for uniqueness) which means they have to be logged in, otherwise I could setup up on loading the ember app. The setup method is called "twice" because of two different scenarios: the one where the user needs to authenticate which means the one in `application.js` wasn't called bc there wasn't a user yet so didn't do anything, and the one where the user was already logged in and we're restoring their authentication which gets called in `application.js`. The key thing to note is `handleAuthentication` isn't called on auth restoration.

addons/api/addon/handlers/indexed-db-handler.js
ZedLi: No, we will have `LegacyNetworkHandler` for the forseeable future. That handles everything for how the old style adapters and serializers work. If you notice, I only hook in currently for `query` requests and even then only for very specific scenarios, otherwise every other request will go through the legacy handler.

addons/api/addon-test-support/helpers/indexed-db.js
```diff
@@ -0,0 +1,30 @@
...
+ * @param hooks
+ */
+export function setupIndexedDb(hooks) {
+  hooks.beforeEach(function () {
+    const indexedDb = this.owner.lookup('service:indexed-db');
+    indexedDb.setup(`test-indexed-db-${Date.now()}`);
+  });
+
+  hooks.afterEach(async function () {
+    const indexedDb = this.owner.lookup('service:indexed-db');
```
DhariniJeeva, line 24: nonblocking: any reason for not defining indexedDb outside the `beforeeach` scope and use that in `aftereach` instead of looking it up again?

DhariniJeeva (approved):

addons/api/addon/services/indexed-db.js
```diff
@@ -0,0 +1,40 @@
...
+  }
+
+  setup(dbName) {
+    // Don't run setup again if we already have one or if we didn't get a name
+    if (this.#db || !dbName) {
+      return;
+    }
+
+    this.#db = new Dexie(dbName);
+    this.#db.version(1).stores(modelIndexes);
```
DhariniJeeva, line 38: curious about this version here, What do they mean? and would there be a case where we access another version?

addons/api/addon-test-support/helpers/indexed-db.js
ZedLi: Can I do that? I thought that because these are test helpers they have to be self contained within the hooks?

addons/api/addon/services/indexed-db.js
ZedLi: Take a look at the [docs](https://dexie.org/docs/Dexie/Dexie.version()) but it's similar to how actual DBs work and having to migrate when any schema changes occur.

The version is to indicate what the schema was like at a particular version and if there are any changes, use a different version. That way users can actually migrate their data to whatever the new schema is but it's a bit overkill for us.

I was hoping there was an easy way to just detect any schema changes and destroy the DB and recreate it but I haven't seen an easy way to do that yet.

addons/api/addon-test-support/helpers/indexed-db.js
DhariniJeeva: Seems like we [can](https://github.com/hashicorp/boundary-ui/blob/3f731867ac36d284ea92e413caad862b691768f3/addons/auth/tests/unit/session-stores/cookie-test.js#L12), we do this in a lot of places

ZedLi: But this is an `addon-test-support` helper, it doesn't have that context right?

lisbet-alvarez (approved):

PR #1858, opened 2023-08-22
chore: 🤖 Change port to be set in package script

ZedLi:
## Description
Made a change to set the port in the package script rather than ember-cli so it doesn't change the default ports when running just `ember test`


cameronperera (approved):

calcaide (approved):
LGTM!

PR #1862, opened 2023-08-23
chore: 🤖 remove prop-type resolution

cameronperera:
## Description

[ember-arg-types updated their project](https://github.com/jkusa/ember-arg-types/pull/99) to use a newer version of prop-type. This means we no longer need this resolution.


package.json
```diff
@@ -76,8 +76,7 @@
     "**/snapdragon/base/cache-base/**/set-value": "^4.0.1",
     "**/jsprim/json-schema": "^0.4.0",
     "**/ember-cli/markdown-it-terminal/markdown-it": "^12.3.2",
-    "**/ember-cli-mirage/@embroider/macros": "^1.0.0",
-    "**/rose/**/ember-arg-types/prop-types": "^15.8.1"
```
ZedLi, line 80: Don't we also now have to update to the newest version of `ember-arg-types` to bring it in?

I'm actually a bit confused since I just took a look at the dependencies and I don't see why we needed a resolution in the first place since it looks like `prop-types` just had to be on `15.8.1`?

cameronperera: Updated `yarn.lock` to address this issue.

ZedLi (approved):

PR #1866, opened 2023-08-25
feat: 🎸 redirect to targets list when canceling

cameronperera:
:tickets: [Jira ticket](https://hashicorp.atlassian.net/browse/ICU-10760)

## Description

redirect to targets list when canceling from the sessions detail screen

✅ Closes: ICU-10760

:desktop_computer: [Desktop preview](https://boundary-ui-desktop-git-icu-10760-route-user-t-ed6945-hashicorp.vercel.app/scopes/global/projects/targets)

When you establish a connection, click "End Session" from the sessions detail screen and user will be redirected to the targets list screen. If you click "Cancel" on a session from the sessions list screen, you should stay in the same location.

### DIFF
--- a/ui/desktop/app/routes/scopes/scope/projects/sessions.js
+++ b/ui/desktop/app/routes/scopes/scope/projects/sessions.js
@@ -93,6 +93,12 @@ export default class ScopesScopeProjectsSessionsRoute extends Route {
   async cancelSession(session) {
     await session.cancelSession();
     await this.ipc.invoke('stop', { session_id: session.id });
+    if (
+      this.router.currentRoute.name ===
+      'scopes.scope.projects.sessions.session.index'
+    ) {
+      await this.router.replaceWith('scopes.scope.projects.targets');
+    }
   }
 
   /**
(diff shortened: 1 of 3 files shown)


lisbet-alvarez (approved):

ui/desktop/app/routes/scopes/scope/projects/sessions.js
```diff
@@ -93,6 +93,12 @@ export default class ScopesScopeProjectsSessionsRoute extends Route {
   async cancelSession(session) {
     await session.cancelSession();
     await this.ipc.invoke('stop', { session_id: session.id });
+    if (
```
DhariniJeeva, line 96: We could make this action take in an argument and pass the route name from the template.

DhariniJeeva (approved):

ZedLi (approved):

ui/desktop/app/routes/scopes/scope/projects/sessions.js
cameronperera: Since we only redirect in one specific case, I feel this is the simplest approach. This is a good idea if we need to make this more complex though!

PR #1868, opened 2023-08-28
feat: 🎸 add Proxy URL info to session detail screen

cameronperera:
## Description

✅ Closes: ICU-10311

:tickets: [Jira ticket](https://hashicorp.atlassian.net/browse/ICU-10311)

:desktop_computer: [Desktop preview](https://boundary-ui-desktop-ebbq8oty9-hashicorp.vercel.app/scopes/global/projects/sessions/1)

proxy-url-example

### DIFF
--- a/addons/core/translations/en-us.yaml
+++ b/addons/core/translations/en-us.yaml
@@ -86,7 +86,3 @@ onboarding:
       learn: Learn about Boundary Desktop app
       cli: Learn about Boundary command-line (CLI)
       api: Learn about Boundary API
-proxy-url:
-  options:
-    address-port: Address & port
-    ssh: SSH
--- a/addons/core/translations/en-us.yaml
+++ b/addons/core/translations/en-us.yaml
@@ -86,3 +86,7 @@ onboarding:
       learn: Learn about Boundary Desktop app
       cli: Learn about Boundary command-line (CLI)
       api: Learn about Boundary API
+proxy-url:
+  options:
+    address-port: Address & port
+    ssh: SSH
--- a/ui/desktop/app/components/proxy-url/index.hbs
+++ b/ui/desktop/app/components/proxy-url/index.hbs
@@ -1,0 +1,34 @@
+<div class='proxy-url-container'>
+  <Hds::Dropdown @listPosition='bottom-left' as |dd|>
+    <dd.ToggleButton @color='secondary' @text={{this.selectedValue}} />
+    <dd.Interactive
+      @text={{t 'proxy-url.options.address-port'}}
+      {{on
+        'click'
+        (fn this.updateContent (t 'proxy-url.options.address-port') dd.close)
+      }}
+    />
+    <dd.Interactive
+      data-test-ssh-option
+      @text={{t 'proxy-url.options.ssh'}}
+      {{on
+        'click'
+        (fn this.updateContent (t 'proxy-url.options.ssh') dd.close)
+      }}
+    />
+  </Hds::Dropdown>
+  {{#if this.isDefault}}
+    <Hds::Copy::Snippet
+      class='copy-address'
+      @textToCopy={{@proxyAddress}}
+      @isTruncated={{true}}
+    />
+    <Hds::Copy::Snippet class='copy-port' @textToCopy={{@proxyPort}} />
+  {{else if this.isSSH}}
+    <Hds::Copy::Snippet
+      class='copy-ssh-command'
+      @textToCopy={{this.sshCommand}}
+      @isTruncated={{true}}
+    />
+  {{/if}}
+</div>
--- a/ui/desktop/app/components/proxy-url/index.hbs
+++ b/ui/desktop/app/components/proxy-url/index.hbs
@@ -17,7 +17,7 @@
       }}
     />
   </Hds::Dropdown>
-  {{#if this.isDefault}}
+  {{#if this.isAddressAndPort}}
     <Hds::Copy::Snippet
       class='copy-address'
       @textToCopy={{@proxyAddress}}
--- a/ui/desktop/app/components/proxy-url/index.js
+++ b/ui/desktop/app/components/proxy-url/index.js
@@ -7,13 +7,17 @@ import Component from '@glimmer/component';
 import { tracked } from '@glimmer/tracking';
 import { action } from '@ember/object';
 import { inject as service } from '@ember/service';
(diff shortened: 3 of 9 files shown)


addons/core/translations/en-us.yaml
```diff
@@ -86,3 +86,7 @@ onboarding:
       learn: Learn about Boundary Desktop app
       cli: Learn about Boundary command-line (CLI)
       api: Learn about Boundary API
+proxy-url:
+  options:
+    address-port: Address & port
+    ssh: SSH
```
cameronperera, line 92: I added these translations here as they are not directly tied to sessions. Resources didn't feel like the best place but happy to take suggestions.

ui/desktop/app/styles/app.scss
```diff
@@ -541,6 +541,25 @@ $chevron-down: url("data:image/svg+xml;utf-8,<svg viewBox='0 0 24 24' fill='%231
...
+
+  .copy-port {
+    width: 120px;
+  }
+
+  .copy-ssh-command {
+    max-width: 75%;
+  }
+}
+
```
joshuaogle, line 562: I think we can get the inline style we were talking about if we just use
```
.proxy-url-container {
   ...container styles

  .hds-copy-snippet {
    width: auto;
  }
}
```

Nice and simple at least 😄

joshuaogle (changes requested):

ui/desktop/app/styles/app.scss
joshuaogle: Sorry, I forgot that doesn't get us the shrinking truncation on the address. Maybe it's more flex-box then.

ui/desktop/app/styles/app.scss
```diff
@@ -541,6 +541,25 @@ $chevron-down: url("data:image/svg+xml;utf-8,<svg viewBox='0 0 24 24' fill='%231
   }
 }
 
+// Proxy URL
+.proxy-url-container {
```
ZedLi, line 545: I'm somewhat not a fan of the ssh command or address/port jumping depending on the size of the content of the dropdown, would it be better to have the dropdown be a consistent width?

cameronperera: How does this look? I am rending double proxy urls at the moment just to test some style changes.
Screen Shot 2023-08-29 at 10 24 05 AM
Screen Shot 2023-08-29 at 10 24 12 AM

ZedLi: I personally like that more, not sure if design or anyone else has preferences

ui/desktop/app/components/proxy-url/index.js
```diff
@@ -0,0 +1,33 @@
...
+
+import Component from '@glimmer/component';
+import { tracked } from '@glimmer/tracking';
+import { action } from '@ember/object';
+import { inject as service } from '@ember/service';
+
+export default class ProxyUrlComponent extends Component {
+  @service intl;
+
+  @tracked selectedValue = this.intl.t('proxy-url.options.address-port');
```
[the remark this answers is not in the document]
cameronperera: I have reached out to product to get confirmation on what we should default to.

cameronperera:
@ZedLi not sure why I cannot reply to this [comment](https://github.com/hashicorp/boundary-ui/pull/1868#discussion_r1308978716) but I will do so here.

This text is specific to the component and more so has to do with the format of session proxy info, not the session itself. Maybe that is an impractical approach though.

addons/core/translations/en-us.yaml
cameronperera: This text is specific to the component and more so has to do with the format of session proxy info, not the session itself. Maybe that is an impractical approach though.

ui/desktop/app/components/proxy-url/index.hbs
```diff
@@ -0,0 +1,34 @@
...
+      data-test-ssh-option
+      @text={{t 'proxy-url.options.ssh'}}
+      {{on
+        'click'
+        (fn this.updateContent (t 'proxy-url.options.ssh') dd.close)
+      }}
+    />
+  </Hds::Dropdown>
+  {{#if this.isAddressAndPort}}
+    <Hds::Copy::Snippet
```
DhariniJeeva, line 21: Are we expecting more types here?  

Can we do  `if this.isAddressAndPort` and `unless this.isAdressAndPort` for the ssh block?

addons/core/translations/en-us.yaml
DhariniJeeva: sounds resource-specific to me

ui/desktop/app/components/proxy-url/index.hbs
```diff
@@ -0,0 +1,34 @@
+<div class='proxy-url-container'>
```
DhariniJeeva, line 1: Non-blocker- Is this a generic component that can be used in other screens or a resource-specific one? 

If it is a resource-specific one, we should consider structuring it something like `session/proxy-url` because right now all the components within `app/components` seem generic and used all over the desktop app

cameronperera: Yeah I can make that change.

ui/desktop/app/components/proxy-url/index.hbs
cameronperera: The intent is to have more options but I can update to this approach until that change comes.

DhariniJeeva (approved):
LGTM!

ui/desktop/app/components/session/proxy-url/index.js
```diff
@@ -0,0 +1,36 @@
...
+import { tracked } from '@glimmer/tracking';
+import { action } from '@ember/object';
+import { inject as service } from '@ember/service';
+import { TYPE_TARGET_SSH } from 'api/models/target';
+
+export default class SessionProxyUrlComponent extends Component {
+  @service intl;
+
+  @tracked selectedValue =
+    this.args.targetType === TYPE_TARGET_SSH
```
ZedLi, line 16: Is this needed? Don't we get a `target` model which means we have access to the `isSSH` helper from that model?

cameronperera: I originally was passing in the whole model in to this component. I personally am not a fan of passing into an entire model if we only need one property from that model. I felt it was unneeded to couple this component to the target model in that way.

PR #1870, opened 2023-08-30
refactor: 💡 Turn on sandboxing for renderer process

ZedLi:
## Description
This lets us keep sandboxing on and `nodeIntegration` off in the renderer process.

We still only have one pty process which is shared across the entire browser window and need to do proper cleanup.

### DIFF
--- a/ui/desktop/electron-app/src/index.js
+++ b/ui/desktop/electron-app/src/index.js
@@ -95,10 +95,7 @@ const createWindow = (partition, closeWindowCB) => {
   // TODO: Figure out how to spawn more than one terminal process in a browser window
   //       while keeping track of handlers
   // Terminal
-  const shell2 =
-    os.platform() === 'win32'
-      ? 'powershell.exe'
-      : process.env.SHELL || '/bin/bash';
+  const shell2 = os.platform() === 'win32' ? 'powershell.exe' : '/bin/bash';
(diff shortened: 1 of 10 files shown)


ui/desktop/electron-app/src/index.js
```diff
@@ -92,8 +92,13 @@ const createWindow = (partition, closeWindowCB) => {
...
   const browserWindow = new BrowserWindow(browserWindowOptions);
 
+  // TODO: Figure out how to spawn more than one terminal process in a browser window
+  //       while keeping track of handlers
   // Terminal
-  const shell2 = os.platform() === 'win32' ? 'powershell.exe' : 'bash';
+  const shell2 =
+    os.platform() === 'win32'
+      ? 'powershell.exe'
+      : process.env.SHELL || '/bin/bash';
```
ZedLi, line 101: Should we just use `bash` or use the user's shell? If they have something like a custom `PS1` it might not play nicely.

cameronperera (approved):
LGTM

ui/desktop/electron-app/src/index.js
DhariniJeeva: I also agree with using bash, at least for now.

DhariniJeeva (approved):

PR #1871, opened 2023-08-30
chore: 🤖 add tabs for details and terminal view

DhariniJeeva:
Add tab for details and terminal view

Screenshot 2023-08-31 at 11 52 06 AM
Screenshot 2023-08-31 at 11 52 21 AM


addons/core/translations/actions/en-us.yaml
```diff
@@ -52,3 +52,4 @@ toggle-color-theme: Toggle Color Theme
 toggle-feature-edition: Toggle Feature Edition
 toggle-licensed-features: Toggle Licensed Features
 more: ', +{remainingItems} more'
+new-shell: New Shell
```
ZedLi, line 55: We can keep this but just wanted to mention this translation only makes sense if we can open multiple terminals

ZedLi (approved):
Looks good!

PR #1875, opened 2023-09-01
Chore remove method

DhariniJeeva:
This method can be used later on to remove a specific channel.

### DIFF
--- a/ui/desktop/electron-app/src/preload.js
+++ b/ui/desktop/electron-app/src/preload.js
@@ -26,6 +26,9 @@ contextBridge.exposeInMainWorld('terminal', {
   create: (vars) => {
     ipcRenderer.send('createTerminal', vars);
   },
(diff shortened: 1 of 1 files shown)


ui/desktop/electron-app/src/preload.js
```diff
@@ -26,6 +26,9 @@ contextBridge.exposeInMainWorld('terminal', {
   create: (vars) => {
     ipcRenderer.send('createTerminal', vars);
   },
+  remove: (callback, id) => {
+    ipcRenderer.removeListener(`removeTerminal-${id}`, callback);
```
ZedLi, line 30: This would still need to be a `send` command and I don't think we'd have a callback

DhariniJeeva: Updated!

ZedLi (approved):

PR #1876, opened 2023-09-01
feat: 🎸 Clean up terminal tab

ZedLi:
## Description
Multiple terminal tabs is difficult with the current HDS tabs component. We have a meeting to discuss this more and other possible design patterns but for now we're just limiting to one terminal.

image

### DIFF
--- a/addons/core/translations/resources/en-us.yaml
+++ b/addons/core/translations/resources/en-us.yaml
@@ -299,7 +299,6 @@ session:
   proxy-url:
     address-port: Address & port
     ssh: SSH
-  #  shell: Shell {shellCount, plural, one {} other {#}}
   shell: Shell
 org:
   title: Org
--- a/addons/core/translations/resources/en-us.yaml
+++ b/addons/core/translations/resources/en-us.yaml
@@ -299,6 +299,8 @@ session:
   proxy-url:
     address-port: Address & port
     ssh: SSH
+  #  shell: Shell {shellCount, plural, one {} other {#}}
+  shell: Shell
 org:
   title: Org
   title_plural: Orgs
--- a/ui/desktop/app/components/session/terminal-tabs/index.hbs
+++ b/ui/desktop/app/components/session/terminal-tabs/index.hbs
@@ -1,0 +1,30 @@
+<Hds::Tabs as |T|>
+  <T.Tab>{{t 'titles.details'}} </T.Tab>
+  <T.Tab @icon='terminal' {{on 'click' this.openTerminal}}>
+    {{t 'resources.session.shell'}}
+  </T.Tab>
+  <T.Panel>
+    {{#if (and @model.proxy_address @model.proxy_port)}}
+      <Session::ProxyUrl
+        @proxyAddress={{@model.proxy_address}}
+        @proxyPort={{@model.proxy_port}}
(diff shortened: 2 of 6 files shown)


ui/desktop/app/components/session/terminal-tabs/index.hbs
```diff
@@ -0,0 +1,30 @@
...
+            'resources.session.messages.connected.description'
+            targetDisplayName=@model.target.displayName
+          }}
+        />
+      </Hds::ApplicationState>
+    {{/if}}</T.Panel>
+  <T.Panel>
+    <div id='terminal-container'></div>
+  </T.Panel>
+</Hds::Tabs>
```
cameronperera, line 30: Did all of this have to move into this component? Seems like we could have had just the `<div id='terminal-container'></div>` in here?

addons/core/translations/resources/en-us.yaml
```diff
@@ -299,6 +299,8 @@ session:
   proxy-url:
     address-port: Address & port
     ssh: SSH
+  #  shell: Shell {shellCount, plural, one {} other {#}}
```
cameronperera, line 302: Do we need to keep this translation still?

ui/desktop/app/components/session/terminal-tabs/index.hbs
ZedLi: This was working originally under the assumption of having multiple terminal tabs. Either way I think the right abstraction is having _all_ the tabs under one component since it has to be aware of all possible terminals

addons/core/translations/resources/en-us.yaml
ZedLi: No, this was pending on a discussion on whether we're going to move forward with multiple terminals

DhariniJeeva (approved):

ui/desktop/app/components/session/terminal-tabs/index.hbs
cameronperera: With the shift to just one terminal for now, I don't think this should be a concern. If we keep all of this template code here, I would argue that the name should change to reduce confusion. However, I would like to see components like this be a little smaller and more focused. Right now this component has template code for tabs for the session details page, along with the ProxyUrl, CredentialsPanel, and the Hds::ApplicationState component. That seems like too much for this one component.

ZedLi: The concern is still present. The main issue is the `openTerminal` and state that's needed. I want a component to encapsulate this HDS tab component including the ability to create the terminal which can't live in a component that is just the `panel` and I don't want it in a route. I originally called it `hds-tabs` but changed it to `terminal-tabs` but didn't have a better name, I'm open to suggestions for the name for sure.

If the issue is you think there's too much happening in the details tab, that could be a separate component then which could happen in a followup PR.

cameronperera: That all sounds good and I am happy for these topics to be addressed in a later PR so we aren't blocking any work. As these tabs are not directly specific to the terminals but the session as a whole, how does `session/tabs` sound? It would follow the same approach as `Session::ProxyUrl` and just be `Session::Tabs`.

PR #1881, opened 2023-09-06
feat: 🎸 Clean up killing terminals

ZedLi:
## Description
Cleaned up the event handlers on the browser and in electron to properly dispose of terminals.

### DIFF
--- a/ui/desktop/electron-app/src/ipc/handlers.js
+++ b/ui/desktop/electron-app/src/ipc/handlers.js
@@ -127,19 +127,36 @@ ipcMain.on('createTerminal', (event, payload) => {
     cwd: process.env.HOME,
     env: process.env,
   });
+  const incomingDataChannel = `terminalIncomingData-${id}`;
+  const keystrokeChannel = `terminalKeystroke-${id}`;
+  const resizeChannel = `resize-${id}`;
+  const removeChannel = `removeTerminal-${id}`;
 
   // This sends to the renderer and xterm whatever the ptyProcess (host terminal) outputs.
   ptyProcess.on('data', function (data) {
-    sender.send(`terminalIncomingData-${id}`, data);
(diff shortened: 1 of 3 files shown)


ui/desktop/electron-app/src/ipc/handlers.js
```diff
@@ -127,19 +127,36 @@ ipcMain.on('createTerminal', (event, payload) => {
...
+  ipcMain.on(resizeChannel, (event, { cols, rows }) => {
     ptyProcess.resize(cols, rows);
   });
+
+  // We use `ipcMain.once` as we want this listener cleaned up
+  // after killing the ptyProcess as well.
+  ipcMain.once(removeChannel, () => {
+    // Just let the error bubble up since we can't do anything
+    try {
+      ptyProcess.kill();
```
ZedLi, line 155: Should we do anything if there's an error killing the process?

calcaide: @ZedLi 
That's actually a pretty good question. What we can do? I don't think we need to report this to the user. But if this can cause problems with other terminals, maybe we can force a refresh of the app?

calcaide (approved):
LGTM!! Nice work!

ui/desktop/electron-app/src/ipc/handlers.js
DhariniJeeva: I like @calcaide's idea of force refreshing the app in case of multiple terminals.

DhariniJeeva (approved):

ui/desktop/electron-app/src/ipc/handlers.js
ZedLi: Wouldn't that be too disruptive for the user? Does restarting the app guarantee the killing of the spawned processes?

PR #1887, opened 2023-09-07
feat: 🎸 allow for connection for target with single host

cameronperera:
## Description

✅ Closes: ICU-10431

:tickets: [Jira ticket](https://hashicorp.atlassian.net/browse/ICU-10431)

Allow for connection for target with single host. Next we will [show a modal of host](https://hashicorp.atlassian.net/browse/ICU-10432) to pick from when trying to establish a connection. This is only the happy path and tests will be updated once the flow for multiple hosts has been started.

## How to Test

Mirage does not have targets with only a single host so please refer to gif below.

single-host

## Checklist:

### DIFF
--- a/ui/desktop/app/routes/scopes/scope/projects/targets/target.js
+++ b/ui/desktop/app/routes/scopes/scope/projects/targets/target.js
@@ -44,18 +44,18 @@ export default class ScopesScopeProjectsTargetsTargetRoute extends Route {
       );
 
       // Extract host ids from all host sets
-      const hostIds = hostSets.map(({ host_ids }) => host_ids);
+      const hostIds = hostSets.flatMap(({ host_ids }) => host_ids);
 
       // Load unique hosts
-      const uniqueHostIds = new Set(hostIds.flat());
+      const uniqueHostIds = new Set(hostIds);
 
       const hosts = await Promise.all(
-        [...uniqueHostIds].map(
-          async (hostId) => await this.store.findRecord('host', hostId)
+        [...uniqueHostIds].map((hostId) =>
+          this.store.findRecord('host', hostId)
         )
       );
 
-      if (isConnecting && hosts.length < 2) {
+      if (isConnecting && hosts.length === 1) {
         await this.connect(model);
       }
 
@@ -70,7 +70,7 @@ export default class ScopesScopeProjectsTargetsTargetRoute extends Route {
   @action
   @loading
   async preConnect(target) {
-    if (target.address || target.hosts.length < 2) {
+    if (target.address || target.hosts.length === 1) {
       await this.connect(target);
     }
   }
--- a/ui/desktop/app/routes/scopes/scope/projects/targets/target.js
+++ b/ui/desktop/app/routes/scopes/scope/projects/targets/target.js
@@ -23,19 +23,44 @@ export default class ScopesScopeProjectsTargetsTargetRoute extends Route {
    * Load a target
    * @param {object} params
    * @param {string} params.target_id
-   * @param {bool} params.isConnecting
    * @return {TargetModel}
    */
-  async model({ target_id, isConnecting }) {
-    const target = await this.store.findRecord('target', target_id, {
+  model({ target_id }) {
+    return this.store.findRecord('target', target_id, {
       reload: true,
     });
+  }
 
-    if (isConnecting) {
-      await this.preConnect(target);
+  async afterModel(model, transition) {
+    const { isConnecting } = transition.to.queryParams;
+
+    if (isConnecting && model.address) {
+      await this.connect(model);
+    } else {
+      const hostSets = await Promise.all(
+        model.host_sources.map(({ host_source_id }) =>
+          this.store.findRecord('host-set', host_source_id)
+        )
+      );
+
+      // Extract host ids from all host sets
+      const hostIds = hostSets.map(({ host_ids }) => host_ids);
+
+      // Load unique hosts
+      const uniqueHostIds = new Set(hostIds.flat());
+
+      const hosts = await Promise.all(
+        [...uniqueHostIds].map(
+          async (hostId) => await this.store.findRecord('host', hostId)
+        )
+      );
+
+      if (isConnecting && hosts.length < 2) {
+        await this.connect(model);
+      }
+
+      model.set('hosts', hosts);
     }
-
-    return target;
   }
 
   /**
@@ -45,7 +70,7 @@ export default class ScopesScopeProjectsTargetsTargetRoute extends Route {
   @action
   @loading
   async preConnect(target) {
-    if (target.address) {
+    if (target.address || target.hosts.length < 2) {
       await this.connect(target);
     }
   }
--- a/ui/desktop/app/routes/scopes/scope/projects/targets/target.js
+++ b/ui/desktop/app/routes/scopes/scope/projects/targets/target.js
@@ -23,22 +23,18 @@ export default class ScopesScopeProjectsTargetsTargetRoute extends Route {
    * Load a target
    * @param {object} params
    * @param {string} params.target_id
-   * @return {TargetModel}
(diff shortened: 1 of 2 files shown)


ui/desktop/app/routes/scopes/scope/projects/targets/target.js
```diff
@@ -23,19 +23,44 @@ export default class ScopesScopeProjectsTargetsTargetRoute extends Route {
...
+      const uniqueHostIds = new Set(hostIds.flat());
 
-    return target;
+      const hosts = await Promise.all(
+        [...uniqueHostIds].map(
+          async (hostId) => await this.store.findRecord('host', hostId)
+        )
+      );
+
+      if (isConnecting && hosts.length < 2) {
```
DhariniJeeva, line 58: Should this be `hosts.length === 1`, if we have 0 hosts, this condition would pass.

ui/desktop/app/routes/scopes/scope/projects/targets/target.js
```diff
@@ -23,19 +23,44 @@ export default class ScopesScopeProjectsTargetsTargetRoute extends Route {
...
-    if (isConnecting) {
-      await this.preConnect(target);
-    }
+  async afterModel(model, transition) {
+    const { isConnecting } = transition.to.queryParams;
+
+    if (isConnecting && model.address) {
+      await this.connect(model);
+    } else {
+      const hostSets = await Promise.all(
```
DhariniJeeva, line 40: Do you need to await this here?

ui/desktop/app/routes/scopes/scope/projects/targets/target.js
```diff
@@ -23,19 +23,44 @@ export default class ScopesScopeProjectsTargetsTargetRoute extends Route {
...
+  model({ target_id }) {
+    return this.store.findRecord('target', target_id, {
       reload: true,
     });
+  }
 
-    if (isConnecting) {
-      await this.preConnect(target);
-    }
+  async afterModel(model, transition) {
```
ZedLi, line 34: I actually feel most of this fetching should happen in the `model` hook which is where it seems most appropriate. When i had originally mentioned you could use the `afterModel` hook it was specifically for when you check for if you should directly connect (lines 58-60) since that causes a transition and that makes sense to have it in the afterModel hook. That way the actual model returned from the `model` hook is accurate

ui/desktop/app/routes/scopes/scope/projects/targets/target.js
```diff
@@ -23,19 +23,44 @@ export default class ScopesScopeProjectsTargetsTargetRoute extends Route {
...
+      await this.connect(model);
+    } else {
+      const hostSets = await Promise.all(
+        model.host_sources.map(({ host_source_id }) =>
+          this.store.findRecord('host-set', host_source_id)
+        )
+      );
+
+      // Extract host ids from all host sets
+      const hostIds = hostSets.map(({ host_ids }) => host_ids);
```
ZedLi, line 47:
```suggestion
      const hostIds = hostSets.flatMap(({ host_ids }) => host_ids);
```

ui/desktop/app/routes/scopes/scope/projects/targets/target.js
```diff
@@ -23,19 +23,44 @@ export default class ScopesScopeProjectsTargetsTargetRoute extends Route {
...
+
+      // Load unique hosts
+      const uniqueHostIds = new Set(hostIds.flat());
```
ZedLi, line 50:
```suggestion
      const uniqueHostIds = new Set(hostIds);
```

calcaide, line 50: I am not understanding why we use `new Set` here. As per my understanding `.flat()` will return a new array, but `new Set` will return an object.

Later on (below) you use `uniqueHostIds` with `array.map()` which needs and array. Maybe we can avoid uniqueHostsIds types transformations and just operate with array?

ui/desktop/app/routes/scopes/scope/projects/targets/target.js
cameronperera: Yes, thanks for catching that. Updating now.

ui/desktop/app/routes/scopes/scope/projects/targets/target.js
```diff
@@ -23,19 +23,44 @@ export default class ScopesScopeProjectsTargetsTargetRoute extends Route {
...
 
-    return target;
+      const hosts = await Promise.all(
+        [...uniqueHostIds].map(
+          async (hostId) => await this.store.findRecord('host', hostId)
```
ZedLi, line 54: You shouldn't need to `await` these since you're awaiting the array of promises with `Promise.all` anyways

ui/desktop/app/routes/scopes/scope/projects/targets/target.js
```diff
@@ -23,19 +23,44 @@ export default class ScopesScopeProjectsTargetsTargetRoute extends Route {
...
+        )
+      );
+
+      if (isConnecting && hosts.length < 2) {
+        await this.connect(model);
+      }
+
+      model.set('hosts', hosts);
```
ZedLi, line 62: Should we return this separately rather than setting a property on the model for a field we didn't previously declare? 

Alternatively, does it make sense to have the target model do the retrieving and loading of the hosts themselves?

ui/desktop/app/routes/scopes/scope/projects/targets/target.js
cameronperera: Yes. This tells the top level promise to not resolve until all of the `this.store.findRecord('host-set', host_source_id)` calls resolve. Similar to line 54... expect we don't actually need the `async/await` on the individual `findRecord` call. So I have removed those.

ui/desktop/app/routes/scopes/scope/projects/targets/target.js
cameronperera: Yeah I would like to do some testing with this. I simply added back the host fetching logic we had for the `targets/target-id/hosts` route we had in the past. The `Set()` is to remove any duplicate hostIds. However, I am not sure if we can have duplicates so I would need to confirm if that is even a concern. I would like to avoid transforming the data back and forth if possible. I will look into this to see if this was done for performance reasons.

ui/desktop/app/routes/scopes/scope/projects/targets/target.js
cameronperera: Gotcha. That makes sense. I will update that now.

ui/desktop/app/routes/scopes/scope/projects/targets/target.js
```diff
@@ -45,7 +70,7 @@ export default class ScopesScopeProjectsTargetsTargetRoute extends Route {
   @action
   @loading
   async preConnect(target) {
-    if (target.address) {
+    if (target.address || target.hosts.length < 2) {
```
DhariniJeeva, line 73: Same comment as above, should it be `target.hosts.length === 1`

ui/desktop/app/routes/scopes/scope/projects/targets/target.js
cameronperera: Yep. I caught that after the push. Removed.

ui/desktop/app/routes/scopes/scope/projects/targets/target.js
cameronperera: Agreed. Updated.

ZedLi (approved):
Looks great!

ui/desktop/app/routes/scopes/scope/projects/targets/target.js
calcaide: @cameronperera gotcha. I think we could do some pair programming next week? I am missing a bit of context and a quick pair session will help to understand better the purpose of this 😄

PR #1889, opened 2023-09-08
feat: 🎸 add host details to session details screen if present

cameronperera:
## Description

:tickets: [Jira ticket](https://hashicorp.atlassian.net/browse/ICU-10282)

Before:
Screen Shot 2023-09-08 at 2 49 38 PM

After:
Screen Shot 2023-09-08 at 2 36 49 PM

## How to Test

Not testable via mirage yet. See screenshot above

## Checklist:

### DIFF
--- a/ui/desktop/app/styles/app.scss
+++ b/ui/desktop/app/styles/app.scss
@@ -518,7 +518,7 @@ $chevron-down: url("data:image/svg+xml;utf-8,<svg viewBox='0 0 24 24' fill='%231
   padding: 1.5rem 2.5rem;
   margin-top: 3.5rem;
 
-  .host-detail {
+  .detail-group:not(:last-child) {
     margin-bottom: 2.25rem;
   }
 
--- a/ui/desktop/app/styles/app.scss
+++ b/ui/desktop/app/styles/app.scss
@@ -518,7 +518,7 @@ $chevron-down: url("data:image/svg+xml;utf-8,<svg viewBox='0 0 24 24' fill='%231
   padding: 1.5rem 2.5rem;
   margin-top: 3.5rem;
 
-  .detail-group:first-child {
+  .host-detail {
     margin-bottom: 2.25rem;
   }
 
--- a/ui/desktop/app/styles/app.scss
+++ b/ui/desktop/app/styles/app.scss
@@ -512,11 +512,15 @@ $chevron-down: url("data:image/svg+xml;utf-8,<svg viewBox='0 0 24 24' fill='%231
   }
 }
 
-.details-sidebar-target-details {
+.details-sidebar {
   background-color: var(--token-color-palette-neutral-50);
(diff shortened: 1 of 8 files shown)


ui/desktop/app/styles/app.scss
```diff
@@ -512,12 +512,16 @@ $chevron-down: url("data:image/svg+xml;utf-8,<svg viewBox='0 0 24 24' fill='%231
...
 }
 
-.details-sidebar-target-details {
+.details-sidebar {
   background-color: var(--token-color-palette-neutral-50);
   border-radius: 0.313rem;
   padding: 1.5rem 2.5rem;
   margin-top: 3.5rem;
 
+  .detail-group:first-child {
```
ZedLi, line 521: Doesn't this add margin if there's only one child, e.g. in the case where we just show a target with no host details?

calcaide (approved):
LGTM! left a none blocking comment

ui/desktop/app/styles/app.scss
cameronperera: Yeah. Let me update this.

cameronperera: Fixed.

ui/desktop/app/styles/app.scss
```diff
@@ -512,12 +512,16 @@ $chevron-down: url("data:image/svg+xml;utf-8,<svg viewBox='0 0 24 24' fill='%231
...
 }
 
-.details-sidebar-target-details {
+.details-sidebar {
   background-color: var(--token-color-palette-neutral-50);
   border-radius: 0.313rem;
   padding: 1.5rem 2.5rem;
   margin-top: 3.5rem;
 
+  .host-detail {
```
ZedLi, line 521: This works for this use-case but if we ever add another section to the sidebar we'd have to refactor this again. I feel like using  `:not(:first-child)` or `:not(:last-child)` would work better for this scenario, what do you think?

cameronperera: I was wondering if we should keep it more dynamic because I believe we will add "Active Sessions" to the sidebar for the `targets/target_id` route as well. Let me play around with this.

cameronperera: Updated. Let me know what you think.

ZedLi: Yup I think this looks good. One last question is do we plan on using the `detail-group` class for anything else?

If not should we just do something like this since it's only used for this margin right now:
```
  > :not(:last-child) {
    margin-bottom: 2.25rem;
  }
```

ZedLi (approved):

ui/desktop/app/styles/app.scss
cameronperera: Nope, it really is just needed for spacing. This is nice. I will update to this.

PR #1890, opened 2023-09-08
Chore terminal tweaks

calcaide:
## Description

A quick change to use user shell by default.

## Checklist:

### DIFF
--- a/ui/desktop/electron-app/src/ipc/handlers.js
+++ b/ui/desktop/electron-app/src/ipc/handlers.js
@@ -118,6 +118,8 @@ handle('closeWindow', () => app.quit());
 ipcMain.on('createTerminal', (event, payload) => {
   const { id, cols, rows } = payload;
   const { sender } = event;
+  console.log('Carlos');
+  console.log(process.env.SHELL);
   const terminalShell =
     os.platform() === 'win32' ? 'powershell.exe' : '/bin/bash';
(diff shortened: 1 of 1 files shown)


ui/desktop/electron-app/src/ipc/handlers.js
```diff
@@ -119,7 +119,7 @@ ipcMain.on('createTerminal', (event, payload) => {
   const { id, cols, rows } = payload;
   const { sender } = event;
   const terminalShell =
-    os.platform() === 'win32' ? 'powershell.exe' : '/bin/bash';
+    os.platform() === 'win32' ? 'powershell.exe' : process.env.SHELL;
```
ZedLi, line 122: I think we should still default to bash just in case, even though any common unix system will have the variable defined. But we can't assume it always will be for some custom flavor of unix or if the user modified it for whatever reason. 
```suggestion
    os.platform() === 'win32' ? 'powershell.exe' : process.env.SHELL || '/bin/bash';
```

ZedLi (approved):

ui/desktop/electron-app/src/ipc/handlers.js
cameronperera: Second this change.

cameronperera (approved):

PR #1891, opened 2023-09-08
Icu 10926 add dynamic credential fields in sb model

DhariniJeeva:
:tickets: [Jira ticket](https://hashicorp.atlassian.net/browse/ICU-10926)

## Description

This PR adds support for dynamic credentials in storage bucket creation.

Screenshot 2023-09-08 at 2 45 37 PM

 
 
## Checklist:


addons/api/addon/generated/models/storage-bucket.js
```diff
@@ -79,6 +79,13 @@ export default class GeneratedStorageBucketModel extends BaseModel {
   })
   secrets_hmac;
 
+  @attr('boolean', {
+    isNestedAttribute: true,
+    description: '',
```
cameronperera, line 84: Do we have a description we can provide here?

cameronperera (approved):
LGTM. Left a non-blocking comment.

addons/api/tests/unit/serializers/storage-bucket-test.js
```diff
@@ -57,6 +65,44 @@ module('Unit | Serializer | storage bucket', function (hooks) {
...
+      worker_filter: 'workerfilter',
+      region: 'eu-west-1',
+      access_key_id: '',
+      secret_access_key: '',
+      secrets_hmac: 'hmac',
+      disable_credential_rotation: true,
+      role_arn: 'arn',
+      role_external_id: null,
+      role_session_name: null,
+      role_tags: [],
```
ZedLi, line 86: Can we have these fields be non null so we can show that they get serialized correctly?

ZedLi (approved):
Looks good!

addons/api/tests/unit/serializers/storage-bucket-test.js
calcaide: Agreed with Zed's comment. We should make sure we serialize correctly those files with and without values.

calcaide (approved):
LGTM!! great work! left a none blocking comment

addons/api/tests/unit/serializers/storage-bucket-test.js
DhariniJeeva: updated the attr type to use `object-as-array` transform as that is the expected backend format from the request.

PR #1897, opened 2023-09-13
feat: 🎸 add support for dynamic credentials in storage bucket

DhariniJeeva:
:tickets: [Jira ticket](https://hashicorp.atlassian.net/browse/ICU-10927)

## Description

This PR adds support for dynamic credentials in the storage bucket form.

**NOT INCLUDED**: Update state of the credentials (will be in a separate PR to keep this one simple)

Model change: [credential type is defaulted to static]

Screenshot 2023-09-13 at 1 49 02 PM

Before: 
Screenshot 2023-09-13 at 1 44 11 PM

After:
[edit mode]
 
Screenshot 2023-09-13 at 1 56 18 PM

[new form]
 
Screenshot 2023-09-13 at 1 56 07 PM

## How to Test

:technologist: https://boundary-c93ixys59-hashicorp.vercel.app/scopes/global/storage-buckets/new

## Checklist:


addons/api/addon/serializers/storage-bucket.js
```diff
@@ -8,6 +8,11 @@ import ApplicationSerializer from './application';
 export default class StorageBucketSerializer extends ApplicationSerializer {
   serialize(snapshot) {
     const serialized = super.serialize(...arguments);
+    //Dynamic credentials do not allow credential rotation, and the API requires this field to be
+    //sent along with the dynamic credential fields
+    if (serialized.attributes.role_arn) {
```
DhariniJeeva, line 13: FYI- for now, I have it here, but I will be refactoring these for the update functionality to separate the fields based on the type selection using `for` option in the generated model

ZedLi: Can we do it like this? We're calling this credential type as "type" and switching off of it but this isn't really true is it?

We already have a type and it's `aws`, how would all of this work when we add another type like `azure`?

DhariniJeeva: good point, this may not work if we have more than one `type`. In that case, I could do something like how we do in [host-catalog ](https://github.com/hashicorp/boundary-ui/blob/main/addons/api/addon/serializers/host-catalog.js#L45)where we split the fields based on the record type and the credential type

addons/api/addon/models/storage-bucket.js
```diff
@@ -13,6 +13,14 @@ export const TYPES_STORAGE_BUCKET_PLUGIN = Object.freeze([
   TYPE_STORAGE_BUCKET_PLUGIN_AWS_S3,
 ]);
 
+export const TYPE_CREDENTIAL_STATIC = 'static';
+export const TYPE_CREDENTIAL_DYNAMIC = 'dynamic';
+
+export const TYPES_CREDENTIALS = Object.freeze([
```
calcaide, line 19: None blocking comment: What is the purpose of TYPES_CREDENTIALS being an `object.freeze`?

DhariniJeeva: This is the pattern we are following when creating new type constants to prevent modifications

addons/api/addon/serializers/storage-bucket.js
DhariniJeeva: This has been refactored in [update](https://github.com/hashicorp/boundary-ui/pull/1901) pr

calcaide (approved):
Nice work!! Thank you for this!

ui/admin/tests/acceptance/storage-buckets/create-test.js
```diff
@@ -105,6 +107,44 @@ module('Acceptance | storage-buckets | create', function (hooks) {
...
+    assert.dom('.hds-form-radio-card').exists({ count: 2 });
+
+    await click(DYNAMIC_CREDENTIAL_SELECTOR);
+    await fillIn(ROLE_ARN_SELECTOR, 'test-arn-id');
+    await click(SAVE_BTN_SELECTOR);
+    const storageBucket = this.server.schema.storageBuckets.findBy({
+      name: NAME_FIELD_TEXT,
+    });
+    assert.strictEqual(storageBucket.name, NAME_FIELD_TEXT);
+    assert.strictEqual(getStorageBucketCount(), storageBucketCount + 1);
```
lisbet-alvarez, line 126: one small thing: i think a assert can be added here to make sure that `storageBucket.credentialType` is equal to `TYPE_CREDENTIAL_STATIC`

lisbet-alvarez (approved):
Looks great 🙌

ui/admin/tests/acceptance/storage-buckets/create-test.js
DhariniJeeva: ah yes, good point, I'll include these

DhariniJeeva: @lisbet-alvarez - So, the model properties(shown in the desc screenshot) aren't exposed here. However, the unit test on the SB model covers the logic for `credentialtype`.  I added some extra UI tests to ensure we get the correct cred type.

lisbet-alvarez: ahh okay makes sense, thanks for looking into that!

PR #1898, opened 2023-09-14
feat: 🎸 create Host Modal for connecting to specific modal

cameronperera:
## Description

✅ Closes: ICU-10432

:tickets: [Jira ticket](https://hashicorp.atlassian.net/browse/ICU-10432)

Before: `N/A`
After:

https://github.com/hashicorp/boundary-ui/assets/29386339/420c1e6a-269d-441f-9455-160e42375c26

## How to Test

Click on `Connect` or `Quick Connect` buttons for the first target. Either from the targets list screen or the target details screen.

:desktop_computer: [Desktop preview](https://boundary-ui-desktop-lko7pe5xo-hashicorp.vercel.app/scopes/global/projects/targets/t_l5fhet6083?isConnecting=true)

## Checklist:

### DIFF
--- a/addons/api/addon/models/host.js
+++ b/addons/api/addon/models/host.js
@@ -5,6 +5,7 @@
 
 import GeneratedHostModel from '../generated/models/host';
 import { attr } from '@ember-data/model';
+import { pluginTypes } from './host-catalog';
 
 export default class HostModel extends GeneratedHostModel {
   // =attributes
@@ -56,6 +57,14 @@ export default class HostModel extends GeneratedHostModel {
    */
   get isPlugin() {
     return this.type === 'plugin';
+  }
+
+  /**
+   * True if the host catalog is an unknown type.
+   * @type {boolean}
+   */
+  get isUnknown() {
+    return this.isPlugin && !pluginTypes.includes(this.plugin?.name);
   }
 
   /**
--- a/ui/desktop/app/styles/app.scss
+++ b/ui/desktop/app/styles/app.scss
@@ -652,3 +652,35 @@ $chevron-down: url("data:image/svg+xml;utf-8,<svg viewBox='0 0 24 24' fill='%231
     border-radius: 0 0 5px 5px;
   }
 }
+
+// Host Modal
+.host-list-container {
+  border: 1px solid var(--token-color-border-primary);
+  padding: 0.25rem;
+  border-radius: 6px;
+}
+
+.host-container {
+  button {
+    background: unset;
+    border: unset;
+    display: flex;
+    justify-content: space-between;
+    align-items: center;
+    width: 100%;
+    padding: 0.5rem;
+
+    .host-info-container {
+      display: flex;
+      gap: 0.75rem;
+
+      .host-info {
+        text-align: left;
+      }
+    }
+  }
+
+  .host-arrow {
+    color: var(--token-color-foreground-faint);
+  }
+}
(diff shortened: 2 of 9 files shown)


addons/api/addon/models/host.js
```diff
@@ -5,6 +5,7 @@
 
 import GeneratedHostModel from '../generated/models/host';
 import { attr } from '@ember-data/model';
+import { pluginTypes } from './host-catalog';
```
calcaide, line 8: None blocking comment: is this what you are looking to refactor in the future with the `TYPE` thread we discuss yesterday on slack? [link here](https://hashicorp.slack.com/archives/C012GTH1C4E/p1694620307660139)

cameronperera: Yes. Just with it being a larger refactor, I didn't want to include it in this PR as it would grow to 20+ files changed.

calcaide: Definetly and totally onboard with your opinion, I was just wondering after our chat 😝 . Good work!

ui/desktop/app/styles/app.scss
```diff
@@ -652,3 +652,35 @@ $chevron-down: url("data:image/svg+xml;utf-8,<svg viewBox='0 0 24 24' fill='%231
...
+// Host Modal
+.host-list-container {
+  border: 1px solid var(--token-color-border-primary);
+  padding: 0.25rem;
+  border-radius: 6px;
+}
+
+.host-container {
+  button {
+    background: unset;
```
calcaide, line 665: None blocking comment: what is the purpose/reason we use `unset` here?

calcaide (approved):
LGTM! Good work

ui/desktop/app/styles/app.scss
cameronperera: I wanted the actions and capabilities of a button without it looking like a button so I decided to just unset this style. I can update this to match the background color of the modal if that approach is more clear.

calcaide: No problem, I was just curious.

ui/desktop/app/styles/app.scss
```diff
@@ -653,6 +653,38 @@ $chevron-down: url("data:image/svg+xml;utf-8,<svg viewBox='0 0 24 24' fill='%231
...
 
+// Host Modal
+.host-list-container {
+  border: 1px solid var(--token-color-border-primary);
+  padding: 0.25rem;
+  border-radius: 6px;
+}
+
+.host-container {
+  button {
```
ZedLi, line 664: This is a design question but I wonder if there should be some interaction when hovering over the buttons? I don't get any feedback when hovering over them which feels like it doesn't make it obvious they're buttons

ZedLi (approved):
Looks good!

One random question I have is can we go back in electron? I notice in the browser going back from the modal is broken but if we can't go back it's a moot point

cameronperera:
> Looks good!
> 
> One random question I have is can we go back in electron? I notice in the browser going back from the modal is broken but if we can't go back it's a moot point

I do not believe we can. I tried going back while using the actual electron build and it just does nothing.

ui/desktop/app/styles/app.scss
cameronperera: That is a great feedback. Yeah I can tweak this to give it some indication that you are hovering over a button. I will move that to a new PR since this is already large.

PR #1907, opened 2023-09-19
refactor: 💡sb dynamic credential type secret fields and how we serialize

DhariniJeeva:
This PR refactors the dynamic credential serializer and secrets within the storage bucket form

## Checklist:


addons/api/addon/serializers/storage-bucket.js
```diff
@@ -49,17 +49,22 @@ export default class StorageBucketSerializer extends ApplicationSerializer {
         json.attributes[key] = null;
     }
 
+    //json.secrets is only present, if there are any updates to the secret fields
+    if (options.isNestedSecret && json.secrets) {
+      if (!fieldsByCredentialType[credentialType].includes(key)) {
+        delete json.secrets[key];
+      }
```
ZedLi, line 56: Do we have to worry about having an empty secrets object if we delete all the keys?

ZedLi:
Also I was wondering if we should persist the `worker_filter` field when toggling between the radio cards since it's the same field instead of clearing it?

Or do we not care since it's below the radio cards and it would be weirder

addons/api/tests/unit/serializers/storage-bucket-test.js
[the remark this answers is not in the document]
DhariniJeeva: thanks for the feedback. I thought about this originally but decided to go with model data since the secrets were getting cleared, but on second thought, using the user action seems like the cleanest way. Updated!

addons/api/addon/serializers/storage-bucket.js
DhariniJeeva: to be on the safe side - I'm getting rid of the entire secrets now

DhariniJeeva:
> Also I was wondering if we should persist the `worker_filter` field when toggling between the radio cards since it's the same field instead of clearing it?
> 
> Or do we not care since it's below the radio cards and it would be weirder

good point - since we are persisting other common fields, this should be persisted as well.

addons/api/tests/unit/serializers/storage-bucket-test.js
```diff
@@ -199,13 +221,14 @@ module('Unit | Serializer | storage bucket', function (hooks) {
...
         role_tags: { Project: 'Automation', foo: 'bar' },
+        secrets_hmac: null,
       },
       version: 1,
     };
     assert.deepEqual(serializedRecord, expectedResult);
   });
 
-  test('it normalizes correctly', function (assert) {
+  test.skip('it normalizes correctly', function (assert) {
```
ZedLi, line 231: Was this meant to be skipped?

DhariniJeeva: ah nope, reverted this

PR #1909, opened 2023-09-20
Icu-11111 don't display empty modal when a target has no target address and no host sources

lisbet-alvarez:
:tickets: [Jira ticket](https://hashicorp.atlassian.net/browse/ICU-11111 )

## Description

If a target is missing an address and has no host sources then an empty modal was displayed. This is a bug. To fix this, the connect buttons were disabled and a new app state message was added for this case.

Before:
Screenshot 2023-09-19 at 1 52 37 PM

After:
Screenshot 2023-09-19 at 3 36 49 PM

Screenshot 2023-09-20 at 1 18 40 PM

## How to Test
1. Run boundary in dev mode (use latest binary).
2. Create a target in admin UI with no hosts and no address.
3. Run desktop client using `ENABLE_MIRAGE=false yarn start:desktop` and while on this branch.
4. Open desktop client and validate that button in target list view is disabled and button in details view is disabled. Validate correct message is displayed. 

## Checklist:

### DIFF
--- a/addons/api/addon/models/target.js
+++ b/addons/api/addon/models/target.js
@@ -333,4 +333,8 @@ export default class TargetModel extends GeneratedTargetModel {
   get isSSH() {
     return this.type === TYPE_TARGET_SSH;
   }
+
+  get isConnectable() {
+    return this.address || this.host_sources?.length;
+  }
 }
--- a/addons/api/addon/models/target.js
+++ b/addons/api/addon/models/target.js
@@ -333,8 +333,4 @@ export default class TargetModel extends GeneratedTargetModel {
   get isSSH() {
     return this.type === TYPE_TARGET_SSH;
   }
-
-  get isConnectable() {
-    return this.address || this.host_sources?.length;
-  }
 }
--- a/ui/desktop/app/routes/scopes/scope/projects/targets.js
+++ b/ui/desktop/app/routes/scopes/scope/projects/targets.js
@@ -52,14 +52,6 @@ export default class ScopesScopeProjectsTargetsRoute extends Route {
     // Recursively query all targets within the current scope
     let targets = await this.store.query('target', queryOptions);
 
-    // Load all targets in order to populate host_sources field.
-    // This is meant to be a temporary solution.
-    await Promise.all(
-      targets.map(async ({ id: target_id }) => {
-        await this.store.findRecord('target', target_id);
-      })
-    );
-
     // Filter out targets to which users do not have the connect ability
     targets = targets.filter((t) => this.can.can('connect target', t));
 
--- a/ui/desktop/app/routes/scopes/scope/projects/targets.js
+++ b/ui/desktop/app/routes/scopes/scope/projects/targets.js
@@ -52,6 +52,14 @@ export default class ScopesScopeProjectsTargetsRoute extends Route {
     // Recursively query all targets within the current scope
     let targets = await this.store.query('target', queryOptions);
 
+    // Load all targets in order to populate host_sources field.
+    // This is meant to be a temporary solution.
+    await Promise.all(
+      targets.map(async ({ id: target_id }) => {
+        await this.store.findRecord('target', target_id);
+      })
+    );
+
     // Filter out targets to which users do not have the connect ability
     targets = targets.filter((t) => this.can.can('connect target', t));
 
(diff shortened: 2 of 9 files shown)


DhariniJeeva (approved):

cameronperera (approved):
LGTM

ui/desktop/app/templates/scopes/scope/projects/targets/index.hbs
```diff
@@ -167,6 +167,10 @@
                   @route='scopes.scope.projects.targets.target'
                   @model={{B.data.id}}
                   @query={{hash isConnecting=true}}
+                  disabled={{and
+                    (not B.data.address)
+                    (not B.data.host_sources)
+                  }}
```
cameronperera, line 173: Non-blocking comment: For the sake of time I am happy with this. After giving it some more thought, we could add this logic to the target model then called `B.data.isConnectable()` (or a better name than that 😆). It would make this easier to understand at a glance.

lisbet-alvarez: oh great idea, I'll clean this up to use this

ZedLi: Yup I agree with this, it would make it easier to read since we use this check in quite a few places

cameronperera (approved):

dvonanderson (approved):
Looks good Lis

ui/desktop/app/templates/scopes/scope/projects/targets/index.hbs
calcaide: I totally agree with this approach, nice thinking!!

calcaide (approved):
Good work and very nice catch with this @lisbet-alvarez !!

ui/desktop/app/routes/scopes/scope/projects/targets.js
```diff
@@ -52,6 +52,14 @@ export default class ScopesScopeProjectsTargetsRoute extends Route {
     // Recursively query all targets within the current scope
     let targets = await this.store.query('target', queryOptions);
 
+    // Load all targets in order to populate host_sources field.
+    // This is meant to be a temporary solution.
+    await Promise.all(
+      targets.map(async ({ id: target_id }) => {
+        await this.store.findRecord('target', target_id);
```
ZedLi, line 59: You don't need this inner await if you're using `Promise.all`

ui/desktop/app/templates/scopes/scope/projects/targets/index.hbs
```diff
@@ -160,14 +160,16 @@
...
               {{#if (can 'connect target' B.data)}}
-                <Hds::Button
-                  data-test-targets-connect-button
-                  @text={{t 'resources.session.actions.connect'}}
-                  @color='secondary'
-                  @route='scopes.scope.projects.targets.target'
-                  @model={{B.data.id}}
-                  @query={{hash isConnecting=true}}
-                />
+                {{#if B.data.isConnectable}}
```
ZedLi, line 163: What happens if we don't include this check on the list page and we let the user try and connect? Will it go to the target page and just error?

lisbet-alvarez: it will just error

ui/desktop/app/routes/scopes/scope/projects/targets.js
lisbet-alvarez: oops, thx ill fix that!

ui/desktop/app/templates/scopes/scope/projects/targets/index.hbs
ZedLi: But by that time we'll have a target loaded right, can we prevent the modal from popping up?

I think we really need to not load all the targets individually, this will cause a massive strain for our current users who have a lot of targets

addons/api/addon/models/target.js
```diff
@@ -333,4 +333,8 @@ export default class TargetModel extends GeneratedTargetModel {
   get isSSH() {
     return this.type === TYPE_TARGET_SSH;
   }
+
+  get isConnectable() {
```
ZedLi, line 337: I also think this should probably be an ability

ZedLi (approved):

PR #1925, opened 2023-09-26
fix: 🐛 Stop sending ssh command when session initiated in CLI

ZedLi:
✅ Closes: ICU-11209

:tickets: [Jira ticket](https://hashicorp.atlassian.net/browse/ICU-11209)

## Description
When a user initiates a session from the CLI, we won't be aware of the proxy details so can't automatically initiate an SSH connection. We need to prevent trying to automatically connect in these scenarios.

## How to Test
Initiate a connection from the CLI and then open the shell for that session in the desktop client.

## Checklist:

### DIFF
--- a/ui/desktop/electron-app/src/cli/path.js
+++ b/ui/desktop/electron-app/src/cli/path.js
@@ -5,11 +5,16 @@
 
 const path = require('path');
 const { isWindows } = require('../helpers/platform.js');
+const isDev = require('electron-is-dev');
 
 module.exports = {
   // Returns boundary cli path
   path: () => {
     const name = isWindows() ? 'boundary.exe' : 'boundary';
-    return path.resolve(process.resourcesPath, 'cli', name);
+
+    // The CLI will be located in a different location once packaged up in an ASAR
+    return isDev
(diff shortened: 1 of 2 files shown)


ui/desktop/electron-app/src/cli/path.js
```diff
@@ -5,11 +5,16 @@
...
 module.exports = {
   // Returns boundary cli path
   path: () => {
     const name = isWindows() ? 'boundary.exe' : 'boundary';
-    return path.resolve(process.resourcesPath, 'cli', name);
+
+    // The CLI will be located in a different location once packaged up in an ASAR
+    return isDev
+      ? path.resolve(__dirname, '..', '..', 'cli', name)
+      : path.resolve(process.resourcesPath, 'cli', name);
```
ZedLi, line 18: This was a bug I created when fixing the CLI path in production.

I forgot about dev and broke normal development 😓

cameronperera: Good catch. This is why I am only seeing the issue when I am in dev.

cameronperera (approved):

lisbet-alvarez (approved):
🙌

PR #1931, opened 2023-09-27
test: 💍 Add delete resources e2e UI test

stasryzhov:
✅ Closes: ICU-10897

:tickets: [ICU-10897](https://hashicorp.atlassian.net/browse/ICU-10897)

## Description
This PR adds an e2e UI test to verify different boundary resources can be deleted using UI.
The boundary resources covered within this test are:

- target SSH
- target TCP
- username-password credentials
- static credential store
- vault credential store
- static host set
- static host
- static host catalog
- dynamic aws host catalog
- user
- group
- role in global scope
- role in org scope
- role in project scope
- password account
- auth method
- worker
- project
- org

Note:
The test prepares the test data (i.e. the boundary resources to be deleted) using CLI. The reasons:
- resources creation is not a SUT in this test
- it reduces the test execution time

[ICU-10897]: https://hashicorp.atlassian.net/browse/ICU-10897?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


ui/admin/tests/e2e/tests/delete-resources.spec.js
```diff
@@ -0,0 +1,197 @@
...
+    await page.goto(`/scopes/${projectId}/targets/${sshTargetId}`);
+    await deleteResource(page);
+
+    // Delete TCP target
+    await page.goto(`/scopes/${projectId}/targets/${tcpTargetId}`);
+    await deleteResource(page);
+
+    // Delete username-password credentials
+    await page.goto(
+      `/scopes/${projectId}/credential-stores/${vaultCredentialStoreId}/credentials/${usernamePasswordCredential}`
```
moduli, line 111: Wasn't this created in the static credential store?
```suggestion
      `/scopes/${projectId}/credential-stores/${staticCredentialStoreId}/credentials/${usernamePasswordCredential}`
```

ui/admin/tests/e2e/tests/delete-resources.spec.js
```diff
@@ -0,0 +1,197 @@
...
+    );
+    await deleteResource(page);
+
```
moduli, line 114: Do we need to add the other credential types in here?

Static Credential Store
- User Password Credential
- Key Pair Credential
- JSON Credential

Vault Credential Store
- User Password Credential
- Key Pair Credential

ui/admin/tests/e2e/helpers/boundary-cli.js
```diff
@@ -66,18 +67,471 @@ exports.connectToTarget = async (targetId) => {
...
+  }
+  return hostCatalog.id;
+};
+
+/**
+ * Uses the boundary CLI to create a new dynamic AWS host catalog.
+ * @param {string} projectId ID of the project under which the host catalog will be created.
+ * @returns {string} new host catalog's ID
+ */
+exports.createDynamicAWSHostCatalogCli = async (projectId) => {
```
moduli, line 316:
```suggestion
exports.createDynamicAwsHostCatalogCli = async (projectId) => {
```

moduli (changes requested):

ui/admin/tests/e2e/tests/delete-resources.spec.js
stasryzhov: We do. Out of scope if this PR and ticket.

ui/admin/tests/e2e/helpers/boundary-cli.js
```diff
@@ -66,18 +67,471 @@ exports.connectToTarget = async (targetId) => {
...
+  return newAuthMethod.id;
+};
+
+/**
+ * Uses the boundary CLI to make an auth-method primary.
+ * @param {string} scopeId ID of the scope that the auth-method belongs to.
+ * @param {string} authMethodId ID of the auth method that will be made primary.
+ * @returns {object} make auth method primary response.
+ */
+exports.makeAuthMethodPrimary = async (scopeId, authMethodId) => {
```
calcaide, line 177: None blocking comment: all the methods within this file that uses methodNameCli, I am missing the CLI in this one.

ui/admin/tests/e2e/helpers/boundary-cli.js
calcaide: +1 to this suggestion!

moduli (changes requested):

ui/admin/tests/e2e/tests/delete-resources.spec.js
```diff
@@ -0,0 +1,197 @@
...
+    );
+    let vaultCredentialStoreId = await createNewVaultCredentialStoreCli(
+      projectId,
+      secretPolicyName,
+      boundaryPolicyName
+    );
+    let usernamePasswordCredentialId =
+      await createNewUsernamePasswordCredentialCli(staticCredentialStoreId);
+    let tcpTargetId = await createNewTcpTarget(projectId);
+    let sshTargetId = await createNewSshTarget(projectId);
```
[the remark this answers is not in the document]
stasryzhov: How do you suggest to organize this? Create a `delete-resources-ent.spec.js` file with enterprise related resourses covered?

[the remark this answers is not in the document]
moduli: Similar to the BE test suite. I think it will need to live in `boundary-ui-enterprise`. I don't know if the FE team has a specific naming convention for enterprise-only files.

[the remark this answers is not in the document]
stasryzhov: @calcaide @ZedLi Do you agree with this approach? If so, are there naming conventions for enterprise-only files?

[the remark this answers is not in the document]
moduli: Oh... I did not know there were no enterprise-only files... so how is `boundary-ui-enterprise` used? 

If we're keeping it all in this repo, then maybe one option is to use the tag feature: https://playwright.dev/docs/test-annotations#tag-tests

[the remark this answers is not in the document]
ZedLi: Might be good to sync on this but I don't think it needs to live in the `boundary-ui-enterprise` repo. We haven't had any "enterprise repo" specific files besides github workflows yet.

Our approach to the repos are also different from the backend. We actually have all of the enterprise features in `boundary-ui` gated behind feature flags. The frontend is somewhat different since it doesn't have all the features kept in separate repos as the true gatekeeping happens from access to the boundary enterprise binary.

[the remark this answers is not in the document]
calcaide: @moduli 
Long story short. The UI is baked with all the features and then we use flags to determinate what we "turn" OFF or ON. In fact, in `boundary-ui` we define features specific to OSS, Enterprise and HCPb (is not the cloud-ui you see on the portal but the UI that runs the binary deployed within HCP). [Here is the features file](https://github.com/hashicorp/boundary-ui/blob/main/ui/admin/config/features.js) as example.

As @ZedLi mentioned, I think would be good to do a quick sync about this.

ui/admin/tests/e2e/README.md
```diff
@@ -136,6 +136,11 @@ PWDEBUG=console yarn playwright test --headed --config ./tests/e2e/playwright.co
 PWDEBUG=console yarn playwright test --headed --config ./tests/e2e/playwright.config.js login.spec.js --debug
 ```
 
+To run the tests with a specific tag (e.g. @enterprise), use these commands.
+```bash
+PWDEBUG=console yarn playwright test --grep @enterprise --headed --config ./tests/e2e/playwright.config.js delete-resources.spec.js
+```
```
moduli, line 142: Would would be our use cases be? I see two major ones:
- run all tests applicable to Community Edition
- run all tests applicable to Enterprise (which may include tests from the first set)

How would we be able to achieve those?

ui/admin/tests/e2e/helpers/boundary-cli.js
```diff
@@ -66,18 +67,471 @@ exports.connectToTarget = async (targetId) => {
...
+        `boundary scopes update \
+         -id=${scopeId} \
+         -primary-auth-method-id ${authMethodId} \
+         -format json`
+      )
+    ).item;
+  } catch (e) {
+    console.log(`${e.stderr}`);
+  }
+  return makeAuthMethodPrimaryResponse;
```
moduli, line 191: Wonder if we should change this to return a `bool` on whether or not it's successful? Not sure if we really want to add logic about interpreting CLI responses into tests.

ui/admin/tests/e2e/README.md
stasryzhov: Yeah, in these cases we'd do:

**Community Edition tests:**
`PWDEBUG=console yarn playwright test --grep-invert @enterprise --headed --config ./tests/e2e/playwright.config.js delete-resources.spec.js
`
That will run all tests except those tagged @hashicorp/team-grc-compliance 

**Enterprise tests (i.e. all tests)**
`PWDEBUG=console yarn playwright test --headed --config ./tests/e2e/playwright.config.js delete-resources.spec.js`
That will run all tests.

ui/admin/tests/e2e/helpers/boundary-cli.js
stasryzhov: Discussed with @moduli. For now, not returning anything from the method. 
If there is an error in the CLI command used in this method, it will be caught in the catch block and logged. 
The test will fail on a remove primary auth-method attempt.

moduli (approved):

PR #1934, opened 2023-09-28
update breadcrumbs to use hds::breadcrumbs

lisbet-alvarez:
:tickets: [Jira ticket](https://hashicorp.atlassian.net/browse/ICU-9424)

## Description

Add `@bagaar/ember-breadcrumbs`
Add component that uses HDS::Breadcrumbs
Remove previous custom breadcrumbs
Remove controller files that were only being used for breadcrumbs
Remove unnecessary unit test files
Clean-up any tests

## How to Test

1. Go to routes and verify correct breadcrumbs are displayed.

## Checklist:


ZedLi:
I notice double breadcrumbs for orgs/projects
image

cameronperera (commented):
Please ignore if this was discussed already but I wanted to bring attention to how we are using icons in our breadcrumbs. We have an icon for the resource type but not for the individual resource name which leads to our breadcrumbs looking like the the attached image. In the HDS [docs](https://helios.hashicorp.design/components/breadcrumb#icon-placement) this is called out as something to avoid. 
Screenshot 2023-10-11 at 3 42 29 PM

lisbet-alvarez:
> Please ignore if this was discussed already but I wanted to bring attention to how we are using icons in our breadcrumbs. We have an icon for the resource type but not for the individual resource name which leads to our breadcrumbs looking like the the attached image. In the HDS [docs](https://helios.hashicorp.design/components/breadcrumb#icon-placement) this is called out as something to avoid. Screenshot 2023-10-11 at 3 42 29 PM



> Please ignore if this was discussed already but I wanted to bring attention to how we are using icons in our breadcrumbs. We have an icon for the resource type but not for the individual resource name which leads to our breadcrumbs looking like the the attached image. In the HDS [docs](https://helios.hashicorp.design/components/breadcrumb#icon-placement) this is called out as something to avoid. Screenshot 2023-10-11 at 3 42 29 PM

good catch, thank you! I removed all icon usage

cameronperera (approved):
LGTM. Thanks for taking on this large refactor!

ui/admin/tests/unit/controllers/scopes/scope/targets/target/add-injected-application-credential-sources-test.js
```diff
@@ -0,0 +1,57 @@
...
+import { module, test } from 'qunit';
+import { setupTest } from 'ember-qunit';
+
+module(
+  'Unit | Controller | scopes/scope/targets/target/add-injected-application-credential-sources',
+  function (hooks) {
+    setupTest(hooks);
+
+    test('it exists', function (assert) {
+      let controller = this.owner.lookup(
```
calcaide, line 15: None blocking comment: Can we use `const` here as we did within the test below?

ui/admin/tests/unit/controllers/scopes/scope/targets/target/add-brokered-credential-sources-test.js
```diff
@@ -0,0 +1,57 @@
...
+import { module, test } from 'qunit';
+import { setupTest } from 'ember-qunit';
+
+module(
+  'Unit | Controller | scopes/scope/targets/target/add-brokered-credential-sources',
+  function (hooks) {
+    setupTest(hooks);
+
+    test('it exists', function (assert) {
+      let controller = this.owner.lookup(
```
calcaide, line 15: None blocking comment: Can we use `const` here as we did in the tests below?

ui/admin/tests/unit/controllers/scopes/scope/targets/new-test.js
```diff
@@ -6,11 +6,11 @@
 import { module, test } from 'qunit';
 import { setupTest } from 'ember-qunit';
 
-module('Unit | Controller | scopes/scope/users/new', function (hooks) {
+module('Unit | Controller | scopes/scope/targets/new', function (hooks) {
   setupTest(hooks);
 
   test('it exists', function (assert) {
-    let controller = this.owner.lookup('controller:scopes/scope/users/new');
+    let controller = this.owner.lookup('controller:scopes/scope/targets/new');
```
calcaide, line 13: None blocking comment: Can we use `const` here as we did in the tests below?

ui/admin/app/templates/scopes/scope/workers.hbs
```diff
@@ -4,4 +4,9 @@
 }}
 
 {{page-title (t 'resources.worker.title_plural')}}
+<Breadcrumbs::Item
```
calcaide, line 7: None blocking comment: Is this a new way/pattern to define the breadcrumb? I like it 😄 if so, would you mind to present or document somehow (or provide a link if is already documented somewhere) the new way of working with this? (not asking this as part of the PR)

I see we need to define `<Breadcrumbs::Container>` and then the `<Breadcrumbs::Item>` ?

ui/admin/app/routes/scopes/scope/targets/target/enable-session-recording/index.js
```diff
@@ -14,15 +13,6 @@ export default class ScopesScopeTargetsTargetEnableSessionRecordingIndexRoute ex
 
   // =methods
 
-  /**
```
calcaide, line 17: N-I-C-E

calcaide (approved):
Amazing work @lisbet-alvarez !!! I left some none blocking comments. This new pattern, IMHO is cleaner and provides a better developer experience. Great job!!

ui/admin/tests/unit/controllers/scopes/scope/targets/target/add-injected-application-credential-sources-test.js
lisbet-alvarez: Good catch! Thank you!

ui/admin/app/components/breadcrumbs/container/index.js
```diff
@@ -0,0 +1,30 @@
...
+  @action
+  registerContainer(element) {
+    let { itemClass, linkClass } = this.args;
+
+    // A child `ol` is rendered in the Hds::Breadcrumb container and this is
+    // the element that the in-element helper should drop breadcrumbs into.
+    this.container = {
+      element: element.querySelector('ol'),
+      itemClass,
+      linkClass,
```
ZedLi, line 19: What would we use these classes for? It seems like this is for the package's breadcrumb implementation but we use our own HDS breadcrumbs?

ui/admin/app/components/breadcrumbs/item/index.hbs
```diff
@@ -0,0 +1,16 @@
+{{#each this.breadcrumbsService.containers as |container|}}
```
ZedLi, line 1: When we would have more than one breadcrumb container?

ui/admin/app/templates/scopes/scope/workers.hbs
lisbet-alvarez: This pattern is currently used in cloud-ui and uses this [addon](https://github.com/Bagaar/ember-breadcrumbs). I will add this to tech time topics (to give overview of new pattern). I can also add to docs (a README perhaps), any preference of where it should go?

ui/admin/app/templates/scopes/scope.hbs
```diff
@@ -6,13 +6,43 @@
...
+    @model={{@model.id}}
+  />
 {{/if}}
 
 {{#if @model.isProject}}
   {{page-title (t 'resources.org.title_plural')}}
   {{page-title this.scopes.selectedOrg.displayName}}
   {{page-title (t 'resources.project.title_plural')}}
   {{page-title @model.displayName}}
+  <Breadcrumbs::Item
```
ZedLi, line 26: Is this the only way to display these breadcrumbs? How did it work previously instead of us manually enumerating the scope breadcrumbs?

ui/admin/package.json
```diff
@@ -46,6 +46,7 @@
     "@babel/core": "^7.18.5",
     "@babel/eslint-parser": "^7.18.2",
     "@babel/plugin-proposal-decorators": "^7.18.6",
+    "@bagaar/ember-breadcrumbs": "^4.1.0",
```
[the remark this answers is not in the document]
lisbet-alvarez: That's a really good point, i'll refactor this PR to only use the service

ui/admin/app/components/breadcrumbs/container/index.js
lisbet-alvarez: good catch, ill remove this.

ui/admin/app/templates/scopes/scope.hbs
lisbet-alvarez: This is how it worked before: https://github.com/hashicorp/boundary-ui/blob/main/ui/admin/app/components/bread-crumbs.hbs  
https://github.com/hashicorp/boundary-ui/blob/main/ui/admin/app/controllers/scopes/scope/scopes.js

ui/admin/app/components/breadcrumbs/item/index.hbs
lisbet-alvarez: oop nope, ill refactor this as well, thank you!

ZedLi: Hmm I wasn't exactly saying this needs to be refactored but I was wondering when one would have more than one container, what's the use case since it looks like the service supports more than one container and multiple ones can be registered

lisbet-alvarez: The use case I can think of for having multiple containers is for rendering breadcrumbs with different attributes applied to diff containers. This won't apply to our use case tho. Since the container we are registering is always going to be the` <ol>` element of the `<Hds::Breadcrumb/>` component. So we will at most have 1 container registered at all times.

ZedLi: Yup or if for some reason you needed multiple sets of breadcrumbs on the same page. I don't think we'd ever have multiple containers and it seems like the implementation makes that assumption [here](https://github.com/hashicorp/boundary-ui/blob/6116b9aba8a05762deb682a1354cda3e09fb18d9/ui/admin/app/components/breadcrumbs/item/index.js#L17) anyways and would always grab the first and only container

ui/admin/app/templates/scopes/scope/workers.hbs
calcaide: @lisbet-alvarez I think mentioning it in techTime it's enough since you already provided [the documentation](https://github.com/Bagaar/ember-breadcrumbs). Amazing!

ui/admin/app/components/breadcrumbs/container/index.js
```diff
@@ -0,0 +1,32 @@
+import Component from '@glimmer/component';
+import { inject as service } from '@ember/service';
+import { action } from '@ember/object';
+
+/*
+  The `containers` property from the `breadcrumbsService` is tracked
+  so registering and unregistering a container is was causes the DOM
```
ZedLi, line 7:
```suggestion
  so registering or unregistering a container will cause the DOM
```

ZedLi (approved):
Looks great, thanks for this huge amount of work!

PR #1936, opened 2023-09-28
fix: 🐛 host modal renders whenever there is an error during connection

lisbet-alvarez:
:tickets: [Jira ticket](https://hashicorp.atlassian.net/browse/ICU-11231)

## Description
If the `connect` action were to cause an error then an error modal would render (when connecting from target list view). However, the choose hosts modal would also, incorrectly, render. 

Before: 
image
After:
image

## How to Test

1. Start a boundary dev instance.
2. Create a ssh target with a valid address but no hosts and no injected credentials.
3. Attempt to connect to the ssh target using the "connect" button in target list view (targets table).
4. Only an error modal should render. The choose hosts modal should not render.

:desktop_computer: [Desktop preview](https://boundary-ui-desktop-git-icu-11231-host-modal-r-a2c005-hashicorp.vercel.app/scopes/global/projects/targets)

## Checklist:

### DIFF
--- a/ui/desktop/app/routes/scopes/scope/projects/targets/target.js
+++ b/ui/desktop/app/routes/scopes/scope/projects/targets/target.js
@@ -7,7 +7,6 @@ import Route from '@ember/routing/route';
 import { inject as service } from '@ember/service';
 import { action } from '@ember/object';
 import { loading } from 'ember-loading';
-import { tracked } from '@glimmer/tracking';
 
 export default class ScopesScopeProjectsTargetsTargetRoute extends Route {
   // =services
@@ -20,7 +19,7 @@ export default class ScopesScopeProjectsTargetsTargetRoute extends Route {
 
   // =attributes
 
-  @tracked connectionError = false;
+  isConnectionError = false;
 
   // =methods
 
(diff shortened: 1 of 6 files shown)


ui/desktop/app/routes/scopes/scope/projects/targets/target.js
```diff
@@ -17,6 +18,10 @@ export default class ScopesScopeProjectsTargetsTargetRoute extends Route {
   @service router;
   @service session;
 
+  // =attributes
+
+  @tracked connectionError = false;
```
ZedLi, line 23: Do we need this to be tracked?

Should we also call it `isConnectionError` since it's a boolean?

ui/desktop/app/routes/scopes/scope/projects/targets/target.js
```diff
@@ -65,6 +70,22 @@ export default class ScopesScopeProjectsTargetsTargetRoute extends Route {
     }
   }
 
+  /**
+   * Sets the 'isConnecting' queryParam to false if connection failed.
+   * @returns {boolean}
+   */
+  @action
+  didTransition() {
+    if (this.connectionError) {
```
ZedLi, line 79: Should we set the connection error back to false here after we change the query param value? Otherwise it will always be true since it never gets reset

ZedLi (approved):
Thanks for fixing this!

dvonanderson (approved):

PR #1939, opened 2023-10-04
fix: 🐛 Fix edge case with showing host modal

ZedLi:
:tickets: [Jira ticket](https://hashicorp.atlassian.net/browse/ICU-11285)

## Description
If a user clicks retry when an error occurs during connection, we never reset our error connection flag which will prevent any further host modals from appearing or 

## How to Test

Setup two targets with hosts where one will error out when connecting. Hit retry and try connecting to the other target and the host modal should still appear.

## Checklist:


lisbet-alvarez (approved):
Thanks for fixing this! Works locally for me as well!

cameronperera (approved):

DhariniJeeva (approved):

PR #1947, opened 2023-10-10
Upgrade addons/auth to Ember 4.12

dvonanderson:
:tickets: [Jira ticket](https://hashicorp.atlassian.net/browse/ICU-11332)

## Description

Ember `addon/auth` upgrade v4.4.1...v4.12.2

## Checklist:


addons/auth/.eslintrc.js
ZedLi, line 1: I think we should take the updates to the babel options, e.g. the `ecmaVersion` and plugin change

addons/auth/.prettierrc.js
```diff
@@ -4,6 +4,12 @@
  */
 
 module.exports = {
-  singleQuote: true,
-  printWidth: 80,
+  overrides: [
```
ZedLi, line 7: This probably doesn't need to get changed, we might have committed this previously too

addons/auth/package.json
ZedLi, line 1: I think there are changes to use `concurrently` for some of our scripts. We should probably switch to it since `npm-run-all` hasn't been updated for 5 years.

dvonanderson: @ZedLi ok sounds good, I will make changes and test

addons/auth/.eslintrc.js
dvonanderson: @ZedLi ?
```
ecmaVersion: 'latest',
babelOptions: {
      plugins: [
        ['@babel/plugin-proposal-decorators', { decoratorsBeforeExport: true }],
      ],
```

cameronperera (approved):

ZedLi (approved):
Looks good! I think just can just be merged into main now

PR #1948, opened 2023-10-10
fix: 🐛 Fix electron not inheriting path variable in production

ZedLi:
## Description
It turns out electron does not properly inherit the path variable from process.env.PATH in production. This presents a problem as users won't have access to many of their binaries that they normally would have on their path.

If you're wondering why I didn't use the `fix-path` package directly, it's because they switched over to ESM and we can't directly use it. Their previous version doesn't fix linux so I had to manually use it.

Before:
image

After:
image

## How to Test
Create a production build locally and confirm your path gets loaded correctly when opening a shell in boundary.

### DIFF
--- a/ui/desktop/electron-app/package.json
+++ b/ui/desktop/electron-app/package.json
@@ -28,6 +28,7 @@
     "node-html-parser": "^5.3.3",
     "node-pty": "^1.0.0",
     "semver": "^7.5.3",
+    "shell-env": "^3.0.1",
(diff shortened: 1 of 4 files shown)


cameronperera (approved):
Nice fix!

DhariniJeeva (approved):
looks good! Thanks for the fix!

lisbet-alvarez (approved):
THank you!!

PR #1949, opened 2023-10-11
Refactor key value list component

DhariniJeeva:
:tickets: [Jira ticket](https://hashicorp.atlassian.net/browse/ICU-11370)

## Description

This PR allows a single input field with an action button in the key value list component.

## Checklist:


ui/admin/app/components/form/field/key-value-list/index.hbs
```diff
@@ -73,15 +78,17 @@
...
-                @type='text'
-                disabled={{@disabled}}
-                aria-labelledby={{valueHeaderID}}
-                {{on 'input' (set-from-event this 'newOptionValue')}}
-              />
-            </B.Td>
+            {{#unless @hasNoValue}}
+              <B.Td>
+                <Hds::Form::TextInput::Field
+                  @value={{this.newOptionValue}}
```
ZedLi, line 84: Do we need to handle the backing js file for not having a value now?

DhariniJeeva: I didn't change the js file because the `value` can be empty

DhariniJeeva: Screenshot 2023-10-11 at 4 16 31 PM

But open to changing this if this might cause confusion

calcaide (approved):
LGTM!! Thanks for the work!

ui/admin/app/components/form/field/key-value-list/index.hbs
ZedLi: I personally think it's an easy to change to not send an empty string so would probably do that

ZedLi (approved):
Looks good!

ui/admin/app/components/form/field/key-value-list/index.hbs
DhariniJeeva: I was looking at an incorrect example with a `value` input field in the action params. If we don't send a `value` field to the `@addOption` action, we won't have the entry for value. So there is no need to change anything in the backing js file. 
Screenshot 2023-10-12 at 2 19 11 PM

ZedLi: Right my point was we're sending a value field to the callback with just an empty string

DhariniJeeva: good point, I will clean this up in the following PR

ZedLi: I think it's fine if it's been merged in, it's not that big of a deal. It's just an extra field that won't get used

PR #1956, opened 2023-10-18
fix: 🐛 Move host fetch to session domain

calcaide:
## Description

Move host fetch from connect method within target domain to session domain.

### DIFF
--- a/ui/desktop/app/routes/scopes/scope/projects/targets/target.js
+++ b/ui/desktop/app/routes/scopes/scope/projects/targets/target.js
@@ -154,17 +154,14 @@ export default class ScopesScopeProjectsTargetsTargetRoute extends Route {
       // we don't have to manually persist the proxy details.
       session.proxy_address = address;
       session.proxy_port = port;
+      // Add retrieved credentials to the session
       if (credentials) {
         credentials.forEach((cred) => session.addCredential(cred));
-      }
-      if (session.host_id) {
-        const host = await this.store.findRecord('host', session.host_id);
-        session.addHost(host);
       }
 
       await this.router.transitionTo(
         'scopes.scope.projects.sessions.session',
-        session
+        session_id
       );
     } catch (e) {
       this.isConnectionError = true;
--- a/ui/desktop/app/routes/scopes/scope/projects/sessions/session.js
+++ b/ui/desktop/app/routes/scopes/scope/projects/sessions/session.js
@@ -19,9 +19,21 @@ export default class ScopesScopeProjectsSessionsSessionRoute extends Route {
    * @param {string} params.session_id
    * @return {SessionModel}
    */
-  model({ session_id }) {
-    return this.store.findRecord('session', session_id, {
+  async model({ session_id }) {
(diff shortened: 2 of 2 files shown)


cameronperera (approved):
LGTM, thanks for the fix

ui/desktop/app/routes/scopes/scope/projects/targets/target.js
```diff
@@ -154,17 +154,14 @@ export default class ScopesScopeProjectsTargetsTargetRoute extends Route {
       // we don't have to manually persist the proxy details.
       session.proxy_address = address;
       session.proxy_port = port;
+      // Add retrieved credentials to the session
       if (credentials) {
         credentials.forEach((cred) => session.addCredential(cred));
       }
-      if (session.host_id) {
```
ZedLi, line 160: I don't think the issue was here, it was most likely this [fetch](https://github.com/hashicorp/boundary-ui/blob/262df46fac2282e4e2409a6aeff0ead38a205350/ui/desktop/app/routes/scopes/scope/projects/targets/target.js#L42) since we didn't use to fetch the host sets before. (Plus if you notice, the user had to add `host-set` read permissions, not `host`)

If it errored out in the actual connect, we would be getting the error popup instead since it would have been caught in the try-catch. We should verify this by modifying our own grants in boundary.

I think the main issue here is we technically have a regression here. Before the changes, we didn't require the user to retrieve any host sets before being able to connect since they could access the "quick connection" function we're currently doing without having to specifically select a host. Now, we're trying to retrieve the hosts sets just to be able to visit a target. I think we need to decide how to handle the possibility of a user not having permissions now.

ZedLi (changes requested):

ui/desktop/app/routes/scopes/scope/projects/targets/target.js
calcaide: @ZedLi 

Gotcha. I updated the slack DM conversation, please take a look at that.

Knowing this is not the solution, but I think could be an improvement? IMHO in session domain is where the host information is need, so to me, makes more sense to fetch it there. Perhaps I can improve this wrapping it on a try/catch, so if the user does not have permissions, we will inform. 

Let me know your thoughts, happy to close this PR if you think otherwise.

ZedLi: I missed this comment but yes we also need to handle the extra host request which is permissioned as well. 
I agree this makes sense in the session model but we also shouldn't error out when trying to retrieve the info as well

calcaide: @ZedLi 
Ok, then this PR, as per my understanding is already covering that and is NOT erroring out when trying to retrieve the info. If you are ok with it, please approve and I will merge, otherwise let me know what you think we are missing.
Thank you!

calcaide:
Closing this PR since this issue has been fixed [here](https://github.com/hashicorp/boundary-ui/pull/1968)

PR #1957, opened 2023-10-18
chore: 🤖 refactor key value list component, text input support

DhariniJeeva:
[jira](https://hashicorp.atlassian.net/browse/ICU-11435), [jira](https://hashicorp.atlassian.net/browse/ICU-11271) 
## Description

This PR refactors the key value list component to add support for single text input and replaces the rose key value list component used in the host set form.

**Azure:**
Before:
Screenshot 2023-10-17 at 8 57 22 PM

After
 
Screenshot 2023-10-26 at 6 13 53 PM

 
 

**AWS:**
Before:
Screenshot 2023-10-17 at 9 03 31 PM

After:
 
Screenshot 2023-10-26 at 6 13 06 PM

 
## How to Test

## Checklist:


addons/api/mirage/config.js
```diff
@@ -431,9 +431,16 @@ function routes() {
       return hostSets.where({ hostCatalogId });
     }
   );
+
   this.post('/host-sets', function ({ hostCatalogs, hostSets }) {
     const attrs = this.normalizedRequestAttrs();
     const hostCatalog = hostCatalogs.find(attrs.hostCatalogId);
+    if (hostCatalog.plugin) {
```
DhariniJeeva, line 438: currently there's an issue when you update a newly created plugin type host set due to not attaching `plugin` info when we create a plugin host set.

addons/api/mirage/factories/host-set.js
```diff
@@ -27,11 +27,10 @@ export default factory.extend({
   },
   preferred_endpoints() {
     if (this.type === 'plugin') {
-      const enpointsAmount = faker.number.int({ min: 1, max: 5 });
       let result = [];
-      for (let i = 0; i < enpointsAmount; ++i) {
-        result.push(faker.internet.ip());
-      }
+
+      result.push(faker.internet.ip());
```
DhariniJeeva, line 32: I noticed a weird behavior when testing this, [the](https://github.com/hashicorp/boundary-ui/pull/1957/files#diff-4cbf41fe5ec71ecf02bccd70b53f754ffef7911f0c6f228b139ff350010abb08R176-R179) `promise.all` doesn't wait for all the buttons to be clicked so I had to use a single entry for this array here. But this only happens in the tests and is not an issue when I manually test the code.

ui/admin/app/components/form/field/list-wrapper/index.hbs
```diff
@@ -0,0 +1,180 @@
...
+                    {{on 'input' (set-from-event option 'value')}}
+                  />
+                </B.Td>
+
+                <B.Td>
+                  {{#if @removeOptionByIndex}}
+                    <Hds::Button
+                      @text={{t 'actions.remove'}}
+                      @color='critical'
+                      data-test='remove-button'
```
lisbet-alvarez, line 138:
```suggestion
                      data-test-remove-button
```

ui/admin/app/components/form/field/list-wrapper/index.hbs
```diff
@@ -0,0 +1,180 @@
...
+                  disabled={{@disabled}}
+                  aria-labelledby={{keyHeaderID}}
+                  {{on 'input' (set-from-event this 'newRow')}}
+                />
+              </B.Td>
+
+              <B.Td>
+                <Hds::Button
+                  @text={{t 'actions.add'}}
+                  data-test='add-button'
```
lisbet-alvarez, line 163:
```suggestion
                  data-test-add-button
```

lisbet-alvarez (approved):
🙌

lisbet-alvarez (approved):
🙌

ui/admin/tests/acceptance/host-catalogs/host-sets/update-test.js
```diff
@@ -150,49 +172,51 @@ module('Acceptance | host-catalogs | host sets | update', function (hooks) {
 
     const name = 'aws host set';
     await fillIn('[name="name"]', name);
-    // Remove all the preferred endpoints
+
     await Promise.all(
-      findAll('form fieldset:nth-of-type(1) [title="Remove"]').map((element) =>
-        click(element)
+      findAll('[name="preferred_endpoints"] [data-test="remove-button"]').map(
```
DhariniJeeva, line 177:
```suggestion
      findAll('[name="preferred_endpoints"] [data-test-remove-button]').map(
```

ui/admin/tests/acceptance/host-catalogs/host-sets/update-test.js
```diff
@@ -150,49 +172,51 @@ module('Acceptance | host-catalogs | host sets | update', function (hooks) {
...
+        (element) => click(element)
       )
     );
-    await fillIn('[name="preferred_endpoints"]', 'endpoint');
-    await click('form fieldset:nth-of-type(1) [title="Add"]');
+    assert.strictEqual(
+      findAll('[name="preferred_endpoints"] [data-test="remove-button"]')
```
DhariniJeeva, line 182:
```suggestion
      findAll('[name="preferred_endpoints"] [data-test-remove-button]')
```

ui/admin/tests/acceptance/host-catalogs/host-sets/update-test.js
```diff
@@ -150,49 +172,51 @@ module('Acceptance | host-catalogs | host sets | update', function (hooks) {
...
-      findAll('form fieldset:nth-of-type(2) [title="Remove"]').map((element) =>
+      findAll('[name="filters"] [data-test="remove-button"]').map((element) =>
         click(element)
       )
     );
-    await fillIn('[name="filters"]', 'filter');
-    await click('form fieldset:nth-of-type(2) [title="Add"]');
+
+    assert.strictEqual(
+      findAll('[name="filters"] [data-test="remove-button"]').length,
```
DhariniJeeva, line 198:
```suggestion
      findAll('[name="filters"] [data-test-remove-button]').length,
```

ui/admin/tests/acceptance/host-catalogs/host-sets/update-test.js
```diff
@@ -150,49 +172,51 @@ module('Acceptance | host-catalogs | host sets | update', function (hooks) {
...
+      0
+    );
+
+    await fillIn('[name="preferred_endpoints"] input', 'sample endpoint');
+    await click('[name="preferred_endpoints"] button');
+
     // Remove all the filters
     await Promise.all(
-      findAll('form fieldset:nth-of-type(2) [title="Remove"]').map((element) =>
+      findAll('[name="filters"] [data-test="remove-button"]').map((element) =>
```
DhariniJeeva, line 192:
```suggestion
      findAll('[name="filters"] [data-test-remove-button]').map((element) =>
```

ui/admin/tests/acceptance/host-catalogs/host-sets/update-test.js
```diff
@@ -203,20 +227,26 @@ module('Acceptance | host-catalogs | host sets | update', function (hooks) {
     await fillIn('[name="name"]', name);
     // Remove all the preferred endpoints
     await Promise.all(
-      findAll('form fieldset:nth-of-type(1) [title="Remove"]').map((element) =>
-        click(element)
+      findAll('[name="preferred_endpoints"] [data-test="remove-button"]').map(
```
DhariniJeeva, line 230:
```suggestion
      findAll('[name="preferred_endpoints"] [data-test-remove-button]').map(
```

ui/admin/tests/acceptance/host-catalogs/host-sets/update-test.js
```diff
@@ -203,20 +227,26 @@ module('Acceptance | host-catalogs | host sets | update', function (hooks) {
...
-      findAll('form fieldset:nth-of-type(1) [title="Remove"]').map((element) =>
-        click(element)
+      findAll('[name="preferred_endpoints"] [data-test-remove-button]').map(
+        (element) => click(element)
       )
     );
-    await fillIn('[name="preferred_endpoints"]', 'endpoint');
-    await click('form fieldset:nth-of-type(1) [title="Add"]');
+    assert.strictEqual(
+      findAll('[name="preferred_endpoints"] [data-test="remove-button"]')
```
DhariniJeeva, line 235:
```suggestion
      findAll('[name="preferred_endpoints"] [data-test-remove-button]')
```

addons/api/mirage/config.js
calcaide: None blocking comment: Can we add a comment to explain why we are doing this? It will help maintenance. I.e: Adding attributes to avoid error at updating time

ui/admin/app/components/form/field/list-wrapper/index.hbs
```diff
@@ -0,0 +1,180 @@
...
+<Hds::Form::Fieldset
+  @layout={{@layout}}
+  @isRequired={{@isRequired}}
+  @isOptional={{@isOptional}}
+  as |F|
+>
+
+  {{yield F}}
+
+  {{#let (unique-id) (unique-id) as |keyHeaderID valueHeaderID|}}
```
calcaide, line 15: Asking for learning purposes: when we do, (unique-id) (unique-id) as |keyHeaderID valueHeaderID|, it's because we are generating two new bindings (keyHeader & valueHeader) with a unique ID values, is that right?

ui/admin/app/components/form/field/list-wrapper/index.hbs
```diff
@@ -0,0 +1,180 @@
...
+    <F.Control>
+      {{!if key/value pair}}
```
calcaide, line 17: Not blocking comment: do we want this comment here?

ui/admin/app/components/form/field/list-wrapper/index.js
```diff
@@ -16,6 +16,8 @@ export default class MappingListComponent extends Component {
    */
   @tracked newOptionKey = '';
 
+  @tracked newRow = '';
```
calcaide, line 19: None blocking comment: I do understand newRow as a temporary variable to persist the values of the new row before "adding" the row (addRow) is that correct?

ui/admin/app/components/form/host-set/azure/index.hbs
```diff
@@ -69,96 +69,34 @@
...
-    @link={{doc-url 'host-set.preferred-endpoints'}}
+  <Form::Field::ListWrapper
+    @layout='horizontal'
+    @isOptional={{true}}
+    @name='preferred_endpoints'
+    @options={{@model.preferred_endpoints}}
+    @addRow={{fn @addRow 'preferred_endpoints'}}
+    @removeOptionByIndex={{fn @removeOptionByIndex 'preferred_endpoints'}}
+    @disabled={{form.disabled}}
+    @hasSingleTextInputField={{true}}
```
calcaide, line 80: None blocking question: What is the purpose of declaring hasSingleTextInputField? it is for having just 1 input file to add options to the list?

ui/admin/tests/integration/components/form/field/list-wrapper-test.js
```diff
@@ -31,33 +31,33 @@ module('Integration | Component | key-value-list', function (hooks) {
...
   });
 
-  test('it renders multiple options', async function (assert) {
+  test('it renders multiple options with key value pair', async function (assert) {
     this.options = [
       { key: 'one', value: 'two' },
       { key: 'three', value: 'four' },
     ];
 
     // Template block usage:
```
calcaide, line 40: None blocking comment: can we delete this comment?

ui/admin/tests/integration/components/form/field/list-wrapper-test.js
```diff
@@ -31,33 +31,33 @@ module('Integration | Component | key-value-list', function (hooks) {
...
 
     assert.dom('tbody tr').exists({ count: 3 });
   });
 
-  test('it renders multiple options with only keys', async function (assert) {
-    this.options = [{ key: 'one' }, { key: 'three' }];
+  test('it renders multiple options with a single text field', async function (assert) {
+    this.options = [{ value: 'one' }, { value: 'three' }];
 
     // Template block usage:
```
calcaide, line 54: None blocking comment: can we delete this comment?

addons/api/mirage/factories/host-set.js
calcaide: Nice catch! I understand there is a testing issue, if I have time today, I will try to understand it better and check if there is a solution to it.

I am a bit concern, if we want to test i.e removing all filters, having just 1 item in the list might not cover all the usage of the component.

calcaide (approved):
LGTM!!! Nice work @DhariniJeeva 🙌

addons/api/mirage/factories/host-set.js
ZedLi: We shouldn't change the mirage data just to make a test pass, we should figure out a way to fix the test

ui/admin/app/components/form/field/list-wrapper/index.hbs
ZedLi, line 1: I think this component should be redesigned. I think if we're calling it a `list-wrapper` now the API surface should change and we shouldn't be adding all these if statements for different possibilities and calling things like `@hasKeyValuePair` or `hasSingleTextInputField`, especially as you have TODOs for other possibilities at the bottom of this file. 

I don't think this approach is tenable and when I had made this component initially, it was specifically _just_ for key values so I had made some assumptions about the usage of this component. 

Let me know if you have any thoughts on this.

ZedLi (changes requested):

addons/api/mirage/config.js
DhariniJeeva: sure I can add a comment, but basically the creation of a new dynamic plugin in mirage doesn't work as expected. In mirage, we aren't attaching the plugin information similar to actual APIs

ui/admin/app/components/form/field/list-wrapper/index.hbs
DhariniJeeva: yup that is correct

DhariniJeeva:
Making changes to the design of this component to organize it better, moving this to draft for now.

addons/api/mirage/factories/host-set.js
ZedLi: I took a look at this but I'm surprised this worked before or that we didn't have an issue previously. The `Promise.all` works fine but the underlying issue is all the promises are run at the same time (aka all the buttons are clicked at once) which causes a race condition because the remove function uses the index of the array to remove them so it might not remove the right one.

Using a `for...of` loop to go over the promises sequentially fixes the issue

ui/admin/app/components/form/host-set/index.hbs
```diff
@@ -9,6 +9,6 @@
   edit=@edit
   submit=@submit
   cancel=@cancel
-  removeItemByIndex=@removeItemByIndex
-  addStringItem=@addStringItem
```
ZedLi, line 13: Did these actions need to be removed as well?

addons/api/mirage/factories/host-set.js
DhariniJeeva: thanks for looking into it, that makes sense. I will try your suggestion.

ui/admin/app/components/form/field/list-wrapper/index.hbs
DhariniJeeva: Thanks for the input! As we discussed, I updated the component to yield different components based on the UI requirement. This reduces code repetition in consumer forms and makes it more readable and organized. 

One downside is adding different input types such as select, and textarea based on UI requirements, but I think that is unavoidable given sometimes fields have different transforms which makes it harder to generalize and dynamically rendering the consumer code would lead to code repetition and make the wrapper pointless. I think this is a good balance. 

This is still a win and would get rid of a lot of repetition in the codebase.

ui/admin/app/components/form/field/list-wrapper/index.js
DhariniJeeva: This isn't needed anymore, but you are right. That was my original intention.

ui/admin/app/components/form/field/list-wrapper/index.hbs
ZedLi: I was thinking the wrapper would mainly include the `Fieldset` component since that's the same between all of them. And we would either yield the content (or `yield to` the header and body).

It seems like currently we don't even call the wrapper component?

ui/admin/app/components/form/field/list-wrapper/text-input/index.hbs
```diff
@@ -0,0 +1,65 @@
...
+  @isOptional={{@isOptional}}
+  name={{@name}}
+  as |F|
+>
+
+  {{yield F}}
+
+  <F.Control>
+
+    {{#each @options as |option index|}}
```
ZedLi, line 18: Should we have been consistent and used the table or changed it for both of them? I think originally we made it a table for semantic meaning, i.e. we're dealing with tabular data. 

I could be convinced it's either a table or unordered list but I don't think we should just have divs.

ui/admin/app/components/form/field/list-wrapper/text-input/index.hbs
```diff
@@ -0,0 +1,65 @@
...
+
+      <Hds::SegmentedGroup class='list-wrapper-input' as |S|>
```
ZedLi, line 20: Should we be using segmented group? The looks are now inconsistent, plus the add button and delete button don't line up

ui/admin/app/components/form/field/list-wrapper/text-input/index.hbs
```diff
@@ -0,0 +1,65 @@
...
+        <S.TextInput
+          @value={{option.value}}
+          @type='text'
+          disabled={{@disabled}}
+          aria-describedby={{F.ariaDescribedBy}}
```
ZedLi, line 25: Shouldn't these be aria-labeledby?

ui/admin/app/components/form/field/list-wrapper/key-value/index.hbs
```diff
@@ -14,13 +14,11 @@
 
   {{#let (unique-id) (unique-id) as |keyHeaderID valueHeaderID|}}
     <F.Control>
-      <Hds::Table class='key-value-list' name={{@name}}>
+      <Hds::Table class='list-wrapper' name={{@name}}>
```
ZedLi, line 17: Should this class still be `list-wrapper` here?

ui/admin/app/components/form/field/list-wrapper/text-input/index.hbs
DhariniJeeva: I think we can fix the style for the existing component if we can use segmented group. I reached out to Joshua to see if we can use this component, technically it doesn't have the label and satisfies the requirements.

ui/admin/app/components/form/field/list-wrapper/index.hbs
ZedLi: I was thinking of having the wrapper just handle the fieldset portions of the components, what do you think about this? We'd have to slightly change how we call the component but it's pretty similar
```
<Hds::Form::Fieldset
  @layout={{@layout}}
  @isRequired={{@isRequired}}
  @isOptional={{@isOptional}}
  as |F|
>
  {{yield F to='fieldset'}}
  <F.Control>
    {{yield
      (hash
        KeyValue=(component
          'form/field/list-wrapper/key-value' disabled=@disabled
        )
        TextInput=(component
          'form/field/list-wrapper/text-input' disabled=@disabled
        )
      )
      to='field'
    }}
  </F.Control>
</Hds::Form::Fieldset>
```

DhariniJeeva: I was thinking the same too, we can even render the <F.error> and the legends here instead of doing those in the consumer.

ui/admin/app/components/form/field/list-wrapper/text-input/index.hbs
DhariniJeeva: turns out we can't use the segmented group for this scenario, link [here](https://hashicorp.slack.com/archives/C7KTUHNUS/p1698339724997629). For now, I'm following the existing tabular form to keep things simple

ui/admin/app/components/form/field/list-wrapper/key-value/index.hbs
DhariniJeeva: updated!

ui/admin/app/components/form/field/list-wrapper/index.hbs
DhariniJeeva: updated

ui/admin/app/components/form/field/list-wrapper/index.js
ZedLi, line 1: I think this file doesn't need to exist anymore?

ui/admin/app/templates/scopes/scope/host-catalogs/host-catalog/host-sets/new.hbs
```diff
@@ -23,8 +23,6 @@
       @model={{@model}}
       @submit={{route-action 'save' @model}}
       @cancel={{route-action 'cancel' @model}}
-      @removeItemByIndex={{route-action 'removeItemByIndex' @model}}
-      @addStringItem={{route-action 'addStringItem' @model}}
```
ZedLi, line 27: Were these actions removed?

ui/admin/app/components/form/field/list-wrapper/index.hbs
```diff
@@ -0,0 +1,32 @@
...
+}}
+
+<Hds::Form::Fieldset
+  @layout={{@layout}}
+  @isRequired={{@isRequired}}
+  @isOptional={{@isOptional}}
+  as |F|
+>
+  <F.Legend>{{@label}}</F.Legend>
+  <F.HelperText>{{@helper}}</F.HelperText>
```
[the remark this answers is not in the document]
DhariniJeeva: updated it to be consistent with our patter

ui/admin/app/components/form/credential-library/vault-ssh-certificate/index.hbs
```diff
@@ -236,55 +236,64 @@
...
-    </M.HelperText>
-    {{#if @model.errors.critical_options}}
-      <M.Error as |E|>
-        {{#each @model.errors.critical_options as |error|}}
-          <E.Message>{{error.message}}</E.Message>
-        {{/each}}
-      </M.Error>
-    {{/if}}
-  </Form::Field::KeyValueList>
+  <Form::Field::ListWrapper @layout='horizontal' @isOptional={{true}}>
```
ZedLi, line 239: I think we still need to pass `@disabled={{form.disabled}}` through the wrapper

ZedLi (approved):
Looks great, just had a small comment with using the `@disabled`.

PR #1960, opened 2023-10-19
refactor: 💡 update how connection errors are displayed

cameronperera:
✅ Closes: ICU-10962

:tickets: [Jira ticket](https://hashicorp.atlassian.net/browse/ICU-10962)

Before:
Screenshot 2023-10-19 at 5 19 33 PM

After:
Screenshot 2023-10-19 at 5 19 02 PM

## How to Test

Create an SSH target with no injected credentials will cause a failure when trying to connect.

## Checklist:


ui/desktop/app/routes/scopes/scope/projects/targets/target.js
```diff
@@ -167,16 +186,16 @@ export default class ScopesScopeProjectsTargetsTargetRoute extends Route {
         session
       );
     } catch (e) {
+      const targetErrors = model.get('errors');
+      targetErrors.add('connect', e.message);
+
+      /* eslint-disable-next-line ember/no-controller-access-in-routes */
+      const controller = this.controllerFor(
```
calcaide, line 193: @cameronperera What is the purpose of setting `isConnection` to false within the controller?

I don't think we should disable this eslint rule, perhaps we can look for an alternative? Happy to help

ui/desktop/app/routes/scopes/scope/projects/targets/target.js
```diff
@@ -86,6 +85,26 @@ export default class ScopesScopeProjectsTargetsTargetRoute extends Route {
...
   }
 
+  /**
+   * Removes any errors from target model when transitioning away
+   * from this route and sets isConnectionError to false if true.
+   * @returns {boolean}
+   */
+  @action
+  willTransition() {
+    /* eslint-disable-next-line ember/no-controller-access-in-routes */
```
calcaide, line 95: As mentioned below, we shouldn't avoid this eslint rule, perhaps looks for an alternative way. What we are trying to achieve here? Happy to help to look for an alternative.

calcaide (changes requested):
Looking good! I left a couple of comments related to accessing the controller within the route, that's a bad pattern and we should avoid it, [as per this info](https://github.com/ember-cli/eslint-plugin-ember/blob/master/docs/rules/no-controller-access-in-routes.md). 
Happy to help finding an alternative to this, let's chat on slack about it?

ui/desktop/app/routes/scopes/scope/projects/targets/target.js
```diff
@@ -167,16 +186,16 @@ export default class ScopesScopeProjectsTargetsTargetRoute extends Route {
         session
       );
     } catch (e) {
+      const targetErrors = model.get('errors');
+      targetErrors.add('connect', e.message);
```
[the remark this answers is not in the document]
cameronperera: How would we go about making the connect error it's own thing? We don't have a connect model to assign the error to.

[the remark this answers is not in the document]
ZedLi: Why does it need to be on a model?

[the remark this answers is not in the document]
cameronperera: That is a good question. I am not sure how you might go about that without moving this work into a component (which I believe we discussed doing at one point).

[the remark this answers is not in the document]
ZedLi: It's no different than just putting it on the controller like how how we've already done for other things isn't it?

ui/desktop/app/controllers/scopes/scope/projects/targets/target.js
```diff
@@ -13,6 +13,7 @@ export default class ScopesScopeProjectsTargetsTargetController extends Controll
   queryParams = [{ isConnecting: { type: 'boolean' } }];
 
   @tracked isConnecting = false;
+  @tracked connectError;
```
ZedLi, line 16: If we are now using a `connectError`, do we still need `isConnectionError` in the route?

ui/desktop/app/routes/scopes/scope/projects/targets/target.js
```diff
@@ -86,6 +85,24 @@ export default class ScopesScopeProjectsTargetsTargetRoute extends Route {
     return true;
   }
 
+  /**
+   * Removes any errors from target model when transitioning away
+   * from this route and sets isConnectionError to false if true.
+   * @returns {boolean}
+   */
+  @action
+  willTransition() {
```
ZedLi, line 94: Is the proper event `willTransition` or `resetController`? It seems like it's more of the latter?

addons/core/translations/resources/en-us.yaml
```diff
@@ -554,6 +554,9 @@ target:
     cannot-connect:
```
ZedLi, line 554: I feel like the name of this translation should change and be more specific since we have one that's now `could-not-connect`

ui/desktop/app/routes/scopes/scope/projects/targets/target.js
calcaide: I think `resetController` makes more sense [as per documentation](https://guides.emberjs.com/v4.12.0/routing/query-params/), but do we also set `this.isConnectionError` with the transition, as per [previous comment](https://github.com/hashicorp/boundary-ui/pull/1960/files#r1376410464) we might not need it. Then it makes even more sense the usage of `resetController`

ui/desktop/app/routes/scopes/scope/projects/targets/target.js
```diff
@@ -167,16 +184,14 @@ export default class ScopesScopeProjectsTargetsTargetRoute extends Route {
         session
       );
     } catch (e) {
+      /* eslint-disable-next-line ember/no-controller-access-in-routes */
```
calcaide, line 187: As per [previous comment](https://github.com/hashicorp/boundary-ui/pull/1960/files#r1378174976) if we end up using `resetController`, what do you guys think of here not touching the controller but calling `resetController`?

Within `resetController`, [as per documentation](https://guides.emberjs.com/v4.12.0/routing/query-params/), it receives the param `isExiting`, so within `resetController`, we can deal with all the logic knowing if we are exiting or not, which I think, covers the usages of reseting controller we are doing within this route.

Thoughts? @ZedLi @cameronperera

ui/desktop/app/controllers/scopes/scope/projects/targets/target.js
calcaide: None blocking comment: Indeed with @ZedLi comment, in case we do not need it, I am fine getting that change on this PR or on a separate as tech debt (for pragmatism sake and promote staging ASAP).

calcaide (approved):
Left some comments. Good job on this!!

cameronperera:
Closing as connection logic has changed and this is no longer a valid solution.

PR #1965, opened 2023-10-23
chore: 🤖 Upgrade `addons/core` to Ember 4.12.2

dvonanderson:
✅ Closes: ICU-11341

:tickets: [Jira ticket](https://hashicorp.atlassian.net/browse/ICU-11341)

## Description
- Upgrade `addons/core` to Ember 4.12.2
- Fixed `template-lint-disable no-passed-in-event-handlers` rule

## Checklist:


addons/core/package.json
```diff
@@ -45,44 +45,47 @@
...
     "ember-page-title": "^7.0.0",
-    "ember-qunit": "^6.0.0",
-    "ember-resolver": "^8.0.3",
-    "ember-source": "~4.4.0",
-    "ember-template-lint": "^4.14.0",
+    "ember-qunit": "^6.2.0",
+    "ember-resolver": "^10.0.0",
+    "ember-source": "~4.12.0",
+    "ember-source-channel-url": "^3.0.0",
+    "ember-template-lint": "^4.18.2",
```
dvonanderson, line 72: @calcaide after some testing `ember-template-lint: "^4.18.2"` and `prettier: "^2.7.1"` are our allowed upgrades for our template linting.  Pushing to > `ember-template-lint: "5.0.0"` has some changes that I will add to tech time to discuss.

calcaide: Thanks for the heads up!!

addons/core/addon/components/form/authenticate/index.hbs
```diff
@@ -3,12 +3,10 @@
   SPDX-License-Identifier: MPL-2.0
 }}
 
-{{! //TODO fix the issue and delete disabling the rule }}
-{{! template-lint-disable no-passed-in-event-handlers }}
 {{#if @model.type}}
   {{component
     (concat 'form/authenticate/' @model.type)
-    submit=@submit
+    onSubmit=@submit
```
calcaide, line 9: None blocking comment: We use `onSubmit` without the `@` i.e: `@onSubmit` because is a component declaration?

calcaide (approved):
LGTM!! Awesome work upgrading to Ember 4.12.x!!! 🙌

addons/core/addon/components/form/authenticate/index.hbs
dvonanderson: @calcaide yes that is correct

addons/api/package.json
```diff
@@ -28,11 +28,11 @@
     "ember-cli-htmlbars": "^6.2.0",
     "ember-cli-mirage": "^2.4.0",
     "ember-data": "~4.12.0",
+    "@babel/core": "^7.21.0",
```
ZedLi, line 31: Why was this added to dependencies from dev?

addons/core/package.json
ZedLi: Can you give a quick explanation what the issues are?

dvonanderson: @ZedLi I don't think its really an issue its just an idea of how we use quotes and also different indentation in our test files.  I am still trying to compare the releases and find out what changed from `ember-template-lint: "4.18.2"` to `ember-template-lint: "5.7.2"` which is what ember addons are using. Currently its only targeting the `tests` files.  So we can either let the linter do its work or we can use overrides to cancel linting on test.  Thoughts

addons/api/package.json
dvonanderson: @ZedLi fixed

addons/core/package.json
ZedLi: I assume the issue is that we're now linting the HBS code that were embedded which was [included](https://github.com/ember-template-lint/ember-template-lint/releases/tag/v5.0.0) in v5? And that's why it's only in tests?

The quotes should always be single quotes though

addons/api/.eslintrc.js
```diff
@@ -9,14 +9,12 @@ module.exports = {
   root: true,
   parser: '@babel/eslint-parser',
   parserOptions: {
-    ecmaVersion: 'latest',
+    ecmaVersion: 2018,
```
ZedLi, line 12: Why the change back to `2018` and legacy?

addons/core/package.json
dvonanderson: @ZedLi yes that is correct. Ok I will push it back to 5.72 and put a message out in boundary-ui channel

addons/api/.eslintrc.js
dvonanderson: @ZedLi  So this one I couldn't find much of a explanation for it, but it seems to be part of the `@babel/plugin-proposal-decorators`.   Here is the failing [job](https://github.com/hashicorp/boundary-ui/actions/runs/6618893361/job/17978429268).   My guess is its because the `parserOptions` are not the same across for every addon and admin-ui?  Any thoughts here would be great

ZedLi: Does it happen locally?

dvonanderson: @ZedLi no

ZedLi: It would be good to investigate why the CI fails then and what's the difference between local and the job

dvonanderson: @ZedLi looks like the previous failed CI jobs are clean now.   Still not 100% what the issue was, but possibly the mismatch versions of `@babel/eslint-parser`.

addons/core/vendor/.gitkeep
ZedLi, line 1: Do you know if this changed anything? It seems like it was added to track an empty folder for git but isn't needed anymore?

ZedLi (approved):

PR #1968, opened 2023-10-25
Host set permission issue

calcaide:
## Description

- Target does not fetch a 'host' anymore. This is happening now within session.
- Within session, when fetching a 'host', if the user does not have permission we do NOT show an error and let the user quick connect.
- Decouple getting target host sources into host-sets, now isn't part of `model()` but has it's own method `getTargetHostSets`.
- When fetching `host-set`, if the user does not have permissions, we do NOT show it and allow the user for quick connect.
- Decouple getting host from host-sets, now isn't part of `model()` btu has it's own method `getHostsFromHostSets`.
- When fetching `host`, if the user does not have permissions, we do NOT show it and allow the user for quick connect.

## How to Test

- Create a target with address and port.
- Create a host catalog with one host-set and one host.
- Create a host catalog with two host-set and 3 host (2 host in 1 host set, 1 remaining host in other host-set).
- Create a user with specific grants.
- Play around changing `host` and `host-set` grants, allowing JUST list and the app works as expected. Allows quick connect and shows no errors related to grants).
- Play around changing `host` and `host-set` grants, allowing list & read and the app works as expected. Allows quick connect when no host-sets are present, performs quick connect when just one host-set and one host present, shows choose host modal when more than 1 host is present on the target.

Grants used:

- `id=*;type=session;actions=list,read:self,cancel:self`.
- `id=*;type=target;actions=list,authorize-session,read`.
- `id=*;type=host-set;actions=list`.
- `id=*;type=host;actions=list`.
- `id=*;type=host-catalog;actions=list,read`


ui/desktop/app/routes/scopes/scope/projects/targets/target.js
```diff
@@ -63,7 +98,7 @@ export default class ScopesScopeProjectsTargetsTargetRoute extends Route {
     const { isConnecting } = transition.to.queryParams;
 
     if (isConnecting) {
-      if (model.target.address || model.hosts.length === 1) {
+      if (model.target.address || model.hosts.length <= 1) {
```
ZedLi, line 101: Do we need the `model.target.address` if we're checking for `model.hosts.length <= 1`? Seems like that covers the address case now

ui/desktop/app/routes/scopes/scope/projects/targets/target.js
```diff
@@ -23,6 +23,56 @@ export default class ScopesScopeProjectsTargetsTargetRoute extends Route {
...
 
+  /**
+   * Given an array of hostSources returns an array of host sets
+   * @param {array} hostSources
+   * @returns {[HostSetModel]} hostSets
+   */
+  async getTargetHostSets(hostSources) {
+    let hostSets = [];
+
+    for (const hostSource of hostSources) {
```
ZedLi, line 34: I'm not sure I understand why you called these with a `for...of` loop which is not concurrent but also we stop at the first error anyways. Is this to try and return any hosts which might have succeeded?

Why not just try catch the `Promise.all` which fails fast since if there's any forbidden it's unlikely any would succeed?

ui/desktop/app/routes/scopes/scope/projects/targets/target.js
```diff
@@ -23,6 +23,56 @@ export default class ScopesScopeProjectsTargetsTargetRoute extends Route {
...
+  /**
+   * Given an array of hostSets returns an array of all the hosts within the
+   * host-set
+   * @param {array} hostSets
+   * @returns {[HostModel]} hosts
+   */
+  async getHostsFromHostSets(hostSets) {
+    const hosts = [];
+
+    if (hostSets.length > 1) {
```
ZedLi, line 58: Why is this length check needed?

ui/desktop/app/routes/scopes/scope/projects/targets/target.js
calcaide: @ZedLi 
Ok, I will implement `Promise.all` in both methods.

ui/desktop/app/routes/scopes/scope/projects/targets/target.js
calcaide: @ZedLi 
The purpose of this was to check hostSets has actually content. But you are right, despite I think the check should stay, the way of checking is incorrect. If there is just 1 hostSet this will not enter to fetch it, so it's incorrect.
I will change the check you just `if (hostSets)`. If hostSets is empty, we will return an empty array (so will not fail) but if hostSets has any value we will try to fetch them.

ui/desktop/app/routes/scopes/scope/projects/targets/target.js
calcaide: @ZedLi 
That is correct, with the new condition the `model.target.address` is already covered.

ui/desktop/app/routes/scopes/scope/projects/targets/target.js
```diff
@@ -36,24 +77,9 @@ export default class ScopesScopeProjectsTargetsTargetRoute extends Route {
       reload: true,
     });
 
-    if (target.host_sources) {
-      const hostSets = await Promise.all(
```
ZedLi, line 40: It's fine that you split this out into functions but to be honest, all that was needed was a try catch around these `Promise.all` (or the entire thing) and it would've just returned an empty array of hosts

addons/api/mirage/factories/target.js
```diff
@@ -46,17 +45,7 @@ export default factory.extend({
...
-        randomlySelectedCredentials;
-      randomlySelectedHostSets = server.schema.hostSets
-        // BLERG:  fun fact, for no reason at all, the element passed
-        // into a where function is not a full model instance, as you might
-        // expect at this level of abstraction, but appears to be a serialized
-        // representation of the model instance.  It's very confusing since
-        // the result set of `where` _is a collection of model instances_.
-        .where((hostSet) => hostSet.scopeId === target.scope.id)
-        .models.filter(() => randomBoolean(hostSetChance));
+      let randomlySelectedCredentialLibraries, randomlySelectedCredentials;
```
ZedLi, line 48: Any reason we removed the random hosts? Don't we still want a random number of hosts for good testing coverage and mirage usage?

I would've expected if we wanted addresses we would have a different trait for that

cameronperera: Good point. I wasn't thinking about mirage when I made the change. I will revert this and add a `withAddress` trait.

cameronperera: Fixed.

ui/desktop/tests/acceptance/scopes-test.js
```diff
@@ -282,53 +282,4 @@ module('Acceptance | scopes', function (hooks) {
       'Cannot retry'
     );
   });
-
-  test('handles cli error on connect', async function (assert) {
```
ZedLi, line 286: Are these tests not valid anymore, shouldn't we still be testing these scenarios?

Can we actually add a few tests for testing bare bones permissions as well?

cameronperera: These tests are duplicates of exactly what we are doing in the `target-test.js` file. Plus we used to be able to connect from this screen, but that is not true anymore. It all happens on the target-details route now.

ZedLi (approved):
Looks great, would love one last test for confirming the host retrieval now on the sessions model!

PR #1974, opened 2023-10-30
fix: 🐛 Add read:self permission for sessions in ability check

ZedLi:
## Description
We don't currently support the subaction grant for sessions `read:self` which could be a common permission for end users.

Before:
image

After:
image

## How to Test
Create a new user with a role that has read:self permissions instead of read.

## Checklist:


dvonanderson (approved):

calcaide (approved):
LGTM! Thanks for the work

PR #1975, opened 2023-10-31
Session permissions issue

cameronperera:
## How to Test

## Checklist:

### DIFF
--- a/ui/desktop/app/components/session/tabs/index.hbs
+++ b/ui/desktop/app/components/session/tabs/index.hbs
@@ -37,100 +37,92 @@
       <bc.Sidebar>
         <div class='details-sidebar'>
           {{#if @model.host}}
-            <div>
-              <div
-                class='title hds-typography-display-200 hds-font-weight-semibold'
-              >
-                {{t 'resources.host.details'}}
-              </div>
-              <Rose::MetadataList @orientation='vertical' as |list|>
-                {{! Host name }}
-                {{#if @model.host.name}}
-                  <list.Item @icon='monitor' @color='faint'>
-                    {{@model.host.name}}
-                  </list.Item>
-                {{/if}}
-                {{! Host description }}
-                {{#if @model.host.description}}
-                  <list.Item @icon='message-circle-fill' @color='faint'>
-                    {{@model.host.description}}
-                  </list.Item>
-                {{/if}}
-                {{! Host address }}
-                {{#if @model.host.address}}
-                  <list.Item @icon='network-alt' @color='faint'>
-                    {{@model.host.address}}
-                  </list.Item>
-                {{/if}}
-                {{! Host ID }}
-                <list.Item @icon='hash' @color='faint'>
-                  <Copyable
-                    @text={{@model.host.id}}
-                    @buttonText={{t 'actions.copy-to-clipboard'}}
-                    @acknowledgeText={{t 'states.copied'}}
-                  >
-                    {{@model.host.id}}
-                  </Copyable>
-                </list.Item>
-              </Rose::MetadataList>
+            <div
+              class='title hds-typography-display-200 hds-font-weight-semibold'
+            >
+              {{t 'resources.host.details'}}
             </div>
+            <Rose::MetadataList @orientation='vertical' as |list|>
+              {{! Host name }}
+              {{#if @model.host.name}}
+                <list.Item @icon='monitor' @color='faint'>
+                  {{@model.host.name}}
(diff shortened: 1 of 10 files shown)


ui/desktop/app/routes/scopes/scope/projects/targets/target.js
```diff
@@ -156,14 +162,35 @@ export default class ScopesScopeProjectsTargetsTargetRoute extends Route {
...
-      const session = await this.store.findRecord('session', session_id);
+      try {
+        session = await this.store.findRecord('session', session_id);
+      } catch (error) {
+        /**
+         * if the user cannot read or fetch the session we add the important
+         * information returned from the connect command to allow the user
+         * to still continue their work with the information they need
+         */
+        this.store.pushPayload({
```
ZedLi, line 175: I think we need to specify the type in this call otherwise it will only use the application serializer (even though we don't have any normalization specific for sessions).

ui/desktop/app/components/session/tabs/index.hbs
```diff
@@ -75,57 +75,63 @@
...
-              {{! Project name }}
-              <list.Item @icon='grid' @color='faint'>
-                {{@model.target.project.displayName}}
-              </list.Item>
-              {{! Target description }}
-              {{#if @model.target.description}}
-                <list.Item @icon='message-circle-fill' @color='faint'>
-                  {{@model.target.description}}
+          {{#if @model.target}}
+            <div>
```
ZedLi, line 79: Was this div necessary or is it extra?

ui/desktop/tests/acceptance/projects/sessions/session-test.js
```diff
@@ -160,7 +164,29 @@ module('Acceptance | projects | sessions | session', function (hooks) {
...
+
+    assert.strictEqual(currentURL(), urls.session);
+  });
+
+  test('visiting a session that does not have read permissions but a successful connect', async function (assert) {
+    assert.expect(2);
+    // verifies second call to sessions/:id is also a 403
+    QUnit.onUncaughtException = (err) => {
+      assert.true(err.errors[0].isForbidden);
+    };
```
ZedLi, line 177: To give a little explanation as to why this was necessary:

We [re-retrieve](https://github.com/hashicorp/boundary-ui/blob/e5263ba52c910cc7f19402f184bc81b5e2cefaab/ui/desktop/app/routes/scopes/scope/projects/sessions/session.js#L23) the session model when we enter the sessions route. We either previously preloaded by manually pushing to the store or successfully retrieved the model during connection. This retrieval will first pull from the cache in ember data store and _then_ retrieve the model again in the background (unless we set `backgroundReload: false`). 

This background refresh is the issue as it will still return a `403` but we can't catch it anywhere as it happens in the background inside ember internals. So eventually the test will fail without anyway for us to catch the error.

We _could_ turn off the background refresh as we don't really currently use any data that could be updated; the only thing that would get updated is the state of the session (i.e. when a session goes from pending to active once we connect). But we never really turn it off in our application and I think it could be unexpected in the future if we ever did need it, the only real negative to this would be in the situation where a user doesn't have permissions, we would be making an extra call that would fail silently in the background.

ui/desktop/tests/acceptance/projects/sessions/session-test.js
```diff
@@ -160,7 +164,29 @@ module('Acceptance | projects | sessions | session', function (hooks) {
...
-    await click('[data-test-target-detail-connect-button]');
+    await click(TARGET_CONNECT_BUTTON);
+
+    assert.strictEqual(currentURL(), urls.session);
+  });
+
+  test('visiting a session that does not have read permissions but a successful connect', async function (assert) {
+    assert.expect(2);
+    // verifies second call to sessions/:id is also a 403
+    QUnit.onUncaughtException = (err) => {
```
ZedLi, line 175: We've talked about this offline but I verified that this will set it globally so we need to reset this back to what it was at the end of this test.

ZedLi (changes requested):

ZedLi (approved):
Great work!

ui/desktop/tests/acceptance/projects/sessions/session-test.js
cameronperera: Resolving so I can merge.

calcaide (commented):
Sorry I am late to this PR. Reviewed and very nice job, really like the detailed code comments, well done!!

PR #1980, opened 2023-11-01
feat: 🎸 Add checks for cancel permissions in desktop sessions

ZedLi:
## Description
Added checks for canceling sessions. 
In cases where the user has no session read permissions, I forced a refresh on the list page in scenarios where the state has changed for sessions so they can still be canceled. Unfortunately we can't do anything on the details page since we can't retrieve the singular session without the permission so that would still error out.

Before:
Screenshot 2023-11-01 at 2 42 29 PM
Screenshot 2023-11-01 at 2 42 33 PM

After:
Screenshot 2023-11-01 at 2 42 02 PM
Screenshot 2023-11-01 at 2 42 05 PM

## How to Test
Only grant list and read permissions to sessions.

## Checklist:


addons/api/mirage/factories/session.js
```diff
@@ -15,7 +16,8 @@ export const pickRandomStatusString = () =>
 
 export default factory.extend({
   id: () => generateId('ss_'),
-
+  authorized_actions: () =>
+    permissions.authorizedActionsFor('session') || ['read', 'cancel'],
```
DhariniJeeva, line 20: not a blocker- should we also include read:self and cancel:self here?

addons/api/tests/unit/abilities/session-test.js
```diff
@@ -108,4 +108,74 @@ module('Unit | Ability | session', function (hooks) {
 
     assert.true(canService.can('read session', session));
   });
+
+  test('can cancel a session that is active', function (assert) {
+    assert.expect(1);
+
+    const session = store.createRecord('session', {
+      authorized_actions: ['cancel'],
```
DhariniJeeva, line 116: same as above, should we add test cases for `read:self` and `cancel:self`?

addons/api/mirage/factories/session.js
ZedLi: That's somewhat redundant since read covers `read:self` and same with cancel

addons/api/tests/unit/abilities/session-test.js
ZedLi: I did add test cases for them

addons/api/addon/abilities/session.js
```diff
@@ -14,13 +14,25 @@ export default class SessionAbility extends ModelAbility {
...
+
+  /**
+   * Only available sessions may be canceled.
+   * @type {boolean}
+   */
+  get canCancel() {
+    return (
+      this.model.isAvailable &&
+      (this.hasAuthorizedAction('cancel:self') ||
+        this.hasAuthorizedAction('cancel'))
```
DhariniJeeva, line 35: not a blocker: should this be symmetrical to super.CanRead by introducing `canCancel` in the model abilities?

ZedLi: I don't think it can be symmetrical, cancel is a specific permission for sessions which can't be generalized to the base model

ui/desktop/app/routes/scopes/scope/projects/sessions.js
```diff
@@ -84,6 +85,28 @@ export default class ScopesScopeProjectsSessionsRoute extends Route {
 
   // =actions
 
+  /**
+   * A hack to force a refresh when coming from a child route back to the list route.
+   * This is needed because these model hooks are on the `sessions` route which is a
+   * parent to the `sessions.session` route so we never leave the route which means
+   * the model hooks won't rerun when going back to the list route.
+   *
+   * This could be alternatively solved by moving all the model fetching to
```
DhariniJeeva, line 94: out of curiosity- why didn't we implement the alternate solution?

ZedLi: I tried it but ran into errors, it might be worth looking into refactoring it later

addons/api/addon/abilities/session.js
DhariniJeeva: you are right, forgot only sessions can cancel now

DhariniJeeva (approved):

cameronperera (approved):
LGTM

ui/desktop/app/routes/scopes/scope/projects/sessions.js
ZedLi: I actually re-did it and it works now, I might have had an unrelated error from permissions testing, do we have a preference?

addons/api/addon/abilities/session.js
calcaide: none blocking comment: In a comment below, you mention for read, `read` and `read:self` are redundant. If so, shouldn't we just check here for `cancel`? As per my understanding of the comment, that will cover `cancel:self`, isn't?

ui/desktop/app/routes/scopes/scope/projects/sessions.js
ZedLi: I think I will actually move it out into the index even with this weird extra error because otherwise we force a user to have list permissions as well when trying to connect with just `read,cancel` permissions for sessions. If it's in the index, the list route won't get called during connect as it won't be in the direct hierarchy

addons/api/addon/abilities/session.js
ZedLi: An end user might only have `cancel:self`, when I said it's redundant i mean having _both_ `cancel` and `cancel:self`

ui/desktop/app/templates/scopes/scope/projects/sessions/index.hbs
```diff
@@ -201,11 +201,12 @@
               <SessionStatus @model={{B.data}} />
             </B.Td>
             <B.Td>
-              {{#if B.data.isAvailable}}
+              {{#if (can 'cancel session' B.data)}}
```
calcaide, line 204: None blocking comment: Didn't check with design perhaps this is their desired behaviour. But I am thinking, is a user has no grants to cancel a session, shouldn't we disable the button instead of hidding it?

ui/desktop/app/templates/scopes/scope/projects/sessions/session.hbs
```diff
@@ -15,13 +15,15 @@
...
 
   <page.actions>
-    <Hds::Button
-      data-test-session-detail-cancel-button
-      @text={{t 'resources.session.actions.end'}}
-      @icon='x'
-      @color='secondary'
-      {{on 'click' (route-action 'cancelSession' @model)}}
-    />
+    {{#if (can 'cancel session' @model)}}
```
calcaide, line 18: None blocking comment: same comment as above about disabling the button rather than hidding it.

calcaide (approved):
Left none blocking comment. Awesome catch and work 🙌

ui/desktop/app/templates/scopes/scope/projects/sessions/index.hbs
ZedLi: I had seen a [recommendation](https://hashicorp.atlassian.net/wiki/spaces/DES/pages/2678685874/Hiding+Disabling) before that we should be hiding things users don't have permissions for rather than disabling so was just going off of that

calcaide: That makes sense!! Thanks for pointing out the recommendation!

ui/desktop/app/routes/scopes/scope/projects/sessions.js
ZedLi: I figured out the bug and was able to move the model hook over, take another look.

addons/rose/app/styles/rose/components/layout/_body-content.scss
```diff
@@ -8,7 +8,7 @@
 .rose-layout-body-content {
   display: grid;
   grid-template-areas: 'body body sidebar';
-  grid-template-columns: 1fr 1fr 300px;
+  grid-template-columns: 1fr 1fr fit-content(300px);
```
ZedLi, line 11: This was to originally handle the case where we might not have any target info but we _should_ always have some target info due to the user hitting the target list route first.

I kept this in here though because it still makes the grid more responsive and doesn't otherwise change anything.

ui/desktop/app/components/session/tabs/index.hbs
```diff
@@ -34,98 +34,105 @@
...
-            >
-              {{t 'resources.host.details'}}
-            </div>
-            <Rose::MetadataList @orientation='vertical' as |list|>
-              {{! Host name }}
-              {{#if @model.host.name}}
-                <list.Item @icon='monitor' @color='faint'>
-                  {{@model.host.name}}
+
+      {{#if (or @model.host @model.target)}}
```
cameronperera, line 38: good catch! Thanks

cameronperera (approved):
LGTM

DhariniJeeva (approved):

PR #1981, opened 2023-11-02
Update auth methods form

DhariniJeeva:
:tickets: [Jira ticket](https://hashicorp.atlassian.net/browse/ICU-11272)

## Description

This PR updates the rose key value list components with the list wrapper component. 

**NOTE**: For simplicity and easy review, I did not include the select list wrapper in this PR. We will still see old route actions and the rose key value components. These will be included in a separate PR

Before:
Screenshot 2023-11-02 at 11 21 55 AM

LDAP:

Screenshot 2023-11-02 at 11 22 33 AM

After:
LDAP:
Screenshot 2023-11-02 at 10 52 01 AM

OIDC:
Screenshot 2023-11-02 at 10 51 49 AM

Screenshot 2023-11-02 at 10 51 44 AM

## How to Test

## Checklist:


ui/admin/app/components/form/field/list-wrapper/text-input/index.hbs
```diff
@@ -16,7 +15,7 @@
               @value={{option.value}}
               @type='text'
               disabled={{@disabled}}
-              aria-labelledby={{valueHeaderID}}
+              aria-label={{valueHeaderID}}
```
DhariniJeeva, line 18: Note: since these fields don't have a visible label, we should use `aria-label` for accessibility conformance. `aria-labelled by` is used when we have visible label fields like the key-value component

ui/admin/app/components/form/auth-method/ldap/index.hbs
```diff
@@ -76,73 +76,39 @@
...
+      {{/if}}
+    </:fieldset>
+
+    <:field as |F|>
+      <F.Textarea
+        @name='certificates'
+        @options={{@model.certificates}}
+        @addOption={{fn @addOption 'certificates'}}
+        @removeOptionByIndex={{fn @removeOptionByIndex 'certificates'}}
+        @disabled={{form.disabled}}
```
lisbet-alvarez, line 107: I think this `@disabled` arg can be removed since it is already being passed in on line 82

addons/core/translations/form/en-us.yaml
```diff
@@ -103,7 +103,7 @@ claim:
   label: Claim
 claims_scopes:
   label: Claims Scopes
-  help: Optional claims scopes that will be requested during authentication.
+  help: claims scopes that will be requested during authentication.
```
lisbet-alvarez, line 106:
```suggestion
  help: Claims scopes that will be requested during authentication.
```

ui/admin/app/components/form/field/list-wrapper/text-input/index.hbs
lisbet-alvarez: great catch!!

lisbet-alvarez (approved):
So much neater!! Looks great 🙌

ui/admin/app/components/form/auth-method/ldap/index.hbs
DhariniJeeva: thanks for the catch, I updated and cleaned up some of the other files with redundant `@disabled` arg

ui/admin/app/components/form/auth-method/index.hbs
```diff
@@ -10,6 +10,8 @@
     edit=@edit
     submit=@submit
     cancel=@cancel
+    addOption=this.addOption
+    removeOptionByIndex=this.removeOptionByIndex
```
ZedLi, line 14: I just had a thought, if we're using a wrapper component that doesn't just handle `value` like currently for the two components, won't we need to have multiple functions for them?

I wonder if we should discuss trying to make the components handle these so we can remove the need for `addOption` or `removeOptions` on the consumer. I had originally only left it to the consumer because I didn't want to try and generalize handling the tracked array but it seems like we should now?

ui/admin/app/components/form/field/list-wrapper/text-input/index.hbs
ZedLi: We shouldn't be using a GUID then. The label should be human readable otherwise we would be reading back a GUID to the user.

We were treating the table header as a "label" which was human readable and linked it with the ID and `aria-labeledby`, if we're just using `aria-label`, we need to just give it real text.

You can confirm this with your screen reader, give it a try. We should get more into the habit of checking with a screen reader

ui/admin/app/components/form/field/list-wrapper/textarea/index.hbs
```diff
@@ -0,0 +1,63 @@
...
+    <:body as |B|>
+
+      {{#each @options as |option index|}}
+        <B.Tr>
+          <B.Td>
+            <Hds::Form::Textarea::Field
+              @value={{option.value}}
+              @type='text'
+              disabled={{@disabled}}
+              aria-label={{valueHeaderID}}
```
ZedLi, line 18: Same here, don't use an ID for the aria-label

ui/admin/tests/acceptance/auth-methods/update-test.js
```diff
@@ -115,21 +132,28 @@ module('Acceptance | auth methods | update', function (hooks) {
     await select('form fieldset:nth-of-type(1) select', 'RS384');
     await click('form fieldset:nth-of-type(1) [title="Add"]');
     // Remove all allowed audiences
-    await Promise.all(
-      findAll('form fieldset:nth-of-type(2) [title="Remove"]').map((element) =>
-        click(element)
-      )
+    const allowedAudiencesList = await Promise.all(
```
ZedLi, line 135: We don't need to be awaiting or using promise.all since we're just finding html elements

ui/admin/app/components/form/auth-method/index.hbs
DhariniJeeva: could you elaborate on what you mean by multiple functions?

ui/admin/app/components/form/field/list-wrapper/textarea/index.hbs
DhariniJeeva: Agree with this. I will make the changes

ui/admin/app/components/form/auth-method/index.hbs
ZedLi: What would it look like if we also used the key value list or the yet to be made select list wrapper? I assume they won't be the same function as the current `addOption`?

DhariniJeeva: that is a good point, passing the model to the components can help remove these functions from consumers in case of different types. something like this, let me know what you think.

```
    const field = this.args.name;
    const existingArray = this.args.model[field] ?? [];
    const newArray = [
      ...existingArray,
      { key: this.newOptionKey, value: this.newOptionValue },
    ];
    set(this.args.model, field, newArray);

    this.newOptionKey = '';
    this.newOptionValue = '';
```

ZedLi: Yup I think it's worth making the change since we use nearly use it the same way everywhere (except I think the worker tags?)

We could just leave the current way in and check if we have a function to use that first, otherwise just use the model/name (maybe also leave the `@options` as optional).

ui/admin/app/components/form/field/list-wrapper/index.hbs
```diff
@@ -7,6 +7,7 @@
   @layout={{@layout}}
   @isRequired={{@isRequired}}
   @isOptional={{@isOptional}}
+  @disabled={{@disabled}}
```
ZedLi, line 10: I don't think the fieldset can take a disabled

ui/admin/app/components/form/field/list-wrapper/key-value/index.hbs
```diff
@@ -10,15 +10,14 @@
         <H.Th id={{keyHeaderID}}>{{t 'form.key.label'}}</H.Th>
         <H.Th id={{valueHeaderID}}>{{t 'form.value.label'}}</H.Th>
         <H.Th>
-          {{#if @removeOptionByIndex}}
+          {{#if this.removeOptionByIndex}}
```
ZedLi, line 13: I don't think you need these if statements anymore since you'll always have it now

ui/admin/app/components/form/field/list-wrapper/key-value/index.hbs
```diff
@@ -41,15 +40,15 @@
...
+            {{#if this.removeOptionByIndex}}
               <Hds::Button
                 data-test-remove-button
                 @text={{t 'actions.remove'}}
                 @color='critical'
                 @icon='trash'
                 @isIconOnly={{true}}
                 disabled={{@disabled}}
-                {{on 'click' (fn @removeOptionByIndex index)}}
+                {{on 'click' (fn this.removeOptionByIndex @name index)}}
```
ZedLi, line 51: Why pass in the name here? You already have access to it in the function

ui/admin/app/components/form/field/list-wrapper/key-value/index.js
```diff
@@ -21,27 +21,54 @@ export default class MappingListComponent extends Component {
...
+
+  /**
+   * Removes an option by index. We recreate a new array after
+   * splicing out the item so that ember is aware that the array has been modified.
+   * @param field {string}
+   * @param index {number}
+   */
+  @action
+  removeOptionByIndex(field, index) {
+    const newArray = this.args.model[field].filter((_, i) => i !== index);
```
ZedLi, line 71: We should do the same check we did in `addOption` here as well and check if a user is passing it in and call it for them if they are

ui/admin/app/components/form/field/list-wrapper/text-input/index.js
```diff
@@ -20,22 +20,42 @@ export default class MappingListTextInputComponent extends Component {
...
+  /**
+   * Removes an option by index. We recreate a new array after
+   * splicing out the item so that ember is aware that the array has been modified.
+   * @param field {string}
+   * @param index {number}
+   */
+
+  @action
+  removeOptionByIndex(field, index) {
+    console.log(field, 'fielddd', this.args.model[field]);
```
ZedLi, line 57: extra console log here

ui/admin/app/components/form/field/list-wrapper/key-value/index.hbs
DhariniJeeva: yup not needed!

ui/admin/app/components/form/field/list-wrapper/text-input/index.js
```diff
@@ -20,22 +20,45 @@ export default class MappingListTextInputComponent extends Component {
...
+        this.args.addOption({
+          value: this.newOptionValue,
+        });
+      }
+    } else {
+      const field = this.args.name;
+      const existingArray = this.args.model[field] ?? [];
+      const newArray = [
+        ...existingArray,
+        { key: this.newOptionKey, value: this.newOptionValue },
```
ZedLi, line 40: Does this one take a key?

ui/admin/app/components/form/field/list-wrapper/textarea/index.js
```diff
@@ -0,0 +1,64 @@
...
+        this.args.addOption({
+          value: this.newOptionValue,
+        });
+      }
+    } else {
+      const field = this.args.name;
+      const existingArray = this.args.model[field] ?? [];
+      const newArray = [
+        ...existingArray,
+        { key: this.newOptionKey, value: this.newOptionValue },
```
ZedLi, line 40: Same here, do we have a key?

ZedLi (approved):
Looks good!

ui/admin/app/components/form/field/list-wrapper/textarea/index.js
DhariniJeeva: oops copy pasta. removed!

PR #1984, opened 2023-11-03
refactor: 💡 allow target connection even with no read ability

cameronperera:
## Description

If a user can connect to a target but not read, we will allow them to connect now.
Connect logic has moved back to the `scopes.scope.projects.targets.index` route and we call the main `connect` method on that route as well from the child route. Connect no longer handles the error states, each route which has unique error handling implements that logic in it's own route. As we move away from `route-actions` all of this logic will likely move to the controllers for each route.

## How to Test

If you have a user (not admin) that you can remove the `read` permission from for targets, you can test this build locally.

## Checklist:

- [ ] ~My changes generate no new warnings~

### DIFF
--- a/ui/desktop/app/controllers/scopes/scope/projects/targets/index.js
+++ b/ui/desktop/app/controllers/scopes/scope/projects/targets/index.js
@@ -1,0 +1,103 @@
+/**
+ * Copyright (c) HashiCorp, Inc.
+ * SPDX-License-Identifier: MPL-2.0
+ */
+
+import Controller from '@ember/controller';
+import { inject as service } from '@ember/service';
+import { action } from '@ember/object';
+import { loading } from 'ember-loading';
+
+export default class ScopesScopeProjectsTargetsIndexController extends Controller {
+  // =services
+
+  @service confirm;
+  @service ipc;
+  @service router;
+  @service session;
+  @service store;
+
+  // =methods
+
+  /**
+   * Quick connect method used to call main connect method and handle
+   * connection errors unique to this route
+   * @param {TargetModel} target
+   */
+  @action
+  async quickConnect(target) {
+    try {
+      await this.connect(target);
+    } catch (error) {
+      this.confirm
+        .confirm(error.message, { isConnectError: true })
+        // Retry
+        .then(() => this.quickConnect(target));
+    }
+  }
+
+  /**
+   * Establish a session to current target.
+   * @param {TargetModel} target
+   * @param {HostModel} host
+   */
+  @action
+  @loading
+  async connect(target, host) {
+    // Check for CLI
+    const cliExists = await this.ipc.invoke('cliExists');
+    if (!cliExists) throw new Error('Cannot find Boundary CLI.');
+
+    const options = {
+      target_id: target.id,
+      token: this.session.data.authenticated.token,
+    };
+
+    if (host) options.host_id = host.id;
+
+    // Create target session
+    const connectionDetails = await this.ipc.invoke('connect', options);
+
+    // Associate the connection details with the session
+    let session;
+    const { session_id, address, port, credentials } = connectionDetails;
+    try {
+      session = await this.store.findRecord('session', session_id);
+    } catch (error) {
+      /**
+       * if the user cannot read or fetch the session we add the important
+       * information returned from the connect command to allow the user
+       * to still continue their work with the information they need
+       */
+      this.store.pushPayload('session', {
+        sessions: [
+          {
+            id: session_id,
+            proxy_address: address,
+            proxy_port: port,
+            target_id: target.id,
+          },
+        ],
+      });
+      session = this.store.peekRecord('session', session_id);
+    }
+
+    // Flag the session has been open in the desktop client
+    session.started_desktop_client = true;
+    /**
+     * Update the session record with proxy information from the CLI
+     * In the future, it may make sense to push this off to the API so that
+     * we don't have to manually persist the proxy details.
+     */
+    session.proxy_address = address;
+    session.proxy_port = port;
+    if (credentials) {
+      credentials.forEach((cred) => session.addCredential(cred));
+    }
+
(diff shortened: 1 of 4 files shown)


ui/desktop/app/routes/scopes/scope/projects/targets/target.js
```diff
@@ -120,9 +145,9 @@ export default class ScopesScopeProjectsTargetsTargetRoute extends Route {
   /**
    * Determine if we show host modal or quick connect based on target attributes.
    * @param {TargetModel} model
+   * @param {function} toggleModal
    */
   @action
-  @loading
   async preConnect(model, toggleModal) {
```
DhariniJeeva, line 151: Do we not need this `@loading` here anymore?

cameronperera: No. The main `connect` has `@loading` on it so it handles that for all instances of "connect". I tested this to make sure.

DhariniJeeva (approved):
LGTM

ui/desktop/tests/acceptance/projects/targets/index-test.js
```diff
@@ -225,7 +230,7 @@ module('Acceptance | projects | targets', function (hooks) {
       .doesNotExist();
   });
 
-  test('user is redirected to target details page when unable to connect from list view', async function (assert) {
+  test('user is redirected to target details page when unable to connect from list view if they have read and authorize-seesion permissions', async function (assert) {
```
DhariniJeeva, line 233:
```suggestion
  test('user is redirected to target details page when unable to connect from list view if they have read and authorize-session permissions', async function (assert) {
```

ui/desktop/app/controllers/scopes/scope/projects/targets/index.js
```diff
@@ -0,0 +1,103 @@
...
+  }
+
+  /**
+   * Establish a session to current target.
+   * @param {TargetModel} target
+   * @param {HostModel} host
+   */
+  @action
+  @loading
+  async connect(target, host) {
```
calcaide, line 46: None blocking comment: we moved `connect` from the route file to the controller file. Can you briefly explain why? I do like more this approach tho. Makes more sense on what a controller is, but I would like to hear your thoughts.

ui/desktop/app/controllers/scopes/scope/projects/targets/index.js
```diff
@@ -0,0 +1,103 @@
...
+import { action } from '@ember/object';
+import { loading } from 'ember-loading';
+
+export default class ScopesScopeProjectsTargetsIndexController extends Controller {
+  // =services
+
+  @service confirm;
+  @service ipc;
+  @service router;
+  @service session;
```
calcaide, line 17: Are we using the service `session` within this file? I can not find a reference to it.

ui/desktop/app/controllers/scopes/scope/projects/targets/index.js
cameronperera: Yes! I actually tried keeping it in the route as that is the standard in our app. However, when trying to utilize the connect `route-action` in the child route, we lost the ability to await the connect. This way, we can properly await the connect without showing that the route has changed in the event of a successful connection.

ui/desktop/app/controllers/scopes/scope/projects/targets/index.js
cameronperera: Yes. It is used on line 53 below to get the authenticated token.

ui/desktop/app/controllers/scopes/scope/projects/targets/index.js
```diff
@@ -0,0 +1,103 @@
...
+     * In the future, it may make sense to push this off to the API so that
+     * we don't have to manually persist the proxy details.
+     */
+    session.proxy_address = address;
+    session.proxy_port = port;
+    if (credentials) {
+      credentials.forEach((cred) => session.addCredential(cred));
+    }
+
+    await this.router.transitionTo(
```
ZedLi, line 98: I don't think the `transitionTo` returns a promise?

ui/desktop/app/controllers/scopes/scope/projects/targets/index.js
```diff
@@ -0,0 +1,103 @@
...
+    }
+  }
+
+  /**
+   * Establish a session to current target.
+   * @param {TargetModel} target
+   * @param {HostModel} host
+   */
+  @action
+  @loading
```
ZedLi, line 45: Should the `loading` decorator be on this or the connect functions that use these? Or can we decorate them all with `loading`?

ZedLi (approved):

ui/desktop/app/controllers/scopes/scope/projects/targets/index.js
calcaide: You are right, I miss that one, and is an important one 😅

ui/desktop/app/controllers/scopes/scope/projects/targets/index.js
calcaide: +1 to Zed's Comment, [as per documentation](https://api.emberjs.com/ember/release/classes/RouterService/methods/transitionTo?anchor=transitionTo), it returns and object.

calcaide (approved):
LGTM!! Thanks for this work 🙌

ui/desktop/app/controllers/scopes/scope/projects/targets/index.js
cameronperera: removed.

ui/desktop/app/controllers/scopes/scope/projects/targets/index.js
cameronperera: That is a good point. I will document this for when we refactor the error messages part of this work.

PR #1989, opened 2023-11-09
feat: 🎸 Add handler to make search requests to client daemon

ZedLi:
✅ Closes: https://hashicorp.atlassian.net/browse/ICU-11691

## Description
Added a handler just to the desktop client to integrate with the client daemon.

## How to Test
Build a binary from the `llb-client-daemon` branch and use that binary in your local dev environment. You can confirm in the DC that we're now not making any requests to the controller in the network tab and see the request logs in the main thread in electron.

## Checklist:

### DIFF
--- a/addons/api/addon/handlers/client-daemon-handler.js
+++ b/addons/api/addon/handlers/client-daemon-handler.js
@@ -1,0 +1,64 @@
+import { pluralize } from 'ember-inflector';
+
+/**
+ * Not all types are yet supported by the client daemon so we'll
+ * just whitelist the ones we need to start with.
+ * @type {string[]}
+ */
+const supportedTypes = ['target', 'session'];
+
+/**
+ * Handler to sit in front of the API request layer
+ * so we can request from the daemon first
+ */
+const ClientDaemonHandler = {
+  async request(context, next) {
+    switch (context.request.op) {
+      case 'query': {
+        const { store, data } = context.request;
+        const { type, query } = data;
+
+        if (!supportedTypes.includes(type)) {
+          return next(context.request);
+        }
+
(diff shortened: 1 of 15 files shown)


addons/api/addon/handlers/client-daemon-handler.js
ZedLi, line 1: I originally had this in the desktop store service since it's only used in the desktop and really will only be used there but I wasn't sure where to put it. I just put it back in a handlers folder in the API addon but am open to suggestions

addons/api/addon/handlers/client-daemon-handler.js
```diff
@@ -0,0 +1,67 @@
...
+  async request(context, next) {
+    switch (context.request.op) {
+      case 'query': {
+        const { store, data } = context.request;
+        const { type, query } = data;
+        const isClientDaemonRunning = await this.ipc.invoke(
+          'isClientDaemonRunning'
+        );
+
+        if (!supportedTypes.includes(type) || !isClientDaemonRunning) {
```
ZedLi, line 24: Should I just default back to our normal calls if the daemon isn't running for any reason? Or should we error out?

cameronperera: That is a great question. I would imagine we want to default back to our normal calls. We wouldn't be able to do any search and filtering though right? Should we hide the bar if so?

cameronperera (approved):

addons/api/addon/handlers/client-daemon-handler.js
ZedLi: Yes exactly, we wouldn't be able to properly filter, it probably would make sense to hide the filters if that's the route we go down but we haven't talked too much about the failure case

PR #1990, opened 2023-11-10
Upgrade desktop UI to Ember 4.12.2

dvonanderson:
:tickets: [Jira ticket](https://hashicorp.atlassian.net/browse/ICU-11747)

## Description

- Upgrade `desktop` Ui to Ember 4.12.2
- Fix `stylelint` errors for `media-feature-range-notation`
- fix `template-lint` errors for `Non-translated string used`
- 
## Checklist:


addons/api/package.json
```diff
@@ -69,9 +69,6 @@
     "qunit-dom": "^2.0.0",
     "webpack": "^5.78.0"
   },
-  "peerDependencies": {
```
ZedLi, line 72: Why were peer dependencies removed? Don't we want them?

ui/desktop/package.json
```diff
@@ -36,7 +36,7 @@
     "start:desktop": "ember electron",
     "clean:coverage": "rm -rf coverage_*",
     "pretest": "yarn clean:coverage",
-    "test": "npm-run-all test:*",
+    "test": "concurrently \"npm:lint\" \"npm:test:*\" --names \"lint,test:\"",
```
ZedLi, line 39: Why did we add linting to this, I didn't notice it in admin UI but we should keep our linting and testing commands separate. Can you change this to just ember exam in both desktop and UI?

package.json
```diff
@@ -22,14 +22,14 @@
     "build:ui:desktop:win": "docker-compose -f docker-compose-embedding.yml run build-desktop-production && yarn build:ui:desktop:app",
     "build:ui:desktop:production": "yarn --cwd ui/desktop build:production",
     "build:ui:desktop:app": "yarn --cwd ui/desktop build:desktop",
-    "lint": "npm-run-all -p lint:*",
+    "lint": "concurrently --kill-others-on-fail \"yarn lint:addons-api\" \"yarn lint:addons-auth\" \"yarn lint:addons-core\" \"yarn lint:addons-rose\" \"yarn lint:ui-admin\" \"yarn lint:ui-desktop\"",
```
ZedLi, line 25: Why are these commands listing all the packages to run instead of using the wild card syntax to run all the ones in this package.json?

addons/api/package.json
calcaide: Peer dependencies shouldn't be removed from package.json. Also, the dependencies declared within peerDependencies, in this case `ember-source` need to be installed at developer responsability.

In fact (**not to be resolved within this PR**) currently, yarn is complaining about `ember-source` been umet as peer dependency within api, core and rose, I attach screenshot of my terminal after running `yarn install`

Screenshot 2023-11-10 at 1 21 28 PM

ui/admin/app/styles/_notify.scss
```diff
@@ -86,7 +86,7 @@
   }
 }
 
-@media only screen and (max-width: 680px) {
```
calcaide, line 89: Was this a manual change or an automated one? if manual, can you elaborate on why we change?

ui/desktop/.eslintrc.js
```diff
@@ -36,6 +38,7 @@ module.exports = {
       files: [
         './.eslintrc.js',
         './.prettierrc.js',
+        './.stylelintrc.js',
         './.template-lintrc.js',
         '.stylelintrc.js',
```
calcaide, line 43: Isn't line 41 and line 43 a duplicate? perhaps we should choose one and delete the other one?

ui/admin/app/styles/_notify.scss
calcaide: Why we are performing this change within `ui/admin` if this is desktop upgrade?

ui/admin/package.json
```diff
@@ -102,7 +102,7 @@
     "sinon": "^14.0.0",
     "stylelint": "^15.10.1",
     "stylelint-config-prettier-scss": "^0.0.1",
-    "stylelint-config-standard-scss": "^6.0.0",
```
calcaide, line 105: I will suggest to not perform `ui/admin` upgrades on a PR that is expected to perform `ui/desktop` unless there is a reason for it, can you ellaborate why we have `admin` changes within this PR?

addons/api/package.json
dvonanderson: @ZedLi I guess a question of learning would be why would we need a peer-dependency for Ember source 4.0.0 when we have it already for addons and ui?

ui/desktop/package.json
dvonanderson: @ZedLi absolutely we can

package.json
dvonanderson: @ZedLi good catch, will fix

ui/desktop/.eslintrc.js
dvonanderson: @calcaide good catch, fix incoming

ui/admin/app/styles/_notify.scss
dvonanderson: @calcaide updating the `stylelint-standard-config-scss` to the correct version that matches with `stylelint` and not give us an unmet peer dependency, has a new modern way of writing media query ranges.  `prefix` allows the use of patterns like `min-width` where as `context` wants more the modern take which is using an evaluator to express it.

https://stylelint.io/user-guide/rules/media-feature-range-notation/

ui/admin/package.json
dvonanderson: @calcaide will do

ui/admin/app/styles/_notify.scss
dvonanderson: @calcaide I will create a ticket to upgrade both `stylelint-standard-config-scss` dependencies for `admin` and `desktop`.  May also be a discussion point for #tech-time, because using `context` patterns for media queries is a pattern change.

addons/rose/package.json
```diff
@@ -96,7 +96,7 @@
     "react-syntax-highlighter": "^15.5.0",
     "stylelint": "^15.10.1",
     "stylelint-config-prettier-scss": "^0.0.1",
-    "stylelint-config-standard-scss": "^6.0.0",
+    "stylelint-config-standard-scss": "^11.0.0",
```
calcaide, line 99: This change belongs to `rose` is this necessary within this PR? I think we want to keep the scope of the upgrade within Desktop?

ui/admin/package.json
```diff
@@ -31,7 +31,7 @@
     "start": "ember serve",
     "clean:coverage": "rm -rf coverage_*",
     "pretest": "yarn clean:coverage",
-    "test": "concurrently \"npm:lint\" \"npm:test:*\" --names \"lint,test:\"",
+    "test": "concurrently \"npm:test:*\" --names \"test:\"",
```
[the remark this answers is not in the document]
calcaide: Will approve, but I was thinking on getting this change within Admin Upgrade to ember. All good.

calcaide (approved):
LGTM!! Thanks for the massive work!! 🙌 💪 

As a note for the next upgrade (I think admin is left?) to narrow down the scope of the PR to 1 UI or to 1 addon, facilitates understanding and help us to get small wins on the way 😉

PR #1991, opened 2023-11-10
chore: 🤖 Add support for pagination in query method

DhariniJeeva:
:tickets: [Jira ticket](https://hashicorp.atlassian.net/browse/ICU-11688)

Backend introduced pagination for all resources, and they have a default page_size of 1000, which means if there are more than 1k resources, the UI will not get the full list unless we pass in the `refresh_token` of the current result list. 

The goal of this PR is to simply return the full result set from the API

Screenshot 2023-11-10 at 9 37 45 AM

## How to Test
[Note: you need to run the `llb/pagination` to test this]
- Add 4 targets, and in the routes/target set a page_size of 1(this is to mock the API default page_size limit). You should still see all 4 on the UI 

## Checklist:

- ~[] I have added before and after screenshots for UI changes~

### DIFF
--- a/addons/api/addon/adapters/application.js
+++ b/addons/api/addon/adapters/application.js
@@ -118,12 +118,41 @@ export default class ApplicationAdapter extends RESTAdapter.extend(
 
   /**
    * Intercepts "empty" responses and adds an empty `items` array.
+   * This query method now supports pagination by checking the responseType and
+   * passing in the refresh token if the responseType indicates more data
    * @override
    * @method query
    * @return {Promise} promise
    */
-  query() {
-    return super.query(...arguments).then(prenormalizeArrayResponse);
+  async query(store, schema, query) {
+    let result;
+    let data = [];
+
+    //run this loop as long as the response_type is delta,
+    //which indicates that there are more items in the list
+    do {
+      try {
+        result = await super.query(store, schema, query);
+
+        //add the result items to a data array
+        if (result.items) {
+          data.push(...result.items);
+        }
+
+        //pass in the refresh token for subsequent calls to fetch the remaining list iems
+        query.refresh_token = result.refresh_token;
+
+        //break the loop as soon the response_type becomes complete
+        if (result.response_type === 'complete') {
+          break;
+        }
+      } catch (err) {
+        console.error(err);
+      }
+    } while (result.response_type === 'delta');
+
+    result.items = data;
+    return result;
   }
 
   /**
--- a/addons/api/addon/adapters/application.js
+++ b/addons/api/addon/adapters/application.js
@@ -128,8 +128,8 @@ export default class ApplicationAdapter extends RESTAdapter.extend(
     let result;
     let data = [];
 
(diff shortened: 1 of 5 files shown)


lisbet-alvarez (approved):
🎊Awesome work for pagination support kick off!🙌

addons/api/addon/adapters/application.js
```diff
@@ -118,12 +118,41 @@ export default class ApplicationAdapter extends RESTAdapter.extend(
...
    * @return {Promise} promise
    */
-  query() {
-    return super.query(...arguments).then(prenormalizeArrayResponse);
+  async query(store, schema, query) {
+    let result;
+    let data = [];
+
+    //run this loop as long as the response_type is delta,
+    //which indicates that there are more items in the list
```
ZedLi, line 132:
```suggestion
    // Run this loop as long as the response_type is delta,
    // which indicates that there are more items in the list
```

addons/api/addon/adapters/application.js
```diff
@@ -118,12 +118,41 @@ export default class ApplicationAdapter extends RESTAdapter.extend(
...
+        }
+
+        //pass in the refresh token for subsequent calls to fetch the remaining list iems
+        query.refresh_token = result.refresh_token;
+
+        //break the loop as soon the response_type becomes complete
+        if (result.response_type === 'complete') {
+          break;
+        }
+      } catch (err) {
```
ZedLi, line 149: Should we just let it throw for now?

addons/api/addon/adapters/application.js
```diff
@@ -118,12 +118,41 @@ export default class ApplicationAdapter extends RESTAdapter.extend(
...
+   * This query method now supports pagination by checking the responseType and
+   * passing in the refresh token if the responseType indicates more data
    * @override
    * @method query
    * @return {Promise} promise
    */
-  query() {
-    return super.query(...arguments).then(prenormalizeArrayResponse);
+  async query(store, schema, query) {
+    let result;
```
ZedLi, line 128: Should this start as { items: [] }? And we can push to the array rather than using a data field? Do we need it separated for any reason?

DhariniJeeva: I separated to avoid adding duplicates to the results items array. Otherwise, I had to do another check to ensure I wasn't adding duplicated results to the item array, which seemed more readable.

ZedLi: What do you mean by duplicates in this case?

calcaide: After a chat with @DhariniJeeva over zoom I understand this issue better and I agree we will need 2 data structures. One `result` to keep the entire result of the query and `data` to keep building an array of `result.items` per iteration.

DhariniJeeva: Since we are reassigning the `result`, if we don't have another data structure to store the total result, we will only be overwriting the previous entry in the result, and pushing items to that result can lead to duplicates as a side effect

PR #1995, opened 2023-11-15
Icu 11748 add free search for targets

cameronperera:
:tickets: [Jira ticket](https://hashicorp.atlassian.net/browse/ICU-11748)

## Description

This adds the ability to type in a search field and query against specific target properties. If the query has no match, it will return an empty array and show a unique message.
_This work does not include filtering work... yet. The whole segmented group is present but only the text input is working for this PR. Tests will be added later as well while we plan an easy way to mock the client daemon responses in our tests._

https://github.com/hashicorp/boundary-ui/assets/29386339/182fae88-6c5e-4b5a-95ee-208eabec817e

## How to Test

Build a binary from the `llb-client-daemon` branch and use that binary in your local dev environment. You will need to move that binary into the `ui/desktop/electron-app/cli` folder as well. Be sure to start the desktop client using this command: `ENABLE_MIRAGE=false BYPASS_CLI_SETUP=true yarn start:desktop`

## Checklist:


addons/core/translations/actions/en-us.yaml
```diff
@@ -52,3 +52,8 @@ toggle-color-theme: Toggle Color Theme
 toggle-feature-edition: Toggle Feature Edition
 toggle-licensed-features: Toggle Licensed Features
 more: ', +{remainingItems} more'
+search: Search
+filtering:
```
DhariniJeeva, line 56: not a blocker- since this file doesn't have a nested structure, I'm wondering if we need to introduce a new pattern here-  Can we move `yes, no, narrow-results` to titles and keep search and filter here? would that work?

ui/desktop/app/controllers/scopes/scope/projects/targets/index.js
```diff
@@ -17,6 +18,13 @@ export default class ScopesScopeProjectsTargetsIndexController extends Controlle
   @service session;
   @service store;
 
+  // =attributes
+
+  @tracked targets = this.model;
+  @tracked freeSearch;
+  @tracked searchItems = ['Project 1', 'Project 2', 'Project 3'];
```
DhariniJeeva, line 25: What are these `searchItems` for?  Is this for testing purposes?

cameronperera: Yes, it was for testing purposes originally and what I will be working on next. I had converted that dropdown to be more like what it will look like in it's final iteration which will be a dynamic list of projects available for filtering. I decided to leave it in for now instead of hardcoding it in the template.

addons/core/translations/actions/en-us.yaml
cameronperera: Yeah that works. Keep it simple. I will update and push up.

DhariniJeeva (approved):
LGTM

ui/desktop/app/controllers/scopes/scope/projects/targets/index.js
calcaide: not blocking comment: Could we add a short comment saying this is temporary and we will rely on real data later?

ui/desktop/app/templates/scopes/scope/projects/targets/index.hbs
```diff
@@ -13,189 +13,149 @@
...
+          />
+          <S.Dropdown as |DD|>
+            <DD.ToggleButton
+              @color='secondary'
+              @text={{t 'resources.scope.title'}}
+            />
+            <DD.Header @hasDivider={{true}}>
+              <Hds::Form::TextInput::Base
+                @type='search'
+                placeholder={{t 'actions.narrow-results'}}
```
lisbet-alvarez, line 34: ? for learning: why doesn't this input need an `aria-label` but the input in line 20 does?

ui/desktop/app/controllers/scopes/scope/projects/targets/index.js
```diff
@@ -100,4 +108,37 @@ export default class ScopesScopeProjectsTargetsIndexController extends Controlle
       session_id
     );
   }
+
+  /**
```
calcaide, line 112: None blocking and not mandatory: Would be nice to have a comment like in the other methods, perhaps something like sets freeSearch value and triggers the search? you know better the context. Thoughts?

calcaide (approved):

ui/desktop/electron-app/src/services/client-daemon-manager.js
```diff
@@ -24,7 +24,7 @@ class ClientDaemonManager {
     const data = spawnSync(daemonStatusCommand);
 
     const status = jsonify(data);
-    this.#socketPath = status?.socket_address;
+    this.#socketPath = status?.item?.socket_address;
```
lisbet-alvarez, line 27: ? for learning: what is this change for?

lisbet-alvarez (approved):
🙌✨

ui/desktop/app/controllers/scopes/scope/projects/targets/index.js
```diff
@@ -17,6 +18,13 @@ export default class ScopesScopeProjectsTargetsIndexController extends Controlle
   @service session;
   @service store;
 
+  // =attributes
+
+  @tracked targets = this.model;
```
ZedLi, line 23: Do you think it's possible that we have a code-smell here by having to save the model and not just using the model directly returned from the route? Don't we have a convention to do the data fetching in routes?

How will we handle loading or error cases with this?

ui/desktop/app/controllers/scopes/scope/projects/targets/index.js
```diff
@@ -100,4 +108,37 @@ export default class ScopesScopeProjectsTargetsIndexController extends Controlle
...
+
+  /**
+   * builds a query string for all searchable target properties
+   * @returns {string}
+   */
+  buildQuery() {
+    return this.searchableProps
+      .map((prop) => `${prop} % '${this.freeSearch}'`)
+      .join(' or ');
+  }
```
ZedLi, line 130: I think we talked about this before but does it make sense for this to live on the controller? Do you think the responsibility of building the query string should be on this component/controller?

ui/desktop/app/templates/scopes/scope/projects/targets/index.hbs
```diff
@@ -13,189 +13,149 @@
...
-
-        <toolbar.secondary>
-          <ToolbarRefresher @onClick={{route-action 'refreshTargets'}} />
-        </toolbar.secondary>
-
-      </Rose::Toolbar>
-    {{/if}}
-
+  <page.body class='targets'>
     {{#if @model}}
```
ZedLi, line 17: Why do we have a check for `@model` here but `this.targets` below? It seems like these can fall out of sync now?

ZedLi (changes requested):

ui/desktop/app/controllers/scopes/scope/projects/targets/index.js
```diff
@@ -100,4 +108,37 @@ export default class ScopesScopeProjectsTargetsIndexController extends Controlle
       session_id
     );
   }
+
+  /**
+   * @param {object} event
+   */
+  @action
+  async handleInput(event) {
```
ZedLi, line 116: This will also need to be debounced

cameronperera: This will be address in another PR, ticket [here](https://hashicorp.atlassian.net/browse/ICU-11851)

ui/desktop/app/templates/scopes/scope/projects/targets/index.hbs
cameronperera: Answer for learning: It actually does need `aria-label`. I just missed it here. Looking [here](https://helios.hashicorp.design/components/form/text-input?tab=code#formtextinputbase) at the HDS docs for base text input, we are responsible for making sure it is added. Thanks for catching. Fixed now.

ui/desktop/app/controllers/scopes/scope/projects/targets/index.js
cameronperera: Updated. A lot is going on so comments should be added to each action so we are fully aware of what is going on.

ui/desktop/electron-app/src/services/client-daemon-manager.js
cameronperera: Answer for learning: The BE team updated this API response to nest `socket_address` inside of `item` to match their other API responses.

ui/desktop/app/controllers/scopes/scope/projects/targets/index.js
cameronperera: Great suggestion. Updated!

ui/desktop/app/controllers/scopes/scope/projects/targets/index.js
cameronperera: Moved out.

ZedLi (approved):

DhariniJeeva (approved):

addons/api/addon/utils/mqlQuery.js
```diff
@@ -0,0 +1,128 @@
...
+    generateMQLFilterExpression(filters),
+  ].filter((item) => item);
+
+  return and(expressions);
+}
+
+const isEmpty = (input) =>
+  input === null || input === undefined || input === '';
+
+// Escape any double quotes or backspaces
```
lisbet-alvarez, line 101:
```suggestion
// Escape any double quotes or backslashes
```

ZedLi: haha woops good catch!

lisbet-alvarez (approved):
✨ Awesome work!

PR #2000, opened 2023-11-20
chore: 🤖 Update stylelint config dependency

dvonanderson:
✅ Closes: https://hashicorp.atlassian.net/browse/ICU-11818

:tickets: [Jira ticket](https://hashicorp.atlassian.net/browse/ICU-11818)

## Description
- Updated `styelint-config-standard-scss` to fix incorrect peer dependency warning for `styelint`
- Upgrading to `11.0.0`  introduced a new rule for writing media ranges for media queries.  Instead of using `prefix` style  notation(`max-width: 668px`), the new configuration wants us to use `context` style notation(`width <= 668px`)
- Tested the new style of notation by adding a white background to the `.ember-notify-cn` attribute and it works as expected.

#### Warning
Screenshot 2023-11-20 at 11 28 53

#### Test for sizes under `668px`
https://github.com/hashicorp/boundary-ui/assets/24277002/cf035cd6-cbc4-4759-a9cb-6cd3f2d8454c

#### Test for sizes above `668px`
https://github.com/hashicorp/boundary-ui/assets/24277002/df11308e-e256-4889-8413-b3f45ca4736a

## How to Test
- [ ] Start up desktop  
- [ ] Click on sessions
- [ ] Click the "Cancel" action on any active session and make sure that the notification looks correctly for the action and the styles applied in the media range

## Checklist:

### DIFF
--- a/ui/admin/app/styles/_notify.scss
+++ b/ui/admin/app/styles/_notify.scss
@@ -86,7 +86,7 @@
   }
 }
 
-@media only screen and (max-width: 680px) {
+@media only screen and (width <= 680px) {
   .ember-notify-cn {
     box-sizing: border-box;
     padding-left: 20px;
--- a/ui/desktop/app/styles/_notify.scss
+++ b/ui/desktop/app/styles/_notify.scss
@@ -90,7 +90,6 @@
 
 @media only screen and (width <= 680px) {
   .ember-notify-cn {
-    background-color: #fff;
     box-sizing: border-box;
     padding-left: 20px;
     width: 100%;
--- a/ui/desktop/app/styles/_notify.scss
+++ b/ui/desktop/app/styles/_notify.scss
@@ -88,8 +88,9 @@
   }
 }
 
-@media only screen and (max-width: 680px) {
+@media only screen and (width <= 680px) {
   .ember-notify-cn {
+    background-color: #fff;
     box-sizing: border-box;
     padding-left: 20px;
     width: 100%;
--- a/addons/rose/app/styles/rose/components/_message.scss
+++ b/addons/rose/app/styles/rose/components/_message.scss
@@ -16,7 +16,7 @@
   width: sizing.rems(l) * 15; // 16 rems/320 px
   margin-bottom: sizing.rems(l);
 
-  @media (max-width: media.width(small)) {
+  @media (width <= (media.width(small))) {
     width: auto;
   }
 
--- a/addons/rose/app/styles/rose/components/dialog/_dialog.scss
+++ b/addons/rose/app/styles/rose/components/dialog/_dialog.scss
@@ -22,7 +22,8 @@ $xxxxs: sizing.rems(xxxxs); // 1
 
   border-radius: $xxxs;
   background-color: var(--subtle);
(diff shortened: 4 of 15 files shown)


ui/admin/app/styles/_notify.scss
```diff
@@ -86,7 +86,7 @@
   }
 }
 
-@media only screen and (max-width: 680px) {
+@media only screen and (width <= 680px) {
```
DhariniJeeva, line 89: what about other places where we use sizing variables? `(max-width: media.width(small)` Do we need to take care of these as well?

dvonanderson: @DhariniJeeva that's a great point.  Those didn't show up on linting, but the pattern should be the same .  Let me do some testing

calcaide: To try answer @DhariniJeeva question, here my 2 cents: (@dvonanderson please, let me know if I am missing something)

- The change from `max-widdth: X` to `width <= X` it just applies to media-queries. Not sure the motivation behind this change, perhaps linter suggestion. I do personally think, is easier to understand now.
- The change, responds to a new syntax introduced within Media queries Level 4, [here the documentation](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_media_queries/Using_media_queries#syntax_improvements_in_level_4).
- This change, should NOT be applied outside of `@media-queries`.

ui/desktop/app/styles/_notify.scss
```diff
@@ -88,8 +88,9 @@
   }
 }
 
-@media only screen and (max-width: 680px) {
+@media only screen and (width <= 680px) {
   .ember-notify-cn {
+    background-color: #fff;
```
calcaide, line 93: None blocking comments:
- What is the purpose of adding white as background color?
- Shouldn't we use a HDS color token rather than hex color value? perhaps `foreground-high-contrats` [color palette link](https://helios.hashicorp.design/foundations/colors?tab=palette).

dvonanderson: @calcaide totally forgot to remove that, just doing some testing

ui/admin/app/styles/_notify.scss
calcaide: After a chat with @dvonanderson and as I mentioned above, Yes, we have other `@media` queries (the ones mentioned by @DhariniJeeva ) that need to be change. Devon is working on it and will provide those changes within this PR 😉

dvonanderson: @calcaide @DhariniJeeva just pushed up new commit that has all `@media` queries switched over to use `prefix` notation.  Used a red background to test.

https://github.com/hashicorp/boundary-ui/assets/24277002/0d573314-96d5-43ce-860f-3cd35d38595c

ZedLi (approved):

lisbet-alvarez (approved):
🙌 tested locally, looks great! ✨

cameronperera (approved):

DhariniJeeva (approved):
Looks great! ty

calcaide (approved):

PR #2001, opened 2023-11-20
chore: 🤖 Add custom key handler for windows

calcaide:
:tickets: [Jira ticket](https://hashicorp.atlassian.net/browse/ICU-11474)

## Description

Add custom key handler only for Windows OS so we achieve Ctrl+C to copy within the embedded terminal.

Used `attachCustomKeyEventHandler` from xterm.js. Info [here](https://github.com/xtermjs/xterm.js/blob/4375b3cdb9e36cd2746ff9a101fbeb48f7d0b794/typings/xterm.d.ts#L473-L482) and [here](https://xtermjs.org/docs/guides/hooks/)

## Screenshots:

Selecting text + (ctrl+c):
Screenshot 2023-11-20 152608

Ctrl+v:
Screenshot 2023-11-20 152618

## How to Test

- Run desktop client within windows machine.
- Stablish an ssh or tcp session with a target.
- Go to the shell tab and: select text + (ctrl+c) + (ctrl+v).
- Expected result: the selected text is pasted on the terminal.

## Checklist:
- ~~[ ] I have added JSON response output for API changes~~
- ~~[ ] My changes generate no new warnings~~

### DIFF
--- a/ui/desktop/app/components/session/tabs/index.js
+++ b/ui/desktop/app/components/session/tabs/index.js
@@ -61,11 +61,11 @@ export default class SessionTerminalTabsComponent extends Component {
      * Custom key event handler for windows to achieve Ctrl+C as on the other OS's
      */
     if (isWindows) {
-      xterm.attachCustomKeyEventHandler((arg) => {
+      xterm.attachCustomKeyEventHandler(async (arg) => {
         if (arg.ctrlKey && arg.code === 'KeyC' && arg.type === 'keydown') {
           const selection = xterm.getSelection();
           if (selection) {
-            navigator.clipboard.writeText(selection);
+            await navigator.clipboard.writeText(selection);
             return false;
           }
         }
--- a/ui/desktop/app/components/session/tabs/index.js
+++ b/ui/desktop/app/components/session/tabs/index.js
@@ -50,11 +50,28 @@ export default class SessionTerminalTabsComponent extends Component {
       return;
     }
 
+    const { isWindows } = await this.ipc.invoke('checkOS');
     const xterm = new Terminal(terminalOptions);
     this.terminal = xterm;
     const fitAddon = new FitAddon();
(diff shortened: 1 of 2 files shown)


ui/desktop/electron-app/src/ipc/handlers.js
```diff
@@ -117,6 +117,15 @@ handle('closeWindow', () => app.quit());
  */
 handle('checkCommand', async (command) => which(command, { nothrow: true }));
 
+/**
```
calcaide, line 120: All the changes on this file, except this code block has been added automatically for linting tools.

ui/desktop/app/components/session/tabs/index.js
```diff
@@ -50,12 +50,29 @@ export default class SessionTerminalTabsComponent extends Component {
...
 
+    /**
+     * Custom key event handler for windows to achieve Ctrl+C as on the other OS's
+     */
+    if (isWindows) {
+      xterm.attachCustomKeyEventHandler((arg) => {
+        if (arg.ctrlKey && arg.code === 'KeyC' && arg.type === 'keydown') {
+          const selection = xterm.getSelection();
+          if (selection) {
+            navigator.clipboard.writeText(selection);
```
lisbet-alvarez, line 68: non-blocking question for learning: Since `writeText` returns a promise should we use `await`OR in this case do we not use `await` since `writeText` is inside an event handler OR do we not want to use `await` since we are handling a user action and we want to return immediately?

calcaide: @lisbet-alvarez Good catch! Didn't though about handling the promise since I expected this handler to return immediately (so writing on the clipboard should be instantly). But thinking twice, I think, is a good idea, just in case it takes longer than expected to write on the clipboard.

ZedLi (approved):
Looks good!

PR #2002, opened 2023-11-21
Icu 11825 replace `store.findAll` method with `store.query`

lisbet-alvarez:
:tickets: [Jira ticket](https://hashicorp.atlassian.net/browse/ICU-11825)

## Description

Refactor `store.findAll` to use `store.query`. Most list endpoints use store.query to retrieve `items` array. The only resources that use `findAll` are scopes. Scopes resource is also paginated so instead of modifying the adapter findAll method to handle pagination we decided to remove it and use query instead. The reason is that findAll is being deprecated by Ember v5. 

## How to Test

Initiate a boundary dev instance using the `llb-list-pagination` branch and make sure no new errors or warnings are generated. 

## Checklist:

### DIFF
--- a/addons/api/tests/unit/adapters/application-test.js
+++ b/addons/api/tests/unit/adapters/application-test.js
@@ -67,7 +67,7 @@ module('Unit | Adapter | application', function (hooks) {
       'user',
       '1',
       mockSnapshot,
-      'findRecord'
+      'findRecord',
     );
     assert.strictEqual(findRecordURL, '/v1/users/1:my-custom-method');
     const findAllURL = adapter.buildURL('user', null, mockSnapshot, 'findAll');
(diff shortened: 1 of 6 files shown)


addons/api/tests/unit/adapters/application-test.js
```diff
@@ -67,7 +67,7 @@ module('Unit | Adapter | application', function (hooks) {
       'user',
```
lisbet-alvarez, line 67: All the comma additions are due to linting.

cameronperera (approved):

dvonanderson (approved):

ZedLi (approved):
Looks good!

Just as an FYI for anyone: technically `findAll` and `query` have [slightly different](https://github.com/emberjs/ember.js/issues/15256#issuecomment-302894768) results. Query doesn't return a live record array but that shouldn't matter since we don't depend on that.

To get exactly `findAll` functionality, it's [recommended](https://stackoverflow.com/a/73594666) to call a `peekAll` after as well, but we shouldn't need that.

PR #2004, opened 2023-11-22
chore: 🤖 handle removed_ids from pagionation api

DhariniJeeva:
## Description

PR to handle removed_ids before returning the result set.

## How to Test

## Checklist:

### DIFF
--- a/addons/api/addon/adapters/application.js
+++ b/addons/api/addon/adapters/application.js
@@ -124,6 +124,10 @@ export default class ApplicationAdapter extends RESTAdapter.extend(
       result = await super.query(store, schema, query);
       //add the result items to a data array
       if (result && result.items) {
+        //filter out the removed_ids before returning the result set
+        if (result.removed_ids) {
+          result = result.items.filter((i) => !result.removed_ids.includes(i));
+        }
         data.push(...result.items);
         //pass in the refresh token for subsequent calls to fetch the remaining list items
         query.refresh_token = result.refresh_token;
--- a/addons/api/addon/adapters/application.js
+++ b/addons/api/addon/adapters/application.js
@@ -139,7 +139,7 @@ export default class ApplicationAdapter extends RESTAdapter.extend(
           break;
         }
       }
-    } while (result && result.response_type === 'complete');
+    } while (result && result.response_type === 'delta');
 
     result.items = data;
     return prenormalizeArrayResponse(result);
--- a/addons/api/addon/adapters/application.js
+++ b/addons/api/addon/adapters/application.js
@@ -124,9 +124,9 @@ export default class ApplicationAdapter extends RESTAdapter.extend(
       result = await super.query(store, schema, query);
       //add the result items to a data array
       if (result && result.items) {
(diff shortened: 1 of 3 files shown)


addons/api/addon/adapters/application.js
```diff
@@ -124,9 +124,16 @@ export default class ApplicationAdapter extends RESTAdapter.extend(
       result = await super.query(store, schema, query);
       //add the result items to a data array
       if (result && result.items) {
+        //filter out the removed_ids before returning the result set
+        //This is due to items being removed from the previous iteration
+        //while the user is still paginating
```
ZedLi, line 129:
```suggestion
        // Filter out the removed_ids before returning the result set.
        // This is due to items being removed from the previous iteration
        // while the user is still paginating.
```

addons/api/addon/adapters/application.js
```diff
@@ -124,9 +124,16 @@ export default class ApplicationAdapter extends RESTAdapter.extend(
...
+        if (result.removed_ids) {
+          data = data.items.filter((i) => !result.removed_ids.includes(i));
+        }
         data.push(...result.items);
         //pass in the refresh token for subsequent calls to fetch the remaining list items
         query.refresh_token = result.refresh_token;
+
         //break the loop as soon the response_type becomes complete
         if (result.response_type === 'complete') {
```
ZedLi, line 138: Not related to this PR but why is this `break` needed? Isn't it already handled by your while loop?

addons/api/addon/adapters/application.js
ZedLi, line 1: I seem to be getting this error while testing it, did you ever see this?

image

addons/api/addon/adapters/application.js
```diff
@@ -124,9 +124,16 @@ export default class ApplicationAdapter extends RESTAdapter.extend(
       result = await super.query(store, schema, query);
       //add the result items to a data array
       if (result && result.items) {
+        //filter out the removed_ids before returning the result set
+        //This is due to items being removed from the previous iteration
+        //while the user is still paginating
+        if (result.removed_ids) {
+          data = data.items.filter((i) => !result.removed_ids.includes(i));
```
ZedLi, line 131: `data.items` doesn't exist yet.

addons/api/addon/adapters/application.js
ZedLi, line 1: It seems like the API also returns the removed IDs for each subsequent call, so we would be trying to remove an already removed item every call.

We should collate all the removed IDs in a set and then remove them from the final results at the end.

addons/api/addon/adapters/application.js
DhariniJeeva: addressed [here](https://github.com/hashicorp/boundary-ui/pull/2032)

DhariniJeeva:
Closing this as the API is changed and the FE doesn't need to handle`removed_ids` anymore as they are only returned when the pagination is complete and when the user clicks on the initial query again, so by then, we would only have valid resources that aren't deleted.

PR #2007, opened 2023-11-22
feat: 🎸 Allow prompts in oidc auth method form

DhariniJeeva:
:tickets: [Jira ticket](https://hashicorp.atlassian.net/browse/ICU-11834)

## Description

This PR allows users to select a prompt with their OIDC auth method. There is no default choice; it is optional, and the user can select 0 or more options. 

 

Screenshot 2023-11-27 at 1 35 36 PM

## How to Test

## Checklist:

- ~[] I have added steps to reproduce and test for bug fixes in the description~


addons/core/translations/form/en-us.yaml
```diff
@@ -119,6 +119,9 @@ to_claim:
 signing_algorithms:
   label: Signing Algorithms
   help: The signing algorithms allowed for the auth method.
+prompts:
+  label: Prompts
+  help: The prompts allowed for an oidc auth method
```
calcaide, line 124: none blocking and very picky comment 😅 . I just notive the other `help` values have a . at the end. Can we add that here too?

calcaide:
I was manually testing this, with the console open and I am getting at uncaught exception within updatePrompt.

How to reproduce:
- Checkout this branch and run the Admin UI.
- Go to Auth methods > new > OIDC.
- When in the creation form, open console and play with Promps.
- You should see an uncaught exception per event when selecting different promps.


My thoughts about the issue:
- I think the issue is that `previousSelection` is undefined when creating a form for first time, so we are trying to `.filter()` over an undefined var.
- Perhaps we should account for that case, i.e: `const previousSelection = this.args.model.prompts || [];`
- Please shareyour thoughts about it, thank you

calcaide (changes requested):
Nice job!! this is looking great!! I left a comment regarding an issue I found, but seems minor 😉

DhariniJeeva:
> I was manually testing this, with the console open and I am getting at `uncaught exception` within `updatePrompt`.
> 
> How to reproduce:
> 
> * Checkout this branch and run the Admin UI.
> * Go to Auth methods > new > OIDC.
> * When in the creation form, open browser console and play with Promps.
> * You should see an uncaught exception per event when selecting different promps.
> 
> My thoughts about the issue:
> 
> * I think the issue is that `previousSelection` is undefined when creating a form for first time, so we are trying to `.filter()` over an undefined var.
> * Perhaps we should account for that case, i.e: `const previousSelection = this.args.model.prompts || [];`
> * Please shareyour thoughts about it, thank you

thanks for the catch, fixed

ui/admin/app/components/form/auth-method/oidc/index.js
```diff
@@ -27,4 +30,48 @@ export default class FormAuthMethodOidcComponent extends Component {
...
+   * @param {object} event
+   */
+  @action
+  updatePrompt(option, event) {
+    const currentSelection = option.split(',').map((value) => ({ value }));
+
+    const previousSelection = this.args.model.prompts || [];
+
+    //when the checkbox is unchecked, remove the option form the exisiting list
+    const removeSelection = previousSelection.filter((i) => i.value !== option);
```
calcaide, line 61: None blocking comment: totally personal opinion, and I recognize a picky comment, that's why I mention is none blocking and feel free to ignore it if you disagree 😉 But I would like the var `i` is a bit more descriptive, perhaps `item`?

calcaide (approved):
LGTM!! Thank you for this work, looking good on the UI 😄 🙌

joshuaogle:
This looks good. Correct me if I'm reading this wrong, but it sounds like these prompts are specifically talking about what to do on their return visit, right? I wasn't able to find much info on the prompts in our docs, are we able to add helper text for the options here?  So something like:
• `None`: Skip prompts if the user is already signed in and has granted consent
• `Consent`: Always require the user to explicitly consent to scopes requested
• `Login`: Always require the user to log in to this auth provider
• `Select Account`: Allow the user to select from multiple accounts from this provider

ui/admin/app/components/form/auth-method/oidc/index.js
ZedLi, line 61: Why is this needed before you've even checked if the checkbox is checked? I think I understand why it's needed but reading this initially it was quite confusing based on the name

ui/admin/app/components/form/auth-method/oidc/index.js
```diff
@@ -27,4 +30,48 @@ export default class FormAuthMethodOidcComponent extends Component {
...
+          ...previousSelection,
+          ...currentSelection,
+        ]);
+      } else {
+        set(this.args.model, 'prompts', [...currentSelection]);
+      }
+    } else {
+      set(this.args.model, 'prompts', [...removeSelection]);
+    }
+  }
```
ZedLi, line 76: Couldn't this have been simplified by doing an if check on `checked` and if it's true, adding it to the model and if it's false, removing it from the model?

Do we also need to use ember `set`? I know we've used it previously but is it needed anymore?

DhariniJeeva: set seemed like the way to add the value of a property to the model,  What would we use in place of `set`?

ui/admin/app/components/form/auth-method/oidc/index.js
DhariniJeeva: moving this to the else condition for better readability

ui/admin/app/components/form/auth-method/oidc/index.js
```diff
@@ -27,4 +30,48 @@ export default class FormAuthMethodOidcComponent extends Component {
...
+  updatePrompt(option, event) {
+    const currentSelection = option.split(',').map((value) => ({ value }));
+
+    const previousSelection = this.args.model.prompts || [];
+
+    //when the checkbox is unchecked, remove the option form the exisiting list
+    const removeSelection = previousSelection.filter((i) => i.value !== option);
+
+    //add the selected option to the model, only when the checkbox is in checked state
+    if (event.target.id && event.target.checked) {
```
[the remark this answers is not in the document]
DhariniJeeva: not needed, I think I initially thought it was needed to make sure the ID has a checked property.

ui/admin/app/components/form/auth-method/oidc/index.js
ZedLi: Can't you just set it directly? i.e. `this.args.model.prompts = `

ui/admin/app/components/form/auth-method/oidc/index.js
```diff
@@ -27,4 +30,48 @@ export default class FormAuthMethodOidcComponent extends Component {
...
+
+  //actions
+
+  /**
+   * @param {string} option
+   * @param {object} event
+   */
+  @action
+  updatePrompt(option, event) {
+    const currentSelection = option.split(',').map((value) => ({ value }));
```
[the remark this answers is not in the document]
DhariniJeeva: I used it to add the string value to an object, but now I simplified this to [{value}]

ui/admin/app/components/form/auth-method/oidc/index.js
```diff
@@ -27,4 +29,41 @@ export default class FormAuthMethodOidcComponent extends Component {
...
+  @action
+  updatePrompt(value, event) {
+    const currentSelection = [{ value }];
+    const previousSelection = this.args.model.prompts || [];
+
+    //add the selected option to the model, only when the checkbox is in checked state
+    if (event.target.checked) {
+      this.args.model.prompts = [...previousSelection, ...currentSelection];
+    } else {
+      //when the checkbox is unchecked, remove the option form the exisiting list
```
ZedLi, line 62:
```suggestion
    const currentSelection = { value };
    const previousSelection = this.args.model.prompts || [];

    // Add the selected option to the model, only when the checkbox is in checked state
    if (event.target.checked) {
      this.args.model.prompts = [...previousSelection, currentSelection];
    } else {
      // When the checkbox is unchecked, remove the option form the exisiting list
```

ui/admin/app/components/form/auth-method/oidc/index.js
```diff
@@ -27,4 +29,41 @@ export default class FormAuthMethodOidcComponent extends Component {
...
+   */
+  prompts = options.oidc.prompts;
+
+  @tracked selectedPrompts = this.parsePromptsArray();
+
+  /**
+   * @returns {string}
+   */
+  parsePromptsArray() {
+    return this.args.model.prompts.map((item) => item.value) ?? [];
```
ZedLi, line 44:
```suggestion
    return this.args.model?.prompts.map((item) => item.value) ?? [];
```

ZedLi (commented):
Looks good but we should also add/update some tests for this new field

ZedLi (approved):

PR #2009, opened 2023-11-23
chore: 🤖 Update DC main element padding

calcaide:
✅ Closes: https://hashicorp.atlassian.net/browse/ICU-10497

:tickets: [Jira ticket](https://hashicorp.atlassian.net/browse/ICU-10497)

## Description

Update Desktop Client main element padding to be closer to HDS.

## Screenshots:

Before the padding change:
before

After the padding change:
after

## Checklist:

- ~~[ ] I have added JSON response output for API changes~~
- ~~[ ] I have added steps to reproduce and test for bug fixes in the description~~
- ~~[ ] I have added tests that prove my fix is effective or that my feature works~~


calcaide:
@joshuaogle Adding you to review this PR since it is a design change and I will appreciate your feedback here 😄 Thank you!!

ui/desktop/app/styles/app.scss
```diff
@@ -37,6 +37,11 @@
     overflow: auto;
 
     main {
+      /** This overrides the main defined within Rose.
+       * HDS has this set to 1.5rem top/bottom 3rem left/right. HDS assumes the usage of breadcrumbs
+       * but the DC have no breadcrumbs, so we will keep it at 3rem all around.
```
ZedLi, line 42: What is meant by `HDS` has set this to `1.5rem`? What's this referring to?

calcaide: @ZedLi Sorry, I might need to re-work that comment and improve my explanation.

HDS with app-frame adds a top/bottom padding of 1.5rem and 3rem on the side. The 1.5rem top padding instead of 3rem is to leave that extra 1.5rem space for the breadcrumbs of the UI's that use this component.
Desktop Client does not use app-frame (but it will in the future). Since currently we do not use the app-frame but want to give a look like we are, we want to add same padding as HDS with app-frame which will be 1.5rem but desktop client does not have breadcrumbs, so instead we define it at 3rem and looks like HDS and gives a feel of consistency within our UI's.

@cameronperera explain it better [within the ticket](https://hashicorp.atlassian.net/browse/ICU-10497).

We tfind out this during connection workflow work and we though it will be a very easy thing to provide consistency until we tackle a refactor to use HDS app-frame. But happy to discuss if you think otherwise.

cameronperera: Well stated @calcaide. That is exactly right. This was just something we noticed when working on post connection workflow changes and thought it would move the desktop client closer to matching HDS styling.

cameronperera (approved):

ui/desktop/app/styles/app.scss
ZedLi: Do you have a link to these styles? I can't seem to find them

calcaide: The app-frame component docs [are here](https://helios.hashicorp.design/layouts/app-frame).

Also, sorry for the confussion, the padding is applied within `cloud-ui/addons/core` but it within app-frame, here is the code [that adds the padding](https://github.com/hashicorp/cloud-ui/blob/main/addons/core/addon/styles/components/core/layout/main/index.scss#L17)

ZedLi: I'm still confused, that looks to be cloud's implementation of their own layout. I also see no recommendation for padding for the app frame HDS component nor any CSS that we need to override in their styles. 

Is this us trying to make it look more like cloud? Is there an actual recommendation from the design team for this change when we say "matching HDS styling"?

calcaide:
Sorry for the confusion @ZedLi .

You are correct. There is no prescription on padding/margin when using appFrame within HDS documentation.

The only usage we saw was within HCP (cloud-ui). There is no recommendation from anyone related to design (@cameronperera please, correct me if I am wrong), this is something Cameron mentioned while we were working on connection workflow and I personally agreed that will be nice to have the same look as cloud ui, specially, on the desktop client feels the padding was too small (personal opinion).

Since this change has a very low cost, we decided to proceed with it adding @joshuaogle  to the PR to get design feedback.

I do not have a strong opinion and I am totally fine closing this PR and not merging this, there is no prescription neither design asking us for this changes, was something proactive and totally assuming this might not get merged.

If you still confused or want to keep this discussion going on, feel free to open a slack thread or bring it to tech time 😃 I apologize for not providing a better explanation for this change.

joshuaogle (approved):
Thank you for making this change. So much nicer!

You're right, there is no requirement for the padding, but I've been using increased padding like this in future designs and it helps readability quite a bit

calcaide:
> Looks good, thanks for the explanation!
> 
> My last question is should this only be for desktop client only or is it also applicable for admin UI as well?

@ZedLi that's a good point. I will bring that question to design and follow up if need with ticket creation and execution.

PR #2012, opened 2023-11-27
feat: 🎸 Create `time-remaining` helper

dvonanderson:
✅ Closes: https://hashicorp.atlassian.net/browse/ICU-10519

## Description
- Create `time-remaining` helper to track remaining session time.

### View of time remaining:
https://github.com/hashicorp/boundary-ui/assets/24277002/9a729457-c2e0-4118-a922-1c8dac54b5f0

### View of seconds remaining:
https://github.com/hashicorp/boundary-ui/assets/24277002/aff8f1ed-b806-4a0c-aeae-f19c49584f7b

## How to Test
- [ ] Start up Boundary Dev instance
- [ ] Use `ENABLE_MIRAGE=false yarn start:desktop` to start desktop UI
- [ ] Add the `{{time-remaining @model.expirationTime}}` to this [session.hbs ](https://github.com/hashicorp/boundary-ui/blob/a891f7a1a5f5d3adb139a2d8600bfa035579196b/ui/desktop/app/templates/scopes/scope/projects/sessions/session.hbs#L13)file

## Checklist:

~~- [] I have added steps to reproduce and test for bug fixes in the description~~


ui/desktop/app/helpers/time-remaining.js
```diff
@@ -0,0 +1,44 @@
+import Helper from '@ember/component/helper';
```
dvonanderson, line 1: @calcaide @cameronperera should this helper be located in `core` or is it ok in the `ui/desktop` location?

ui/desktop/app/helpers/time-remaining.js
```diff
@@ -0,0 +1,44 @@
...
+import { inject as service } from '@ember/service';
+
+export default class extends Helper {
+  // =services
+  @service clockTick;
+  @service intl;
```
dvonanderson, line 7: @calcaide @cameronperera do we need to use translations the strings hours, minutes, and seconds?

ui/desktop/app/helpers/time-remaining.js
```diff
@@ -0,0 +1,44 @@
...
+    if (hours > 0) timeRemaining += `${hours}:`;
+    if (minutes > 0) timeRemaining += `${minutes}:`;
+    if (hours === 0 && minutes === 0) {
+      if (seconds > 0) timeRemaining += `${seconds} seconds remaining`;
+    } else {
+      timeRemaining += `${seconds} remaining`;
+    }
+
+    // Return the time remaining
+    return timeRemaining;
```
ZedLi, line 42: I don't think we should be manually formatting this. We already use `intl-unofficial-duration-unit-format`

ui/desktop/app/helpers/time-remaining.js
ZedLi: This is true but we've also had quite a few date/time formatters already in core only used in one UI so we should be consistent

ui/desktop/app/helpers/time-remaining.js
calcaide: Indeed with @ZedLi comment, we should not manually format this. Take a look at old time duration helper and follow the same pattern when formating, [here the code](https://github.com/hashicorp/boundary-ui/blob/main/addons/core/addon/helpers/format-time-duration.js#L22-L28) it uses `intl-unofficial-duration-unit-format`

ui/desktop/app/helpers/time-remaining.js
```diff
@@ -0,0 +1,44 @@
...
+  @service intl;
+
+  compute([expiration_time]) {
+    if (!expiration_time) return;
+
+    // Parse the expiration time
+    const expirationTime = new Date(expiration_time).getTime();
+
+    // Get the curent time
+    const currentTime = this.clockTick.now;
```
calcaide, line 16: I don't think we should take the current time but use the `created_time` we receive within the payload. I will try to explain better:

- Once a session is created we receive the session payload, wich includes `created_time` and `expiration_time`.
- The remainig time of a session is not: current time - expiration time. I think it is: creation time - expiration time.
- I do think the creation time does not need to be necessary the current time. For instance, a user can create a session, do not cancel it, navigate around the desktop client and later on came back to the session detail, then, if we are using current time the remaining time will be incorrect.

Let me know your thoughts on this one or if you want to discuss further

calcaide (changes requested):
Left some comments to discuss. Let me know if you want to discuss any of those on slack/zoom or they are clear 😉

ui/desktop/app/helpers/time-remaining.js
```diff
@@ -0,0 +1,44 @@
+import Helper from '@ember/component/helper';
+import { inject as service } from '@ember/service';
+
+export default class extends Helper {
+  // =services
+  @service clockTick;
+  @service intl;
+
+  compute([expiration_time]) {
```
calcaide, line 9: For [my comment below](https://github.com/hashicorp/boundary-ui/pull/2012/files#r1406864854) I do think, is best if this helper receives: `creation_time` and `expiration_time` and do the math with these two.

Also, this approach will give make this helper reusable everywhere we need to give time remaining, as long you pass the created time and the expiration time.

ui/desktop/app/helpers/time-remaining.js
calcaide: I do agree with @ZedLi 

It is true that this helper, right now is use JUST on desktop. But a time-remaning helper, if we do it pure, so it receives 2 params, and returns the subtract of them, can be reusable everywhere.

Also, @ZedLi  mentions we have time related helpers within core, even tho we just use them in Admin.

I second @ZedLi  comment, for consistency’s sake we should have it within core.

ui/desktop/app/helpers/time-remaining.js
calcaide: I deleted the previous comment. I was totally wrong.

Because expiration time is a timestamp to get remaining time your implementation is totally correct. remaining time = current time - expiration time.

ui/desktop/app/helpers/time-remaining.js
cameronperera: I agree with that reasoning. It should be in core with the other time related helpers. Thanks!

dvonanderson: Moved the time-remaining helper to `addons/core`

ui/desktop/app/helpers/time-remaining.js
dvonanderson: Translations have been added using the `intl` service

ui/desktop/app/helpers/time-remaining.js
```diff
@@ -0,0 +1,44 @@
...
+    if (!expiration_time) return;
+
+    // Parse the expiration time
+    const expirationTime = new Date(expiration_time).getTime();
```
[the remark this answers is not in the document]
dvonanderson: @ZedLi thank you for that recommendation.  It worked out perfectly

ui/desktop/app/helpers/time-remaining.js
dvonanderson: Manual format has been updated, and now the helper is using `DurationUnitFormat` to format the output.

calcaide:
@dvonanderson After the PR update, we are missing the helper code itself

addons/core/addon/helpers/time-remaining.js
```diff
@@ -0,0 +1,40 @@
...
+
+    // Calculate the difference in milleseconds
+    const difference = expirationTime - currentTime;
+
+    // Calculate the difference in seconds
+    const differenceInSeconds = Math.floor(difference / 1000);
+
+    // If the difference is negative, return 0:0:00 remaining
+    if (differenceInSeconds <= 0)
+      return `0:0:00 ${this.intl.t('resources.session.remaining')}`;
```
calcaide, line 27: I am not sure we need a return here. How about if differenceInSeconds <= 0 we do differenceInSeconds = 0 ? then, your implementation below will format correctly `0:00:00` so we avoid manual formatting? thoughts?

calcaide: Perhaps line 23 can be something like: `const differenceInSeconds = (difference > 0) ? Math.floor(difference / 1000) : 0;` and then we avoid line 26 and 27 ?

dvonanderson: @calcaide awesome that works.  

https://github.com/hashicorp/boundary-ui/assets/24277002/801d98c1-4c8f-42b5-bc40-4a331a0e0cd4

calcaide (changes requested):
One test of the helper is failing

calcaide (changes requested):
PR still having a failing test

calcaide (approved):
LGTM!! all tests are passing! the helper looks ready to go 😃

cameronperera (approved):
LGTM. Thanks for this work @dvonanderson!

PR #2016, opened 2023-11-29
feat: 🎸 Add pagination support to client daemon handler

ZedLi:
✅ Closes: https://hashicorp.atlassian.net/browse/ICU-11743

## Description
Add support for pagination from the client daemon handler.

## Checklist:

### DIFF
--- a/addons/api/addon/handlers/client-daemon-handler.js
+++ b/addons/api/addon/handlers/client-daemon-handler.js
@@ -1,5 +1,6 @@
 import { pluralize } from 'ember-inflector';
 import { generateMQLExpression } from '../utils/mqlQuery';
+import ArrayProxy from '@ember/array/proxy';
 
 /**
  * Not all types are yet supported by the client daemon so we'll
@@ -21,7 +22,7 @@ const ClientDaemonHandler = {
     switch (context.request.op) {
       case 'query': {
         const { store, data } = context.request;
-        const { type, query } = data;
+        const { type, query, options: { pushToStore = true } = {} } = data;
         const isClientDaemonRunning = await this.ipc.invoke(
           'isClientDaemonRunning',
         );
@@ -36,7 +37,7 @@ const ClientDaemonHandler = {
         // TODO: Remove usages of recursive from callers and scope_id since
         //  all calls to daemon are recursive, scope_id can be part of search query instead
         // eslint-disable-next-line no-unused-vars
-        let { recursive, scope_id, ...remainingQuery } = query;
+        let { recursive, scope_id, page, pageSize, ...remainingQuery } = query;
         let searchQuery = '';
         if (remainingQuery.query) {
           const { search, filters } = remainingQuery.query;
@@ -63,7 +64,7 @@ const ClientDaemonHandler = {
         const [results] = Object.values(
           await this.ipc.invoke('searchClientDaemon', remainingQuery),
         );
-        const payload = { items: results ?? [] };
+        const payload = { items: paginateResults(results, page, pageSize) };
 
         const schema = store.modelFor(type);
         const serializer = store.serializerFor(type);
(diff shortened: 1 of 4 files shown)


addons/api/addon/handlers/client-daemon-handler.js
```diff
@@ -21,7 +22,7 @@ const ClientDaemonHandler = {
     switch (context.request.op) {
       case 'query': {
         const { store, data } = context.request;
-        const { type, query } = data;
+        const { type, query, options: { pushToStore = true } = {} } = data;
```
ZedLi, line 25: This is just an extra escape hatch I added for scenarios where we want to retrieve data without pushing to the store.

addons/api/addon/handlers/client-daemon-handler.js
```diff
@@ -75,14 +76,48 @@ const ClientDaemonHandler = {
           'query',
         );
 
-        // TODO: Add some pagination
+        // Return the raw data if we don't push to the store.
+        let records = normalizedPayload.data;
```
ZedLi, line 80: This unfortunately means we wouldn't be returning models, do we feel OK with that?

addons/api/addon/handlers/client-daemon-handler.js
```diff
@@ -75,14 +76,48 @@ const ClientDaemonHandler = {
           'query',
         );
 
-        // TODO: Add some pagination
+        // Return the raw data if we don't push to the store.
```
ZedLi, line 79: This unfortunately means we wouldn't be returning models, do we feel OK with that?

cameronperera: I don't exactly understand why we need the ArrayProxy here. Could you explain that further please?

ZedLi: Is this related to my original comment about returning models or `ArrayProxy` only? Do you mean why do we need `ArrayProxy` at all?

cameronperera: Yes, why do we need it at all. I read the code comments below and still wasn't exactly sure.
Also not a fan of not getting the model back, however if we know that going in then it's not a huge deal.

ZedLi: It's to return a meta object, basically to imitate [this](https://guides.emberjs.com/release/models/handling-metadata/) so we have a way to return the total number of items in the result set when normally it would just be an array of the paginated data.

cameronperera (approved):

addons/api/addon/handlers/client-daemon-handler.js
cameronperera: Thanks! That page was helpful.

ZedLi:
> One more question that isn't blocking. In cases where we do push the data to the store, we would still be able to fetch a singular target from the store when visiting a specific target correct? So the list of targets from the client daemon would be a list of POJOs but when visiting a single target we would be able to retrieve that from the store.

Hmm I'm not exactly sure I'm understanding your question but visiting a specific resource should not be any different as it is currently since our handler only looks at `query` requests, not `findRecord` ones. It will just do the normal `findRecord` logic which depends on how we call it if it returns immediately from the store. It will always make the network request if needed and will never go through the daemon. 

 > So the list of targets from the client daemon would be a list of POJOs
I just want to make sure we're on the same page: the client daemon returns models back as if it were from the API currently, just paginated (which means real materialized ember models). The _only_ time we will get back POJOs is if we use the escape hatch to not push data into the ember data store.

addons/api/addon/handlers/client-daemon-handler.js
```diff
@@ -75,14 +76,48 @@ const ClientDaemonHandler = {
...
+ * number of results to return to the caller.
+ *
+ * @param array
+ * @param page
+ * @param pageSize
+ * @returns {[*]}
+ */
+const paginateResults = (array, page, pageSize) => {
+  const length = array?.length;
+  if (!array || length === 0) {
```
DhariniJeeva, line 109: do we need to check both array and array length here? can this be just `if (length === 0) { return []}`

ui/desktop/app/routes/scopes/scope/projects/targets/index.js
```diff
@@ -37,20 +37,21 @@ export default class ScopesScopeProjectsTargetsIndexRoute extends Route {
   /**
    * Loads all targets under current scope.
```
DhariniJeeva, line 38: Can this comment be updated to reflect the new returns?

addons/api/addon/handlers/client-daemon-handler.js
ZedLi: Yeah, array could be undefined/null which I want to catch

ui/desktop/app/routes/scopes/scope/projects/targets/index.js
ZedLi: I already had updated the `@return`, are you referring to something else?

DhariniJeeva: `Loads all targets under current scope` should we mention the total items being returned here?

PR #2018, opened 2023-11-29
feat: 🎸 Implement time-remaining helper into desktop UI

dvonanderson:
✅ Closes: https://hashicorp.atlassian.net/browse/ICU-10487

:tickets: [Jira ticket](https://hashicorp.atlassian.net/browse/ICU-10487)

## Description
- Implement `time-remainig` helper to display duration of a session for desktop app.
-  Moved the header elements into a `div` to enforce current styling for items.
- 

https://github.com/hashicorp/boundary-ui/assets/24277002/78ca5ffd-f8a9-4ee3-8177-e98852662d3e

## How to Test
---
#### PreCondition for user with all session permissions
- [ ] Spin up a Boundary dev instance and start the desktop application using `ENABLE_MIRAGE=false yarn start:desktop`
- [ ] Default permissions that come with default binary dev instance will allow list, read capabilities of a session and if you open developer console and look at network tab on the desktop app there will be no `403 forbidden`  responses
---
#### Testing user with all session permissions
- [ ] Click on "Targets" in the sidebar navigation and click "Connect" on any of generated targets from the Boundary dev instance targets.
- [ ] User gets redirected to "session details"  page.
- [ ] Next to the ID of the session in the header, you should see the time-remaining helper counting down
- [ ] After the connect workflow, click back on "Sessions" in the sidebar navigation and click the active session and you should see the time remaining helper, with state of the helper persisting and time is continuously going down. 

---
#### PreCondition for user with list session list permissions
- [ ] Spin up a Boundary dev instance and start the desktop application using `ENABLE_MIRAGE=false yarn start:desktop`
- [ ] Create a user and provide them with grants with the ability to authorize-session and the permission to only list sessions and not read them.
#### Grants used:
- `id=*;type=session;actions=list,cancel:self`
- `id=*;type=target;actions=list,authorize-session`
---

### Testing user with just list permissions
- [ ] Click on "Targets" in the sidebar navigation and click "Connect" on any of generated targets from the Boundary dev instance targets.
- [ ] User gets redirected to "session details"  page.
- [ ] Next to the ID of the session in the header, you should see the time-remaining helper counting down
- [ ] If you clicked on the target with host sources, user should not see the host details on the content sidebar
- [ ] Users with list permissions will still be able to connect to session, but if you open a your dev console their should be a 403 forbidden error response in the network tab of your console.
- [ ] After the connect workflow, click back on "Sessions" in the sidebar navigation and click the active session and you should see the time remaining helper, with state of the helper persisting and time is continuously going down. 

## Checklist:

### DIFF
--- a/ui/desktop/app/controllers/scopes/scope/projects/targets/index.js
+++ b/ui/desktop/app/controllers/scopes/scope/projects/targets/index.js
@@ -60,7 +60,8 @@ export default class ScopesScopeProjectsTargetsIndexController extends Controlle
 
     // Associate the connection details with the session
     let session;
-    const { session_id, address, port, credentials } = connectionDetails;
+    const { session_id, address, port, credentials, expiration_time } =
+      connectionDetails;
     try {
       session = await this.store.findRecord('session', session_id);
     } catch (error) {
@@ -76,9 +77,11 @@ export default class ScopesScopeProjectsTargetsIndexController extends Controlle
             proxy_address: address,
             proxy_port: port,
             target_id: target.id,
+            expiration_time,
           },
         ],
       });
+
       session = this.store.peekRecord('session', session_id);
     }
 
--- a/ui/desktop/app/controllers/scopes/scope/projects/targets/index.js
+++ b/ui/desktop/app/controllers/scopes/scope/projects/targets/index.js
@@ -60,8 +60,13 @@ export default class ScopesScopeProjectsTargetsIndexController extends Controlle
 
     // Associate the connection details with the session
     let session;
-    const { session_id, address, port, credentials, expiration_time } =
-      connectionDetails;
+    const {
+      session_id,
+      address,
+      port,
+      credentials,
+      expiration_time: expiration,
+    } = connectionDetails;
     try {
       session = await this.store.findRecord('session', session_id);
     } catch (error) {
@@ -77,7 +82,7 @@ export default class ScopesScopeProjectsTargetsIndexController extends Controlle
             proxy_address: address,
             proxy_port: port,
             target_id: target.id,
-            expiration_time,
+            expiration_time: expiration,
           },
         ],
       });
--- a/ui/desktop/app/controllers/scopes/scope/projects/targets/index.js
+++ b/ui/desktop/app/controllers/scopes/scope/projects/targets/index.js
@@ -60,13 +60,8 @@ export default class ScopesScopeProjectsTargetsIndexController extends Controlle
 
     // Associate the connection details with the session
     let session;
-    const {
-      session_id,
-      address,
-      port,
-      credentials,
-      expiration: expiration_time,
-    } = connectionDetails;
+    const { session_id, address, port, credentials, expiration } =
+      connectionDetails;
     try {
       session = await this.store.findRecord('session', session_id);
     } catch (error) {
--- a/ui/desktop/app/controllers/scopes/scope/projects/targets/index.js
+++ b/ui/desktop/app/controllers/scopes/scope/projects/targets/index.js
@@ -65,7 +65,7 @@ export default class ScopesScopeProjectsTargetsIndexController extends Controlle
       address,
       port,
       credentials,
-      expiration_time: expiration,
+      expiration: expiration_time,
     } = connectionDetails;
     try {
       session = await this.store.findRecord('session', session_id);
--- a/ui/desktop/app/styles/app.scss
+++ b/ui/desktop/app/styles/app.scss
@@ -505,14 +505,7 @@ $chevron-down: url("data:image/svg+xml;utf-8,<svg viewBox='0 0 24 24' fill='%231
 .session-details-info {
   display: flex;
   align-items: center;
-
-  > * {
-    margin-right: 1rem;
-  }
-
-  .hds-badge {
-    margin-left: 1rem;
-  }
+  gap: 1rem;
 }
 
 .target-details-header {
--- a/ui/desktop/app/styles/app.scss
+++ b/ui/desktop/app/styles/app.scss
@@ -502,15 +502,16 @@ $chevron-down: url("data:image/svg+xml;utf-8,<svg viewBox='0 0 24 24' fill='%231
   }
 }
 
-.session-details-header {
-  &,
-  h1 {
-    display: flex;
-    align-items: center;
+.session-details-info {
+  display: flex;
+  align-items: center;
 
-    > * {
-      margin-right: 1rem;
-    }
+  > * {
+    margin-right: 1rem;
+  }
+
+  .hds-badge {
+    margin-left: 1rem;
   }
 }
 
--- a/ui/desktop/app/styles/app.scss
+++ b/ui/desktop/app/styles/app.scss
@@ -502,7 +502,18 @@ $chevron-down: url("data:image/svg+xml;utf-8,<svg viewBox='0 0 24 24' fill='%231
   }
 }
 
-.session-details-header,
+.session-details-header {
+  &,
+  h1 {
+    display: flex;
+    align-items: center;
+
+    > * {
+      margin-right: 1rem;
+    }
+  }
+}
+
 .target-details-header {
   display: flex;
   align-items: center;
--- a/ui/desktop/app/templates/scopes/scope/projects/sessions/session.hbs
+++ b/ui/desktop/app/templates/scopes/scope/projects/sessions/session.hbs
@@ -8,10 +8,16 @@
 <Rose::Layout::Page as |page|>
 
   <page.header>
-    <h1 class='session-details-header'>
-      <Hds::IconTile @color='boundary' @icon='terminal' />
-      {{@model.displayName}}
-    </h1>
+    <div class='session-details-header'>
+      <h1>
+        <Hds::IconTile @color='boundary' @icon='terminal' />
+        {{@model.displayName}}
+      </h1>
+      <Hds::Badge
+        @text={{time-remaining @model.expiration_time}}
+        @icon='loading'
+      />
+    </div>
   </page.header>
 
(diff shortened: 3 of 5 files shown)


ui/desktop/app/controllers/scopes/scope/projects/targets/index.js
```diff
@@ -60,7 +60,8 @@ export default class ScopesScopeProjectsTargetsIndexController extends Controlle
 
     // Associate the connection details with the session
     let session;
-    const { session_id, address, port, credentials } = connectionDetails;
+    const { session_id, address, port, credentials, expiration_time } =
```
ZedLi, line 63: The expiration field returned from a connection is actually called `expiration`

ZedLi (changes requested):

ui/desktop/app/controllers/scopes/scope/projects/targets/index.js
dvonanderson: @ZedL can you elaborate a little further, so when we destructor those details in this file, are we getting those details from the `connect` method in the `ipc`.  In there the field is called `expiration_time`

ZedLi: Can you try connecting and seeing what JSON you get back from the binary?

ui/desktop/app/templates/scopes/scope/projects/sessions/session.hbs
```diff
@@ -8,10 +8,16 @@
 <Rose::Layout::Page as |page|>
 
   <page.header>
-    <h1 class='session-details-header'>
-      <Hds::IconTile @color='boundary' @icon='terminal' />
-      {{@model.displayName}}
-    </h1>
+    <div class='session-details-header'>
```
ZedLi, line 11: Do you need this extra div?

ui/desktop/app/styles/app.scss
```diff
@@ -502,7 +502,18 @@ $chevron-down: url("data:image/svg+xml;utf-8,<svg viewBox='0 0 24 24' fill='%231
   }
 }
 
-.session-details-header,
+.session-details-header {
```
ZedLi, line 505: Can we consolidate these styes to a more generic class name and just apply the class multiple times rather than using `&, h1` since they're exactly the same styles

ZedLi (changes requested):

ui/desktop/app/controllers/scopes/scope/projects/targets/index.js
cameronperera: I just looked into what the binary returns from the `connect` command and it is `expiration`. We will have to update the destructor here and assign it as `expiration_time: expiration` for the push to the store.

cameronperera (changes requested):

calcaide:
Within the manual testing procedure, we are missing the case when a user has NO session read permissions but can perform a connection. We MUST test this, since is one of the edge cases directly affected by the connect method.

ui/desktop/app/controllers/scopes/scope/projects/targets/index.js
calcaide: That is correct, the binary returns `expiration` object. This should mimic that behaviour.

ui/desktop/app/styles/app.scss
```diff
@@ -502,7 +502,19 @@ $chevron-down: url("data:image/svg+xml;utf-8,<svg viewBox='0 0 24 24' fill='%231
   }
 }
 
-.session-details-header,
+.session-details-info {
+  display: flex;
+  align-items: center;
+
+  > * {
+    margin-right: 1rem;
```
calcaide, line 510: Do we need `margin-right` if the badge has `margin-left`?

ui/desktop/app/templates/scopes/scope/projects/sessions/session.hbs
```diff
@@ -8,9 +8,13 @@
 <Rose::Layout::Page as |page|>
 
   <page.header>
-    <h1 class='session-details-header'>
+    <h1 class='session-details-info'>
       <Hds::IconTile @color='boundary' @icon='terminal' />
       {{@model.displayName}}
```
calcaide, line 13: `{{@model.displayName}}` is a text that is not wrapped into any HTML element. Since here we are using the `H1` as a container for: badge + text + time remaining helper I would suggest to wrap `{{@model.displayName}}` within a `span`.

Then within the styles, on line 509: `> * { margin-right: 1rem }` you are givin margin-right to every element, then line 513, 514 and 515 are not need. 

If you have a container `h1` in this case, and elements within the container, wrapping elements makes easier to controll them.

ui/desktop/app/styles/app.scss
```diff
@@ -502,7 +502,19 @@ $chevron-down: url("data:image/svg+xml;utf-8,<svg viewBox='0 0 24 24' fill='%231
...
+  }
+
+  .hds-badge {
```
calcaide, line 513: As [per this comment](https://github.com/hashicorp/boundary-ui/pull/2018/files#r1414485581), if you apply the approch described there, you do not need to specify a `margin-left` for `hds-badge`

DhariniJeeva (approved):
looks good!

ui/desktop/app/controllers/scopes/scope/projects/targets/index.js
```diff
@@ -60,7 +60,13 @@ export default class ScopesScopeProjectsTargetsIndexController extends Controlle
 
     // Associate the connection details with the session
     let session;
-    const { session_id, address, port, credentials } = connectionDetails;
+    const {
+      session_id,
+      address,
+      port,
+      credentials,
+      expiration_time: expiration,
```
ZedLi, line 68:
```suggestion
      expiration: expiration_time,
```

I think you have it mixed up

ui/desktop/app/templates/scopes/scope/projects/sessions/session.hbs
ZedLi: Should we be using the new [text](https://helios.hashicorp.design/components/text) HDS components now? I think we mentioned them in our last tech time

ui/desktop/app/controllers/scopes/scope/projects/targets/index.js
```diff
@@ -60,7 +60,13 @@ export default class ScopesScopeProjectsTargetsIndexController extends Controlle
 
     // Associate the connection details with the session
     let session;
-    const { session_id, address, port, credentials } = connectionDetails;
+    const {
+      session_id,
+      address,
+      port,
+      credentials,
+      expiration: expiration_time,
```
calcaide, line 68: The `connectionDetails` are returned from the CLI ,desktop client leverages CLI to perform a connection, [here the code](https://github.com/hashicorp/boundary-ui/blob/main/ui/desktop/electron-app/src/models/session.js#L54-L66). 
Therefore, if you look at `connectionDetails` you are not receiving `expiration_time`, you are receiving `expiration`. You can test this also connecting to a target using CLI.


That said, What are we actually trying here? we are defining `expiration: expiration_time` at destructuring time but then assigning `expiration_time: expiration`? 
- This is redundant, you could have done `expiration_time` at destructuring and assigning time.
- That is making the linter to not pass.
- As described above, the assignments are incorrect.

ui/desktop/app/controllers/scopes/scope/projects/targets/index.js
```diff
@@ -76,9 +82,11 @@ export default class ScopesScopeProjectsTargetsIndexController extends Controlle
             proxy_address: address,
             proxy_port: port,
             target_id: target.id,
+            expiration_time: expiration,
```
calcaide, line 85: [If you fix the comment above](https://github.com/hashicorp/boundary-ui/pull/2018/files#r1417747186), this should be OK, Since from the CLI we are receiving `expiration`, but as you can see, here we are pushing to our store (UI) and there, we have model `expiration` as `expiration_time`.

ui/desktop/app/templates/scopes/scope/projects/sessions/session.hbs
```diff
@@ -8,10 +8,16 @@
 <Rose::Layout::Page as |page|>
 
   <page.header>
-    <h1 class='session-details-header'>
+    <span class='session-details-info'>
```
calcaide, line 11: FYI a `<span>` is not allow to be the parent of a `<h1>`, [more info about span](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/span)

Can we use the `<page.header>` as container? what other patterns are we following in our UI's to achieve this use case?

ui/desktop/app/templates/scopes/scope/projects/sessions/session.hbs
```diff
@@ -8,10 +8,16 @@
...
       <Hds::IconTile @color='boundary' @icon='terminal' />
-      {{@model.displayName}}
-    </h1>
+      <Hds::Text::Display @tag='h1' @size='500'>
```
calcaide, line 13: Did you check with Joshua if this is our desired size?

calcaide (changes requested):
Within how to test (within PR description) I am missing 2 very important tests:
- Opening an active session. The current test just covers target -> connect -> check the redirection to session detail has the time remaining but I am missing: sessions -> click any active session -> check the session detail screen has the time remaining.
- The 2 tests mentioned above connect to a target -> redirect to session detail + opening an active session, have important permissions related to this work. I am missing a test procedure for a user with NO session read permissions. Can you please provide a test procedure for this case? If you have questions about this last one, let's open a thread on #boundary-ui 😉

cameronperera (approved):
LGTM

calcaide (approved):

PR #2021, opened 2023-11-30
feat: 🎸 Do not stop client daemon if it was already started

ZedLi:
✅ Closes: https://hashicorp.atlassian.net/browse/ICU-11932

## Description
This PR is to handle the scenario where the client daemon was already running when the desktop client is started. We don't want to stop the client daemon if we the desktop client wasn't the one to start up the client daemon.

## How to Test
Start the daemon before running the DC and then close the desktop client. Confirm that the daemon is still running. Likewise, if the daemon wasn't started beforehand, confirm that the daemon is stopped once the DC is closed.

## Checklist:


cameronperera (approved):

lisbet-alvarez (approved):
🙌

PR #2022, opened 2023-11-30
Icu 11802 filter targets by project scope

cameronperera:
:tickets: [Jira ticket](https://hashicorp.atlassian.net/browse/ICU-11802)

https://github.com/hashicorp/boundary-ui/assets/29386339/fe12c878-2e9b-461a-bce4-b96b9ada00ac

Added handling for the fact that orgs and projects can not have names.
Screenshot 2023-12-06 at 3 04 37 PM

## How to Test

Build a binary from the llb-client-daemon branch and use that binary in your local dev environment. You will need to move that binary into the ui/desktop/electron-app/cli folder as well. Be sure to start the desktop client using this command: ENABLE_MIRAGE=false BYPASS_CLI_SETUP=true yarn start:desktop

To test this change, you will need to create some new projects. I would recommend creating a project that has no targets in it to verify that it does not show up. You should only see projects that are included in the list of targets.

## Checklist:


ui/desktop/app/routes/scopes/scope/projects/targets/index.js
```diff
@@ -52,6 +62,14 @@ export default class ScopesScopeProjectsTargetsIndexRoute extends Route {
       this.can.can('connect target', target),
     );
 
-    return { targets: targets, totalItems };
+    // Query all targets for defining filtering values
+    let allTargets = await this.store.query('target', { query: null });
```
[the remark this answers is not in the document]
cameronperera: Oh right. I want to use the options with `pushToStore` set to false correct?

ui/desktop/app/routes/scopes/scope/projects/targets/index.js
```diff
@@ -52,6 +62,15 @@ export default class ScopesScopeProjectsTargetsIndexRoute extends Route {
       this.can.can('connect target', target),
     );
 
-    return { targets: targets, totalItems };
+    // Query all targets for defining filtering values
+    const options = { pushToStore: true };
```
ZedLi, line 66: Shouldn't this be false?

cameronperera: It for sure should be false. Thanks for catching that. Fixing now.

ui/desktop/app/controllers/scopes/scope/projects/targets/index.js
```diff
@@ -44,6 +46,31 @@ export default class ScopesScopeProjectsTargetsIndexController extends Controlle
...
+      targetScopeIds.includes(project.id),
+    );
+
+    if (this.scopeSearchTerm) {
+      availableScopes = availableScopes.filter((item) => {
+        const isNameMatch = item.name
+          .toLowerCase()
+          .includes(this.scopeSearchTerm);
+        const isIdMatch =
+          item.id.toLowerCase() === this.scopeSearchTerm.toLowerCase();
```
cameronperera, line 67: I decided to use an exact match here for project Id. Since we only see project name in the dropdown, I figured a partial match could be confusing on the off chance that two or more projects share the same search term. However, I wanted to include the project Id as a searchable attribute just in case someone were to copy the idea out of admin and try to paste it into the desktop client. Thoughts on this?

ui/desktop/app/templates/scopes/scope/projects/targets/index.hbs
```diff
@@ -23,24 +23,51 @@
...
             </DD.Header>
-            {{#each this.searchItems as |item|}}
-              <DD.Checkbox @value={{item}} data-test-checkbox={{item}}>
-                {{item}}
-              </DD.Checkbox>
-            {{/each}}
+            {{#unless this.availableScopes}}
+              <DD.Description @text={{t 'titles.no-results'}} />
+            {{/unless}}
+            {{#let (group-by this.availableScopes 'scopeModel') as |orgs|}}
```
ZedLi, line 45: It's probably OK for scopes but I just want to mention we probably need to handle any resource that can be unbounded like the discussion of [targets](https://hashicorp.slack.com/archives/C0647GX0TLJ/p1701197068542169?thread_ts=1701196245.008619&cid=C0647GX0TLJ) in this thread initially.

We most likely need to cap the display just in case

ui/desktop/app/routes/scopes/scope/projects/targets/index.js
```diff
@@ -35,14 +38,21 @@ export default class ScopesScopeProjectsTargetsIndexRoute extends Route {
...
-   * Loads all targets and the number of targets under current scope.
+   * Loads queried targets, the number of targets under current scope, and
+   * projects for filtering options.
    *
-   * @returns {Promise<{totalItems: number, targets: [TargetModel]}>}
+   * @returns {Promise<{totalItems: number, targets: [TargetModel], projects: [ScopeModel]}>}
    */
-  async model({ search = '' }) {
+  async model({ search = '', scopes }) {
     // TODO: Filter targets by scope we're in manually
```
ZedLi, line 47: Will this be a separate PR?

ui/desktop/app/routes/scopes/scope/projects/targets/index.js
```diff
@@ -23,6 +23,9 @@ export default class ScopesScopeProjectsTargetsIndexRoute extends Route {
       refreshModel: true,
       replace: true,
     },
+    scopes: {
+      refreshModel: true,
```
ZedLi, line 27: We should include `replace: true` here too

ui/desktop/app/controllers/scopes/scope/projects/targets/index.js
ZedLi, line 1: I think we really do need to extract this filter out to a component and try to genericize it. It already seems to explode the amount of logic in the controller just for one filter and this will be the same in sessions

ui/desktop/app/controllers/scopes/scope/projects/targets/index.js
```diff
@@ -21,15 +21,36 @@ export default class ScopesScopeProjectsTargetsIndexController extends Controlle
...
+  @tracked scopeSearchTerm;
+  @tracked scopes = [];
+  @tracked selectedScopes = [];
   @tracked page = 1;
   @tracked pageSize = 10;
-  @tracked searchItems = ['Project 1', 'Project 2', 'Project 3'];
+
+  constructor() {
+    super(...arguments);
+    this.getAllTargets();
```
ZedLi, line 36: I don't think this can be done here, aren't controllers singletons?

ui/desktop/app/controllers/scopes/scope/projects/targets/index.js
```diff
@@ -21,15 +21,36 @@ export default class ScopesScopeProjectsTargetsIndexController extends Controlle
...
+  }
 
   // =methods
 
+  async getAllTargets() {
+    // Query all targets for defining filtering values
+    const options = { pushToStore: false };
+    let allTargets = await this.store.query('target', { query: null }, options);
```
DhariniJeeva, line 44: Why are we passing null? What happens when we don't pass anything to query?

ui/desktop/app/controllers/scopes/scope/projects/targets/index.js
```diff
@@ -21,15 +21,36 @@ export default class ScopesScopeProjectsTargetsIndexController extends Controlle
...
+
+    // Filter out targets to which users do not have the connect ability
+    allTargets = allTargets.filter((target) =>
+      target.authorized_actions.includes('authorize-session'),
```
DhariniJeeva, line 48: can this be combined with the query method? 

`const options = { pushToStore: false, authorized_actions: ['authorize-session'] };`

ui/desktop/app/routes/scopes/scope/projects/targets/index.js
cameronperera: Yes that was the plan. I wanted to tackle this after getting the scope filter to work.

ui/desktop/app/controllers/scopes/scope/projects/targets/index.js
```diff
@@ -139,4 +184,52 @@ export default class ScopesScopeProjectsTargetsIndexController extends Controlle
...
     this.page = 1;
   }
+
+  /**
+   * Sets scopeSearchTerm to be used for filtering availableScopes
+   * @param {object} event
+   */
+  @action
+  filterItems(event) {
+    let { value } = event.target;
```
DhariniJeeva, line 194:
```suggestion
    const { value } = event.target;
```

ui/desktop/app/controllers/scopes/scope/projects/targets/index.js
```diff
@@ -139,4 +184,52 @@ export default class ScopesScopeProjectsTargetsIndexController extends Controlle
...
+  }
+
+  /**
+   * Handles checkbox event changes for selectedScopes but does not
+   * update the scopes query param
+   * @param {object} event
+   */
+  @action
+  selectItem(event) {
+    let { checked, value } = event.target;
```
DhariniJeeva, line 205:
```suggestion
    const { checked, value } = event.target;
```

ui/desktop/app/controllers/scopes/scope/projects/targets/index.js
cameronperera: I agree. I know we talked about it before. Are you still good with doing this in a separate PR?

ui/desktop/app/controllers/scopes/scope/projects/targets/index.js
cameronperera: Updated to be just `{}`. Much nicer looking.

ui/desktop/app/controllers/scopes/scope/projects/targets/index.js
ZedLi: Not currently but you can ask backend to see if it's something that is possible to be searched on

cameronperera: I am not sure if we need to. I am only doing this because I don't believe I can use the `can` abilities without it being a TargetModel.

ui/desktop/app/controllers/scopes/scope/projects/targets/index.js
ZedLi: Yup, but for the future I think this is something that really should be part of the original PR so we can see the design

ui/desktop/app/templates/scopes/scope/projects/targets/index.hbs
```diff
@@ -22,24 +22,51 @@
...
             @text={{t 'resources.scope.title'}}
+            @count={{this.scopes.length}}
           />
           <DD.Header @hasDivider={{true}}>
             <Hds::Form::TextInput::Base
               @type='search'
+              @value={{this.scopeSearchTerm}}
               placeholder={{t 'actions.narrow-results'}}
               aria-label={{t 'actions.narrow-results'}}
+              {{on 'search' this.filterItems}}
```
ZedLi, line 37: Shouldn't this be on `input`?

cameronperera: This is the search input inside the dropdown.

ZedLi: Right, why isn't it on input as well? And I don't think [search](https://developer.mozilla.org/en-US/docs/Web/API/HTMLInputElement/search_event) can be used in production.

ui/desktop/app/controllers/scopes/scope/projects/targets/index.js
```diff
@@ -139,4 +184,52 @@ export default class ScopesScopeProjectsTargetsIndexController extends Controlle
     this.search = value;
     this.page = 1;
   }
+
+  /**
+   * Sets scopeSearchTerm to be used for filtering availableScopes
+   * @param {object} event
+   */
+  @action
```
ZedLi, line 192: We should also probably debounce this

ui/desktop/app/templates/scopes/scope/projects/targets/index.hbs
ZedLi: I actually take back that it's probably OK for scopes. I don't think we can ever assume what the user's usage will be and if it can be unbounded, we need to handle it.


https://github.com/hashicorp/boundary-ui/assets/5783847/b4b6a605-7d98-465f-8d04-240ab9a6b047

ui/desktop/app/controllers/scopes/scope/projects/targets/index.js
cameronperera: Removed. Moved fetch back to the model hook.

addons/rose/addon/components/rose/filterable-dropdown/index.js
```diff
@@ -0,0 +1,84 @@
...
+  @tracked selectedItems = [];
+  @tracked searchTerm;
+
+  // =methods
+
+  /**
+   * Takes itemOptions and filters them by searchTerm if there is one
+   * @returns {[object]}
+   */
+  get itemOptions() {
```
ZedLi, line 24: It doesn't matter too much but this shouldn't be under the methods section you labeled on line 18

ui/desktop/app/templates/scopes/scope/projects/targets/index.hbs
```diff
@@ -64,77 +74,79 @@
           }}
           @valign='middle'
         >
-          <:body as |B|>
-            <B.Tr>
-              <B.Td>
+          <:body as |b|>
```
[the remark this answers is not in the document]
cameronperera: I am confused as well. I looked around the HDS docs they have a mix of both. We should standardize this if we haven't already.

addons/rose/addon/components/rose/filterable-dropdown/index.js
cameronperera: Fixed.

ui/desktop/tests/acceptance/scopes-test.js
```diff
@@ -180,7 +180,16 @@ module('Acceptance | scopes', function (hooks) {
...
-    this.stubClientDaemonSearch('targets', 'targets', 'targets', 'targets');
+    this.stubClientDaemonSearch(
+      'targets',
+      'targets',
+      'targets',
+      'targets',
+      'targets',
+      'targets',
+      'targets',
+      'targets',
```
DhariniJeeva, line 191: Can you provide some context on why we do this?

ui/desktop/app/routes/scopes/scope/projects/targets/index.js
```diff
@@ -41,19 +45,25 @@ export default class ScopesScopeProjectsTargetsIndexRoute extends Route {
...
+   * Loads queried targets, the number of targets under current scope, and
+   * projects for filtering options.
    *
-   * @returns {Promise<{totalItems: number, targets: [TargetModel]}>}
+   * @returns {Promise<{totalItems: number, targets: [TargetModel], projects: [ScopeModel]}>, allTargets: [TargetModel]}
    */
-  async model({ search, page, pageSize }) {
+  async model({ search, scopes, page, pageSize }) {
     // TODO: Filter targets by scope we're in manually
     // const { id: scope_id } = this.modelFor('scopes.scope');
```
DhariniJeeva, line 55: is this needed for later?

addons/rose/tests/integration/components/rose/filterable-dropdown/index-test.js
```diff
@@ -0,0 +1,29 @@
...
+import { hbs } from 'ember-cli-htmlbars';
+
+module(
+  'Integration | Component | rose/filterable-dropdown/index',
+  function (hooks) {
+    setupRenderingTest(hooks);
+
+    test('it renders', async function (assert) {
+      // Set any properties with this.set('myProperty', 'value');
+      // Handle any actions with this.set('myAction', function(val) { ... });
```
DhariniJeeva, line 13: Can this be removed?

addons/rose/addon/components/rose/filterable-dropdown/index.js
```diff
@@ -0,0 +1,84 @@
...
+    let items = this.args.itemOptions;
+    if (this.searchTerm) {
+      items = this.args.itemOptions.filter((item) => {
+        const isNameMatch = item.name
```
DhariniJeeva, line 26: Not a blocker- to improve readability, you can do the lowerCase conversions outside the filter loop

ZedLi (approved):

ui/desktop/tests/acceptance/scopes-test.js
cameronperera: Absolutely. Each `'targets'` in this stub represents a query to the client daemon to fetch targets. In this specific tests, we query targets eight times.  Once when we do the initial visit on line 193. Then three more times with each click in the test. We could pull each click and assert into their own tests but it would still need four `'targets'` for each test. Does that help explain why this is needed?

ui/desktop/app/routes/scopes/scope/projects/targets/index.js
cameronperera: Yes. This is something we will be fixing in a different PR.

addons/rose/tests/integration/components/rose/filterable-dropdown/index-test.js
cameronperera: Yes.

addons/rose/addon/components/rose/filterable-dropdown/index.js
cameronperera: Good idea. I like that change.

DhariniJeeva (approved):
LGTM

ui/desktop/app/templates/scopes/scope/projects/targets/index.hbs
cameronperera: This was discussed over in meeting and will be addressed in a later PR to limit scope of those changes.

PR #2023, opened 2023-11-30
feat: 🎸 Add pagination component to targets

ZedLi:
✅ Closes: https://hashicorp.atlassian.net/browse/ICU-11905

## Description
This adds the pagination component for targets in the desktop client. I created a thin wrapper around the HDS pagination so for the `paginationQueryParams` as it should be the same for every usage we have.

Still working on having a test harness for our client daemon now.

https://github.com/hashicorp/boundary-ui/assets/5783847/6749bb35-097b-4716-a938-811ad36dcf20

## Checklist:

### DIFF
--- a/addons/api/addon-test-support/handlers/client-daemon-search.js
+++ b/addons/api/addon-test-support/handlers/client-daemon-search.js
@@ -1,0 +1,62 @@
+import sinon from 'sinon';
+
+/**
+ * This test helper can be used to help setup your sinon stubs in your tests.
+ * Must be called after setupMirage.
+ *
+ * ```js
+ * import setupStubs from 'api/test-support/handlers/client-daemon-search';
+ *
+ * module('Acceptance | my test', function(hooks) {
+ *  setupApplicationTest(hooks);
+ *  setupMirage(hooks);
+ *  setupStubs(hooks);
+ *
+ *   // add your actual tests here
+ * });
+ * ```
+ *
+ * This will setup a stub for the IPC service and include helper methods to
+ * help setup what the client daemon search command returns.
+ *
+ * @module Test Helpers
+ * @public
+ */
+export default function setupStubs(hooks) {
+  hooks.beforeEach(function () {
+    const ipcService = this.owner.lookup('service:ipc');
+    this.ipcStub = sinon.stub(ipcService, 'invoke');
+
+    /**
+     * We aim to still use mirage data when stubbing out the client daemon
+     * search. However, the handler expects the search data to be raw JSON from
+     * the API which will get serialized. We add back the scope object for it to
+     * get correctly normalized and other fields such as "attributes" from the
+     * API will already be hoisted and will not be re-normalized from mirage data.
+     *
+     * This convenience method can take in multiple types,
+     * e.g. this.stubClientDaemonSearch('targets', 'sessions').
+     * This will have sinon stub out each call so the first call to search will
+     * provide all the models from the first parameter you specify,
+     * the second call will provide all the models from the second argument, etc.
+     * @param types
+     */
+    this.stubClientDaemonSearch = (...types) => {
+      types.forEach((type, i) => {
+        this.ipcStub
+          .withArgs('searchClientDaemon')
+          .onCall(i)
+          .returns({
+            [type]: this.server.db[type].map((model) => ({
+              ...model,
+              scope: this.server.db.scopes.find(model.scopeId),
+            })),
+          });
+      });
+    };
+  });
+
+  hooks.afterEach(function () {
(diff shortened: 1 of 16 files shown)


addons/api/addon-test-support/handlers/client-daemon-search.js
```diff
@@ -0,0 +1,62 @@
...
+     * API will already be hoisted and will not be re-normalized from mirage data.
+     *
+     * This convenience method can take in multiple types,
+     * e.g. this.stubClientDaemonSearch('targets', 'sessions').
+     * This will have sinon stub out each call so the first call to search will
+     * provide all the models from the first parameter you specify,
+     * the second call will provide all the models from the second argument, etc.
+     * @param types
+     */
+    this.stubClientDaemonSearch = (...types) => {
```
ZedLi, line 44: This _should_ be enough for our testing but in the future if we need to customize which models need to be returned from mirage instead of all, I imagine we can provide callbacks instead of strings which should be easy to handle

ui/desktop/tests/acceptance/scopes-test.js
```diff
@@ -190,6 +180,7 @@ module('Acceptance | scopes', function (hooks) {
 
   test('can navigate among org scopes via header navigation', async function (assert) {
     assert.expect(3);
+    this.stubClientDaemonSearch('targets', 'targets', 'targets', 'targets');
```
cameronperera, line 183: From your comment, this will all for 4 calls that will each return targets to the client daemon. Will this search or filter the response according to query params if any are provided?

cameronperera (approved):
LGTM

ui/desktop/tests/acceptance/scopes-test.js
ZedLi: No this won't handle query params, you have to mock the data that should be returned from search/filtering. We're not aiming to test the query params since that's not in our control and we have to assume that the client daemon returns the correct data (though pagination will still be handled because that's not from the search call).

I only did this because search is called 4 times, I thought of maybe providing something to just let it mock all search calls rather than having to be explicit about how many times search will be called but I thought this was fine, let me know if you disagree.

addons/api/addon/handlers/client-daemon-handler.js
```diff
@@ -111,17 +110,20 @@ export default class ClientDaemonHandler {
         );
 
         // Return the raw data if we don't push to the store.
-        let records = normalizedPayload.data;
+        let records = normalizedPayload.data.map((record) => ({
```
ZedLi, line 113: I changed the data to be just the attributes now @cameronperera

addons/api/addon/handlers/client-daemon-handler.js
```diff
@@ -111,17 +110,20 @@ export default class ClientDaemonHandler {
...
-        return ArrayProxy.create({
-          content: records,
-          meta: { totalItems: results?.length ?? 0 },
-        });
+        // Set a meta property on the array to store the total items of the results.
+        // This isn't conventional but is better than returning an ArrayProxy
+        // or EmberArray since the ember store query method asserts it has to be an array
+        // so we can't just return an object.
+        records.meta = { totalItems: results?.length ?? 0 };
+        return records;
```
ZedLi, line 126: I changed the return type to not use an array proxy now since there's not much of a benefit to it versus just attaching it to an array directly besides the fact it can be a gotcha @lisbet-alvarez

cameronperera: So what exactly will the response look like now?

ZedLi: Should be used the same, it's just a normal array instead of a proxy array now. The data will look the same

ui/desktop/app/templates/scopes/scope/projects/targets/index.hbs
```diff
@@ -13,47 +13,45 @@
...
-      </div>
+          </DD.Header>
+          {{#each this.searchItems as |item|}}
+            <DD.Checkbox @value={{item}} data-test-checkbox={{item}}>
+              {{item}}
+            </DD.Checkbox>
+          {{/each}}
+        </S.Dropdown>
+        <S.Dropdown as |DD|>
+          <DD.ToggleButton @color='secondary' @text='Active session' />
```
DhariniJeeva, line 45: We should use translated string here and also on L50

ZedLi: These are the example filters that were previously added and will be replaced in future PRs

lisbet-alvarez (approved):
🙌

PR #2024, opened 2023-11-30
fix: 🐛 Fix add expiration_time when user has no grants to read

calcaide:
## Description

This fixes the issue mentioned within [this slack thread](https://hashicorp.slack.com/archives/C012GTH1C4E/p1701384535719469?thread_ts=1701372005.384369&cid=C012GTH1C4E).

The test checks that a user with NO permission to read a session, but with permission to connect can establish a connection to the target, then been redirect succesfully to Session detail page. Regardless the permissions, the user expect a minimum information on the session detail page.

Behaviour:
- User clicks connect to a specific target.
- When the user stablish the connection, we read important information of the session and aggregate that to the session store.
- Once the connection is stablished, user gets redirected to the session page.
- Because the user can not read the session, relies only on the important information we aggregate at connection time to redenr the session detail page.

Issue this PR solves:
- The test uses a `stub` but was not returning the `expiration_time`, so the store had `expiration_time: undefined`.
- The process of aggregating session data from the connection was not taking care of `expiration_time`. So in the edge case of a user with no permission to read the session, will never get `expiration_time` on the model.

### DIFF
--- a/ui/desktop/tests/acceptance/projects/sessions/session-test.js
+++ b/ui/desktop/tests/acceptance/projects/sessions/session-test.js
@@ -191,11 +191,11 @@ module('Acceptance | projects | sessions | session', function (hooks) {
       address: 'a_123',
       port: 'p_123',
       protocol: 'tcp',
+      expiration_time: '2022-02-05T09:55:39.216Z',
     });
 
     await visit(urls.target);
     await click(TARGET_CONNECT_BUTTON);
-
     assert.strictEqual(currentURL(), urls.session);
   });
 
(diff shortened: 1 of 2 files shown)


ui/desktop/tests/acceptance/projects/sessions/session-test.js
```diff
@@ -191,11 +191,11 @@ module('Acceptance | projects | sessions | session', function (hooks) {
       address: 'a_123',
       port: 'p_123',
       protocol: 'tcp',
+      expiration_time: '2022-02-05T09:55:39.216Z',
```
calcaide, line 194: @dvonanderson review this value, since I put the first timestamp I had handy.

dvonanderson (approved):
thank you carlos

PR #2026, opened 2023-12-01
refactor: 💡 Use a native class for the client daemon handler

ZedLi:
## Description
This refactors the client daemon handler to be a native class instead of just a simple function that is bound with the store's context. It turns out the handlers can take any object that just has a `request` function on it. This enables us to move the services to the class and have the directly injected with ember aware of their usage.

This fixes the issue where tests would error out when we try to use any service from within the handler (e.g. the IPC invoke methods) even when we try to mock the service out. This enables us to be able to properly mock out the client daemon searching. 

## Checklist:


cameronperera (approved):
LGTM

lisbet-alvarez (approved):
🙌

PR #2037, opened 2023-12-08
PoC use hds pageHeader

calcaide:
## Description

THIS PR WILL NOT BE MERGED it is just for demonstration and discussion purposes.
THIS changes have been checked with @joshuaogle .

Goal of this PR:
- Use `<Hds::PageHeader>` component, [documentation here](https://helios.hashicorp.design/components/page-header), within desktop client.
- Because we're still relying on `rose-layout` for Boundary UI's, we will keep using `<Rose::Layout::Page>` and its childs elements, such as `<page.header>` as containers, but overriding/cleaning up the rose styles of those elements to keep them as pure containers as possible.
- This PR provides an overrid of `rose-layout-age` JUST for desktop client. If we like this approach an decide to use `<Hds::PageHeader>` this override will be deleted from desktop and ported to Rose, as a stage 1 deprecation process. So all the `<page.header>` within our UI's are simple containers to wrap `<Hds::PageHeader>` until we use `appFrame` as base layout.

Next steps (if we like the idea and approach):
- I will create tasks to start using `<Hds::PageHeader>` within desktop.
- I will create tasks to start using `<Hds::PageHeader>` within Admin UI.
- I will create  a task to clean up all the `.rose-layout-page-header` within `rose` and perhaps (TBD) clean up/deprecate `<DocLink>` rose component. It will not be necessary.

## Screenshots:

Before the changes:
Screenshot 2023-12-08 at 12 01 05 PM

After:
Screenshot 2023-12-08 at 11 58 24 AM

### DIFF
--- a/ui/desktop/app/styles/app.scss
+++ b/ui/desktop/app/styles/app.scss
@@ -10,6 +10,17 @@
 @import 'rose';
 @import 'notify';
 
+// Overrides to use HDS appHeader. Once we stop relying on rose-layout page
+// for layouting, we can delete these styles.
+.rose-layout-page {
+  display: grid;
+  grid-template-areas:
+    'header header header'
+    'body body body';
+}
+
+// End of Rose override to fit HDS components /////////////
+
 .rose-layout-centered {
   > .rose-layout-page {
     width: sizing.rems(l) * 18;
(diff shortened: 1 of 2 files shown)


lisbet-alvarez (approved):
I like this approach! Thank you for making this PoC!

PR #2038, opened 2023-12-11
refactor: 💡 Refactor workers list to use `Hds::Table`

dvonanderson:
✅ Closes: https://hashicorp.atlassian.net/browse/ICU-11734

## Description
- Refactor workers list to use `Hds::Table` instead of `Rose::Table`.
- Moved `tags` into their own cell
- Removed background-color form `Rose::Toolbar` instead of adding additional custom margin to `Hds::Table`.  Referenced in the thread below
- Rearranged location of header and data for `workers.id` to the end of the table

### Before
---
Screenshot 2023-12-08 at 14 23 56

### After
---
Screenshot 2023-12-11 at 11 02 21

## How to Test
- [ ] Start up Admin UI with `yarn start`
- [ ] Click on "Workers" resource in the left sidebar navigation.

## Checklist:

### DIFF
--- a/addons/rose/app/styles/rose/components/toolbar/_toolbar.scss
+++ b/addons/rose/app/styles/rose/components/toolbar/_toolbar.scss
@@ -12,7 +12,6 @@ $xxxxs: sizing.rems(xxxxs); // 1
 
 .rose-toolbar {
   display: flex;
-  background-color: var(--ui-gray-subtler-6);
   padding: $xxs + $xxxxs;
 
   .rose-toolbar-primary,
--- a/ui/admin/app/templates/scopes/scope/workers/index.hbs
+++ b/ui/admin/app/templates/scopes/scope/workers/index.hbs
@@ -90,67 +90,61 @@
 
       </Rose::Toolbar>
 
-      <Rose::Table as |table|>
-        <table.header as |header|>
-          <header.row as |row|>
-            <row.headerCell>{{t 'resources.worker.title'}}</row.headerCell>
-            <row.headerCell>{{t 'form.id.label'}}</row.headerCell>
-            <row.headerCell>{{t
-                'resources.worker.table.session_count'
-              }}</row.headerCell>
-            <row.headerCell>{{t
-                'resources.worker.table.release_version'
-              }}</row.headerCell>
-            <row.headerCell>{{t
-                'resources.worker.table.ip_address'
-              }}</row.headerCell>
-          </header.row>
-        </table.header>
-        <table.body as |body|>
-          {{#each @model as |worker|}}
-            <body.row as |row|>
-              <row.headerCell>
-                {{#if (and (can 'read worker' worker) worker.isPki)}}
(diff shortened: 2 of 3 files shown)


addons/rose/app/styles/rose/components/toolbar/_toolbar.scss
```diff
@@ -12,7 +12,6 @@ $xxxxs: sizing.rems(xxxxs); // 1
 
 .rose-toolbar {
   display: flex;
-  background-color: var(--ui-gray-subtler-6);
```
DhariniJeeva, line 15: Are we ok with not having background color in other instances where we use `rose::toolbar`?

DhariniJeeva: That sounds good, can we add some screenshots of other instances which is affected by this change for reference?

calcaide: @dvonanderson Can you provide a link to the slack thread you mention please?

dvonanderson: @calcaide it's in the description

calcaide: Indeed with @DhariniJeeva comment. After getting the context of [the slack thread](https://hashicorp.slack.com/archives/C012GTH1C4E/p1702063578595999), seems like taking the background can be good enough until we refactor this toolbar.

That said, we already cover this exact use case within auth-methods, where we change the rose table for hds table, [here the template code ](https://github.com/hashicorp/boundary-ui/blob/main/ui/admin/app/templates/scopes/scope/auth-methods/index.hbs#L116-L127) and we add 1rem of margin, [here the css code](https://github.com/hashicorp/boundary-ui/blob/main/ui/admin/app/styles/app.scss#L742-L744).

I am not judging what solution is better. But the fact that we are implementing different solutions to the same problem. Can we choose one and keep it consistent?

If you ask me and IMHO one solutions isn't exclusive. Adding margin seems good to me, since what we want is to increase the distance between these 2 elements. Deleting the background make it look cleaner.

If this specific change is causing friction or raising doubts, I will suggest to stick the PR to rafctor the `HDS::Table` and bring the toolbar issue to tech time. Toolbar is used in different places of our UI, so can be good to dedicate a specific ticket/PR to "normalize" the toolbar.

dvonanderson: @DhariniJeeva @calcaide I think I like Carlo's recommendation of just using the existing margin for the `auth-methods` list.  I will make this part of tech-time to get some feedback.  Here is what it looks like for the `auth-method` list, clearly we could remove the `margin-top` that is set on it, but I think we should wait until those toolbar changes happen.

Screenshot 2023-12-11 at 20 31 47

calcaide (approved):

PR #2040, opened 2023-12-11
fix: 🐛 target cred sources association

DhariniJeeva:
https://hashicorp.atlassian.net/browse/ICU-12046
## Description

This PR fixes the error caused when the `add/remove credential sources` process was repeated.

Note: When we `return` within the `await.all`, the peekAll method on L53& L54 adds duplicate stale IDs from the previous iteration with no `type`, causing the error. 

## How to Test

Add an injected credential source to a target
Remove the credential source from the target
Repeat 1-2 more times
you should not see any error

## Checklist:

### DIFF
--- a/ui/admin/app/routes/scopes/scope/targets/target/add-brokered-credential-sources.js
+++ b/ui/admin/app/routes/scopes/scope/targets/target/add-brokered-credential-sources.js
@@ -40,11 +40,11 @@ export default class ScopesScopeTargetsTargetAddBrokeredCredentialSourcesRoute e
     await all(
       credentialStores.map(({ id: credential_store_id, isStatic }) => {
         if (isStatic) {
-          return this.store.query('credential', {
+          this.store.query('credential', {
             credential_store_id,
           });
         } else {
-          return this.store.query('credential-library', {
+          this.store.query('credential-library', {
             credential_store_id,
           });
         }
--- a/ui/admin/app/routes/scopes/scope/targets/target/add-injected-application-credential-sources.js
+++ b/ui/admin/app/routes/scopes/scope/targets/target/add-injected-application-credential-sources.js
@@ -40,11 +40,11 @@ export default class ScopesScopeTargetsTargetAddInjectedApplicationCredentialSou
     await all(
       credentialStores.map(({ id: credential_store_id, isStatic }) => {
         if (isStatic) {
-          return this.store.query('credential', {
+          this.store.query('credential', {
             credential_store_id,
           });
         } else {
-          return this.store.query('credential-library', {
+          this.store.query('credential-library', {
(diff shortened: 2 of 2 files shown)


lisbet-alvarez (approved):
🙌 LGTM!

ui/admin/app/routes/scopes/scope/targets/target/add-brokered-credential-sources.js
```diff
@@ -40,11 +40,11 @@ export default class ScopesScopeTargetsTargetAddBrokeredCredentialSourcesRoute e
     await all(
       credentialStores.map(({ id: credential_store_id, isStatic }) => {
         if (isStatic) {
-          return this.store.query('credential', {
+          this.store.query('credential', {
```
ZedLi, line 43: I'm slightly confused, if you don't return the query we're not getting promises back for the array on the `Promise.all` so what will it be awaiting on?

(Unrelated but do we need to use rsvp `all` instead of `Promise.all`?)

ZedLi (commented):
Why does this fix things? Why would the peekall return anything different since we're not using the actual return from the `all`? Is it perhaps just not waiting for the results?

DhariniJeeva:
> Why does this fix things? Why would the peekall return anything different since we're not using the actual return from the `all`? Is it perhaps just not waiting for the results?

yes, I believe it is due to not waiting for the results, I did not investigate much due to time constraints right now, so I followed the pattern we have in other places such as add-accounts, add host-sources

ui/admin/app/routes/scopes/scope/targets/target/add-brokered-credential-sources.js
DhariniJeeva: Update: After an offline investigation, we found out that this could be a side effect of the ember-data upgrade, but due to time constraints, we are going to merge this fix as is, which is consistent with other instances where peekAll() is used and doesn't throw an error. This will be added to the backlog for further investigation.

calcaide: None blocking comment: still confuse about the `await` Id o believe we need to `await` when we are using `store.query` to wait for the results?

calcaide (approved):
LGTM! tested manually with chrome and firefox on mac and I do not see any errors, thanks for taking care of this!

ui/admin/app/routes/scopes/scope/targets/target/add-brokered-credential-sources.js
DhariniJeeva: yes, we will need to await on the query method, sorry for the confusion regarding my comment.

PR #2042, opened 2023-12-13
Icu 11916 add flyout for session in prgoress field

lisbet-alvarez:
:tickets: [Jira ticket](https://hashicorp.atlassian.net/browse/ICU-11916)

## Description
Add sessions flyout to targets list view.
Populate flyout with table of all pending and active sessions of a particular target.

https://github.com/hashicorp/boundary-ui/assets/107949262/aac1524b-dbd2-48b7-984b-7419d7fa7d69

## How to Test

1. Use llb-pagination-2 be branch and start `boundary dev`
3. Checkout this branch and run `ENABLE_MIRAGE=false BYPASS_CLI_SETUP=true yarn start:desktop`

## Checklist:


cameronperera (approved):
Nice work!

ui/desktop/app/routes/scopes/scope/projects/sessions/session.js
```diff
@@ -20,7 +20,9 @@ export default class ScopesScopeProjectsSessionsSessionRoute extends Route {
    * @return {SessionModel}
    */
   async model({ session_id }) {
-    const session = await this.store.findRecord('session', session_id);
+    const session = await this.store.findRecord('session', session_id, {
+      reload: true,
```
ZedLi, line 24: This technically shouldn't have a forced reload here since it should be using the cached version if it exists due to the [logic](https://github.com/hashicorp/boundary-ui/blob/0487cbc808fa45e78c6d2e854175da4b8fe8c926/ui/desktop/app/controllers/scopes/scope/projects/targets/index.js#L129) we have in case a user doesn't have permissions to read a session.

We'd be making an extra call that would fail otherwise.

ui/desktop/app/routes/scopes/scope/projects/targets/index.js
```diff
@@ -91,6 +91,7 @@ export default class ScopesScopeProjectsTargetsIndexRoute extends Route {
             ],
           },
         },
+        force_refresh: true,
```
ZedLi, line 94: As an FYI, I'm going to be changing how we retrieve these sessions as it's probably worth just making one call to retrieve all sessions (and even load them into the store) rather than making separate calls for each target. I need to do this regardless for the active session filter

ui/desktop/app/templates/scopes/scope/projects/targets/index.hbs
```diff
@@ -108,10 +108,11 @@
               <B.Td>
                 {{#if B.data.sessions.length}}
                   <Hds::Button
+                    data-test-targets-sessions-flyout-button={{B.data.target.id}}
                     @text={{t 'actions.yes'}}
                     @color='tertiary'
                     @icon='info'
-                    {{on 'click' this.toggleSessionsFlyout}}
+                    {{on 'click' (fn this.selectTargetSessions B.data)}}
```
ZedLi, line 115: Why did we pass all the targets instead of just the target we needed here?

ZedLi (changes requested):

addons/api/addon-test-support/handlers/client-daemon-search.js
```diff
@@ -47,7 +47,7 @@ export default function setupStubs(hooks) {
           .withArgs('searchClientDaemon')
           .onCall(i)
           .returns({
-            [type]: this.server.db[type].map((model) => ({
+            [type]: this.server.db[type]?.map((model) => ({
```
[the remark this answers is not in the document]
lisbet-alvarez: Added this for tests since there's a scenario that I pass in an empty string in place of a `sessions` string to get back no results since the goal was to test canceling a session. Let me know if you know of a better way to do this. (ref ln 29 in index-test.js)

ui/desktop/app/controllers/scopes/scope/projects/targets/index.js
```diff
@@ -63,6 +66,16 @@ export default class ScopesScopeProjectsTargetsIndexController extends Controlle
     return availableScopes;
   }
 
+  /**
+   * Returns the target and sessions associated with that target
+   * @returns {object}
+   */
+  get selectedTargetWithSessions() {
```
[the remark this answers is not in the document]
lisbet-alvarez: oh good point, i'll update this accordingly.

ui/desktop/app/routes/scopes/scope/projects/sessions/session.js
lisbet-alvarez: Ah okay that makes sense, tho the only reason i added it was due to a test failing because a badge on the sessions detail page wasn't correctly displaying text (time remaining). I'll remove this.

ui/desktop/app/routes/scopes/scope/projects/targets/index.js
lisbet-alvarez: good call, that would simplify this.

ui/desktop/app/templates/scopes/scope/projects/targets/index.hbs
lisbet-alvarez: `B.data` is an object that includes a single target and associated sessions.

ui/desktop/app/routes/scopes/scope/projects/sessions/session.js
ZedLi: Hmm I wonder why the cached data doesn't have the proper expiration_time actually...

lisbet-alvarez: The weird thing is that I did a couple of `console.log`s to look into this and the model showed to have `expiration_time` defined so I am unsure why it's not displaying properly. It's also only an issue in tests 🤔

ZedLi: I bet it's something weird happening between the mirage data and when it gets normalized in the client daemon handler since without the reload we're just pulling the data from that when it gets pushed to the store...

ui/desktop/app/templates/scopes/scope/projects/targets/index.hbs
```diff
@@ -80,82 +80,86 @@
...
-                {{#if B.data.sessions.length}}
-                  <Hds::Button
-                    @text={{t 'actions.yes'}}
-                    @color='tertiary'
-                    @icon='info'
-                    {{on 'click' this.toggleSessionsFlyout}}
-                  />
-                {{/if}}
+                {{#let
+                  (call (fn this.getAvailableSessions B.data))
```
ZedLi, line 111: Is this needed? You only call this to sort the sessions but it only gets used on line 114 it looks like so do you really need it sorted?

Couldn't you just call `{{#if B.data.isActive}}` on line 114 instead?

ui/desktop/app/controllers/scopes/scope/projects/targets/index.js
```diff
@@ -174,4 +183,41 @@ export default class ScopesScopeProjectsTargetsIndexController extends Controlle
   applyFilter(selectedItems) {
     this.scopes = [...selectedItems];
   }
+
+  @action
+  toggleSessionsFlyout() {
+    this.sessionsFlyoutActive = !this.sessionsFlyoutActive;
```
ZedLi, line 189: Do you need to set this boolean? Couldn't you just be checking if you have a `selectedTarget`? And this method then wouldn't need to exist since on close you could just use `selectTarget` and set the target to null

lisbet-alvarez: yes, good point, ill clean this up.

ui/desktop/app/templates/scopes/scope/projects/targets/index.hbs
lisbet-alvarez: ahh right, that is much simpler!

ui/desktop/app/routes/scopes/scope/projects/sessions/session.js
ZedLi: It's definitely something weird happening...inside the client daemon handler when we normalize the response manually, the issue is happening [here](https://github.com/hashicorp/boundary-ui/blob/cdb73f34876b68d4e4772847e7772882212d5cae/addons/api/addon/serializers/application.js#L246) when we normalize the response, the call to super is setting the fields to null for some reason, it looks like it does it for all date fields 🤔

ui/desktop/app/templates/scopes/scope/projects/targets/index.hbs
```diff
@@ -190,6 +194,78 @@
         <A.Body @text={{t 'resources.target.messages.none.description'}} />
       </Hds::ApplicationState>
     {{/if}}
+    {{#let
+      (call (fn this.getAvailableSessions this.selectedTarget))
```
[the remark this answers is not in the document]
lisbet-alvarez: thank u, for the link! Yes, that makes this simplier.

ui/desktop/app/routes/scopes/scope/projects/sessions/session.js
lisbet-alvarez: could it be because date fields use the date data type? I wonder if it happens for any fields that are not strings... Thx for pointing out where it happens, ill investigate some more on this.

ZedLi:
```
{
    "id": "ss_ldyksmlz79",
    "type": "session",
    "attributes": {
        "target_id": "t_agrcgigmr7",
        "status": "active",
        "created_time": null,
        "updated_time": null,
        "expiration_time": null,
        "version": 41236729692160,
        "scope": {
            "scope": {
                "id": "s_id6ozx3wue",
                "type": "org"
            },
            "id": "s_1sheljmjgm",
            "type": "project",            
            "disabled": false,
            "updated_time": "2023-12-15T08:23:25.639Z",
            "created_time": "2023-12-15T09:55:33.176Z",
            "description": "so xerox",
            "name": "commonly memorable whoa",
            "scope_id": "s_1sheljmjgm",
            "isGlobal": false,
            "isOrg": false,
            "isProject": true
        },
        "authorized_actions": [
            "cancel"
        ]
    },
    "relationships": {}
}
```

The scope ones got normalized fine so I'm not sure 😕. Definitely something to investigate

ZedLi: Ah I think the issue is I was returning date objects instead of the ISO string

ui/desktop/app/controllers/scopes/scope/projects/targets/index.js
```diff
@@ -174,4 +183,40 @@ export default class ScopesScopeProjectsTargetsIndexController extends Controlle
   applyFilter(selectedItems) {
     this.scopes = [...selectedItems];
   }
+
+  @action
+  deselectTarget() {
```
ZedLi, line 188: Is this needed? Couldn't you just use `selectTarget` and pass null?

ui/desktop/app/controllers/scopes/scope/projects/targets/index.js
```diff
@@ -174,4 +183,40 @@ export default class ScopesScopeProjectsTargetsIndexController extends Controlle
...
+    this.selectedTarget = null;
+  }
+
+  /**
+   * Toggle the sessions flyout and initialize variable to store selected target
+   * @param {object} selectedTargetSessions
```
ZedLi, line 194:
```suggestion
   * @param {object} selectedTarget
```

ui/desktop/app/templates/scopes/scope/projects/targets/index.hbs
```diff
@@ -190,6 +189,73 @@
         <A.Body @text={{t 'resources.target.messages.none.description'}} />
       </Hds::ApplicationState>
     {{/if}}
+    {{#if (and this.selectedTarget this.selectedTarget.isActive)}}
```
ZedLi, line 192: Could this just check just be 
```suggestion
    {{#if this.selectedTarget}}
```
We can't select targets that aren't active right? You've already filtered them ones that you can select?

ZedLi (approved):
Looks great!!

ui/desktop/app/templates/scopes/scope/projects/targets/index.hbs
lisbet-alvarez: There could be a scenario where a user has the flyout open, and cancels the only session in the table and that check is there to close the flyout.

ZedLi: Gotcha makes sense!

ui/desktop/tests/acceptance/projects/targets/index-test.js
```diff
@@ -290,4 +297,120 @@ module('Acceptance | projects | targets | index', function (hooks) {
...
+    assert
+      .dom(
+        `[data-test-targets-sessions-flyout-button="${instances.target.id}"]`,
+      )
+      .doesNotExist();
+    assert.strictEqual(currentURL(), urls.targets);
+  });
+
+  test('user cannot cancel a session from inside target sessions flyout without permissions', async function (assert) {
+    assert.expect(4);
```
calcaide, line 358: none blocking comment: totally fine if you want to use it but letting you at the current config, `assert.expect` is not need anymore unless you are using nested assertions, or assertions within a function.

calcaide (approved):
LGTM!! great job!

ui/desktop/tests/acceptance/projects/targets/index-test.js
lisbet-alvarez: ahh right, thanks for catching that!

PR #2045, opened 2023-12-13
fix: 🐛 Fix failing tests from rebase onto main

ZedLi:
## Description
Some tests starting to fail after rebasing onto main. The root cause of the issue is a recent PR added some calls to get sessions in the targets route which adds them to the store so the following call to retrieve the session is returning cached data from the store (which for some reason didn't have expiration_time).

Some of the stubbed client daemon calls were also not being setup correctly as they were returning target data for session calls. 

## Checklist:


lisbet-alvarez (approved):
Thanks for fixing this!

PR #2046, opened 2023-12-13
test(admin/e2e): Add tags to tests

moduli:
This PR updates the Admin UI e2e test suite to make use of [playwright tags](https://playwright.dev/docs/test-annotations) to more easily run all tests for a specific configuration. 
- Community Edition (ce) vs. Enterprise (enterprise)
- AWS infrastructure (aws) vs. Docker infrastructure (docker)
  - Eventually, we will add some tests that will require docker-based infrastructure (i.e. LDAP, ssh certificate injection)
  - We still want to keep some tests using AWS infrastructure. There are some tests that require AWS infrastructure (Dynamic Host Catalogs). This also helps validate using non-secure contexts (using `http` connections).

This is accompanied by some new yarn scripts
```
yarn run e2e:ce:aws
yarn run e2e:ce:docker
yarn run e2e:ent:aws
yarn run e2e:ent:docker
```


ui/admin/tests/e2e/tests/target.spec.js
```diff
@@ -120,7 +120,7 @@ test('Verify session created to target with address, then cancel the session', a
   }
 });
 
-test('Verify TCP target is updated', async ({ page }) => {
+test('Verify TCP target is updated @ce @aws @docker', async ({ page }) => {
```
stasryzhov, line 123: I noticed that in some tests which validate features available in both CE and ENT ( e.g. `Log in with invalid password`) we specify both tags. 

Is it intentional that in these target related cases we specify CE tag only?

moduli: These were the tests that broke in Enterprise because Enterprise has an extra picker to select a TCP or SSH target

stasryzhov (approved):

ui/admin/package.json
```diff
@@ -34,6 +34,10 @@
     "test": "concurrently \"npm:test:*\" --names \"test:\"",
     "test:ember": "COVERAGE=true ember exam --test-port 0 --split=4 --parallel=1 --random",
     "e2e": "yarn playwright test --config ./tests/e2e/playwright.config.js",
+    "e2e:ce:aws": "yarn run e2e --grep \"(?=.*@ce)(?=.*@aws)\"",
```
calcaide, line 37: what is the purpose of `-grep \"(?=.*@ce)(?=.*@aws)\"` ?

moduli: That's currently the only way to do AND with tags (run tests with tag `ce` AND tag `aws`)

From: https://playwright.dev/docs/test-annotations#tag-tests
> Or run tests containing both tags (logical AND operator) using regex lookaheads:
> `npx playwright test --grep "(?=.*@fast)(?=.*@slow)"`

PR #2047, opened 2023-12-13
refactor: 💡 Optimize the retrieval of targets for scopes filter

ZedLi:
## Description
Changed when we retrieve targets so it's only when the user first enters the route. This may need to change in the future but for now it helps greatly with loading times. Ideally, we'd lazy load all the targets when needed which we can evaluate in the future. 

Also fixed a bug where the search box for targets didn't show the value if you changed routes and came back. 

Before:
image

After:
image

You can see the extra call adds about ~1s to searching/filtering.

## How to Test

Setup the search daemon with 20 to 30k targets and see the performance of searching or filtering.

## Checklist:

### DIFF
--- a/ui/desktop/app/templates/scopes/scope/projects/targets/index.hbs
+++ b/ui/desktop/app/templates/scopes/scope/projects/targets/index.hbs
@@ -17,6 +17,7 @@
     {{#if (or @model.targets this.search)}}
       <Hds::SegmentedGroup as |S|>
         <S.TextInput
+          @value={{this.search}}
           @type='search'
           placeholder={{t 'actions.search'}}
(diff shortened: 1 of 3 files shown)


ui/desktop/app/templates/scopes/scope/projects/targets/index.hbs
```diff
@@ -17,6 +17,7 @@
     {{#if (or @model.targets this.search)}}
       <Hds::SegmentedGroup as |S|>
         <S.TextInput
+          @value={{this.search}}
```
cameronperera, line 20: 😵‍💫 
Thanks for catching that.

cameronperera (approved):
LGTM

PR #2048, opened 2023-12-13
Icu 12060 move rose filterable dropdown to core

cameronperera:
:tickets: [Jira ticket](https://hashicorp.atlassian.net/browse/ICU-12060)

## Description

The Rose::FilterableDropdown was moved into the core addon as just `Dropdown`. I added some tests as well to go along with the new component.

## How to Test

Build a binary from the llb-client-daemon branch and use that binary in your local dev environment. You will need to move that binary into the ui/desktop/electron-app/cli folder as well. Be sure to start the desktop client using this command: ENABLE_MIRAGE=false BYPASS_CLI_SETUP=true yarn start:desktop

The dropdown should work as it has in the past. It does have a maxHeight set now via the `@height` param on the dropdown. So if you have many projects, it should only grow so tall and show a scroll bar.

## Checklist:


addons/core/addon/components/dropdown/index.hbs
```diff
@@ -3,7 +3,7 @@
   SPDX-License-Identifier: MPL-2.0
 }}
 
-<Hds::Dropdown @onClose={{this.close}} as |DD|>
+<Hds::Dropdown @height='314px' @onClose={{this.close}} as |DD|>
```
ZedLi, line 6: Why 314 specifically out of curiosity lol

addons/core/package.json
```diff
@@ -42,7 +42,8 @@
     "ember-loading": "^2.0.0",
     "ember-truth-helpers": "^3.0.0",
     "filesize": "^10.0.7",
-    "intl-unofficial-duration-unit-format": "^3.1.0"
+    "intl-unofficial-duration-unit-format": "^3.1.0",
+    "lodash": "^4.17.21"
```
ZedLi, line 46: Should lodash have been removed from admin and desktop now?

addons/core/tests/integration/components/dropdown/index-test.js
```diff
@@ -0,0 +1,121 @@
...
+
+    await click(TOGGLE_DROPDOWN_SELECTOR);
+
+    assert.dom(ITEM_SELECTOR).exists({ count: 2 });
+
+    await fillIn(SEARCH_INPUT_SELECTOR, 'fake');
+
+    // setTimeout allows us to wait for the debounce
+    // to finish before checking the DOM again
+    setTimeout(() => {
```
ZedLi, line 116: Did waitUntil not work?

I would've expected something like this to work
```
    const result = await waitUntil(function() {
      return findAll(ITEM_SELECTOR).length === 1
    })

    assert.true(result)
```

addons/core/addon/components/dropdown/index.hbs
cameronperera: That is the height we have in the wireframes. This can change if we need it to but I felt this was a good height based on the designs.

addons/core/package.json
cameronperera: Yes. I will do that now. Thanks for the reminder.

addons/core/tests/integration/components/dropdown/index-test.js
```diff
@@ -0,0 +1,119 @@
...
+
+    await click(TOGGLE_DROPDOWN_SELECTOR);
+
+    assert.dom(ITEM_SELECTOR).exists({ count: 2 });
+
+    await fillIn(SEARCH_INPUT_SELECTOR, 'fake');
+
+    const result = await waitUntil(function () {
+      return findAll(ITEM_SELECTOR).length === 1;
+    });
```
ZedLi, line 115:
```suggestion
    const result = await waitUntil(() => findAll(ITEM_SELECTOR).length === 1);
```

ZedLi: Optional but this didn't need `function` or braces

ZedLi (approved):
Looks good!

DhariniJeeva (approved):
looks great!

lisbet-alvarez (approved):
✨🙌

PR #2049, opened 2023-12-14
refactor: 💡 Migrate `hosts`list within host-set to `Hds::Table`

dvonanderson:
✅ Closes: https://hashicorp.atlassian.net/browse/ICU-11728

:tickets: [Jira ticket](https://hashicorp.atlassian.net/browse/ICU-11728)

## Description
- Migrate `hosts` list within host-set to `Hds::Table` from `Rose::Table`
- Refactor hosts-tests 

### Before: 
--- 
Screenshot 2023-12-14 at 18 10 06

### After:
---
Screenshot 2023-12-14 at 17 29 43

## How to Test
- [ ] Click on Vercel link for Admin UI or pull down PR and run it locally using `yarn start`
- [ ] Click on any "Orgs" card created by mirage
- [ ] Click on any "Projects" card created by mirage and "Host Catalogs" will appear in the sidebar navigation.
- [ ] Click on "Host Catalogs" in the sidebar navigation.
- [ ] Click on any host catalog from list.
- [ ] Click on the "Hosts-Sets" tab 
- [ ] Click on any host-sets from the list
- [ ] Click on the "Hosts" tab and view list

## Checklist:


ui/admin/app/templates/scopes/scope/host-catalogs/host-catalog/host-sets/host-set/hosts/index.hbs
```diff
@@ -59,51 +59,50 @@
...
+                      @color='critical'
                       {{on
                         'click'
                         (route-action 'removeHost' @model.hostSet host)
                       }}
-                    >
-                      {{t 'actions.remove'}}
-                    </dropdown.button>
-                  </Rose::Dropdown>
+                      data-test-host-set-hosts-dropdown-remove
```
calcaide, line 96: Are we using this `data-test` selector?

ui/admin/tests/acceptance/host-catalogs/host-sets/hosts-test.js
```diff
@@ -51,6 +51,12 @@ module('Acceptance | host-catalogs | host-sets | hosts', function (hooks) {
     newHost: null,
   };
 
+  const HOST_ACTIONS_SELECTOR =
```
calcaide, line 54: I notice we add within the view code a `data-test-` attribute, but I do not see any of the tests using it. Should we delete the `data-test` attribute, or should we add another one for the dropdown and rely of `data-test` attributes rather than long selectors?

calcaide (changes requested):
Left some comments.

ui/admin/tests/acceptance/host-catalogs/host-sets/hosts-test.js
dvonanderson: @calcaide initially I was having some problems grabbing the element in the test environment.  I will remove this.

calcaide: @dvonanderson 
- Can you describe what problems are you having? We are using `data-test-*` selectors without problems within other tests.
- `but it seemed that some used date-test selectors and some don't.` what do you mean by this?
- Feel free to bring any topic to tech time, tech topics are always welcome in such space. That said, the `data-test-*` selectors is something we use to facilitate our lives (and getting some performance win for free) when have to select elements where the pattern to select is too long or too cumbersome. I.e: `tbody .hds-table__tr:nth-child(1) .hds-dropdown-list-item button` is quite long and will be easier for the developer and for the machine that process this test to use a data-test-* selector, i.e: `data-test-host-set-hosts-remove`. I think we should apply common sense, if the selector is  short ie: `main#main-wrapper` makes sense to use such selector but in other cases `data-test` are preferable.
- What do you mean by: `ersonally I think we should adopt the same patterns that cloud-ui uses.`? We actually copy the `data-test-*` approach from `cloud-ui`

calcaide (changes requested):
Left some comments

ui/admin/tests/acceptance/host-catalogs/host-sets/hosts-test.js
dvonanderson: @calcaide 
- So originally I think what was happening with the test was. I had `'[data-test-host-set-host-dropdown-remove-host] button'` as the selector for the `dropdown.interactive`.  I just tested it again without the button and it seemed to work.  Maybe I just needed to restart my test and never did. 
- I used a pattern from the auth-methods create test  to set up the selectors for this test.  I do agree its a bit long and cubersome and `data-test`selectors are much more readable.
- Lastly the comment about how `cloud-ui` use selectors refers more to the using the `selectors.js` to house all of your test-support.  

Incoming commit to change it to a `data-test` selector

ui/admin/app/templates/scopes/scope/host-catalogs/host-catalog/host-sets/host-set/hosts/index.hbs
```diff
@@ -59,51 +59,47 @@
...
-                  @acknowledgeText={{t 'states.copied'}}
-                >
-                  <code>{{host.id}}</code>
-                </Copyable>
-              </row.cell>
-              <row.cell>
+              </B.Td>
+              <B.Td>
+                <Hds::Copy::Snippet @textToCopy={{host.id}} @color='secondary'>
+                  <Hds::Text::Code>{{host.id}}</Hds::Text::Code>
```
calcaide, line 74: Why are we wrapping `Hds::Text::Code` within `Hds::Copy::Snippet` ?

Within [`Hds::Copy::Snippet` documentation,](https://helios.hashicorp.design/components/copy/snippet?tab=code) the only pattern prescribed is to use the attribute `@textToCopy` to pass the text we want the user to see and copy.

The `Hds::Copy::Snippet` does not support this pattern, in fact, it is not "yielding" anything to its content. If you try to nest content inside will be ignored. [You can check the component code](https://github.com/hashicorp/design-system/blob/main/packages/components/addon/components/hds/copy/snippet/index.hbs).

dvonanderson: @calcaide ahh that is correct! fix incoming

ui/admin/tests/acceptance/host-catalogs/host-sets/hosts-test.js
calcaide: @dvonanderson 

- Yeah not sure about that, perhaps was the environment. Or maybe was you need to open the dropdown first for the remove button to appear? `Hds::Dropdown` does not render the options of the dropdown until is not open, therefor the `[data-test-host-set-host-dropdown-remove-host] button` will not exist on the DOM unless you click the dropdown first to open. NOT saying this is the cause, but can be a possibility. You can check this opening the secreen with the browser dev tools, using the element inspector you will see no remove button unless you open the dropdown first.
- Not mandatory at all, just a suggestion: since you mention [the auth-methods create test](https://github.com/hashicorp/boundary-ui/blob/ade1601d8d533d620feddba6469b69df431a2315/ui/admin/tests/acceptance/auth-methods/create-test.js#L22) in there, the selectors that are used more than once are defined as constants at the beginning of the file and then call the constant during the test file.
- Indeed `data-test` attributes need more definition from our part, feel free to bring that topic to tech time, it needs some discussion: when to use `data-test` selectors vs when to not use them, should we use contants in our test when repeting the same selectors over and over? should we have a file containing all the selectors as `cloud-ui` does with the mentioned `selectors.js`?

dvonanderson: @calcaide Awesome thank you for the information!!

calcaide (approved):
LGTM!

ui/admin/app/templates/scopes/scope/host-catalogs/host-catalog/host-sets/host-set/hosts/index.hbs
```diff
@@ -28,26 +28,26 @@
 
   <page.body>
     {{#if @model.hosts}}
-      <Rose::Table as |table|>
-        <table.header as |header|>
-          <header.row as |row|>
+
+      <Hds::Table @model={{@model.hosts}}>
```
calcaide, line 32: Why we are not using table attributes `@model` and `@columns` as we did with the other HDS::Table migration?

calcaide (changes requested):
Left a comment

ui/admin/app/templates/scopes/scope/host-catalogs/host-catalog/host-sets/host-set/hosts/index.hbs
dvonanderson: @calcaide because we are doing conditionals to provide different header cell data

calcaide: @dvonanderson 

Except when using `Hds::Table` for building components, in our templates I see no use of `Hds::Table` without `@columns` attribute.

The pattern I see when building table with dynamic content is checking for the condition once and rendering a table per condition. You can see this pattern in our codebase [here](https://github.com/hashicorp/boundary-ui/blob/main/ui/admin/app/templates/scopes/scope/auth-methods/auth-method/accounts/index.hbs#L27-L124), [here](https://github.com/hashicorp/boundary-ui/blob/main/ui/admin/app/templates/scopes/scope/auth-methods/auth-method/managed-groups/index.hbs#L26-L108) and [here](https://github.com/hashicorp/boundary-ui/blob/main/ui/admin/app/templates/scopes/scope/auth-methods/auth-method/managed-groups/managed-group/members.hbs#L38-L104).

With this implementation, you are evaluating `@model.hostSet.isStatic` 2 times for the header + n times for the host which is a higher computation cost compared to one evaluation.

If the condition is ONLY to provide different headers, you can evaluate once and provide different headers. I.e:
```
<Hds::Table>
    {{#if @model.hostSet.isStatic}}
        <:head>
            head content for condition A
        </:head>
    {{else}}
        <:head>
            head content for condition A
        </:head>
    {{/if}}
```

But I do think. In this specific case, does the condition JUST affects the header or does also affect the body of the table? if so, I will suggest to follow one of the approaches we already have in our codebase provided above.

Let me know your thoughts.

calcaide:
@dvonanderson 

A detail I just realize (applicable for all the already merged `Hds::Table migration` PR's) .

Most of the `Hds::Table` present within templates (excluding the component ones) have a vertical align defined but these don't.

Since this initiative is about to adopt and standardize the usage of `Hds::Table` within Boundary-ui's. Can you please review that?

Examples of `Hds::Table` migration that have been using vertical alignment, [here](https://github.com/hashicorp/boundary-ui/blob/main/ui/admin/app/templates/scopes/scope/groups/index.hbs#L40), [here](https://github.com/hashicorp/boundary-ui/blob/main/ui/admin/app/templates/scopes/scope/groups/group/members.hbs#L41) and [here](https://github.com/hashicorp/boundary-ui/blob/main/ui/admin/app/templates/scopes/scope/storage-buckets/index.hbs#L33).
I am not saying it has to be an specific value, if you have doubts on what value to set, would be good to post on #boundary-ui a question.

dvonanderson:
> @dvonanderson
> 
> A detail I just realize (applicable for all the already merged `Hds::Table migration` PR's) .
> 
> Most of the `Hds::Table` present within templates (excluding the component ones) have a vertical align defined but these don't.
> 
> Since this initiative is about to adopt and standardize the usage of `Hds::Table` within Boundary-ui's. Can you please review that?
> 
> Examples of `Hds::Table` migration that have been using vertical alignment, [here](https://github.com/hashicorp/boundary-ui/blob/main/ui/admin/app/templates/scopes/scope/groups/index.hbs#L40), [here](https://github.com/hashicorp/boundary-ui/blob/main/ui/admin/app/templates/scopes/scope/groups/group/members.hbs#L41) and [here](https://github.com/hashicorp/boundary-ui/blob/main/ui/admin/app/templates/scopes/scope/storage-buckets/index.hbs#L33). I am not saying it has to be an specific value, if you have doubts on what value to set, would be good to post on #boundary-ui a question.

@calcaide totally agree, I will create a ticket to add to the Migration that addresses the alignment middle for all other tables.

ui/admin/app/templates/scopes/scope/host-catalogs/host-catalog/host-sets/host-set/hosts/index.hbs
calcaide: @dvonanderson 

On my previous comment I was wrong: `With this implementation, you are evaluating @model.hostSet.isStatic 2 times for the header + n times for the host which is a higher computation cost compared to one evaluation` the body will keep computing that condition N times.

I like this solution. I think is worth it to bring to tech time the table patterns we are encountering. It will be good to have team consensus.

ui/admin/app/templates/scopes/scope/host-catalogs/host-catalog/host-sets/host-set/hosts/index.hbs
```diff
@@ -28,82 +28,80 @@
...
-              </row.cell>
-              <row.cell>
-                <Copyable
-                  @text={{host.id}}
-                  @buttonText={{t 'actions.copy-to-clipboard'}}
-                  @acknowledgeText={{t 'states.copied'}}
+        <:body as |B|>
+          <B.Tr>
+            <B.Td>
+              {{#if B.data.isStatic}}
```
calcaide, line 53: @dvonanderson 
Wondering if this conditional is actually needed. If a host set is dynamic (plugin, azure or aws), can the host set have hosts that are static? and vice versa?

dvonanderson: @calcaide a question for me is why are we providing a sub-route for dynamic/plugin host?  I removed the conditional and worked fine for the  ` @route='scopes.scope.host-catalogs.host-catalog.host-sets.host-set.hosts.host'`

calcaide: @dvonanderson 
Yep, that's why we need to first validate the question of: Can a host-set with specific type (static, plugin, aws, azure) contain hosts that have different types that the host-set?

This questions is something we need to figure out first, once we know this, will be easier to reason about this case.

Good call on the sub-route, perhaps you can use git history and see when we add this change, usually that brings context. Need to figure out these two things before.

ui/admin/app/templates/scopes/scope/host-catalogs/host-catalog/host-sets/host-set/hosts/index.hbs
dvonanderson: @calcaide such a good point. Really need to start considering the computation complexity of some of the blocks that have significant conditional checks!

ui/admin/app/templates/scopes/scope/host-catalogs/host-catalog/host-sets/host-set/hosts/index.hbs
dvonanderson: @calcaide resolving this conversation

ui/admin/app/templates/scopes/scope/host-catalogs/host-catalog/host-sets/host-set/hosts/index.hbs
dvonanderson: @calcaide after talking with the team and researching how other teams do conditional rendering in table headers, I am going to stick with its current iteration, but I am going to clean it up a bit.  I created an controller that will handle the translations and return the array of headers thats needed to make it look more readable.

ui/admin/app/controllers/scopes/scope/host-catalogs/host-catalog/host-sets/host-set/hosts/index.js
```diff
@@ -0,0 +1,47 @@
...
+export default class ScopesScopeHostsHostCatalogsHostCatalogHostSetsHostSetHostsController extends Controller {
+  // =services
+  @service intl;
+
+  // =attributes
+  /**
+   * Returns the static columns header labels for the table
+   * @type {Array.<object>}
+   */
+  get isStaticColumns() {
```
ZedLi, line 13: These getters could have just been one getter that does the `isStatic` check in it instead of the template. And they shouldn't be named "`is` + Static/DynamicColumns" since they don't return booleans.

ui/admin/app/templates/scopes/scope/host-catalogs/host-catalog/host-sets/host-set/hosts/index.hbs
calcaide: @dvonanderson Resolving this conversation which what outcome? Can you provide an answer to the question stated above?

ui/admin/app/controllers/scopes/scope/host-catalogs/host-catalog/host-sets/host-set/hosts/index.js
dvonanderson: @ZedLi good call out, working on it now

ui/admin/app/templates/scopes/scope/host-catalogs/host-catalog/host-sets/host-set/hosts/index.hbs
dvonanderson: @calcaide yep sorry thought that was more of a search and discovery point, but this change was part of template refactor to deprecate `renderTemplate`

calcaide: @dvonanderson 

I am not asking about when we added this, I can check git history, and yeah, this was part of refactor `renderTemplate` 15 months ago.

What I am still looking for is an answer to the question which I started this conversation, [here the link](https://github.com/hashicorp/boundary-ui/pull/2049#discussion_r1435239638), here the question: `If a host set is dynamic (plugin, azure or aws), can the host set have hosts that are static? and vice versa?`


In my [message here](https://github.com/hashicorp/boundary-ui/pull/2049#discussion_r1435304081) it was saying that first, we need to answer this question. I don't think is a search & discovery point, we are refactoring/migrating tables, this conditional is part of the table, I am not sure is need, in case is not, we are computing a conditional several times for no reason. Since I am not sure (from the domain PoV) this conditional is need, I would like to know that first.

I will appreciate an answer to such question

ui/admin/app/controllers/scopes/scope/host-catalogs/host-catalog/host-sets/host-set/hosts/index.js
```diff
@@ -0,0 +1,43 @@
...
+  // =attributes
+  /**
+   * Returns the columns for the host list.
+   * @type {Array.<object>}
+   */
+  get columns() {
+    if (this.model.hostSet.isStatic) {
+      return [
+        {
+          label: this.intl.t('form.name.label'),
```
DhariniJeeva, line 17: To improve readability - I'd move these properties to an array variable at the top and just return the var name here in the function.

ui/admin/app/templates/scopes/scope/host-catalogs/host-catalog/host-sets/host-set/hosts/index.hbs
```diff
@@ -28,82 +28,67 @@
...
+            <B.Td>
+              <HostCatalogTypeBadge @model={{B.data}} />
+            </B.Td>
+            <B.Td>
+              <Hds::Copy::Snippet
+                @textToCopy={{B.data.id}}
+                @color='secondary'
+              />
+            </B.Td>
+            {{#if (can 'removeHosts hostSet' @model.hostSet)}}
```
calcaide, line 67: The conditional, should be nested inside the `<td>` and not outside, otherwise table rows are going to be different, table should be consistent.

[Here an example](https://github.com/hashicorp/boundary-ui/blob/main/ui/admin/app/templates/scopes/scope/groups/group/members.hbs#L68-L86) of this exact use case within other template we already migrate to HDS table

calcaide (changes requested):
Left comments

ui/admin/app/controllers/scopes/scope/host-catalogs/host-catalog/host-sets/host-set/hosts/index.js
calcaide: Indeed with @DhariniJeeva comment, good call Dharini!!

ui/admin/app/templates/scopes/scope/host-catalogs/host-catalog/host-sets/host-set/hosts/index.hbs
dvonanderson: @calcaide reached out to the BE team, but Dharini was there to answer the question. Host sets have to match the type of host set.   https://hashicorp.slack.com/archives/C014AHXUL3G/p1704482262367159

ui/admin/app/controllers/scopes/scope/host-catalogs/host-catalog/host-sets/host-set/hosts/index.js
dvonanderson: @DhariniJeeva @ZedLi take a look at the changes and let me know what you think.

ui/admin/app/templates/scopes/scope/host-catalogs/host-catalog/host-sets/host-set/hosts/index.hbs
dvonanderson: @calcaide I moved this condition outside so that it doesn't have to render that `td`.  If I move it back outside the `td` will still render with nothing in it and I will have to provide the `dynamic` table a blank header label.  Which I did in this last [commit](https://github.com/hashicorp/boundary-ui/pull/2049/files#diff-1e0487df0f54bc897aae9166eefaf436cd6c19af19b68e6200bde0f2a6f398c2R39).

ui/admin/app/templates/scopes/scope/host-catalogs/host-catalog/host-sets/host-set/hosts/index.hbs
calcaide: @dvonanderson 
Ok. So what we found out is that we have 2 types and are exclusive. So or is type A or is type B.

For that reason, we have 2 table types. I do suggest to:
1. Find in our codebase we have a pattern for templates that have exclusive table types. Like [this one](https://github.com/hashicorp/boundary-ui/blob/main/ui/admin/app/templates/scopes/scope/auth-methods/auth-method/accounts/index.hbs#L25-L168) or [this one](https://github.com/hashicorp/boundary-ui/blob/main/ui/admin/app/templates/scopes/scope/auth-methods/auth-method/managed-groups/index.hbs#L26-L109).
2. Follow the current pattern for this case. If you find the current pattern can be improved, has flaws or can't fit your specific case let's discuss. I strongly recommend leverage tech time so we get team consensus. But feel free to ping me if you want to discuss further.

Why I suggest point 1? If the type is exclusive, we should check the condition just once. If you are type A, render table for type A. If you are type B, render table for type B. Then, the only conditionals left within the table (if any) should be related to permissions.

Thoughts?

ui/admin/app/templates/scopes/scope/host-catalogs/host-catalog/host-sets/host-set/hosts/index.hbs
calcaide: @dvonanderson 
So actions column is JUST available for `static` types? so `dynamic` types have no actions?

ui/admin/app/templates/scopes/scope/host-catalogs/host-catalog/host-sets/host-set/hosts/index.hbs
dvonanderson: @calcaide we discussed in last tech time.  I will ping you tomorrow

ui/admin/app/templates/scopes/scope/host-catalogs/host-catalog/host-sets/host-set/hosts/index.hbs
dvonanderson: @DhariniJeeva correct me if I am wrong, but the `host-set` containing dynamic host can be deleted, but not the dynamic `host` itself?

ui/admin/app/templates/scopes/scope/host-catalogs/host-catalog/host-sets/host-set/hosts/index.hbs
dvonanderson: @calcaide I just pushed a commit that adheres to our current pattern take a look and give me you thoughts?

ui/admin/app/templates/scopes/scope/host-catalogs/host-catalog/host-sets/host-set/hosts/index.hbs
DhariniJeeva: Unfortunately, I do not recall. Have you tried testing this scenario with the controller or the CLI?

ui/admin/app/templates/scopes/scope/host-catalogs/host-catalog/host-sets/host-set/hosts/index.hbs
```diff
@@ -28,82 +28,101 @@
...
-        </table.body>
-      </Rose::Table>
+              </B.Td>
+
+            </B.Tr>
+          </:body>
+        </Hds::Table>
+      {{else}}
+        <Hds::Table
+          @valign='middle'
```
calcaide, line 92: None blocking: Perhaps nit picky, but can we put the `@valign` attribute the last one? so `@model` and `@columns` are first. This is how we have been defining HDS table attributes in the previous work and personally, I think it is good practice to have more important attributes first. Totally debatable and open to it.

ui/admin/app/templates/scopes/scope/host-catalogs/host-catalog/host-sets/host-set/hosts/index.hbs
```diff
@@ -28,82 +28,101 @@
...
-                <HostCatalogTypeBadge @model={{host}} />
-              </row.cell>
-              <row.cell>
-                <Copyable
-                  @text={{host.id}}
-                  @buttonText={{t 'actions.copy-to-clipboard'}}
-                  @acknowledgeText={{t 'states.copied'}}
+      {{#if (get @model.hostSet 'isStatic')}}
+        <Hds::Table
+          @valign='middle'
```
calcaide, line 33: None blocking: Perhaps nit picky, but can we put the `@valign` attribute the last one? so `@model` and `@columns` are first. This is how we have been defining HDS table attributes in the previous work and personally, I think it is good practice to have more important attributes first. Totally debatable and open to it.

ui/admin/app/templates/scopes/scope/host-catalogs/host-catalog/host-sets/host-set/hosts/index.hbs
dvonanderson: @calcaide absolutely agreed

ui/admin/app/templates/scopes/scope/host-catalogs/host-catalog/host-sets/host-set/hosts/index.hbs
dvonanderson: @calcaide absolutely agreed

ui/admin/app/templates/scopes/scope/host-catalogs/host-catalog/host-sets/host-set/hosts/index.hbs
dvonanderson: @DhariniJeeva @calcaide  yea authorized actions for dynamic host are only `no-op` and `read`
image

calcaide: @dvonanderson gotcha!!! then the current implementation is on point. Since we realize that host-set type is exclusive for what table show let me question the current authorized actions. I am glad we clarify that and the current implementation follows it.

calcaide (approved):
LGTM!!

cameronperera (approved):

PR #2050, opened 2023-12-15
fix: 🐛 Fix returning mirage data from client daemon mock

ZedLi:
## Description
This PR fixes two issues in tests:

1. We weren't properly serializing the data (using our mirage serializers to rename some properties to snake case and serializing the scope object)
2. We were returning the data raw which includes any JS objects such as date objects. We need to return this data as standard JSON so our ember data can properly normalize the data it receives.


lisbet-alvarez (approved):
Tested locally and this resolves the failing tests! Thank you!

PR #2053, opened 2023-12-18
feat: 🎸 filter sessions by target

lisbet-alvarez:
:tickets: [Jira ticket](https://hashicorp.atlassian.net/browse/ICU-12066)

## Description

Add target filtering on sessions list view page.

https://github.com/hashicorp/boundary-ui/assets/107949262/8a800dee-9276-483e-b507-c805424ecf0a

## How to test

1. Use llb-pagination-2 be branch and start boundary dev
3. Checkout this branch and run ENABLE_MIRAGE=false BYPASS_CLI_SETUP=true yarn start:desktop

## Checklist:

### DIFF
--- a/addons/core/addon/components/dropdown/index.hbs
+++ b/addons/core/addon/components/dropdown/index.hbs
@@ -3,7 +3,12 @@
   SPDX-License-Identifier: MPL-2.0
 }}
 
-<Hds::Dropdown @height='314px' @onClose={{this.close}} as |DD|>
+<Hds::Dropdown
+  @height='314px'
+  @onClose={{this.close}}
+  @listPosition={{@listPosition}}
+  as |DD|
+>
   <DD.ToggleButton
     @color='secondary'
     @text={{@name}}
--- a/addons/core/addon/components/dropdown/index.hbs
+++ b/addons/core/addon/components/dropdown/index.hbs
@@ -40,7 +40,7 @@
         checked={{includes itemOption.id @checkedItems}}
         {{on 'change' this.selectItem}}
       >
-        {{#if itemOption.name itemOption.name itemOption.id}}
+        {{if itemOption.name itemOption.name itemOption.id}}
       </DD.Checkbox>
     {{/each}}
   {{/if}}
--- a/addons/core/addon/components/dropdown/index.hbs
+++ b/addons/core/addon/components/dropdown/index.hbs
@@ -40,7 +40,11 @@
         checked={{includes itemOption.id @checkedItems}}
         {{on 'change' this.selectItem}}
       >
-        {{itemOption.displayName}}
+        {{#if itemOption.name}}
+          {{itemOption.name}}
+        {{else}}
+          {{itemOption.id}}
+        {{/if}}
       </DD.Checkbox>
     {{/each}}
   {{/if}}
--- a/addons/core/addon/components/dropdown/index.hbs
+++ b/addons/core/addon/components/dropdown/index.hbs
@@ -40,11 +40,7 @@
         checked={{includes itemOption.id @checkedItems}}
         {{on 'change' this.selectItem}}
       >
-        {{#if itemOption.name}}
-          {{itemOption.name}}
-        {{else}}
-          {{itemOption.id}}
-        {{/if}}
+        {{#if itemOption.name itemOption.name itemOption.id}}
       </DD.Checkbox>
     {{/each}}
   {{/if}}
--- a/ui/desktop/app/controllers/scopes/scope/projects/sessions/index.js
+++ b/ui/desktop/app/controllers/scopes/scope/projects/sessions/index.js
@@ -4,25 +4,54 @@
  */
 
 import Controller from '@ember/controller';
+import { tracked } from '@glimmer/tracking';
+import { action } from '@ember/object';
 import orderBy from 'lodash/orderBy';
+import { statusTypes } from 'api/models/session';
 
 export default class ScopesScopeProjectsSessionsIndexController extends Controller {
+  // =services
+
   // =attributes
 
+  queryParams = [{ targets: { type: 'array' } }];
+
+  @tracked targets = [];
+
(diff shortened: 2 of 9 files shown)


ui/desktop/app/templates/scopes/scope/projects/sessions/index.hbs
```diff
@@ -13,110 +13,52 @@
...
-                @name={{filter.name}}
-                @items={{filter.allowedValues}}
-                @selectedItems={{filter.selectedValue}}
-                @onChange={{route-action 'filterBy' filter.name}}
-                as |group|
+  <page.body class='search-filtering'>
+    {{#if this.sortedSessions}}
+      <Hds::SegmentedGroup as |S|>
+        <S.Generic>
+          <Dropdown
```
ZedLi, line 20: I don't think I see anything that would necessitate building your own options, I think this could just be:
```
          <Dropdown
            @name={{t 'resources.target.title'}}
            @itemOptions={{this.availableTargets}}
            @checkedItems={{this.targets}}
            @applyFilter={{fn this.applyFilter 'targets'}}
            @isSearchable={{true}}
            @listPosition='bottom-left'
          />
```

Since the component handles building the items for you if you provide the item options

ui/desktop/app/routes/scopes/scope/projects/sessions/index.js
```diff
@@ -39,53 +34,43 @@ export default class ScopesScopeProjectsSessionsIndexRoute extends Route {
...
+    const queryOptions = { query: { filters }, force_refresh: true };
+    const sessions = await this.store.query('session', queryOptions);
 
-    // If status filters are selected, filter sessions by the selected statuses
-    if (this.status?.length) {
-      sessions = sessions.filter((s) => this.status.includes(s?.status));
+    // Query all sessions and all targets for defining filtering values if entering route for the first time
+    if (from !== 'scopes.scope.projects.sessions.index') {
+      const options = { pushToStore: false };
+      this.allSessions = await this.store.query(
```
ZedLi, line 56: What is `allSessions` used for?

lisbet-alvarez: it is used in the controller file index.js line 37 to make sure that all target filter options are available for existing sessions.

ui/desktop/app/templates/scopes/scope/projects/sessions/index.hbs
lisbet-alvarez: fixed, thank u!

ZedLi: I think you'll need to update the component though since it assumes `displayName` and change it in a few places, we should just manually do the `name || id` display rather than assume it's a model with a `displayName`

addons/core/addon/components/dropdown/index.hbs
```diff
@@ -35,7 +40,11 @@
         checked={{includes itemOption.id @checkedItems}}
         {{on 'change' this.selectItem}}
       >
-        {{itemOption.displayName}}
+        {{#if itemOption.name}}
+          {{itemOption.name}}
+        {{else}}
+          {{itemOption.id}}
+        {{/if}}
```
ZedLi, line 47:
```suggestion
        {{#if itemOption.name itemOption.name itemOption.id}}
```

You also need to update this in the targets controller where we make objects with `displayName`

ui/desktop/app/controllers/scopes/scope/projects/sessions/index.js
```diff
@@ -4,25 +4,59 @@
...
+   * with a limit of 100 targets rendered.
+   * @returns {[TargetModel]}
+   */
+  get availableTargets() {
+    const uniqueSessionTargetIds = new Set(
+      this.model.allSessions.map((session) => session.target_id),
+    );
+    return this.model.allTargets
+      .filter((target) => uniqueSessionTargetIds.has(target.id))
+      .slice(0, 100);
```
ZedLi, line 42: I think you should do this in a separate PR, we should be doing this in the drop down component but this will also prevent you from searching all possible targets right?

lisbet-alvarez: Ah yup you're correct, and good point I'll remove this

ui/desktop/app/controllers/scopes/scope/projects/sessions/index.js
```diff
@@ -4,25 +4,58 @@
...
-  get sorted() {
-    const sessions = this.model;
-    // Sort sessions by created time descending (newest on top)
-    const sortedSessions = orderBy(sessions, 'created_time', 'desc');
-    return [
-      // then move active sessions to the top
-      ...sortedSessions.filter((session) => session.isAvailable),
-      // and all others to the end
-      ...sortedSessions.filter((session) => !session.isAvailable),
-    ];
```
ZedLi, line 26: Do we still need this logic in the sorted sessions or is it not needed anymore?

ZedLi (approved):

ui/desktop/app/controllers/scopes/scope/projects/sessions/index.js
lisbet-alvarez: when initially refactoring this, my thoughts were that this logic was unnecessary since users can use the status filter. However, on second thought it doesn't hurt to have additional sorting for when filters are combined.

ZedLi: I think I agree with you that we probably don't need it but we should just check in with design once they're back. I just wanted to check if this was an explicit decision

lisbet-alvarez: I added ICU-12137 to check back in on this.

cameronperera (approved):
LGTM

PR #2054, opened 2023-12-18
feat: storage policy API and mirage changes

DhariniJeeva:
:tickets: [Jira ticket](https://hashicorp.atlassian.net/browse/ICU-12099) and https://hashicorp.atlassian.net/browse/ICU-12100

## Description

This pr adds API and mirage to support the CRUDL of storage policies. 

Screenshot 2023-12-18 at 2 20 50 PM

## How to Test

## Checklist:


addons/api/addon/generated/models/storage-policy.js
```diff
@@ -0,0 +1,67 @@
...
+  @attr('number', {
+    description: 'Current version number of this resource.',
+  })
+  version;
+
+  @attr('string', {
+    description:
+      'The scope id of this policy. This must be defined for policy creation, but\nis otherwise output only.',
+  })
+  scope_id;
```
ZedLi, line 52: Don't we normally have this on the base model?

addons/api/addon/generated/models/storage-policy.js
```diff
@@ -0,0 +1,67 @@
...
+
+  @attr({
```
ZedLi, line 54: Are these read only fields? I thought we have to set them? If so, you need to give a type for these `attr`

addons/api/addon/serializers/storage-policy.js
```diff
@@ -0,0 +1,22 @@
...
+
+import ApplicationSerializer from './application';
+
+export default class StoragePolicySerializer extends ApplicationSerializer {
+  // =methods
+
+  /**
+   * @return {object}
+   */
+  serialize() {
```
ZedLi, line 14: Why do we have these these serialize and normalize functions? Are they just placeholders? I'd prefer to not have them until we actually need them but if we don't need to do extra serializer/normalize logic we should just delete them

addons/api/mirage/factories/storage-policy.js
```diff
@@ -0,0 +1,35 @@
...
+import factory from '../generated/factories/storage-policy';
+import permissions from '../helpers/permissions';
+import { faker } from '@faker-js/faker';
+import generateId from '../helpers/id';
+
+export default factory.extend({
+  id: () => generateId('pst_'),
+
+  authorized_actions: () =>
+    permissions.authorizedActionsFor('storage-policy') || [
```
ZedLi, line 15: Are there permissions for attaching storage policies to scopes?

addons/api/addon/serializers/storage-policy.js
DhariniJeeva: yes these were placeholders, I'll delete these

addons/api/mirage/factories/storage-policy.js
DhariniJeeva: These are permissions in general for CRUDL operations on storage policies.

ZedLi: Right, I was asking if there is also an additional permission we need to be able to attach these policies to a scope specifically?

ZedLi: Interesting, do you know what permission the backend will check when calling `AttachStoragePolicy` on a scope then?

addons/api/addon/transforms/policy-object.js
```diff
@@ -0,0 +1,37 @@
...
+  /**
+   * Normalize a policy object in year
+   * The API returns the policy deletion and retention information in days format,
+   * but we make the conversion to years format by rounding to the closest integer
+   * @param serialized
+   * @returns {object}
+   */
+  deserialize(serialized) {
+    if (!Object.keys(serialized)) return;
+    const years = Math.round(serialized.days / 365);
```
ZedLi, line 18: Wait why do we convert these to years? Don't we want the data to stay in days since that's what gets returned from the API? If we want to display years that should be in the component

addons/api/addon/transforms/policy-object.js
ZedLi, line 1: This object is also not being tracked properly for ember data it looks like, it really seems like we should start using [ember built in tracked](https://github.com/tracked-tools/tracked-built-ins) rather than doing hacks with transforms going forward

ZedLi (changes requested):

addons/api/addon/transforms/policy-object.js
DhariniJeeva: yes, the conversion is solely for the UI. If we handle this logic in the component, we won't be needing these transform methods and I wasn't sure if we can have empty serialize/deserialize here.

addons/api/mirage/factories/storage-policy.js
DhariniJeeva: checking with the backend team as I couldn't find this information anywhere, but we have a story for attaching storage policy and will be tracked with that [story](https://hashicorp.atlassian.net/browse/ICU-12079)

addons/api/addon/transforms/policy-object.js
DhariniJeeva: so it does seem like [transforms](https://guides.emberjs.com/release/models/defining-models/#toc_transforms) are not required?

calcaide: I agree we shouldn't be conrting to years, after I check the [storage policy CRUDL doc definition](https://docs.google.com/document/d/1yR5qkQztWnh8qTjyMXykEHEqJ0PVM3firn-8DYD38MY/edit), all the /policy/ operations expect days on requests and responses.

That said, I also see the problem of not being a primitive and being a custom type. As Zed mentions [on this comment](https://github.com/hashicorp/boundary-ui/pull/2054/files#r1434405029), could we try to use ember built in tracked here? 

If we go the custom attr route and the transform is mandatory to have, we could just return the values without any transformation. Sounds redundant tho

ZedLi: So to be clear we have two options here: 
1. Use a transform and have the fields be properly tracked (the transform should never just be a passthrough since that's just like having no type in the `attr`)
2. Hoist these fields in our serializer/normalizer like we do for nested attributes and have the model just be normal strings/numbers.

addons/api/package.json
```diff
@@ -28,7 +28,9 @@
     "ember-cli-htmlbars": "^6.2.0",
     "ember-cli-mirage": "^2.4.0",
     "ember-data": "~4.12.0",
-    "ember-fetch": "^8.1.1"
+    "ember-fetch": "^8.1.1",
+    "ember-auto-import": "^2.6.3",
+    "tracked-built-ins": "^3.1.1"
```
DhariniJeeva, line 33: I had `tracked-built-ins` in `devdependecy` first but I was getting module not found when used within the `transforms` methods, I couldn't quite figure out why it was happening, so I moved it to dependency. Let me know if we are fine with this or any other way to get this working as a `devdependecy`.

addons/api/addon/transforms/object.js
```diff
@@ -0,0 +1,36 @@
...
+    }
+  }
+
+  /**
+   * @param deserialized
+   * @returns {object}
+   */
+  serialize(deserialized) {
+    if (typeOf(deserialized) === 'object') {
+      const obj = new TrackedObject(deserialized);
```
ZedLi, line 30: I don't think we need to turn this into a tracked object. I think we should be serializing it back into just a normal object, though it might just work fine even without doing that

addons/api/addon/transforms/object.js
```diff
@@ -0,0 +1,36 @@
...
+import { TrackedObject } from 'tracked-built-ins';
+import { typeOf } from '@ember/utils';
+
+export default class ObjectTransform extends Transform {
+  /**
+   * @param serialized
+   * @returns {object}
+   */
+  deserialize(serialized) {
+    if (typeOf(serialized) === 'object') {
```
ZedLi, line 16: Why's this check needed? Shouldn't it always be an object or just undefined?

addons/api/package.json
ZedLi: Yeah I think it needs to be a dependency if we want to use it in API

ZedLi (approved):
Left a few comments but otherwise looks good!

addons/api/addon/transforms/object.js
DhariniJeeva: yeah you are right, this is not needed I added this as more of a generic check, and now replaced this empty check

addons/api/addon/transforms/object.js
```diff
@@ -0,0 +1,34 @@
...
+import Transform from '@ember-data/serializer/transform';
+import { TrackedObject } from 'tracked-built-ins';
+
+export default class ObjectTransform extends Transform {
+  /**
+   * @param serialized
+   * @returns {object}
+   */
+  deserialize(serialized) {
+    if (Object.keys(serialized).length) {
```
ZedLi, line 15: Won't this error out if it's `undefined`?

And shouldn't we still return a tracked object even if it is?

DhariniJeeva: good catch, it will error out, I'll add a fix for that in a separate PR since this is merged. 

I wasn't sure if that's needed when handling an empty object, do you think we would still need to track it?

ZedLi: Yeah I think so, if the API didn't return anything but a user is updating it we still need it to be tracked when we add to the object

PR #2064, opened 2023-12-28
feat: 🎸 Initial setup of IndexedDB

ZedLi:
✅ Closes: https://hashicorp.atlassian.net/browse/ICU-12162

## Description
Initial setup for using indexedDB and our own caching solution for search and filtering.

## How to Test
Start a boundary dev instance with a binary built from `llb-list-pagination-2`.
Confirm there's data in the indexed db storage in your browser.

## Checklist:

### DIFF
--- a/addons/api/addon/adapters/application.js
+++ b/addons/api/addon/adapters/application.js
@@ -115,6 +115,7 @@ export default class ApplicationAdapter extends RESTAdapter.extend(
    * @return {Promise} promise
    */
   async query(store, schema, query) {
+    const queryObject = structuredClone(query);
     let result;
     let data = [];
 
@@ -128,12 +129,12 @@ export default class ApplicationAdapter extends RESTAdapter.extend(
     // Run this loop as long as the response_type is delta,
     // which indicates that there are more items in the list
     do {
-      result = await super.query(store, schema, query);
+      result = await super.query(store, schema, queryObject);
       //add the result items to a data array
       if (result && result.items) {
         data.push(...result.items);
         //pass in the list token for subsequent calls to fetch the remaining list items
-        query.list_token = result.list_token;
+        queryObject.list_token = result.list_token;
       }
     } while (result && result.response_type === 'delta');
 
@@ -211,7 +212,10 @@ export default class ApplicationAdapter extends RESTAdapter.extend(
   handleResponse(status, headers, payload, requestData) {
     if (this.isInvalid(status, headers, payload)) {
       const detailedMessage = this.generatedDetailedMessage(...arguments);
-      const transformedPayload = this.transformValidationErrors(payload);
+      const transformedPayload = this.transformValidationErrors(
+        payload,
+        status,
+      );
       return new InvalidError(transformedPayload.errors, detailedMessage);
     }
     return super.handleResponse(status, headers, payload, requestData);
@@ -269,15 +273,16 @@ export default class ApplicationAdapter extends RESTAdapter.extend(
    * error format.
    *
    * @param {object} payload
+   * @param status
    * @return {object}
    */
-  transformValidationErrors(payload) {
+  transformValidationErrors(payload, status) {
     const fieldErrors = get(payload, 'details.request_fields') || [];
     // Normalize the primary error message into "base", which is an error
     // that applies to the whole model instance
     const baseError = {
       isInvalid: true,
-      status: payload.source,
+      status,
       code: payload.code,
       detail: payload.message,
       source: { pointer: '/data' },
--- a/addons/api/addon/handlers/indexed-db-handler.js
+++ b/addons/api/addon/handlers/indexed-db-handler.js
@@ -1,0 +1,119 @@
+import { modelIndexes } from 'api/services/indexed-db';
+import { inject as service } from '@ember/service';
+import { getOwner, setOwner } from '@ember/application';
+import { pluralize } from 'ember-inflector';
+
+/**
+ * Handler to sit in front of the API request layer
+ * so we can handle any caching with indexedDb first
+ */
+export default class IndexedDbHandler {
+  @service indexedDb;
+
+  constructor(context) {
+    setOwner(this, getOwner(context));
+  }
+
+  async request(context, next) {
+    switch (context.request.op) {
+      case 'query': {
+        const { store, data } = context.request;
+        const { type, query, options: { pushToStore = true } = {} } = data;
+        const supportedModels = Object.keys(modelIndexes);
+        const { db: indexedDb } = this.indexedDb ?? {};
+        const pluralizedType = pluralize(type);
+
+        // Go through normal flow if we don't yet support the model
+        // or if we don't have an indexedDb instance
+        if (!supportedModels.includes(pluralizedType) || !indexedDb) {
+          return next(context.request);
+        }
+
+        const listToken = await indexedDb.tokens.get(pluralizedType);
+        const adapter = store.adapterFor(type);
+        const schema = store.modelFor(type);
+
+        let payload;
+        try {
+          payload = await adapter.query(store, schema, {
+            ...query,
+            scope_id: 'global',
+            recursive: true,
+            list_token: listToken?.token,
+          });
+        } catch (e) {
+          // If we get any error that's not a rate limiting error,
+          // assume the token is not good anymore. We'll delete the token and
+          // clear the DB and try again without a token.
+          // TODO: Probably should check for 400 and bad token error specifically instead?
+          if (e?.errors[0].status !== 429) {
+            await indexedDb[pluralizedType].clear();
+            await indexedDb.tokens.delete(pluralizedType);
+
+            payload = await adapter.query(store, schema, {
+              ...query,
+              scope_id: 'global',
+              recursive: true,
+              list_token: null,
+            });
+          }
+        }
+
+        // Store the token we just got back from the payload
+        await indexedDb.tokens.put({
+          id: pluralizedType,
+          token: payload.list_token,
+        });
+
+        // Remove any records from the DB if the API indicates they've been deleted
+        if (payload.removed_ids?.length > 0) {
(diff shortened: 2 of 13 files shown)


addons/api/addon/adapters/application.js
```diff
@@ -262,15 +266,16 @@ export default class ApplicationAdapter extends RESTAdapter.extend(
...
-  transformValidationErrors(payload) {
+  transformValidationErrors(payload, status) {
     const fieldErrors = get(payload, 'details.request_fields') || [];
     // Normalize the primary error message into "base", which is an error
     // that applies to the whole model instance
     const baseError = {
       isInvalid: true,
-      status: payload.source,
+      status,
       code: payload.code,
```
ZedLi, line 279: It seems like status and code are always null? We don't get back any status or code in the error payload from the backend so I'm not sure if this is a mistake, the API had changed since we implemented this, or I'm missing something. I needed the status though and piped it through regardless. 

Does anyone know about these fields?

addons/api/addon/services/indexed-db.js
```diff
@@ -0,0 +1,114 @@
...
+   * Data is assumed to be in POJOs and that our normalized data are not
+   * ember objects, otherwise they should also be converted.
+   *
+   * Undo the operation by setting `convertBooleans` to false.
+   *
+   * @param {Object} data
+   * @param {Boolean} convertBooleans
+   * @return {{id, type, attributes: {}, relationships: {}}}
+   */
+  normalizeData(data, convertBooleans) {
```
ZedLi, line 96: Unfortunately indexedDb doesn't support booleans as part of their index so I need this hack to convert all booleans to strings so we can index on them.

I'm not exactly sure how big of a performance hit this is having to convert back and forth every time and if it turns out to be significant, we can either not index on boolean fields or create a custom mapper for each resource where we can specify which fields are booleans and save these values as extra fields to be stored as 0 or 1 instead which we can use to index on instead. 

e.g. This is what the data would look like in that case in indexed db. The annoying part is having to now maintain the mapping of models and changing this anytime we change our models that involve booleans, which is why I didn't start off with this. 
```
id: '123',
attributes: {}, // This is the normal data which would include the actual fields
boolean1: 1,
boolean2: 0
```

ui/admin/ember-cli-build.js
```diff
@@ -34,8 +34,9 @@ module.exports = function (defaults) {
       },
     },
     // TODO: Update to 4.12 when deprecations are resolved
+    //   as multiple things break when forcing compatability on > 4.6
     emberData: {
-      compatWith: '4.4',
+      compatWith: '4.6',
```
ZedLi, line 39: It'd be good to get this to 4.12 at some point since there are supposedly [performance increases](https://blog.emberjs.com/ember-released-4-7/) with ember data in 4.7 and beyond

addons/api/addon/handlers/indexed-db-handler.js
```diff
@@ -0,0 +1,119 @@
...
+            ...query,
+            scope_id: 'global',
+            recursive: true,
+            list_token: listToken?.token,
+          });
+        } catch (e) {
+          // If we get any error that's not a rate limiting error,
+          // assume the token is not good anymore. We'll delete the token and
+          // clear the DB and try again without a token.
+          // TODO: Probably should check for 400 and bad token error specifically instead?
```
cameronperera, line 48: I agree that we should be checking for a bad token error.

ui/admin/app/services/session.js
```diff
@@ -0,0 +1,22 @@
...
+   * Extend ember simple auth's handleAuthentication method
+   * so we can hook in and setup the DB after a successful authentication
+   */
+  handleAuthentication() {
+    super.handleAuthentication(...arguments);
+
+    const userId = this.data?.authenticated?.user_id;
+    const hostUrl = this.window.host;
+    if (userId && hostUrl) {
+      this.indexedDb.setup(formatDbName(userId, hostUrl));
```
cameronperera, line 19: What is the difference for calling setup here versus calling it in the route?

cameronperera (approved):
Overall looks good. I left a couple of comments but nothing blocking.

addons/api/addon/handlers/indexed-db-handler.js
ZedLi: Yeah I'm not sure all the different errors we can get yet so we can come back to this

ui/admin/app/services/session.js
cameronperera: Right. Thanks for the reminder.

addons/api/addon/handlers/indexed-db-handler.js
cameronperera: Totally fine with tackling this later. I would like to ask if we need to worry about those other errors? Wouldn't we only want to clear the token and try again if the error was due to a bad token?

ZedLi: I think we need to confirm the different errors we can get from a bad token.

cameronperera: Gotcha. Thank makes sense.

addons/api/addon/adapters/application.js
DhariniJeeva: I'm curious- what do we get from the error payload? 
As far as I remember, we got the code and status from the error payload

addons/api/addon/handlers/indexed-db-handler.js
```diff
@@ -0,0 +1,119 @@
...
+ * so we can handle any caching with indexedDb first
+ */
+export default class IndexedDbHandler {
+  @service indexedDb;
+
+  constructor(context) {
+    setOwner(this, getOwner(context));
+  }
+
+  async request(context, next) {
```
DhariniJeeva, line 17: Didn't we already add an indexed-db-handler in search and filtering LLB?

addons/api/addon/handlers/indexed-db-handler.js
```diff
@@ -0,0 +1,119 @@
...
+    switch (context.request.op) {
+      case 'query': {
+        const { store, data } = context.request;
+        const { type, query, options: { pushToStore = true } = {} } = data;
+        const supportedModels = Object.keys(modelIndexes);
```
DhariniJeeva, line 22: Are there models that we don't support currently?

ui/admin/app/routes/application.js
```diff
@@ -21,6 +22,8 @@ export default class ApplicationRoute extends Route {
   @service intl;
   @service features;
   @service featureEdition;
+  @service indexedDb;
+  @service('browser/window') window;
```
DhariniJeeva, line 26: Is this a new service?

DhariniJeeva (approved):
Looks good, left a couple of non blocking queries

addons/api/addon/adapters/application.js
ZedLi: This is an example error payload we get:
```
{
  "kind": "InvalidArgument",
  "message": "Error in provided request.",
  "details": {
    "request_fields": [
      {
        "name": "name",
        "description": "This field is required."
      },
      {
        "name": "type",
        "description": "Unknown type provided."
      }
    ]
  }
}
```

I checked some different errors and none of them seemed to have a status or code in the payload

addons/api/addon/handlers/indexed-db-handler.js
ZedLi: Yeah, this is based off of that. I started with that but heavily modified it

addons/api/addon/handlers/indexed-db-handler.js
ZedLi: I'm just starting with a whitelist, we technically can support all of them but this way we can start with only the models we want to use

ui/admin/app/routes/application.js
ZedLi: No this is an old one we've been using so we don't directly access window globals. The primary reason is to help mocking out the `window` object when testing

addons/api/addon/handlers/indexed-db-handler.js
```diff
@@ -0,0 +1,119 @@
...
+              list_token: null,
+            });
+          }
+        }
+
+        // Store the token we just got back from the payload
+        await indexedDb.tokens.put({
+          id: pluralizedType,
+          token: payload.list_token,
+        });
```
lisbet-alvarez, line 66: perhaps we should check if there even is a token first in the case of an empty response?

addons/api/addon/handlers/indexed-db-handler.js
```diff
@@ -0,0 +1,119 @@
...
+            await indexedDb.tokens.delete(pluralizedType);
+
+            payload = await adapter.query(store, schema, {
+              ...query,
+              scope_id: 'global',
+              recursive: true,
+              list_token: null,
+            });
+          }
+        }
```
lisbet-alvarez, line 60: im not completely understanding what should happen in the case of there being a rate limit error. So right now if there's a rate limit error, we just continue but further down in line 85 we do a bulkPut operation. Im thinking that maybe that should be skipped in case of rate limit error, or what is the reason for storing the payload?

ZedLi: Yeah it's not great because if we do just get an error, there's not much we _can_ do depending on the error. This check really should only be for a bad token error which should get updated later but if we do get any other error including rate limiting, we probably want to still get the partial results before being rate limited and still update the DB, but otherwise we most likely get nothing and need to bail.

You're probably right and if we do get an error we should just return immediately here, most likely I should just re-throw any thing that's not a bad token error, thoughts?

lisbet-alvarez (approved):
LGTM 🙌

addons/api/addon/adapters/application.js
ZedLi: I double checked, the API was responding with these fields in `0.1.0` but the API got updated in `0.2.0` so we haven't been passing correctly for a while now...

PR #2068, opened 2024-01-03
feat: 🎸 Add indexedDb parser

ZedLi:
✅ Closes: https://hashicorp.atlassian.net/browse/ICU-12164

## Description
Added a parser to be able to convert our filter and search requests to indexedDb requests.

## How to Test
You'll have to manually make requests to the `queryIndexedDb` function by passing in search and filter params.

## Checklist:


addons/api/addon/utils/indexedDbQuery.js
```diff
@@ -0,0 +1,276 @@
...
+const buildIndexedDbCollection = ({
+  key,
+  value,
+  collection,
+  operation,
+  logicalOperator,
+}) => {
+  switch (operation) {
+    case 'contains':
+      if (logicalOperator === 'and') {
```
DhariniJeeva, line 178: Is it possible to move this check outside, as we do this for every since case?

addons/api/addon/utils/indexedDbQuery.js
```diff
@@ -0,0 +1,276 @@
...
+/**
+ * Cleans the object to remove any keys that have a value of an empty array. This makes
+ * it easier so we don't have to worry about empty arrays getting queried in indexed db.
+ */
+const cleanObject = (obj) => {
+  if (!obj) {
+    return {};
+  }
+
+  return Object.entries(obj).reduce((result, [key, value]) => {
```
DhariniJeeva, line 248: curious why you didn't go for a `for.. in` and just simply delete the keys with empty array 
something like 
`for (const key in obj) { if (Array.isArray(obj[key]) && !obj[key].length) delete obj[key] }`

ZedLi: Mainly because I didn't want to modify the array and prefer to return a new one instead. I like to reduce the side effects in these type of helper functions

addons/api/addon/utils/mql-query.js
```diff
@@ -9,8 +9,8 @@
  *
  * @example
  *   const filter = generateMQLFilterExpression({
- *     name: [{ contains: 'search text' }],
- *     scope: [{ equals: 'scope1'}, { equals: 'scope2' }],
+ *     name: { logicalOperator: 'and', values: [{ notEquals: 'search text' }, { notEquals: 'test' }] },
```
DhariniJeeva, line 12: Non-blocker- can we also include an example when there is no `logicalOperator`?

addons/api/addon/utils/indexedDbQuery.js
```diff
@@ -0,0 +1,267 @@
...
+    const [firstKey, firstFilterArrayOrObject] = entries.shift();
+
+    filterCollection = await entries.reduce(
+      async (currCollection, [key, filterArrayOrObject]) => {
+        // Grab the results from the set of one filters to be `AND` together with the rest of the filters
+        const filterResults = await buildInitialWhereClause({
+          table: indexedDb[resource],
+          key,
+          filterArrayOrObject,
+        }).primaryKeys();
```
DhariniJeeva, line 47: I couldn't find much info on this method, is this an indexedDB method?

addons/api/addon/utils/indexedDbQuery.js
```diff
@@ -0,0 +1,267 @@
...
+
+  if (Object.keys(filters).length) {
+    // Remove the first item to use as the initial collection and then reduce the remaining ones
+    const entries = Object.entries(filters);
+    const [firstKey, firstFilterArrayOrObject] = entries.shift();
+
+    filterCollection = await entries.reduce(
+      async (currCollection, [key, filterArrayOrObject]) => {
+        // Grab the results from the set of one filters to be `AND` together with the rest of the filters
+        const filterResults = await buildInitialWhereClause({
```
DhariniJeeva, line 43: does `buildInitialWhereClause` return a promise?

ZedLi: No it returns a [collection](https://dexie.org/docs/Collection/Collection) which you need to call another method on to return the promise such as `toArray()`

addons/api/addon/utils/indexedDbQuery.js
ZedLi: Yup, the docs are [here](https://dexie.org/docs/Collection/Collection.primaryKeys())

addons/api/addon/utils/mql-query.js
ZedLi: Sure I'll add one

calcaide (approved):
Terrific job!!! Amazing implementation! Congrats!!

None blocking at all: do you think could be a good idea to add some unit test (in the future, not necessary now) for the indexedDbQuery util? I.E a unit test for the function `queryIndexedDb`

addons/api/addon/utils/indexedDbQuery.js
ZedLi: Ahhh if the original question is asking why there's an `await`, it's actually because I call `.primaryKeys` on line 47 which _does_ return a promise

ZedLi:
> None blocking at all: do you think could be a good idea to add some unit test (in the future, not necessary now) for the indexedDbQuery util? I.E a unit test for the function `queryIndexedDb`

Yup that was the plan for the next task

addons/api/addon/utils/indexedDbQuery.js
calcaide: Gotcha!! makes sense :wink:

addons/api/addon/utils/indexedDbQuery.js
```diff
@@ -0,0 +1,267 @@
...
+import { modelIndexes } from '../services/indexed-db';
+
+/**
+ * Takes a POJO representing a filter query and builds a query to
+ * query indexedDb with. Note that the first filter is important as that
+ * will use the actual index in indexedDb while any subsequent ones will be less efficient.
+ *
+ * @example
+ *   const filter = buildIndexedDbQuery(indexedDb, type, {
+ *     search: {
```
lisbet-alvarez, line 16: i think the function in this comment should be renamed to `queryIndexedDb`? Another thing, isn't the `query` param in this comment supposed to be something more like this `{ search: 'generated', filters: {...}` because the `search` object in this comment is created by the `getSearchObject` function.

addons/api/addon/utils/indexedDbQuery.js
```diff
@@ -0,0 +1,267 @@
...
+  const { logicalOperator } = filterArrayOrObject;
+
+  // Short circuit if all the filter types were equals or notEquals since indexedDb
+  // can more efficiently query if we pass all values at once
+  const checkAllFilterType = (type) =>
+    filterValueArray.every((obj) => Object.keys(obj)[0] === type);
+  if (checkAllFilterType('equals')) {
+    return table
+      .where(getKey(key))
+      .anyOf(filterValueArray.map((obj) => Object.values(obj)[0]));
```
lisbet-alvarez, line 100: since we know filter type is going to be `'equals'` wouldn't it be simple to re-write it to `.anyOf(filterValueArray.map((obj) => obj['equals']));`. Feel like that's easier to read.  Same goes for line 105.

addons/api/addon/utils/indexedDbQuery.js
ZedLi: Yup you're right, I'll fix the example

addons/api/addon/utils/indexedDbQuery.js
ZedLi: Yeah good point, I think that is easier to read as well, I'll make the change

lisbet-alvarez (approved):
🥳 🙌

cameronperera (approved):

addons/api/tests/unit/utils/indexed-db-query-test.js
ZedLi, line 1: Can we add a few more tests to confirm that it can search boolean fields as well as some of the other operators such as `>` or `<`?

lisbet-alvarez: of course! great suggestion

addons/api/tests/unit/utils/indexed-db-query-test.js
```diff
@@ -0,0 +1,521 @@
...
+    assert.strictEqual(result.length, 5);
+  });
+
+  test('it filters by single field with "gte" as initial operation', async function (assert) {
+    await seedIndexDb(resource, store, indexedDb, this.server);
+    const query = {
+      filters: {
+        scope_id: {
+          logicalOperator: 'and',
+          values: [{ gte: scope1.id }, { equals: scope2.id }],
```
ZedLi, line 278: What are the expected results when using comparison operators on IDs? Aren't they strings?

lisbet-alvarez: Yeah they are strings scope1.id has the value of `p_123` and scope2.id has the value of `p_456` so scope1.id will always be less than scope2.id. I can remove that second operator since it's not necessary, I thought we should test multiple operations on a single filter. However, that was before adding the other test cases where the operation is second and not first.

ZedLi: Right but I suppose my question is should we be comparing strings since that's what the IDs just are, can we use fields that are actually numbers? I don't want us to be accidentally testing something that's like `"2" > "10" === true`

lisbet-alvarez: Currently, none of the modelIndexes have an index that is a number type. Do we want to add an additional index field or add another resource that will be indexed on a number field ?

ZedLi: I was thinking of just adding another resource, we can just add a TODO to come back to this if you'd like to wait

lisbet-alvarez: yeah i can come back and update these since I'm not sure what fields we are indexing on for the other resources.

addons/api/tests/unit/utils/indexed-db-query-test.js
```diff
@@ -0,0 +1,490 @@
...
+      await queryIndexedDb(indexedDb.db, resource, query);
+    } catch (e) {
+      assert.strictEqual(
+        e.message,
+        'Contains is not supported as the first filter.',
+      );
+    }
+  });
+
+  test('it can filter with "contains" as subsequent operation', async function (assert) {
```
ZedLi, line 212:
```suggestion
  test('it cannot filter with "contains" as subsequent operation', async function (assert) {
```

PR #2069, opened 2024-01-03
fix: 🐛 transform

DhariniJeeva:
fixes object transform

## How to Test

## Checklist:

### DIFF
--- a/addons/api/addon/transforms/object.js
+++ b/addons/api/addon/transforms/object.js
@@ -21,10 +21,9 @@ export default class ObjectTransform extends Transform {
    * @returns {object}
    */
   serialize(deserialized) {
-    if (deserialized && Object.keys(deserialized).length) {
-      return deserialized;
-    } else {
+    if (!deserialized) {
       return {};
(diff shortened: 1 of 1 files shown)


addons/api/addon/transforms/object.js
```diff
@@ -12,20 +12,16 @@ export default class ObjectTransform extends Transform {
...
+    return obj;
   }
 
   /**
    * @param deserialized
    * @returns {object}
    */
   serialize(deserialized) {
-    if (Object.keys(deserialized).length) {
+    if (deserialized && Object.keys(deserialized).length) {
```
ZedLi, line 24: If it's null it might be fine to just return null? Do we need to check at all since I assume the API would want null if there's nothing rather than an empty object

ZedLi (approved):

addons/api/addon/transforms/object.js
DhariniJeeva: yeah that should be fine, I'll update

PR #2073, opened 2024-01-05
Icu 12027 let the user know if the dc isnt compatible with the controller and link them to the previous version

cameronperera:
:tickets: [Jira ticket](https://hashicorp.atlassian.net/browse/ICU-12027)

Screenshot 2024-01-08 at 2 02 08 PM

If the download fails we will show the user an error alert and direct them to the releases page for 1.7.1:
Screenshot 2024-01-08 at 2 02 46 PM

## How to Test

use command: ENABLE_MIRAGE=false yarn start:desktop

llb-list-pagination-2 now has pagination for scopes so if using the most recent changes from that branch will not show this new modal. So use any public version of the binary will allow for the new modal to appear.

## Checklist:


addons/api/addon/adapters/application.js
```diff
@@ -120,6 +120,10 @@ export default class ApplicationAdapter extends RESTAdapter.extend(
 
     // Run this loop as long as the response_type is delta,
     // which indicates that there are more items in the list
```
ZedLi, line 122: This comment goes for the do loop, we should add another comment explaining why this hack is needed

addons/api/mirage/config.js
```diff
@@ -73,7 +73,11 @@ function routes() {
           return scope.scope ? scope.scope.id === scope_id : false;
         });
       }
-      return resultSet.filter(makeBooleanFilter(filter));
+      const filteredResultSet = resultSet.filter(makeBooleanFilter(filter));
+      // add a list_token to the response to simulate pagination
+      // without having list_token, the unsupported controller modal
+      // will show when running with mirage
+      return { items: filteredResultSet.models, list_token: '12345' };
```
ZedLi, line 80: Should this change be made for all routes?

Also can we add some acceptance tests now for this PR as well?

ui/desktop/app/controllers/scopes.js
```diff
@@ -0,0 +1,64 @@
...
+
+  metaData;
+  url;
+  @tracked error = false;
+
+  // =methods
+
+  @action
+  @loading
+  async downloadAndInstallCompatiableVersion() {
```
ZedLi, line 24:
```suggestion
  async downloadCompatibleVersion() {
```

We're not really installing it for them

ui/desktop/app/routes/scopes.js
```diff
@@ -6,14 +6,35 @@
...
 export default class ScopesRoute extends Route {
   // =services
 
   @service store;
+  @service ipc;
 
   // =methods
 
+  async beforeModel() {
+    let isClientDaemonSupported = true;
```
ZedLi, line 20: Should this by default be set to `false`? Since if in the try catch if it errors out now it will be defaulted to `true`? Is that what we want?

ui/desktop/app/routes/scopes.js
```diff
@@ -23,9 +44,16 @@ export default class ScopesRoute extends Route {
    * @param {string} params.scope_id
    * @return {Promise{[ScopeModel]}}
    */
-  model() {
+  async model() {
     // NOTE:  In the absence of a `scope_id` query parameter, this endpoint is
     // expected to default to the global scope, thus returning org scopes.
-    return this.store.query('scope', {}).catch(() => A([]));
+    let scopes = A([]);
```
ZedLi, line 50: Does this need to be an ember array still?

ui/desktop/app/routes/scopes.js
```diff
@@ -23,9 +44,16 @@ export default class ScopesRoute extends Route {
...
+    const isClientDaemonSupported = get(this, 'isClientDaemonSupported');
+
+    if (isClientDaemonSupported) {
```
ZedLi, line 53: I don't understand why this check is needed? Don't we need to retrieve the scopes regardless?

ui/desktop/app/routes/scopes.js
ZedLi, line 20: This variable isn't quite accurate. We're really checking to see if the controller supports pagination and we're inferring if client daemon is supported by that. In the future, we can handle not having the client daemon up but we will always need pagination.

ui/desktop/app/styles/app.scss
```diff
@@ -711,3 +711,14 @@ $chevron-down: url("data:image/svg+xml;utf-8,<svg viewBox='0 0 24 24' fill='%231
...
   }
 }
+
+// Unsupported controller modal
+.unsupported-modal-body {
+  display: grid;
+  gap: 1rem;
+}
+
+.hds-dismiss-button {
```
ZedLi, line 721: Where is this used? If it's internal to HDS we're setting _all_ of them to not display now. 

Is there a reason why it doesn't disappear automatically with `isDismissDisabled` set to true? Should we even be removing this from a design perspective or is it a bug with HDS?

ui/desktop/app/routes/scopes.js
```diff
@@ -23,9 +44,16 @@ export default class ScopesRoute extends Route {
...
+      scopes = await this.store.query('scope', {}).catch(() => A([]));
+    }
+
+    return { scopes, isClientDaemonSupported };
```
ZedLi, line 57: Does this have any effect on further downstream routes that would call `this.modelFor`?

ui/desktop/electron-app/src/index.js
```diff
@@ -127,7 +127,10 @@ const createWindow = (partition, closeWindowCB) => {
   // Opens external links in the host default browser.
   // We just allow boundaryproject.io domain to open on external window (for now).
```
ZedLi, line 128: This comment needs to be updated

ui/desktop/app/controllers/scopes.js
```diff
@@ -0,0 +1,64 @@
...
+      const urlParts = this.url.split('/');
+      const fileName = urlParts[urlParts.length - 1];
+
+      // Create an anchor element to trigger the download
+      const link = document.createElement('a');
+      link.href = window.URL.createObjectURL(blob);
+      link.download = fileName;
+
+      // Simulate click to trigger download
+      link.click();
```
ZedLi, line 53: Is this just to have the download happen in the desktop client for a better UX? Is there a reason we couldn't just have the button link to the release url and it opens in the user's browser?

ZedLi (changes requested):

addons/api/addon/adapters/application.js
cameronperera: Good catch. Updated.

ui/desktop/app/controllers/scopes.js
cameronperera: Absolutely right. Fixed.

ui/desktop/app/controllers/scopes.js
cameronperera: Let me explore that change. That would be more what I would expect as a user.

ui/desktop/app/routes/scopes.js
cameronperera: I have reverted these changes.

ui/desktop/app/routes/scopes.js
cameronperera: Updated to `isPaginationSupported`. Thanks for the feedback.

ui/desktop/app/styles/app.scss
cameronperera: Hds does not hide the button when `isDismissDisabled` is true. I originally decided to hide it to match the designs. However, I was getting a focus-trap error which is due to hiding the dismiss button. Hds docs mention it [here](https://helios.hashicorp.design/components/modal?tab=code#focus-and-focus-trap).
Due to the error, I removed this style for now.

ui/desktop/app/routes/scopes.js
cameronperera: It did but I have reverted this so no longer an issue.

ui/desktop/app/controllers/scopes.js
cameronperera: This process has been updated.

addons/api/mirage/config.js
cameronperera: If we don't make this change here, the tests for desktop will always show the modal. I was thinking we make the change here then update the response on in the test cases that need to test the modal. Happy to discuss alternative methods though.

ui/desktop/app/routes/scopes.js
```diff
@@ -6,14 +6,35 @@
 import Route from '@ember/routing/route';
 import { A } from '@ember/array';
 import { inject as service } from '@ember/service';
+import { set, get } from '@ember/object';
```
ZedLi, line 9: Are these `set` and `get` required instead of just accessing like normal?

ui/desktop/app/routes/scopes.js
```diff
@@ -23,9 +44,53 @@ export default class ScopesRoute extends Route {
...
+        downloadError = true;
+      }
+    }
+
+    controller.set('downloadLink', downloadLink);
+    controller.set('downloadError', downloadError);
+  }
+
+  extractOsSpecificUrl(metaData, os) {
+    return metaData.builds.find((build) => build.os === os).url;
```
ZedLi, line 94: Is the order guaranteed? How do we know if we're grabbing the `zip` or `dmg` for example

addons/api/mirage/config.js
ZedLi: Yup, I meant we should probably make it so all routes return this data and think about how to accomplish that but we can tackle that in the next PR

ui/desktop/app/routes/scopes.js
cameronperera: Yeah that is a good point. I will update this to default to `false`.

ui/desktop/app/routes/scopes.js
```diff
@@ -23,9 +42,59 @@ export default class ScopesRoute extends Route {
...
+    super.setupController(...arguments);
+
+    controller.set('isPaginationSupported', this.isPaginationSupported);
+
+    let downloadLink;
+    let downloadError = false;
+
+    if (!this.isPaginationSupported) {
+      const metaDataUrl =
+        'https://api.releases.hashicorp.com/v1/releases/boundary-desktop/1.7.1';
```
calcaide, line 66: None blocking comment: thinking loud here, take with a pinch of salt.

The version specified in `metaDataUrl` which is `1.7.1` has to be always the latest version of our DC client or as long as is `1.7.1` is fine? asking to know if this is something we should add to our release guide. I.e: Update the version of suggested DC for unsupported pagination purposes.
Thoughts?

ui/desktop/app/templates/scopes.hbs
```diff
@@ -3,4 +3,51 @@
   SPDX-License-Identifier: MPL-2.0
 }}
 
-{{outlet}}
\ No newline at end of file
+{{outlet}}
+{{#unless this.isPaginationSupported}}
```
calcaide, line 7: Comment for sake of visibility:

Already mentioned to @cameronperera in slack and working on it, this comment is for visibility purposes.

In linux, the `<modal>` overlay, does not allow the user to click the close app button on the header, top - right corner (attached screenshot).
Screenshot from 2024-01-11 16-15-07

The user can always go to the desktop dock, right click on Boundary app icon and close it (attached screenshot)
Screenshot from 2024-01-11 16-21-24

Not a big deal. I suggest since we are tight on delivery time, perhaps we can leave this for after release work?

ui/desktop/tests/acceptance/scopes-test.js
```diff
@@ -235,27 +249,95 @@ module('Acceptance | scopes', function (hooks) {
...
+
+    await waitFor('[data-test-unsupported-controller]');
+    assert.dom('[data-test-unsupported-controller]').exists();
+    assert
+      .dom('[data-test-download-link]')
+      .hasAttribute('href', 'windows.fake.download.zip');
+  });
+
+  test('pagination is not supported - mac build', async function (assert) {
+    this.ipcStub.withArgs('checkOS').returns({
```
calcaide, line 276: let's gooo!!! nice use of stubs :wink:

calcaide (approved):
LGTM!!! awesome job @cameronperera 

I left some none blocking comments + information about the issue we found on linux, but not a big deal :wink:

ui/desktop/app/routes/scopes.js
cameronperera: Updated.

ui/desktop/app/routes/scopes.js
cameronperera: Removed use of `set` and `get`.

ui/desktop/app/routes/scopes.js
cameronperera: Great question. I updated this to also grab the correct url by file extension. If it fails to find a url from the metaData, it will throw an error.

ui/desktop/app/templates/scopes.hbs
cameronperera: Thanks for noticing this. I will look into this but like you said, probably not a blocker for release.

ZedLi (approved):

ui/desktop/app/routes/scopes.js
cameronperera: This url will not get updated again. We are using the 1.7.1 release of the boundary desktop because that will be the last desktop app that will support older controllers that do not support pagination. We will likely need to update how the autoUpdater to stop let the user know they need to update if the controller is an older controller as well.

calcaide: @cameronperera Thank you for the clarification. That makes sense :smiley:

PR #2074, opened 2024-01-05
feat: 🎸 Add loading indicators for initial loads of resources

ZedLi:
✅ Closes: https://hashicorp.atlassian.net/browse/ICU-12025

## Description
Added loading indicators for the initial long loads coming from the client daemon. Originally, I was thinking I would have to poll and check the status of the daemon to know if we got any results back since normally the daemon returns cached results immediately but calling with `force_refresh` is blocking and waits enough time to get all the results back.

https://github.com/hashicorp/boundary-ui/assets/5783847/299bbbde-fc13-4c4f-bc5f-55e7d6279560

## How to Test
Use a binary with a client daemon and add at least 10k targets. You should be able to see a load time that's noticeable now. 

## Checklist:

### DIFF
--- a/addons/core/addon/styles/addon.scss
+++ b/addons/core/addon/styles/addon.scss
@@ -38,23 +38,6 @@
   .hds-table {
     margin-bottom: 1rem;
   }
-
-  &-loading {
-    display: flex;
-    flex-direction: column;
-    align-items: center;
-    gap: 1rem;
-    margin-top: 4rem;
-
-    &-title {
-      gap: 1rem;
-      display: flex;
-    }
-
-    > * {
-      width: 350px;
-    }
-  }
 }
 
 // Flyout with table
--- a/addons/core/addon/styles/addon.scss
+++ b/addons/core/addon/styles/addon.scss
@@ -20,3 +20,47 @@
     }
   }
 }
+
+// Search & Filtering
+.search-filtering {
+  .hds-application-state {
+    margin-top: 3rem;
+  }
+
+  .hds-segmented-group {
+    margin-bottom: 1rem;
+
+    .hds-dropdown-toggle-button__text {
+      white-space: nowrap;
+    }
+  }
+
+  .hds-table {
+    margin-bottom: 1rem;
+  }
+
+  &-loading {
+    display: flex;
+    flex-direction: column;
+    align-items: center;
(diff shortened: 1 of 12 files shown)


addons/core/addon/styles/addon.scss
```diff
@@ -20,3 +20,47 @@
     }
   }
 }
+
+// Search & Filtering
```
ZedLi, line 24: I moved this to core since I need these styles in admin anyways and this would help with merge conflicts later

ui/desktop/app/templates/scopes/scope/projects/targets/loading.hbs
```diff
@@ -0,0 +1,14 @@
...
+      {{t 'states.loading.title'}}
+    </Hds::Text::Display>
+  </div>
+  <Hds::Text::Body @tag='p'>
+    {{t
+      'states.loading.description'
+      resource=(t 'resources.target.title_plural')
+    }}
+  </Hds::Text::Body>
+</div>
```
cameronperera, line 14: non-blocking:
Could we use [Hds::ApplicationState](https://helios.hashicorp.design/components/application-state?tab=code#empty-state-with-header-icon) for here and the `/sessions/loading` template?

cameronperera (approved):
Left a comment but as noted... not blocking.

ui/desktop/app/templates/scopes/scope/projects/targets/loading.hbs
ZedLi: ahh good call, totally forgot about the existence of this component, I'll update it

addons/core/translations/states/en-us.yaml
```diff
@@ -14,3 +14,6 @@ active-public: Public
 recording: Recording
 completed: Completed
 failed: Failed
+loading:
```
ZedLi, line 17: I'm not sure what the best place was for this so I just added it to `states`, let me know if you have a better idea for this

DhariniJeeva (approved):

addons/core/translations/states/en-us.yaml
cameronperera: This seems like a good place for this IMO.

PR #2079, opened 2024-01-08
feat: 🎸 create storage policy

DhariniJeeva:
:tickets: [Jira ticket](https://hashicorp.atlassian.net/browse/ICU-12082)

## Description

This PR adds a new form for storage policy creation. It also includes a helper to make the days-to-year conversion. 

Screenshot 2024-01-08 at 10 37 53 AM

Screenshot 2024-01-08 at 10 37 40 AM

Screenshot 2024-01-08 at 10 37 04 AM

Screenshot 2024-01-08 at 10 36 33 AM

## How to Test

## Checklist:

- ~[] I have added before and after screenshots for UI changes~


addons/api/mirage/serializers/policy.js
```diff
@@ -0,0 +1,10 @@
+/**
+ * Copyright (c) HashiCorp, Inc.
+ * SPDX-License-Identifier: MPL-2.0
+ */
+
+import ApplicationSerializer from './application';
+
+export default ApplicationSerializer.extend({
+  modelName: 'policy',
```
DhariniJeeva, line 9: Even though we don't use a serializer for policies, I was getting `a mirage model not registered error` from the mirage, so I had to add this file. let me know if there's another to skip this

ZedLi: I think we still need it for mirage but I haven't dug into that

ui/admin/app/components/form/policy/index.js
```diff
@@ -0,0 +1,172 @@
...
+  @action
+  handleOverridableToggle(field) {
+    this[field] = !this[field];
+  }
+
+  /**
+   * Set retain for input value to a tracked property
+   */
+  @action
+  handleRetainForInput({ target: { value } }) {
```
ZedLi, line 130: Aren't these the same handlers in essence, couldn't we just take in a variable that says which value we should update?

ui/admin/app/components/form/policy/policy-selection/index.hbs
```diff
@@ -0,0 +1,75 @@
...
+    <F.Control>
+      <Hds::SegmentedGroup as |S|>
+        <S.TextInput
+          @value={{@inputValue}}
+          disabled={{@disabled}}
+          aria-label={{t 'resources.policy.label'}}
+          {{on 'input' @onInput}}
+          data-input-type={{@customInputName}}
+        />
+        <S.Generic>
```
ZedLi, line 52: Shouldn't the user be able to select days or years to input?

ui/admin/app/components/form/policy/policy-selection/index.hbs
```diff
@@ -0,0 +1,75 @@
...
+{{#if @isCustomInputVisible}}
+  <Hds::Form::Field class='policy-custom-input' @layout='vertical' as |F|>
+    <F.Label>{{t (concat 'resources.policy.titles.' @customInputName)}}
+    </F.Label>
+    <F.Control>
+      <Hds::SegmentedGroup as |S|>
+        <S.TextInput
+          @value={{@inputValue}}
+          disabled={{@disabled}}
+          aria-label={{t 'resources.policy.label'}}
```
ZedLi, line 48: What translation is this referring to, I don't think this is pointing to the correct label.

ui/admin/app/components/form/policy/index.js
```diff
@@ -0,0 +1,172 @@
...
+
+  //actions
+  /**
+   * Show custom text field when custom option is selected
+   */
+
+  @action
+  handlePolicyTypeSelection({ target: { value: selectedVal, name: policy } }) {
+    switch (policy) {
+      case 'retention_policy':
```
ZedLi, line 92: I feel like this should be just an `if` statement since it can only be two options right?

ZedLi (changes requested):

ui/admin/app/components/form/policy/index.js
```diff
@@ -0,0 +1,172 @@
...
+   */
+  @action
+  cancel() {
+    this.args.cancel();
+    this.is_custom_retention_selected = this.args.model.retain_for?.days > 0;
+    this.is_custom_deletion_selected = this.args.model.delete_after?.days > 0;
+    this.retain_for_overridable = this.args.model.retain_for?.overridable;
+    this.delete_after_overridable = this.args.model.delete_after?.overridable;
+    this.retain_days = this.args.model.retain_for?.days;
+    this.delete_days = this.args.model.delete_after?.days;
```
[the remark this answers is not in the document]
DhariniJeeva: Maybe I misunderstood this, but since this is an object type in which `ember-data` doesn't roll back, I had to store these in tracked properties and then construct the model object at the end. 

Wouldn't the transform only be reached when we hit submit, not during the editing state?

ui/admin/app/components/form/policy/policy-selection/index.hbs
```diff
@@ -0,0 +1,75 @@
...
+        <S.Generic>
+          <Hds::Badge
+            class={{if @disabled 'disabled'}}
+            @text={{t 'resources.policy.titles.days'}}
+            @color='neutral'
+            @type='outlined'
+          />
+        </S.Generic>
+      </Hds::SegmentedGroup>
+      {{#if @isGlobal}}
```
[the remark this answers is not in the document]
DhariniJeeva: We don't show these for projects. It is only available at the global and org level

ui/admin/app/components/form/policy/index.js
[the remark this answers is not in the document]
ZedLi: I meant I thought ember should roll back to its previous reference which was your original tracked object, not that it would transform again

ui/admin/app/components/form/policy/policy-selection/index.hbs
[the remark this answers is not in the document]
ZedLi: Gotcha sounds good then

ui/admin/app/components/form/policy/index.js
DhariniJeeva: yeah I was thinking to update this to an If, will do

addons/core/translations/resources/en-us.yaml
```diff
@@ -925,10 +925,32 @@ policy:
...
     delete: Delete Storage Policy
   messages:
     none:
       title: No storage policies yet
       description: Storage policies define how long to store recordings before they are automatically deleted. We recommend checking your data retention policies for compliance requirements.
       link: Create a new storage policy
+  form:
+    retention_policy:
+      label: Retention Policy
+      help: Specifies how long a recording must be stored
```
lisbet-alvarez, line 939: Small, nitpick is that the help translations usually end with a period (.)
```suggestion
      help: Specifies how long a recording must be stored.
```

ui/admin/app/components/form/policy/index.js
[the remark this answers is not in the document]
DhariniJeeva: You are right; it does maintain previous references. I believe I did not use the models directly in the template earlier, which made me think the objects weren't getting the previous references.

ui/admin/app/components/form/policy/policy-selection/index.hbs
DhariniJeeva: for now this won't be a list, I'm talking with design and backend, I'll update this later based on the decision from that convo

ui/admin/app/components/form/policy/index.js
```diff
@@ -0,0 +1,111 @@
...
+
+const DELETION_POLICY = { do_not_delete: 0, custom: 1 };
+
+export default class FormPolicyComponent extends Component {
+  //attributes
+
+  @tracked retainForOverridable = this.args.model.retain_for?.overridable;
+  @tracked deleteAfterOverridable = this.args.model.delete_after?.overridable;
+  @tracked isCustomRetentionSelected = this.args.model.retain_for?.days > 0;
+  @tracked isCustomDeletionSelected = this.args.model.delete_after?.days > 0;
```
ZedLi, line 24: Are these still needed if you can now directly use the model?

DhariniJeeva: We would still need to keep it there because I'm using these fields to display the custom inputs component for both `delete after` and `retain for`. If I were to move these to the policy-selection component, I won't be able to toggle and use a generic input to show/hide custom inputs.

ZedLi: Hmm fair but don't you already do a check inside `handlePolicyTypeSelection` for the policy type based on name? What's the difference then of just checking which one you need based on name again?

DhariniJeeva: 💯 yes, that is possible; I also moved the toggle checks along with these. we would only have some methods on the main form, and then the generic one will have the actions.

ui/admin/app/components/form/policy/policy-selection/index.js
```diff
@@ -0,0 +1,99 @@
+import Component from '@glimmer/component';
+import { action } from '@ember/object';
+import { tracked } from '@glimmer/tracking';
+
+export default class FormPolicySelectionComponent extends Component {
+  //attributes
+  @tracked isCustomRetentionSelected = this.args.model.retain_for?.days > 0;
+  @tracked isCustomDeletionSelected = this.args.model.delete_after?.days > 0;
+  @tracked retainForOverridable = this.args.model.retain_for?.overridable;
+  @tracked deleteAfterOverridable = this.args.model.delete_after?.overridable;
```
ZedLi, line 10: Also could these have been getters since they're already using a tracked variable?

ui/admin/app/components/form/policy/policy-selection/index.js
```diff
@@ -0,0 +1,99 @@
...
+      ...this.args.model[field],
+      overridable: !this.args.model[field].overridable,
+    };
+  }
+
+  /**
+   * Show custom text field when custom option is selected
+   */
+  @action
+  handlePolicyTypeSelection({ target: { value: selectedVal, name: policy } }) {
```
ZedLi, line 64: I notice when I save the default dropdowns without touching on a new create, I get null
image

but patching the exact same thing without touching it gets me an empty object:
image

and if I change the dropdown and then back I get the strings `"0"`
image

and yet the forever option is the integer `0` and still string `-1` for do not delete
image

What are the correct values we should be sending?

ui/admin/app/components/form/policy/index.hbs
```diff
@@ -0,0 +1,81 @@
...
+    @options={{this.listRententionOptions}}
+    @customInputName='retain_for'
+    @inputValue={{@model.retain_for.days}}
+    @selectedOption={{this.selectRetentionPolicyType}}
+  />
+
+  {{! deletion policy }}
+  <Form::Policy::PolicySelection
+    @name='deletion_policy'
+    @disabled={{form.disabled}}
```
lisbet-alvarez, line 66: One thing i'm noticing from the designs that is missing is that if the retention policy is selected to be "forever" then the deletion policy field should be disabled.
Screenshot 2024-01-11 at 4 42 32 PM

ui/admin/app/components/form/policy/policy-selection/index.js
DhariniJeeva: Interestingly, the API doesn't throw any errors.. but the value they return is in `int` and no object if we don't send a value or an empty obj or null. I'll update to be consistent

ui/admin/app/components/form/policy/index.hbs
DhariniJeeva: ah good catch, I only limited the list, now it should be disabled too

ui/admin/app/styles/app.scss
```diff
@@ -951,3 +951,37 @@
...
+      width: 45%;
+    }
+
+    .hds-form-field--layout-flag {
+      width: 40%;
+      margin-bottom: 0;
+    }
+  }
+
+  .hds-badge {
```
ZedLi, line 976: Could you have used just a normal segmented group select with only one option but made it `readonly` so the user can't change it rather than using a badge and restyling it?

ui/admin/app/components/form/policy/policy-selection/index.js
```diff
@@ -0,0 +1,114 @@
...
+    } else {
+      return false;
+    }
+  }
+  /**
+   * Select options return type is string and we want the `days` in integer format
+   * @type {Number}
+   */
+
+  convertStringToInt(str) {
```
ZedLi, line 61: I feel like this method doesn't do much and you could have just called `Number(str)` directly

ZedLi (approved):
Left a few more comments but looks overall looks good

cameronperera (approved):
LGTM

ui/admin/app/templates/scopes/scope/policies/policy/index.hbs
```diff
@@ -1,4 +1,39 @@
...
+  <page.header>
+    <h2>
+      {{t 'resources.policy.title'}}
+    </h2>
+    <Copyable
+      @text={{@model.id}}
+      @buttonText={{t 'actions.copy-to-clipboard'}}
+      @acknowledgeText={{t 'states.copied'}}
+    >
+      <code>{{@model.id}}</code>
```
lisbet-alvarez, line 20:
```suggestion
      <Hds::Text::Code>{{@model.id}}</Hds::Text::Code>
```

addons/core/addon/helpers/format-day-year.js
```diff
@@ -24,12 +24,14 @@ export default class extends Helper {
    */
 
   compute([days]) {
-    const remainder = days % 365;
-    if (remainder === 0) {
-      const numberOfYears = days / 365;
-      return this.intl.t('titles.year', { numberOfYears });
-    } else {
-      return this.intl.t('titles.days', { days });
+    if (days > 0) {
```
calcaide, line 27: None blocking comment, just checking: I did not see test for this helper within this PR, but [I see it here](https://github.com/hashicorp/boundary-ui/commit/d50bc81c5363f6dfdf4cbf18b0f69892936edc2b#diff-6d1c8a8826f8f8018d7586756dc6bfad32bdfc69263c6d91c138738abfc98a73)

ui/admin/app/styles/app.scss
DhariniJeeva: Hmm, In that case, I would still need to update the style to remove the arrows on the input

addons/core/addon/helpers/format-day-year.js
DhariniJeeva: not sure I follow, the helper has been merged and is already in main. I added an additional check to the helper

DhariniJeeva: Resolving this to get this PR merged. Will address it later if it's still a concern

ui/admin/app/styles/app.scss
DhariniJeeva: resolving this to get this PR merged. will address it later if it's still a concern

ui/admin/app/templates/scopes/scope/policies/new.hbs
```diff
@@ -7,4 +7,25 @@
...
\ No newline at end of file
+/>
+
+<Rose::Layout::Page as |page|>
+  <page.breadcrumbs>
+    <Breadcrumbs::Container />
+  </page.breadcrumbs>
+
+  <page.header>
+    <h2>
```
calcaide, line 18: None blocking comment: we could have used the HDS::Text component here, but I am planning a refactor of the pageHeader all together, [here the poc](https://github.com/hashicorp/boundary-ui/pull/2037). So all good :smile:

ui/admin/app/templates/scopes/scope/policies/policy/index.hbs
```diff
@@ -1,4 +1,39 @@
...
+<Rose::Layout::Page as |page|>
+  <page.breadcrumbs>
+    <Breadcrumbs::Container />
+  </page.breadcrumbs>
+
+  <page.header>
+    <h2>
+      {{t 'resources.policy.title'}}
+    </h2>
+    <Copyable
```
calcaide, line 15: I would suggest to use `Hds::Copy::Snippet` here. If you want, to keep the scope of this PR as it is, perhaps we can add a task to review these [here](https://hashicorp.atlassian.net/browse/ICU-9285)?

calcaide (approved):
Sorry for being late! I left some comments. Thank you for the work!

ui/admin/app/templates/scopes/scope/policies/policy/index.hbs
DhariniJeeva: thanks for the catch. addressed [hrer](https://github.com/hashicorp/boundary-ui/pull/2090)

PR #2080, opened 2024-01-09
chore: 🤖 add attach and detach storage policy permissions

DhariniJeeva:
✅ Closes: https://hashicorp.atlassian.net/browse/ICU-12150

:tickets: [Jira ticket](https://hashicorp.atlassian.net/browse/ICU-12150)

## Description

This PR adds the required permissions in the scope model 

## How to Test

## Checklist:

### DIFF
--- a/addons/api/addon/abilities/scope.js
+++ b/addons/api/addon/abilities/scope.js
@@ -1,0 +1,28 @@
+/**
+ * Copyright (c) HashiCorp, Inc.
+ * SPDX-License-Identifier: MPL-2.0
+ */
+
+import ModelAbility from './model';
+
+/**
+ * Provides abilities for scopes.
+ */
+export default class ScopeAbility extends ModelAbility {
+  // =permissions
+  /**
+   * Permission that checks whether a policy can be attached to a scope
+   * @type {boolean}
+   */
+  get canAttachStoragePolicy() {
+    return this.hasAuthorizedAction('attach-storage-policy');
+  }
+
+  /**
+   * Permission that checks whether a policy can be detached to a scope
+   * @type {boolean}
+   */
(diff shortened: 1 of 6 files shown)


addons/api/addon/abilities/scope.js
ZedLi, line 1: Shouldn't this be in the storage policy ability and not the scope ability? Don't we normally keep it in the pertinent model and we would pass in the scope model?

ui/admin/app/abilities/scope.js
```diff
@@ -47,4 +47,25 @@ export default class OverrideScopeAbility extends OverrideModelAbility {
         return super.canList;
     }
   }
+
+  /**
+   * Attaching a policy is allowed only if the feature flag is enabled
+   * @type {boolean}
+   */
+  get canAttachStoragePolicy() {
+    console.log(this, 'THIS');
```
ZedLi, line 56: console log here

addons/api/addon/abilities/scope.js
DhariniJeeva: hmm, I was looking at the target model and user model where we check for abilities in those directly, so I followed the same pattern, do you have any examples in mind for reference?

ZedLi: For example, list or create permissions are also on the parent resource, which is often a scope but we use the base model for that. Create worker led permissions are still on the worker model and not the scope model

ZedLi: But I also guess this is the first time we've had real permissions check that aren't collection actions on a scope, thinking again about it I suppose it makes sense?

I think the real inconsistency is how we handle authorized actions vs collection actions 🤷

ZedLi (approved):

PR #2083, opened 2024-01-10
feat: 🎸 target list will update according to org scope selected

cameronperera:
✅ Closes: https://hashicorp.atlassian.net/browse/ICU-12205

:tickets: [Jira ticket](https://hashicorp.atlassian.net/browse/ICU-12205)

## Description

If a user selects an org it will now only show targets that belong to that org scope.

https://github.com/hashicorp/boundary-ui/assets/29386339/e09445a0-f280-41af-a5a8-4b6240502ebf

## How to Test

Run `ENABLE_MIRAGE=false BYPASS_CLI_SETUP=true yarn start:desktop` to run against a local binary built off the main branch of the boundary repo. Create a second org with a project and target for listing purposes. After loading all targets when logging in from the global scope, select the second org in the header of the desktop client. The targets listed should only belong to the org you selected from the header of the desktop client.

## Checklist:

### DIFF
--- a/ui/desktop/app/routes/scopes/scope/projects/targets/index.js
+++ b/ui/desktop/app/routes/scopes/scope/projects/targets/index.js
@@ -84,11 +84,18 @@ export default class ScopesScopeProjectsTargetsIndexRoute extends Route {
     transition,
   ) {
     await this.getAllTargets(transition);
+    const projects = this.modelFor('scopes.scope.projects');
+    const projectIds = projects.map((project) => project.id);
 
     const filters = { scope_id: [], id: { values: [] }, type: [] };
     scopes.forEach((scope) => {
       filters.scope_id.push({ equals: scope });
     });
+    if (scopes.length === 0) {
+      projectIds.forEach((projectId) => {
+        filters.scope_id.push({ equals: projectId });
+      });
+    }
     types.forEach((type) => {
       filters.type.push({ equals: type });
     });
@@ -124,7 +131,6 @@ export default class ScopesScopeProjectsTargetsIndexRoute extends Route {
     targets = targets.filter((target) =>
       this.can.can('connect target', target),
     );
-    const projects = this.modelFor('scopes.scope.projects');
 
     return {
       targets,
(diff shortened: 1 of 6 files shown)


ui/desktop/app/routes/scopes/scope/projects/targets/index.js
```diff
@@ -84,11 +84,18 @@ export default class ScopesScopeProjectsTargetsIndexRoute extends Route {
     transition,
```
lisbet-alvarez, line 84: non-blocking question: should the sessions route be updated  as well or will that be in a separate PR ?

lisbet-alvarez (approved):
LGTM

ui/desktop/app/routes/scopes/scope/projects/targets/index.js
```diff
@@ -84,11 +84,18 @@ export default class ScopesScopeProjectsTargetsIndexRoute extends Route {
...
   ) {
     await this.getAllTargets(transition);
+    const projects = this.modelFor('scopes.scope.projects');
+    const projectIds = projects.map((project) => project.id);
 
     const filters = { scope_id: [], id: { values: [] }, type: [] };
     scopes.forEach((scope) => {
       filters.scope_id.push({ equals: scope });
     });
+    if (scopes.length === 0) {
```
ZedLi, line 94: One issue is this assumes that the client daemon is up...which I think is fine for now

ZedLi (approved):

ui/desktop/app/routes/scopes/scope/projects/targets/index.js
cameronperera: In one of your PRs you add the `isClientDaemonRunning`. I could utilize that here to bypass this step.

ui/desktop/app/routes/scopes/scope/projects/targets/index.js
cameronperera: That is a good point. I will add it to sessions now!

ui/desktop/app/routes/scopes/scope/projects/targets/index.js
ZedLi: Sure but you'd need to manually filter it down then

ui/desktop/app/routes/scopes/scope/projects/targets/index.js
cameronperera: Interesting. I added this same logic to the sessions route. However, when you change the scope at the top of the app, it redirects you to the targets route.

cameronperera: Updated!

ZedLi (approved):
Test sessions as well and works great!

PR #2084, opened 2024-01-11
Icu 12137 add additional session sorting and add default filters

lisbet-alvarez:
:tickets: [Jira ticket](https://hashicorp.atlassian.net/browse/ICU-12137)

## Description
Default queryParam values are [pruned](https://github.com/emberjs/ember.js/blob/v5.5.0/packages/%40ember/routing/router.ts#L1032) when a transition from one route to another is made. Therefore, the RouteInfo object for the current route doesn't show the existing default queryParams. In order to work around not being able to access default query param values through the router service, we pass in the filters as an object named `selectedFilters`.

https://github.com/hashicorp/boundary-ui/assets/107949262/cc3cefb8-e243-47b2-869d-c3440481846d

## How to Test
1. Use llb-pagination-2 be branch and start boundary dev
3. Checkout this branch and run ENABLE_MIRAGE=false BYPASS_CLI_SETUP=true yarn start:desktop
4. Make sure filter tags are displayed as expected.

## Checklist:

### DIFF
--- a/addons/core/addon/components/filter-tags/index.js
+++ b/addons/core/addon/components/filter-tags/index.js
@@ -11,9 +11,10 @@ export default class FilterTagsIndexComponent extends Component {
   // =attributes
 
   get filters() {
-    return Object.entries(this.args.filters).flatMap(([key, value]) => {
+    const { allFilters, selectedFilters } = this.args.filters;
+    return Object.entries(allFilters).flatMap(([key, value]) => {
       assert(`Tags must be an array for key ${key}`, Array.isArray(value));
-      const paramsSet = new Set(this.getCurrentQueryParams(key));
+      const paramsSet = new Set(selectedFilters[key]);
       const filters = value.filter((item) => paramsSet.has(item.id));
 
       return filters.map((item) => ({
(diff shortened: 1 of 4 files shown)


addons/core/addon/components/filter-tags/index.js
```diff
@@ -26,24 +27,13 @@ export default class FilterTagsIndexComponent extends Component {
 
   // =methods
 
-  getCurrentQueryParams(key) {
-    let queryParamValue;
-    try {
-      queryParamValue = JSON.parse(this.router.currentRoute.queryParams[key]);
```
ZedLi, line 32: Wow I think I know what the issue is and is just an ember quirk. Setting a "default" or initial query parameter value does what you would expect for the route since it just gets the value directly but I think ember considers that "default" value to be the state as if you had no query parameter set and if you could see the URL you would see in fact there is no query parameters for that default state.

e.g. `page: 1` you would get no query params but if you goto another page you would get `?page=2` and then if you go _back_ to page 1, the query param would disappear since ember considers that the "default" state which is no query params.

For sessions it probably looks like `/sessions` with no query params for the 3 default you have selected but if you change it to just active for example, it would probably be `?status=active` but if then if you change it back to have all 3, the query param disappears as that's the "default" state. 🤯

ui/desktop/app/controllers/scopes/scope/projects/sessions/index.js
```diff
@@ -26,19 +26,25 @@ export default class ScopesScopeProjectsSessionsIndexController extends Controll
...
    */
   get sortedSessions() {
-    return orderBy(this.model.sessions, 'created_time', 'desc');
+    const sortedSessions = orderBy(this.model.sessions, 'created_time', 'desc');
+    return [
+      // then move active sessions to the top
+      ...sortedSessions.filter((session) => session.isAvailable),
+      // and all others to the end
+      ...sortedSessions.filter((session) => !session.isAvailable),
+    ];
```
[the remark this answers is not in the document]
lisbet-alvarez: thank you for the suggestion, that is much cleaner!

addons/core/addon/components/filter-tags/index.js
lisbet-alvarez: thank you for the detailed explanation!

PR #2085, opened 2024-01-11
feat: 🎸 Add fallback if client daemon is down or errors

ZedLi:
✅ Closes: https://hashicorp.atlassian.net/browse/ICU-12026

## Description
Added a fallback to handle the scenario where the client daemon errors out or is down. We will just retrieve everything for _each_ call (which is crazy inefficient but fine as long as it doesn't crash our application). If the daemon is down, I also prevent showing the filters/searching since that can't be done anymore. That can be removed if we add a fallback to the indexed db solution in the future. 

It's actually still quite usable as long as there aren't tens of thousands of resources.

https://github.com/hashicorp/boundary-ui/assets/5783847/78147a08-d378-4968-a670-5293f324c3a2

## How to Test
Start up the desktop client and then kill the client daemon while the DC is still running by running `boundary daemon stop`

## Checklist:

### DIFF
--- a/ui/desktop/app/routes/scopes/scope/projects/sessions/index.js
+++ b/ui/desktop/app/routes/scopes/scope/projects/sessions/index.js
@@ -12,6 +12,7 @@ export default class ScopesScopeProjectsSessionsIndexRoute extends Route {
 
   @service session;
   @service store;
+  @service ipc;
 
   // =attributes
 
@@ -64,6 +65,7 @@ export default class ScopesScopeProjectsSessionsIndexRoute extends Route {
     const from = transition.from?.name;
     const projects = this.modelFor('scopes.scope.projects');
     const projectIds = projects.map((project) => project.id);
+    const { id: scope_id } = this.modelFor('scopes.scope');
 
     const filters = {
       user_id: [{ equals: this.session.data.authenticated.user_id }],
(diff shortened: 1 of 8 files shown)


ui/desktop/app/routes/scopes/scope/projects/sessions/index.js
```diff
@@ -113,6 +123,7 @@ export default class ScopesScopeProjectsSessionsIndexRoute extends Route {
       allSessions: this.allSessions,
       allTargets: this.allTargets,
       totalItems,
+      isClientDaemonRunning: await this.ipc.invoke('isClientDaemonRunning'),
```
ZedLi, line 126: It's so annoying as I didn't realize this but I originally wanted this to be inside a helper but apparently ember can't handle promises inside templates without me using a library like [this](https://github.com/fivetanley/ember-promise-helpers) which I didn't want to pull in. 

I just added this to the model instead and we can remove it in the future.

cameronperera (approved):
Looks great. `scope_id` and `recursive: true` were both added back so we can fetch like we did before correct?
Non-blocking: Would it be easy to add some acceptance tests to verify that the search bars aren't rendered when `isClientDaemonRunning` is false?

ZedLi:
> `scope_id` and `recursive: true` were both added back so we can fetch like we did before correct?

Yup exactly. It gets ignored in the search scenarios.



> Would it be easy to add some acceptance tests to verify that the search bars aren't rendered when `isClientDaemonRunning` is false

Yes that's the plan, I haven't had a chance to add any tests yet.

cameronperera (approved):
Looks great!

PR #2096, opened 2024-01-17
test: 💍 Add tests for MQL expression generator

ZedLi:
## Description
Added tests for generating MQL expressions. There was a small bug with all of the greater/less than operators when working with non strings. We don't use any of them yet but I changed how we sanitize. 

## Checklist:

### DIFF
--- a/addons/api/tests/unit/utils/mql-query-test.js
+++ b/addons/api/tests/unit/utils/mql-query-test.js
@@ -1,0 +1,208 @@
+import {
+  generateMQLSearchExpression,
+  generateMQLFilterExpression,
+  generateMQLExpression,
+} from 'api/utils/mql-query';
+import { module, test } from 'qunit';
+
+module('Unit | Utility | mql-query', function () {
+  module('generateMQLSearchExpression tests', function () {
+    test('it generates search correctly', function (assert) {
+      const result = generateMQLSearchExpression({
+        text: 'search',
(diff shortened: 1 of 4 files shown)


addons/api/tests/unit/utils/mql-query-test.js
```diff
@@ -0,0 +1,208 @@
+import {
+  generateMQLSearchExpression,
+  generateMQLFilterExpression,
+  generateMQLExpression,
+} from 'api/utils/mql-query';
+import { module, test } from 'qunit';
+
+module('Unit | Utility | mql-query', function () {
+  module('generateMQLSearchExpression tests', function () {
```
ZedLi, line 9: I started using nested modules 🤷, I like the organization more

cameronperera: I love it. @calcaide has been looking into this so he may have some input. However, I like the clear use of it in a pretty simple test file.

cameronperera (approved):
LGTM

addons/api/tests/unit/utils/mql-query-test.js
lisbet-alvarez: this new structure makes it easier to read tests (esp for unit tests)

lisbet-alvarez (approved):
LGTM

PR #2098, opened 2024-01-19
fix: 🐛 policy UI bug fix

DhariniJeeva:
:tickets: [Jira ticket](https://hashicorp.atlassian.net/browse/ICU-12299)

## Description

This PR fixes a few things
- UI behavior when the user clears the custom text input
- some minor card updates
- disabled overridable based on a recent API update
- use a current scope while creating a policy within scope settings

https://github.com/hashicorp/boundary-ui/assets/15043878/046bf3ac-ed8e-4913-9c80-5984a3d21837

## How to Test
- create storage policy using staged binary ent

## Checklist:

### DIFF
--- a/ui/admin/app/components/form/policy/index.js
+++ b/ui/admin/app/components/form/policy/index.js
@@ -43,9 +43,13 @@ export default class FormPolicyComponent extends Component {
 
   /**
    * Returns policy type
+   * API doesn't return days for do_not_protect,
+   * so we set option to do_not_protect when there's no retain days
    * @type {string}
    */
   get selectRetentionPolicyType() {
+    if (!this.args.model.retain_for?.days) return 'do_not_protect';
+
     const val = Object.keys(RETENTION_POLICY).find(
       (i) => RETENTION_POLICY[i] === this.args.model.retain_for?.days,
     );
--- a/ui/admin/app/components/form/policy/policy-selection/index.js
+++ b/ui/admin/app/components/form/policy/policy-selection/index.js
@@ -15,13 +15,26 @@ export default class FormPolicySelectionComponent extends Component {
   get showCustomInput() {
     if (this.args.name === 'retention_policy') {
       return this.isCustomRetentionSelected;
-    } else if (
-      this.args.name === 'deletion_policy' &&
-      this.args.model.retain_for?.days >= 0
-    ) {
+    } else if (this.args.name === 'deletion_policy') {
       return this.isCustomDeletionSelected;
     } else {
       return null;
+    }
+  }
+
+  /**
+   * Disable toggle when the form is disabled and when the retain/delete after days are 0
+   * @type {boolean}
+   */
+  get toggleDisabled() {
+    if (
+      this.args.disabled ||
+      this.args.model[this.args.customInputName]?.days === 0 ||
+      !this.args.model[this.args.customInputName]?.days
+    ) {
+      return true;
+    } else {
+      return false;
     }
   }
 
@@ -30,9 +43,7 @@ export default class FormPolicySelectionComponent extends Component {
    * @type {boolean}
    */
   get isCustomRetentionSelected() {
-    //no need to show custom input for the below days
-    const arr = [-1, 0, 2555, 2190];
-    return arr.includes(this.args.model.retain_for?.days) ? false : true;
+    return this.args.selectedOption === 'custom' ? true : false;
   }
 
   /**
@@ -40,14 +51,17 @@ export default class FormPolicySelectionComponent extends Component {
    * @type {boolean}
    */
   get isCustomDeletionSelected() {
-    return this.args.model.delete_after?.days > 0;
+    return this.args.selectedOption === 'custom' ? true : false;
   }
   /**
    * Returns true if the toggle is on
    * @type {boolean}
    */
   get isOverridable() {
-    return this.args.model[this.args.customInputName]?.overridable;
+    return (
+      this.args.model[this.args.customInputName]?.days &&
+      this.args.model[this.args.customInputName]?.overridable
+    );
   }
 
   get isDeleteDisable() {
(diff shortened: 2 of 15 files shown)


ui/admin/app/templates/scopes/scope/policies/policy/index.hbs
```diff
@@ -9,10 +9,11 @@
   </page.breadcrumbs>
 
   <page.header>
-    <Hds::Text::Display @tag='h2'>
+    <h2>
```
DhariniJeeva, line 12: the reason behind not using this new component is due to the lack of margin/padding and I did not want to add it individually, this should be handled in a common place since we will be using this everywhere and will need to test those before adding margins. I'll add an HDS migration ticket for this
Screenshot 2024-01-19 at 12 34 16 PM

ui/admin/app/components/form/policy/policy-selection/index.js
```diff
@@ -15,39 +15,53 @@ export default class FormPolicySelectionComponent extends Component {
...
       return this.isCustomRetentionSelected;
-    } else if (
-      this.args.name === 'deletion_policy' &&
-      this.args.model.retain_for?.days >= 0
-    ) {
+    } else if (this.args.name === 'deletion_policy') {
       return this.isCustomDeletionSelected;
     } else {
       return null;
     }
```
lisbet-alvarez, line 22: since `this.isCustomRetentionSelected` and `this.isCustomDeletionSelected` methods do the same thing now, i think this getter can be simplified and there is no need to separate it into two different methods. Also perhaps the else block on line 20 is not needed anymore ?

DhariniJeeva: 💯 simplified!

ui/admin/app/components/form/policy/policy-selection/index.js
```diff
@@ -13,41 +13,34 @@ export default class FormPolicySelectionComponent extends Component {
...
    */
-  get isCustomRetentionSelected() {
-    //no need to show custom input for the below days
-    const arr = [-1, 0, 2555, 2190];
-    return arr.includes(this.args.model.retain_for?.days) ? false : true;
+  get toggleDisabled() {
+    if (
+      this.args.disabled ||
+      this.args.model[this.args.customInputName]?.days === 0 ||
+      !this.args.model[this.args.customInputName]?.days
```
lisbet-alvarez, line 27: perhaps this if statement can be simplified to just `this.args.disabled || !this.args.model[this.args.customInputName]?.days`

ui/admin/app/components/form/policy/policy-selection/index.js
```diff
@@ -13,41 +13,34 @@ export default class FormPolicySelectionComponent extends Component {
...
-      return this.isCustomRetentionSelected;
-    } else if (
-      this.args.name === 'deletion_policy' &&
-      this.args.model.retain_for?.days >= 0
-    ) {
-      return this.isCustomDeletionSelected;
-    } else {
-      return null;
-    }
+    return this.args.selectedOption === 'custom' ? true : false;
```
ZedLi, line 16: This could be simplified to:
```suggestion
    return this.args.selectedOption === 'custom';
```

ui/admin/app/components/form/policy/policy-selection/index.js
```diff
@@ -13,41 +13,34 @@ export default class FormPolicySelectionComponent extends Component {
...
    */
   get isOverridable() {
-    return this.args.model[this.args.customInputName]?.overridable;
+    return (
+      this.args.model[this.args.customInputName]?.days &&
+      this.args.model[this.args.customInputName]?.overridable
+    );
   }
 
   get isDeleteDisable() {
```
ZedLi, line 46: This could have also been simplified to:
```
    return this.args.name === 'deletion_policy' &&
          this.args.model.retain_for?.days === -1;
```

ui/admin/app/components/form/policy/policy-selection/index.js
DhariniJeeva: hmm, I can't do that here though, because we want to disable the overrides in the UI when we select `do_not_protect` or `do_not_delete` and the API doesn't send me 0 days, it doesn't send anything if the days are 0. So I wanted to capture both behaviors here.

lisbet-alvarez (approved):
LGTM

ZedLi (commented):
Typing 0 in the custom field also causes it to collapse which is jarring, is that intended?

https://github.com/hashicorp/boundary-ui/assets/5783847/8ac2e78e-06a0-4961-af46-faf03000dc3b

ui/admin/app/components/form/policy/policy-selection/index.js
ZedLi: Why is that? Isn't 0 already covered as a falsy value?

DhariniJeeva:
> Typing 0 in the custom field also causes it to collapse which is jarring, is that intended?
> 
>  Screen.Recording.2024-01-19.at.4.33.53.PM.mov

yes we want to show `do not protect` or do not delete when zero is entered

ui/admin/app/components/form/policy/policy-selection/index.js
DhariniJeeva: I forgot 0 was a falsey value, fixed. ty both!

ui/admin/app/components/form/policy/index.js
```diff
@@ -43,9 +43,13 @@ export default class FormPolicyComponent extends Component {
 
   /**
    * Returns policy type
+   * API doesn't return days for do_not_protect,
+   * so we set option to do_not_protect when there's no retain days
    * @type {string}
    */
   get selectRetentionPolicyType() {
+    if (!this.args.model.retain_for?.days) return 'do_not_protect';
```
ZedLi, line 51: I think most of these magic strings should have been consts

ZedLi (approved):
Looks good!

ui/admin/app/components/form/policy/index.js
DhariniJeeva: agree, I can address that separately

PR #2099, opened 2024-01-19
fix: 🐛 Fix errors not getting propagated up from client daemon

ZedLi:
## Description
Client daemon errors weren't correctly getting propagated up to the desktop client which meant they were getting interpreted as search results and erroring out.

## How to Test
A quick way is to start the desktop client, then stop the client daemon and restart it. The DC should now be getting `403` responses from the daemon and it shouldn't error out.

## Checklist:


calcaide (approved):
Awesome work!!

cameronperera (approved):

PR #2103, opened 2024-01-22
fix: 🐛 use filter query param to filter by org scope id

cameronperera:
✅ Closes: https://hashicorp.atlassian.net/browse/ICU-12309

:tickets: [Jira ticket](https://hashicorp.atlassian.net/browse/ICU-12309)

## Checklist:

### DIFF
--- a/ui/desktop/app/routes/scopes/scope/projects/targets/index.js
+++ b/ui/desktop/app/routes/scopes/scope/projects/targets/index.js
@@ -87,17 +87,12 @@ export default class ScopesScopeProjectsTargetsIndexRoute extends Route {
     const { id: scope_id } = this.modelFor('scopes.scope');
     await this.getAllTargets(transition, scope_id);
     const projects = this.modelFor('scopes.scope.projects');
-    const projectIds = projects.map((project) => project.id);
+    const orgScope = this.modelFor('scopes.scope');
 
     const filters = { scope_id: [], id: { values: [] }, type: [] };
     scopes.forEach((scope) => {
       filters.scope_id.push({ equals: scope });
     });
-    if (scopes.length === 0) {
-      projectIds.forEach((projectId) => {
-        filters.scope_id.push({ equals: projectId });
-      });
-    }
     types.forEach((type) => {
       filters.type.push({ equals: type });
     });
@@ -121,14 +116,18 @@ export default class ScopesScopeProjectsTargetsIndexRoute extends Route {
 
     this.addActiveSessionFilters(filters, availableSessions, sessions);
 
-    let targets = await this.store.query('target', {
+    const query = {
       recursive: true,
       scope_id,
       query: { search, filters },
       page,
       pageSize,
       force_refresh: true,
-    });
+    };
+    if (orgScope.isOrg && scopes.length === 0) {
+      query.filter = `("/item/scope/parent_scope_id" == "${orgScope.id}")`;
+    }
+    let targets = await this.store.query('target', query);
     const totalItems = targets.meta?.totalItems;
 
     // Filter out targets to which users do not have the connect ability


ui/desktop/app/routes/scopes/scope/projects/targets/index.js
```diff
@@ -87,17 +87,12 @@ export default class ScopesScopeProjectsTargetsIndexRoute extends Route {
...
-    const projectIds = projects.map((project) => project.id);
+    const orgScope = this.modelFor('scopes.scope');
 
     const filters = { scope_id: [], id: { values: [] }, type: [] };
     scopes.forEach((scope) => {
       filters.scope_id.push({ equals: scope });
     });
-    if (scopes.length === 0) {
-      projectIds.forEach((projectId) => {
-        filters.scope_id.push({ equals: projectId });
```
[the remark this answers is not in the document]
cameronperera: Depends. I actually added this because if you were to filter by scope at the global level followed changing the org scope via the header the filters will clear out. Then if you were to change the scope back to global via the header, the filters you did have would still be there. I only added scope to the the `resetController` but this will be true for any queryParam on the route. We should either clear all filters or just remove the `resetController` entirely. Happy to chat more about this.

ui/desktop/app/routes/scopes/scope/projects/targets/index.js
```diff
@@ -121,14 +116,18 @@ export default class ScopesScopeProjectsTargetsIndexRoute extends Route {
...
       recursive: true,
       scope_id,
       query: { search, filters },
       page,
       pageSize,
       force_refresh: true,
-    });
+    };
+    if (orgScope.isOrg && scopes.length === 0) {
+      query.filter = `("/item/scope/parent_scope_id" == "${orgScope.id}")`;
```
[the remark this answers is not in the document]
cameronperera: That is a good idea. No need to fetch absolutely everything if we know the user wants it narrowed down by the org.

ui/desktop/app/routes/scopes/scope/projects/targets/index.js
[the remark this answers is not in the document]
ZedLi: Yeah I agree with that, I did find it weird that some filters weren't cleared out. I don't particularly have a preference to either clearing all filters or none, but I do agree that it should be one or the other

DhariniJeeva (approved):

ui/desktop/app/routes/scopes/scope/projects/targets/index.js
```diff
@@ -84,29 +84,26 @@ export default class ScopesScopeProjectsTargetsIndexRoute extends Route {
     { search, scopes, availableSessions, types, page, pageSize },
     transition,
   ) {
-    const { id: scope_id } = this.modelFor('scopes.scope');
-    await this.getAllTargets(transition, scope_id);
+    const orgScope = this.modelFor('scopes.scope');
+    // orgFilter used to narrow down resources to only those under
+    // the current org scope if org is not global
+    const orgFilter = `("/item/scope/parent_scope_id" == "${orgScope.id}")`;
```
ZedLi, line 90: I'm not familiar with the filter syntax but are the parentheses needed?

ui/desktop/app/routes/scopes/scope/projects/sessions/index.js
```diff
@@ -133,11 +144,17 @@ export default class ScopesScopeProjectsSessionsIndexRoute extends Route {
   }
 
   resetController(controller, isExiting, transition) {
-    const { to } = transition;
+    const { scopes, targets, status } = transition.to.queryParams;
     // Reset the scopes query param when changing org scope
-    if (!isExiting && to.queryParams.scopes === '[]') {
+    if (isExiting && (scopes || targets || status)) {
```
ZedLi, line 149: These parameters are going to be strings correct? i.e if it's an empty array it will be `"[]"`? Is this the intent?

ui/desktop/app/routes/scopes/scope/projects/targets/index.js
```diff
@@ -147,10 +152,13 @@ export default class ScopesScopeProjectsTargetsIndexRoute extends Route {
 
   resetController(controller, isExiting, transition) {
     const { to } = transition;
-    // Reset the scopes query param when changing org scope
+    // Reset the query params when changing org scope
     if (!isExiting && to.queryParams.scopes === '[]') {
```
ZedLi, line 156: Why do we only check the scopes param?

ZedLi (approved):
Looks good, just had a few questions

ui/desktop/app/routes/scopes/scope/projects/targets/index.js
cameronperera: Actually it seems me don't. I was basing this off of the filter we use for filtering by project type but I have removed it and it doesn't seem to be an issue.

ui/desktop/app/routes/scopes/scope/projects/sessions/index.js
cameronperera: Absolutely correct. Thanks for catching that. I will update now.

ui/desktop/app/routes/scopes/scope/projects/targets/index.js
cameronperera: Sorry, this is a little confusing. I don't think I need to check for the other params in sessions either. Here I am only checking `scopes` because I am looking at the `transition.to.queryParams` and not the `transition.from.queryParams`. When the scope has been changed in the header, it has no queryParam values at the route level and I am using that to reset the values on the controller. Happy to discuss if there is a better way to do this that isn't as confusing.

calcaide (approved):
LGTM! Thanks!!

ui/desktop/app/routes/scopes/scope/projects/sessions/index.js
```diff
@@ -105,7 +110,7 @@ export default class ScopesScopeProjectsSessionsIndexRoute extends Route {
       this.allSessions = await this.store.query(
```
ZedLi, line 110: Just wanted to note we didn't include the parent scope filter for either of the sessions call

cameronperera: Sorry, that was actually part of my changes yesterday and I did not include them. Adding them back now.

PR #2104, opened 2024-01-23
fix: 🐛 delete and retain_until should be readonly and Scope level policies

DhariniJeeva:
## Description

This PR fixes a bug where the old value was retained when the field becomes null/zero from the API, this is due to ember-data not updating the old value if the field was removed from the API, so we explicitly set these `readonly` fields to null. For more context - check [this](https://github.com/hashicorp/boundary-ui/blob/3f4beaada88d62a813a45c978e42ae3e10504b9c/addons/api/addon/serializers/application.js#L302)

And also https://hashicorp.atlassian.net/browse/ICU-12323

## How to Test

1. run ENT staged binary, 
2. and create a few org and global policies, 
3. you should be able to see both global and org at org setting 
4. only global policies at the global scope

## Checklist:

### DIFF
--- a/addons/api/addon/generated/models/session-recording.js
+++ b/addons/api/addon/generated/models/session-recording.js
@@ -112,12 +112,14 @@ export default class GeneratedSessionRecordingModel extends BaseModel {
 
   @attr('date', {
     description: 'The time until a session recording is required to be stored.',
+    readOnly: true,
   })
   retain_until;
 
   @attr('date', {
     description:
       'The time a session recording is scheduled to be automatically deleted.',
+    readOnly: true,
   })
   delete_after;
 }
--- a/addons/core/translations/resources/en-us.yaml
+++ b/addons/core/translations/resources/en-us.yaml
@@ -950,7 +950,6 @@ policy:
   messages:
     detach: Successfully detached
     reapply: Successfully reapplied
-    later: The policy will be reapplied after the recording has completed
     none:
       title: No storage policies yet
       description: A Storage Policy defines how long Session Recordings must be kept and when they should be deleted. We recommend checking your data retention policies for compliance requirements.
--- a/addons/core/translations/resources/en-us.yaml
+++ b/addons/core/translations/resources/en-us.yaml
@@ -950,6 +950,7 @@ policy:
   messages:
     detach: Successfully detached
     reapply: Successfully reapplied
+    later: The policy will be reapplied after the recording has completed
     none:
       title: No storage policies yet
       description: A Storage Policy defines how long Session Recordings must be kept and when they should be deleted. We recommend checking your data retention policies for compliance requirements.
--- a/ui/admin/app/routes/scopes/scope/add-storage-policy/index.js
+++ b/ui/admin/app/routes/scopes/scope/add-storage-policy/index.js
@@ -57,7 +57,6 @@ export default class ScopesScopeAddStoragePolicyIndexRoute extends Route {
   @notifyError(({ message }) => message, { catch: true })
   @notifySuccess('notifications.save-success')
   async attachStoragePolicy(scope) {
-    console.log(scope, 'SCOPE');
     const { storage_policy_id } = scope;
     if (storage_policy_id) {
       await scope.attachStoragePolicy(storage_policy_id);
--- a/ui/admin/app/routes/scopes/scope/add-storage-policy/index.js
+++ b/ui/admin/app/routes/scopes/scope/add-storage-policy/index.js
@@ -22,20 +22,19 @@ export default class ScopesScopeAddStoragePolicyIndexRoute extends Route {
    */
   async afterModel() {
     const { id: scope_id } = this.modelFor('scopes.scope');
-
+    const currentScope = await this.store.query('policy', {
+      scope_id,
+    });
     if (scope_id === 'global') {
-      //Global scope should list both global and org scope policies
-      const globalPolicy = await this.store.query('policy', {
-        scope_id,
-        recursive: true,
-      });
-      this.policyList = globalPolicy;
+      //Global scope should only list policies from its scope
+      this.policyList = currentScope;
     } else {
-      //Org scope should only list policies from its scope
-      const orgPolicy = await this.store.query('policy', {
-        scope_id,
+      //Org scope should list both global and org scope policies
+      const globalScope = await this.store.query('policy', {
+        scope_id: 'global',
       });
-      this.policyList = orgPolicy;
+
+      this.policyList = [...globalScope, ...currentScope];
     }
   }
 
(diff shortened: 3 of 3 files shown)


addons/api/addon/generated/models/session-recording.js
```diff
@@ -112,12 +112,14 @@ export default class GeneratedSessionRecordingModel extends BaseModel {
 
   @attr('date', {
     description: 'The time until a session recording is required to be stored.',
+    readOnly: true,
```
calcaide, line 115: None blocking comment, clarification: I understand setting these attrs to `readOnly` within session-recording is just for show purposes, is that correct? otherwise, we rely on the [policy model](https://github.com/hashicorp/boundary-ui/blob/main/addons/api/addon/generated/models/policy.js#L48-L60).

addons/core/translations/resources/en-us.yaml
```diff
@@ -954,6 +954,7 @@ policy:
   messages:
     detach: Successfully detached
     reapply: Successfully reapplied
+    later: The policy will be reapplied after the recording has completed
```
calcaide, line 957: Where do we use this translation?

calcaide (changes requested):
Left a comment

addons/core/translations/resources/en-us.yaml
DhariniJeeva: this is for a new change I will making in this PR

addons/api/addon/generated/models/session-recording.js
DhariniJeeva: both are different fields, these are only for session recording API

addons/core/translations/resources/en-us.yaml
DhariniJeeva: I'm going to move this change to a separate PR since that is not a bug that can wait a bit

ui/admin/app/routes/scopes/scope/add-storage-policy/index.js
```diff
@@ -23,12 +23,20 @@ export default class ScopesScopeAddStoragePolicyIndexRoute extends Route {
...
     const { id: scope_id } = this.modelFor('scopes.scope');
 
-    //fetch policies from current scope
-    const orgScopePolicies = await this.store.query('policy', {
-      scope_id,
-    });
-
-    this.policyList = orgScopePolicies;
+    if (scope_id === 'global') {
+      //Global scope should list both global and org scope policies
```
DhariniJeeva, line 27: I got this reverse, fixing this

ZedLi:
```suggestion
      // Global scope should list both global and org scope policies
```

Can we add a space to initial comments?

ui/admin/app/routes/scopes/scope/add-storage-policy/index.js
```diff
@@ -22,13 +22,20 @@ export default class ScopesScopeAddStoragePolicyIndexRoute extends Route {
...
   async afterModel() {
     const { id: scope_id } = this.modelFor('scopes.scope');
-
-    //fetch policies from current scope
-    const orgScopePolicies = await this.store.query('policy', {
+    const currentScope = await this.store.query('policy', {
       scope_id,
     });
+    if (scope_id === 'global') {
+      //Global scope should only list policies from its scope
```
DhariniJeeva, line 29:
```suggestion
      // Global scope should only list policies from its scope
```

ui/admin/app/routes/scopes/scope/add-storage-policy/index.js
```diff
@@ -22,13 +22,20 @@ export default class ScopesScopeAddStoragePolicyIndexRoute extends Route {
...
-    //fetch policies from current scope
-    const orgScopePolicies = await this.store.query('policy', {
+    const currentScope = await this.store.query('policy', {
       scope_id,
     });
+    if (scope_id === 'global') {
+      // Global scope should only list policies from its scope
+      this.policyList = currentScope;
+    } else {
+      //Org scope should list both global and org scope policies
```
DhariniJeeva, line 32:
```suggestion
      // Org scope should list both global and org scope policies
```

ui/admin/app/routes/scopes/scope/add-storage-policy/index.js
```diff
@@ -23,12 +23,20 @@ export default class ScopesScopeAddStoragePolicyIndexRoute extends Route {
...
-    this.policyList = orgScopePolicies;
+    if (scope_id === 'global') {
+      //Global scope should list both global and org scope policies
+      const globalPolicy = await this.store.query('policy', {
+        scope_id,
+        recursive: true,
+      });
+      this.policyList = globalPolicy;
+    } else {
+      //Org scope should only list policies from its scope
```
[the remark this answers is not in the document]
DhariniJeeva: yes I had it reversed, the recent commit should fix this

ui/admin/app/routes/scopes/scope/add-storage-policy/index.js
```diff
@@ -22,13 +22,20 @@ export default class ScopesScopeAddStoragePolicyIndexRoute extends Route {
    */
   async afterModel() {
     const { id: scope_id } = this.modelFor('scopes.scope');
-
-    //fetch policies from current scope
-    const orgScopePolicies = await this.store.query('policy', {
+    const currentScope = await this.store.query('policy', {
```
ZedLi, line 25: This is more accurately `currentScopePolicies` right?

ui/admin/app/routes/scopes/scope/add-storage-policy/index.js
```diff
@@ -22,13 +22,20 @@ export default class ScopesScopeAddStoragePolicyIndexRoute extends Route {
...
       scope_id,
     });
+    if (scope_id === 'global') {
+      // Global scope should only list policies from its scope
+      this.policyList = currentScope;
+    } else {
+      // Org scope should list both global and org scope policies
+      const globalScope = await this.store.query('policy', {
```
ZedLi, line 33: Same here, these are policies, not scopes

ui/admin/app/routes/scopes/scope/add-storage-policy/index.js
```diff
@@ -22,13 +22,20 @@ export default class ScopesScopeAddStoragePolicyIndexRoute extends Route {
    */
   async afterModel() {
```
ZedLi, line 23: I just noticed this but how come we do this in the `afterModel` and not just the `model`?

DhariniJeeva: so we want this to be in the `afterModel` method as this is part of the scope model, any `abandon confirmation`, or `cancellation` is based on the scope model and not policies

ZedLi (approved):

calcaide (approved):

PR #2105, opened 2024-01-23
fix: 🐛 Handle when requests to client daemon are returning 403

ZedLi:
## Description
Added some logic to handle the scenario where a user could get `403`s from the client daemon. We can try to re-authenticate with the daemon as part of error handling.

## How to Test
Run the desktop client and then kill the client daemon and restart it. You should still be able to use the desktop client normally and search and filter.

## Checklist:


lisbet-alvarez (approved):
LGTM, tested locally and works as intended!

cameronperera (approved):

DhariniJeeva (approved):

PR #2109, opened 2024-01-24
refactor: 💡 use new baseURL and update topic paths

lisbet-alvarez:
## Description
Confirmed with [web support](https://hashicorp.slack.com/archives/C5FSPUGDS/p1706130252105629) that we can use developer.hashicorp.com links directly. 
Updated all topics to use correct path

## How to Test
Validate that all links are valid. The [web repo](https://github.com/hashicorp/web/tree/main/apps/boundary-io/src/redirects) redirects folder can be a good tool as well to validate.

## Checklist:


cameronperera (approved):
LGTM. I clicked around in the vercel deployments and all links took me to the correct URLs.

addons/core/addon/helpers/doc-url.js
```diff
@@ -23,6 +23,6 @@ export default class extends Helper {
       );
     }
     const path = configuredPath || '';
-    return `${baseURL}${path}`;
+    return /^(http|https):\/\/(\S)+$/.test(path) ? path : `${baseURL}${path}`;
```
ZedLi, line 26: I just wanted to check, is this to check if this is a URL that starts with an http or https scheme to differentiate between a relative url vs an absolute url?

If that's the case, it seems like we just need to check if the string starts with http like `path.startsWith('http')` (do we even have just `http` links?) or keep the regex but just remove the `(\S)` (I don't think you care about the rest of the url nor does it need to be captured)

ui/desktop/electron-app/src/index.js
```diff
@@ -142,12 +142,12 @@ const createWindow = (partition, closeWindowCB) => {
...
 
   // Opens external links in the host default browser.
-  // We allow boundaryproject.io domain to open on external window (for now)
+  // We allow developer.hashicorp.com domain to open on external window
   // and releases.hashicorp.com domain to download the desktop app or
   // link to the release page for the desktop app.
   browserWindow.webContents.setWindowOpenHandler(({ url }) => {
     if (
-      url.startsWith('https://boundaryproject.io/') ||
+      url.startsWith('https://developer.hashicorp.com/boundary/') ||
```
ZedLi, line 150: Do we need to be specific for `/boundary`?

Also do we not have any links at all to `boundaryproject.io` anymore?

addons/core/addon/helpers/doc-url.js
lisbet-alvarez: we only have one https link. I like ur suggestion of using startsWith instead. Ill use startsWith('https') instead since there is no http link. Initially i added http to the regex just in case we do ever add one but now thinking about it that seems unlikely and we could always update if the scenario does come up. Thank you

ui/desktop/electron-app/src/index.js
lisbet-alvarez: All the links we have are in the `/boundary` path so thats why i included it in the base url but it does not need to be like that. Would it be better if we leave out the `boundary` path and add it to the individual path mappings instead in case theres a future scenario where we need to link to a doc that's not in `/boundary`? If so, i agree with that line of reasoning and can update the paths.

Nope i as i was updating i checked that all the links redirect to the developer docs.

ZedLi: I was just a little worried because it's an easy gotcha in the future which we might not notice.  I also think there's no reason we can't trust the hashicorp base domain rather than the specific `/boundary` path and my vote would be to just leave it out

ui/admin/config/environment.js
```diff
@@ -56,62 +56,84 @@ module.exports = function (environment) {
     oidcPollingTimeoutSeconds: 1,
 
     documentation: {
-      baseURL: 'https://boundaryproject.io/help/admin-ui',
+      baseURL: 'https://developer.hashicorp.com',
```
ZedLi, line 59: I'm fine with this as the baseURL but will they always start with `/boundary/docs`?

ZedLi (approved):
Thanks for refactoring this! Way easier to understand now!

ui/admin/config/environment.js
lisbet-alvarez: all the links we have rn start with `/boundary` but not `/docs` some are `/boundary/tutorials`

ZedLi: Gotcha, yeah I wonder if we should assume `/boundary` as part of the base URL but it's not a big deal either way

calcaide (approved):
Tested manually some random links within Admin & Desktop and works like a charm.

ui/admin/config/environment.js
lisbet-alvarez: ah okay im looking at past feedback and realized that i misunderstood something. Ill change this back to use `/boundary`. good catch thank you!

PR #2118, opened 2024-01-29
refactor: 💡 add -addr flag to add-token command

cameronperera:
## Description

This [PR](https://github.com/hashicorp/boundary/pull/4312) added the `-addr` flag to the `boundary daemon add-token` command which allows us to pass it in via a flag and not set it as an env variable.

## Checklist:

### DIFF
--- a/ui/desktop/electron-app/src/services/client-daemon-manager.js
+++ b/ui/desktop/electron-app/src/services/client-daemon-manager.js
@@ -119,7 +119,6 @@ const addTokenCliCommand = (token) => {
   ];
   const sanitizedToken = sanitizer.base62EscapeAndValidate(token);
 
-  // TODO: Is setting an env var for the address the only way? I'm surprised to not see an -addr flag
   const { stdout, stderr } = spawnSync(addTokenCommand, {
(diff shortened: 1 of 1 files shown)


dvonanderson (approved):

DhariniJeeva (approved):
LGTM, had one question - do we still need to pass in the boundary_addr in `postBody` on L83?

cameronperera:
> LGTM, had one question - do we still need to pass in the boundary_addr in `postBody` on L83?

I believe we do. L83 is using the unix socket logic and I do not believe any of that needs to change right now.

lisbet-alvarez (approved):

ui/desktop/electron-app/src/services/client-daemon-manager.js
```diff
@@ -121,7 +122,6 @@ const addTokenCliCommand = (token) => {
   // TODO: Is setting an env var for the address the only way? I'm surprised to not see an -addr flag
   const { stdout, stderr } = spawnSync(addTokenCommand, {
```
lisbet-alvarez, line 123: non-blocking suggestion: i think we can get rid of the "TODO" now

PR #2127, opened 2024-02-01
fix: 🐛 Add error notification if adding token to daemon fails

ZedLi:
✅ Closes: https://hashicorp.atlassian.net/browse/ICU-12488

## Description
We currently don't display any errors or any indication that something has gone wrong if adding a token fails. This change adds a popup toast to at least display the error backend gives. Also need to confirm the error we get back from CLI on windows is correct.

https://github.com/hashicorp/boundary-ui/assets/5783847/d0f3f122-61fd-4038-a150-615782f3771e

## How to Test
Start up a new instance of desktop client. Create a new user that doesn't have read permissions to tokens. Login to this user and we should get an error notification.

## Checklist:

### DIFF
--- a/ui/desktop/app/routes/application.js
+++ b/ui/desktop/app/routes/application.js
@@ -8,6 +8,7 @@ import { inject as service } from '@ember/service';
 import { action } from '@ember/object';
 import { A } from '@ember/array';
 import { getOwner } from '@ember/application';
+import { notifyError } from 'core/decorators/notify';
 
 export default class ApplicationRoute extends Route {
   // =services
@@ -29,6 +30,7 @@ export default class ApplicationRoute extends Route {
    * reported by the main process.  If they differ, update the main process
    * clusterUrl so that the renderer's CSP can be rewritten to allow requests.
    */
+  @notifyError(({ message }) => message, { catch: true })
   async beforeModel() {
     this.intl.setLocale(['en-us']);
     await this.session.setup();
@@ -39,7 +41,7 @@ export default class ApplicationRoute extends Route {
     // Add token to client daemon after a successful authentication restoration
     if (this.session.isAuthenticated) {
       const sessionData = this.session.data?.authenticated;
-      this.ipc.invoke('addTokenToClientDaemon', {
+      await this.ipc.invoke('addTokenToClientDaemon', {
         tokenId: sessionData?.id,
         token: sessionData?.token,
       });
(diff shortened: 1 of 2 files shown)


DhariniJeeva:
I understand they can't do anything with the search bar when there's an error. In that case, should we still show the search toolbar or hide it?

ZedLi:
> I understand they can't do anything with the search bar when there's an error. In that case, should we still show the search toolbar or hide it?

Do you have any suggestions on how we can hide it?

DhariniJeeva:
Don't we usually get the allowed permissions list from the API to control the UI? Or is the flow different since it is an OIDC login?

ZedLi:
> Don't we usually get the allowed permissions list from the API to control the UI? Or is the flow different since it is an OIDC login?

This affects any login actually and is for the daemon but you're right, we theoretically could check for the auth token permission itself by trying to read the token we have and see if we get a `403` 🤔, but I also don't want to add another api request just to check on every request which ends up slowing it down for every other scenario when loading/searching/filtering.

I think this is personally the best tradeoff since we don't keep an auth token model around to check permissions for (or at least an auth token model in the error state since we'd get a 403) unless we do it in a parent route somewhere. Unless there's a better way of handling it so we don't have to read the token every time, maybe a service? I think we should think about this a bit more since I think it's possible but we need to also balance it with performance.

DhariniJeeva:
> > Don't we usually get the allowed permissions list from the API to control the UI? Or is the flow different since it is an OIDC login?
> 
> This affects any login actually and is for the daemon but you're right, we theoretically could check for the auth token permission itself by trying to read the token we have against the API and see if we get a `403` 🤔, but I also don't want to add another api request just to check on every request which ends up slowing it down for every other scenario when loading/searching/filtering.
> 
> I think this is personally the best tradeoff since we don't keep an auth token model around to check permissions for (or at least an auth token model in the error state since we'd get a 403 and we don't get the permissions back on the token after authentication otherwise we could just check localstorage) unless we do it in a parent route somewhere. Unless there's a better way of handling it so we don't have to read the token every time, maybe a service? I think we should think about this a bit more since I think it's possible but we need to also balance it with performance.

hmm yes adding another API request is not ideal right now. we can think about this in the future, this is ok for now

DhariniJeeva (approved):

lisbet-alvarez (approved):
LGTM 🙌

PR #2129, opened 2024-02-05
refactor: 💡 only show scopes with auth methods at authenticate

cameronperera:
✅ Closes: https://hashicorp.atlassian.net/browse/ICU-2355

:tickets: [Jira ticket](https://hashicorp.atlassian.net/browse/ICU-2355)

Before:
Screenshot 2024-02-05 at 2 25 24 PM

After:
Screenshot 2024-02-05 at 2 25 02 PM

## How to Test
Start a `boundary dev` instance and start the desktop client with the cluster URL pointing to `http://localhost:9200`. The only scope you should see in the dropdown is `Global` as the generated org will have no auth methods by default.

## Checklist:

### DIFF
--- a/ui/desktop/app/routes/scopes/scope/authenticate.js
+++ b/ui/desktop/app/routes/scopes/scope/authenticate.js
@@ -32,16 +32,6 @@ export default class ScopesScopeAuthenticateRoute extends Route {
   async model() {
     const { id: scope_id } = this.modelFor('scopes.scope');
 
-    // Fetch auth methods for current scope
-    const authMethods = await this.resourceFilterStore.queryBy(
-      'auth-method',
-      {
-        authorized_actions: [{ contains: 'authenticate' }],
-      },
-      {
-        scope_id,
-      },
-    );
     // Preload all authenticatable auth methods into the store
     const authMethodsForAllScopes = await this.resourceFilterStore.queryBy(
       'auth-method',
@@ -53,12 +43,14 @@ export default class ScopesScopeAuthenticateRoute extends Route {
         recursive: true,
       },
     );
+    const scopeIDs = new Set(authMethodsForAllScopes.map((i) => i.scopeID));
     // Fetch org scopes
     // and filter out any that have no auth methods
-    const scopes = this.modelFor('scopes').filter(
-      ({ id: scope_id, isOrg }) =>
-        isOrg &&
-        authMethodsForAllScopes.filter((i) => i.scopeID === scope_id).length,
+    const scopes = this.modelFor('scopes').filter(({ id: scope_id }) =>
+      scopeIDs.has(scope_id),
+    );
+    const authMethods = authMethodsForAllScopes.filter(
+      (i) => i.scopeID === scope_id,
     );
 
     return hash({
--- a/ui/desktop/app/routes/scopes/scope/authenticate.js
+++ b/ui/desktop/app/routes/scopes/scope/authenticate.js
@@ -5,7 +5,6 @@
 
 import Route from '@ember/routing/route';
 import { inject as service } from '@ember/service';
-import { hash } from 'rsvp';
 
 export default class ScopesScopeAuthenticateRoute extends Route {
   // =services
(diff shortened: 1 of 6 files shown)


ui/desktop/app/routes/scopes/scope/authenticate.js
```diff
@@ -28,13 +29,42 @@ export default class ScopesScopeAuthenticateRoute extends Route {
...
+      {
+        scope_id: 'global',
+        recursive: true,
+      },
+    );
+    // Fetch org scopes
+    // and filter out any that have no auth methods
+    const scopes = this.modelFor('scopes').filter(
+      ({ id: scope_id, isOrg }) =>
+        isOrg &&
```
ZedLi, line 60: I think this is fine but all the scopes from `this.modelFor` should already be only org scopes so this is just a redundant check?

ui/desktop/app/routes/scopes/scope/authenticate.js
```diff
@@ -28,13 +29,42 @@ export default class ScopesScopeAuthenticateRoute extends Route {
...
+        authMethodsForAllScopes.filter((i) => i.scopeID === scope_id).length,
```
ZedLi, line 61: We really shouldn't do a double `filter` here and instead create a set of valid scope IDs from `authMethodsForAllScopes` to do this check since we now know that some users can have a lot of scopes.

ui/desktop/app/routes/scopes/scope/authenticate.js
```diff
@@ -28,13 +29,42 @@ export default class ScopesScopeAuthenticateRoute extends Route {
...
+    // Fetch auth methods for current scope
+    const authMethods = await this.resourceFilterStore.queryBy(
+      'auth-method',
+      {
+        authorized_actions: [{ contains: 'authenticate' }],
+      },
+      {
+        scope_id,
+      },
+    );
```
ZedLi, line 44: Is this better than just doing a `peekAll` and filtering the relevant auth methods since we already retrieve all auth methods below?

ui/desktop/app/routes/scopes/scope/authenticate.js
cameronperera: Good point. Removed.

ui/desktop/app/routes/scopes/scope/authenticate.js
```diff
@@ -28,13 +29,34 @@ export default class ScopesScopeAuthenticateRoute extends Route {
...
+    const scopeIDs = new Set(authMethodsForAllScopes.map((i) => i.scopeID));
+    // Fetch org scopes
+    // and filter out any that have no auth methods
+    const scopes = this.modelFor('scopes').filter(({ id: scope_id }) =>
+      scopeIDs.has(scope_id),
+    );
+    const authMethods = authMethodsForAllScopes.filter(
+      (i) => i.scopeID === scope_id,
```
ZedLi, line 53: Can we call the `i` something like `authMethod` just to have descriptive variables? Same with the other map

ui/desktop/app/routes/scopes/scope/authenticate.js
```diff
@@ -28,13 +29,34 @@ export default class ScopesScopeAuthenticateRoute extends Route {
...
+    );
+
     return hash({
```
ZedLi, line 56: Out of curiosity, do we still need this `hash`?

ZedLi (approved):
Looks good!

ui/desktop/app/routes/scopes/scope/authenticate.js
cameronperera: Updated.

ui/desktop/app/routes/scopes/scope/authenticate.js
cameronperera: Updated.

ui/desktop/app/routes/scopes/scope/authenticate.js
cameronperera: We do not.

ui/desktop/app/templates/scopes/scope/authenticate.hbs
```diff
@@ -57,19 +57,5 @@
       </p>
 
     {{/if}}
-    {{#unless @model.authMethods}}
```
ZedLi, line 60: Should we actually keep this? What if someone for whatever reason just deleted all their auth methods like this 😆 

image

cameronperera: Yeah that is a good point. I will add it back now.

PR #2134, opened 2024-02-09
refactor: 💡 remove route-action usage in accounts

lisbet-alvarez:
✅ Closes: https://hashicorp.atlassian.net/browse/ICU-12616

## Description

Refactor accounts crudl to not use route-actions. These changes should NOT break any tests or generate any new warnings. 

## How to Test

1. Test that set-password functionality works same as before (no new errors).
2. Test that change-password functionality for password type accounts works same as before.
3. Test that delete, edit, save functionality for accounts works same as before.

account set-password
:technologist: [Admin preview](https://boundary-ui-git-icu-12616-remove-route-action-be7d34-hashicorp.vercel.app/account/change-password)

account crudl
:technologist: [Admin preview](https://boundary-ui-git-icu-12616-remove-route-action-be7d34-hashicorp.vercel.app/scopes/global/auth-methods/am_3enprm7y2t/accounts)

## Checklist:


ui/admin/app/controllers/scopes/scope/auth-methods/auth-method/accounts.js
```diff
@@ -0,0 +1,77 @@
...
+        'scopes.scope.auth-methods.auth-method.accounts.account',
+        account,
+      );
+    } else {
+      await this.router.transitionTo(
+        'scopes.scope.auth-methods.auth-method.accounts',
+      );
+    }
+    this.router.refresh();
+  }
```
lisbet-alvarez, line 59: I'm noticing that the existing actions (for the most part) use await for transitionTo() and replaceWith(). Though not in all cases is `await` needed, which is confusing.

calcaide: @lisbet-alvarez 
That's a very good point. I wonder why we are using `await` for `transitionTo`. I am checking documentation of the service [here](https://api.emberjs.com/ember/release/classes/RouterService/methods/transitionTo?anchor=transitionTo) and  redirects [documentation here](https://guides.emberjs.com/release/routing/redirection/). None of them talk about anything being returned or the necessity to `await`.

I will suggest to not use `await` and bring this topic to tech time, just in case other team members have a say about it.

calcaide (approved):
Tested manually following the test instructions, everything works awesome!! Great work :muscle: :woman_dancing: 

P.S: the tests, at least the change password should be tested against a real `boundary dev` instance, does not work using mirage.

ui/admin/app/controllers/scopes/scope/auth-methods/auth-method/accounts.js
ZedLi: I think that's a mistake if we are, I don't think it should need to be `await`ed

ZedLi: Wait I lied, I just looked up the documentation and the [transition](https://api.emberjs.com/ember/4.12/classes/Transition) is actually `thennable` which makes it in practice a promise. These actually do need to be awaited

ui/admin/app/controllers/account/change-password.js
```diff
@@ -5,9 +5,41 @@
...
+  @notifyError(({ message }) => message, { catch: true })
+  @notifySuccess('notifications.save-success')
+  async changePassword(account, currentPassword, newPassword) {
+    await account.changePassword(currentPassword, newPassword);
+    // Transition to index after success.  We ignore errors in the transition,
+    // since the index route is responsible for additional
+    // redirects depending on the authentication context.  These additional
+    // redirects may abort this initial transition, resulting in a
+    // TransitionAborted error, which we do not want to show to the user.
+    await this.router.replaceWith('index').catch(() => {});
```
ZedLi, line 35: I know this is just personal taste but I'm not the biggest fan of mixing both `await` and using `.catch`. I think if we're awaiting we should just handle this in a normal try catch block but I'm not sure what other's opinions are. 

In this simple use case, it doesn't end up mattering since we just swallow the error and I guess it's fine if this is the only time we use it, but I'm personally not the biggest fan because in general you now have to think about it in both a standard synchronous way using `await` but still being mindful of the asynchronous `.catch` callback. This [article](https://www.robinwieruch.de/javascript-async-await-without-try-catch/) might explain it a bit better.

I think it's fine if we say we only use it in this type of scenario but I also don't think it's a big deal to be consistent and just use try catch?

ui/admin/app/controllers/scopes/scope/auth-methods/auth-method/accounts.js
calcaide: Nice digging @ZedLi 

After checking the service documentation and read this:
`returns Transition the transition object associated with this attempted transition` and understood `the transtion` object as a plain object, therefore no need for await. But after reading the documentation about transition that you suggested definetly need the `await` for the `thennable`.

ui/admin/app/controllers/account/change-password.js
lisbet-alvarez: Thank you for the resource and detailed explanation, i agree with updating this. Also using try catch IMO is easier to understand.

cameronperera (approved):

PR #2136, opened 2024-02-12
chore: 🤖 Migrate account-claim-maps in OIDC form to use HDS

dvonanderson:
✅ Closes: https://hashicorp.atlassian.net/browse/ICU-12355

:tickets: [Jira ticket](https://hashicorp.atlassian.net/browse/ICU-12355)

## Description
This is the last of the `Hds::Table` migration work.  In this particular PR we decided to use a similar pattern that we use in the [LDAP](https://github.com/hashicorp/boundary-ui/blob/4be8d103237e44843e72f28db0c0c0e7471e199f/ui/admin/app/components/form/auth-method/ldap/index.hbs#L385) auth method form for the `account_attribute_maps` field.  This PR also closes the migration ticket for `Rose::Form::Fieldset`
- Migrate `account_claim_maps` in OIDC auth method form to use `Hds::Table`
- Fix all test in relation.

### Before:
--- 
Screenshot 2024-02-12 at 09 31 10

### After:
--- 
Screenshot 2024-02-12 at 09 30 07

## How to Test 
- Click on Vercel link for Admin UI or pull down PR and run it locally using yarn start
- Click on "Auth Method" in the sidebar navigation.
### Creating OIDC auth method with `account_claim_maps`
- Click on the "New" button in the top right hand corner of page and select OIDC.
- Ensure in the "Account Claim Maps" section you are able to add a from claim and select the to claim and add them to the table.
- After saving the form ensure that you are able to go back to the created OIDC and see that data in a disabled state.
### Editing OIDC auth method with `account_claim_maps`
- Find the newly created OIDC with `account_claim_maps` click on the "Edit" button at bottom.
- Add or delete additional `account_claim_maps` by using the "Add" button or "trash" icon next to the specific function that you want to test.
- Click on "Save" button at the bottom of the page and ensure that the data that you inputed has been saved.

## Checklist:


ui/admin/app/components/form/auth-method/oidc/index.hbs
```diff
@@ -240,138 +240,137 @@
...
-          {{#each @model.account_claim_maps as |claimMap i|}}
-            <body.row as |row|>
-              <row.cell>
-                <form.input
+  <Hds::Form::Fieldset @isOptional={{true}} as |F|>
+    <F.Legend>{{t 'form.account_claim_maps.label'}}</F.Legend>
+    <F.HelperText>{{t 'form.account_claim_maps.help'}}</F.HelperText>
+    {{#let (unique-id) (unique-id) as |fromHeaderID toHeaderID|}}
+      <F.Control>
+        <Hds::Table class='list-wrapper-field' name='account_claim_maps'>
```
calcaide, line 248: Why we are not using `@columns` to define the table header as we did with all the other HDS tables?

calcaide, line 248: Why we are adding a `name` to the table?

ui/admin/app/components/form/auth-method/oidc/index.hbs
```diff
@@ -240,138 +240,137 @@
...
+          <:head as |H|>
+            <H.Tr>
+              <H.Th id={{fromHeaderID}}>{{t 'form.from_claim.label'}}</H.Th>
+              <H.Th id={{toHeaderID}}>{{t 'form.to_claim.label'}}</H.Th>
+              <H.Th>{{t 'titles.actions'}}</H.Th>
+            </H.Tr>
+          </:head>
+          <:body as |B|>
+            {{#each @model.account_claim_maps as |claimMap i|}}
```
calcaide, line 257: Why don't use `@model` at table definition?

ui/admin/app/components/form/auth-method/oidc/index.hbs
```diff
@@ -240,138 +240,137 @@
...
+                    aria-labelledby={{toHeaderID}}
+                    {{on 'change' (set-from-event claimMap 'to')}}
+                    as |F|
+                  >
+                    <F.Options>
+                      <option disabled={{@disabled}} hidden selected value=''>
+                        {{t 'titles.choose-an-option'}}
+                      </option>
+                      {{#each this.toClaims as |toClaimOption|}}
+                        <option
```
calcaide, line 282: aren't we missing `value` here?

calcaide (changes requested):
Thanks for the work. Left some comments.

ui/admin/tests/acceptance/auth-methods/create-test.js
```diff
@@ -44,6 +44,12 @@ module('Acceptance | auth-methods | create', function (hooks) {
   const CLAIMS_SCOPES_BTN_SELECTOR = '[name="claims_scopes"] button';
   const CLAIMS_SCOPES_INPUT_SELECTOR = '[name="claims_scopes"] input';
   const TOGGLE_SELECTOR = '[name="prompts"]';
+
+  const ACCOUNT_CLAIM_MAPS_ADD_BTN_SELECTOR = '[data-test-add-claim-map]';
+  const ACCOUNT_CLAIM_MAPS_INPUT_SELECTOR = '[data-test-from-claim-input';
```
ZedLi, line 49: Isn't this missing the `]`?

ui/admin/app/components/form/auth-method/oidc/index.hbs
dvonanderson: @calcaide definitely tried both variations of getting this to work, but I think it is because the table ".data" key doesn't have access to that second table row.  We had to use the same pattern here in [LDAP](https://github.com/hashicorp/boundary-ui/blob/786e6130cd61cdd44d23bca9c554c6ca9dc8d092/ui/admin/app/components/form/auth-method/ldap/index.hbs#L394)

ui/admin/app/components/form/auth-method/oidc/index.hbs
dvonanderson: https://github.com/hashicorp/boundary-ui/pull/2136/files#r1489882922

ui/admin/app/components/form/auth-method/oidc/index.hbs
dvonanderson: @calcaide this is a placeholder option.

ui/admin/app/components/form/auth-method/oidc/index.hbs
calcaide: @dvonanderson 
Why are we introducing a placeholder? I am not seeing this placeholder on the old table. Also, it what state are you expecting this placeholder to be shown? I tested your PR on 3 states: new, edit and shown, all states.
Also, if you define `@disabled={{@disabled}}` I don't think it will be shown.

Let me know if I am missing something.

dvonanderson: @calcaide we have used it before in our codebase, and it is currently a part of the `select` component for `Form::Field::ListWrapper`.  I have removed it and I will bring it up for a discussion at tech time.  The general idea is when the user sees the [dropdown](https://github.com/hashicorp/boundary-ui/blob/786e6130cd61cdd44d23bca9c554c6ca9dc8d092/ui/admin/app/components/form/policy/policy-selection/index.hbs#L24) it will show "Choose an Option"  instead of the first item in the options array.  I definitely made a mistake on the `disabled` argument. For now though I have modified the code.

calcaide: I am aware of the place holder functionality and that we have it in our codebase, also examples on HDS documentation. I am not saying we should or shouldn't have it. What I am asking is, in this specific use case and particular component, do we need to have a placeholder?  and since the previous table has no placeholder defined, I am asking what is the reason why we add it here. Did design ask for it? Is an UX improvement? 

I opened Vercel with your implementation:

Edit:
edit

List:
list

New:
new

Any of these states show no placeholder, and therefore my question.

dvonanderson: @calcaide no UX improvement request from design.  the incorrect value for the `disabled` argument was the reason you didn't see it, but its been removed.

ui/admin/app/components/form/auth-method/oidc/index.hbs
calcaide: As per your answer on the previous message, I understand that the issue when using `@model` is that will not allow us to add the last row (attached screenshot), is that correct?

claim_maps

ui/admin/app/components/form/auth-method/oidc/index.hbs
```diff
@@ -240,138 +240,131 @@
...
+          </:head>
+          <:body as |B|>
+            {{#each @model.account_claim_maps as |claimMap i|}}
+              <B.Tr>
+                <B.Td>
+                  <Hds::Form::TextInput::Field
+                    @value={{claimMap.from}}
+                    @type='text'
+                    disabled={{form.disabled}}
+                    aria-labelledby={{fromHeaderID}}
```
calcaide, line 264: Can you explain the usage of `aria-labelledby` in this case? checking HDS documentation about `TextInput` explains about `aria-describedby` but not `aria-labeledby` and since aria is important I would like to understand its usage here.

dvonanderson: @calcaide I think It could go either way.  I guess since we are only providing a name, concise description, or in our case a unique-id we use `labelledby`.  If the `fromHeaderId` provided additional information or a longer description of the current element, we should use `describedby`.

ui/admin/app/components/form/auth-method/oidc/index.hbs
dvonanderson: @calcaide that is correct

ui/admin/app/components/form/auth-method/oidc/index.hbs
```diff
@@ -240,138 +240,137 @@
...
-                <form.input
+  <Hds::Form::Fieldset @isOptional={{true}} as |F|>
+    <F.Legend>{{t 'form.account_claim_maps.label'}}</F.Legend>
+    <F.HelperText>{{t 'form.account_claim_maps.help'}}</F.HelperText>
+    {{#let (unique-id) (unique-id) as |fromHeaderID toHeaderID|}}
+      <F.Control>
+        <Hds::Table class='list-wrapper-field' name='account_claim_maps'>
+          <:head as |H|>
+            <H.Tr>
+              <H.Th id={{fromHeaderID}}>{{t 'form.from_claim.label'}}</H.Th>
```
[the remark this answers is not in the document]
dvonanderson: @calcaide we can remove this, was going to use it in testing and never needed it.

ui/admin/app/components/form/auth-method/oidc/index.hbs
dvonanderson: `name` argument removed

ui/admin/app/components/form/auth-method/oidc/index.hbs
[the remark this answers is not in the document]
calcaide: I saw this new implementation creates `fromHeaderID` and `toHeaderID` in line 246 and I am wondering what's the reason behind these Id's, are just for testing purposes?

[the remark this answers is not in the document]
dvonanderson: @calcaide I have a limited understanding of `(unique-id)` helper, but in this case, because we are not using the standardized way of the table with using `@columns` argument so therefore we loose control over the `a11y` for the inputs and select field labels?

ui/admin/app/components/form/auth-method/oidc/index.hbs
```diff
@@ -240,138 +240,131 @@
...
-          {{#each @model.account_claim_maps as |claimMap i|}}
-            <body.row as |row|>
-              <row.cell>
-                <form.input
+  <Hds::Form::Fieldset @isOptional={{true}} as |F|>
+    <F.Legend>{{t 'form.account_claim_maps.label'}}</F.Legend>
+    <F.HelperText>{{t 'form.account_claim_maps.help'}}</F.HelperText>
+    {{#let (unique-id) (unique-id) as |fromHeaderID toHeaderID|}}
+      <F.Control>
+        <Hds::Table class='list-wrapper-field'>
```
calcaide, line 248: I am getting a bit confused by the answers here. I will try to clarify:

- [Why are we adding ID's to the TH](https://github.com/hashicorp/boundary-ui/pull/2136/files#r1489761821)? So we can identify them, for aria purposes, the TH cell titles are labels to the TR cells inputs. Therefore, they need identification so we can reference them properly.

- [Usage of `unique-id`](https://github.com/hashicorp/boundary-ui/pull/2136/files#r1492633347): [Documentation about unique-id here](https://api.emberjs.com/ember/4.12/classes/Ember.Templates.helpers/methods/each?anchor=unique-id). NONE BLOCKING COMMENT: Do we need unique-id in this use case? it is fine to leave it, but we could also had use a human readable identifier, since the purpose of adding the identifier is aria. I.E: `id=fromClaim`. 

Are these assumptions correct?

dvonanderson: @calcaide that is correct and I totally agree, I have a commit coming to fix this.

ui/admin/app/components/form/auth-method/oidc/index.hbs
```diff
@@ -243,120 +243,102 @@
...
-                  @contextual={{true}}
-                  as |field|
-                >
-                  <field.label>{{t 'form.from_claim.label'}}</field.label>
-                  <field.field />
-                </form.input>
-              </row.cell>
-              <row.cell>
-
+                  @type='text'
```
calcaide, line 264: As we discussed in the pass tech time (Thursday 22 February) and as we have being doing with all the `HDS::Text:inputs` for consistency, if the `type=text` we do not declare it since it's the default.

ui/admin/app/components/form/auth-method/oidc/index.hbs
```diff
@@ -243,120 +243,102 @@
...
+                  disabled={{form.disabled}}
+                  aria-labelledby='fromClaimHeader'
+                  {{on 'input' (set-from-event claimMap 'from')}}
+                />
+              </B.Td>
+              <B.Td>
                 <Hds::Form::Select::Field
```
calcaide, line 271: Don't we need the `@isInvalid` here? wondering why we have it earlier and now we do not.

ui/admin/app/components/form/auth-method/oidc/index.hbs
```diff
@@ -243,120 +243,102 @@
...
               <Hds::Form::Select::Field
                 @value={{this.newToClaim}}
+                @width='100%'
                 disabled={{form.disabled}}
+                aria-labelledby='toClaimHeader'
                 {{on 'change' (set-from-event this 'newToClaim')}}
+                data-test-to-claim-select
                 as |F|
               >
-                <F.Label class='sr-only'>{{t
```
calcaide, line 337: If we are deleting the usage of the class `sr-only`, should we delete the class declaration from the css file?

calcaide (changes requested):
Left some comments.

ui/admin/app/components/form/auth-method/oidc/index.hbs
dvonanderson: @calcaide indeed cleaning up in next commit

ui/admin/app/components/form/auth-method/oidc/index.hbs
dvonanderson: @calcaide cleaning up in next commit

ui/admin/app/components/form/auth-method/oidc/index.hbs
dvonanderson: @calcaide I am adding back the `@isInvalid` argument.  I have been trying to find documentation or patterns on errors on creation and updating.  I have been looking at the [auth_method_service.proto](https://github.com/hashicorp/boundary/blob/main/internal/proto/controller/api/services/v1/auth_method_service.proto), that gives you some detail on what messages are returned on those methods, but it seems we have `@isInvalid` on thing that may not even return an error message?

addons/rose/addon/styles/hds/overrides.scss
```diff
@@ -30,13 +30,6 @@
   .hds-table__td p {
     color: var(--token-form-indicator-optional-color);
   }
-
-  // TODO: This is a temporary hack to fix an issue causing weird whitespace
-  //   and scrollbars from an accessible class in HDS. We have to use !important
-  //   as they already forced it with !important and we can't override it otherwise
-  .sr-only {
```
ZedLi, line 37: Whoa this can't be deleted. Read the comment again. It's not for our usage of `sr-only` but a temporary fix for its usage in HDS.

ZedLi (changes requested):
I had a few comments I forgot to submit previously that looks like they're still relevant

addons/rose/addon/styles/hds/overrides.scss
dvonanderson: @ZedLi t is not used by any other component?

ZedLi: It's used by many HDS components. This is an override for an issue we had in desktop client that hopefully we can remove in the future

dvonanderson: @ZedLi ahh i see it now, its part of the table

ui/admin/app/components/form/auth-method/oidc/index.hbs
```diff
@@ -243,120 +243,105 @@
...
+                  @isIconOnly={{true}}
+                  disabled={{form.disabled}}
+                  @isInvalid={{@model.errors.account_claim_maps}}
                   {{on 'click' (fn @removeItemByIndex 'account_claim_maps' i)}}
-                >
-                  {{t 'actions.remove'}}
-                </Rose::Button>
-              </row.cell>
-            </body.row>
+                  data-test-remove-claim-map={{claimMap.from}}
```
calcaide, line 301: @dvonanderson 
What is the purpose of give a value to the `data-test-` selector?

ui/admin/app/components/form/auth-method/oidc/index.hbs
```diff
@@ -243,120 +243,105 @@
...
+          <B.Tr>
+            <B.Td>
+              <Hds::Form::TextInput::Field
                 @value={{this.newFromClaim}}
-                @label={{t 'resources.auth-method.titles.new-from-claim'}}
+                disabled={{form.disabled}}
+                @isInvalid={{@model.errors.account_claim_maps}}
+                aria-labelledby='fromClaimHeader'
+                {{on 'input' (set-from-event this 'newFromClaim')}}
+                data-test-from-claim-input
```
calcaide, line 314: None blocking comment: as the pattern followed in this file, I will suggest to let the `data-test-` selector name just to describe the field, but it does not need the type.
i.e: `data-test-from-clain-input` or `data-test-to-claim-select` can be `data-test-from-claim` and `data-test-to-claim`. So the type of the field is not coupled to its selector.

ui/admin/app/components/form/auth-method/oidc/index.hbs
dvonanderson: @calcaide during the modification of the test for OIDC form, I was looking at a way of grabbing a specific input field based on the value, but I found a different solution, so this one can actually be removed

calcaide (changes requested):
Left comments

calcaide:
I would suggest to add a test procedure to the PR description on editing the form.

ui/admin/tests/acceptance/auth-methods/update-test.js
```diff
@@ -59,6 +59,12 @@ module('Acceptance | auth-methods | update', function (hooks) {
     '.list-wrapper-field tbody tr:last-child select';
   const SIGNING_ALGORITHMS_ADD_BTN_SELECTOR = '[data-test-add-option-button]';
 
+  const ACCOUNT_CLAIM_MAPS_REMOVE_BTN_SELECTOR = '[data-test-remove-claim-map]';
+  const ACCOUNT_CLAIM_MAPS_ADD_BTN_SELECTOR = '[data-test-add-claim-map]';
+  const ACCOUNT_CLAIM_MAPS_INPUT_SELECTOR = '[data-test-from-claim-input';
```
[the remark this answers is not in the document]
dvonanderson: @ZedLi fixed

ui/admin/app/components/form/auth-method/oidc/index.hbs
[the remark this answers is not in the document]
dvonanderson: @ZedLi I will wait for Dharini's `key-value` upgrade to address this migration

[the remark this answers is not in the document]
calcaide: @dvonanderson can you provide an explanation/context of what have been done here? did we end up applying `key-value` component? if not, why we did not? what are the limitations/implications?

I saw the comment resolve but I can not see any changes on the PR regarding this and wondering what was the result of this comment.

ui/admin/app/components/form/auth-method/oidc/index.hbs
calcaide: @dvonanderson 
- Did you test editing/creating an OIDC Auth Method and adding Account Claim Maps with empty From Claim field?
- I have no knowledge of how the BE manages errors and check our requests when submitting a form, but curious why looking at that proto file, within there is no mention to the fields this form has.

ui/admin/app/components/form/auth-method/oidc/index.hbs
calcaide: @dvonanderson 
Can you provide an explanation of what is the outcome of this comment? You mention that the class will be removed and mark the conversation as resolved, but I do not see any change on CSS file or an answer on your end.

ui/admin/app/components/form/auth-method/oidc/index.hbs
```diff
@@ -243,120 +243,105 @@
...
+        <:body as |B|>
           {{#each @model.account_claim_maps as |claimMap i|}}
-            <body.row as |row|>
-              <row.cell>
-                <form.input
-                  @type='text'
-                  @name='from_claim'
+            <B.Tr>
+              <B.Td>
+                <Hds::Form::TextInput::Field
```
calcaide, line 262: This field used to have an attribute `name` what is the reason to not use it here?

calcaide:
The current implementation breaks the behaviour of the previous one when adding an Account Claim Map.

Steps to reproduce:

- Edit or Create an OIDC Auth Method.
- When on the editing/creation form, Add an Account Claim map.
- When you Click the button "Add", the Claim map is add to the list an the input text to add a new Claim map does not clean up its value, keeping the value of the recent added field.

Here a video to clarify my explanation:

https://github.com/hashicorp/boundary-ui/assets/9775006/00115247-e13f-4c48-942c-bc7437b1d11e

calcaide:
Closing this PR. [We have a PR refactoring key/value comp](https://github.com/hashicorp/boundary-ui/pull/2180) that already does the HDS adoption this PR was attempting.

PR #2144, opened 2024-02-15
chore: 🤖 add credential type select list

DhariniJeeva:
✅ Closes: https://hashicorp.atlassian.net/browse/ICU-12566

:tickets: [Jira ticket](https://hashicorp.atlassian.net/browse/ICU-12566)

create: 

Untitled

edit mode:
Untitled 2

## How
 to Test

## Checklist:


ui/admin/app/components/form/credential-library/vault-generic/index.hbs
```diff
@@ -108,6 +108,38 @@
...
+    as |F|
+  >
+    <F.Label>
+      {{t 'resources.credential-library.form.credential_type.label'}}
+    </F.Label>
+    <F.HelperText>
+      {{t 'resources.credential-library.form.credential_type.help'}}
+    </F.HelperText>
+    <F.Options>
+      <option disabled hidden selected value='' />
```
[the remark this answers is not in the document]
DhariniJeeva: good point 🤔 I'll update this so the users can undo selection if they want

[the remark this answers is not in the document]
DhariniJeeva: updated

ui/admin/app/components/form/credential-library/vault-generic/index.hbs
```diff
@@ -108,6 +108,38 @@
...
+    as |F|
+  >
+    <F.Label>
+      {{t 'resources.credential-library.form.credential_type.label'}}
+    </F.Label>
+    <F.HelperText>
+      {{t 'resources.credential-library.form.credential_type.help'}}
+    </F.HelperText>
+    <F.Options>
+      <option selected value='' />
```
ZedLi, line 126: We can't always have this set as `selected` right? Don't we still need to have it check if it should be selected if it's now a real selection?

DhariniJeeva: theoretically yes? But it looks like the selection seems to work fine when I go back and forth with the options 🤔

ZedLi: You're technically right and it looks like it still works (I checked and the DOM _property_ for `selected` is correctly set to false, which is what matters, even though the DOM _attribute_ for selected is still there) but I'm worried about future bugs in the future. It might be good enough to just not include the attribute at all?

ZedLi (approved):
Tested with a local vault and boundary dev in docker and it works great!

cameronperera (approved):

ui/admin/app/components/form/credential-library/vault-generic/index.hbs
DhariniJeeva: I removed the attr to avoid any bugs, tested and works as expected with no attr.

PR #2156, opened 2024-02-22
Icu 12567 add new field credential mapping overrides

DhariniJeeva:

https://github.com/hashicorp/boundary-ui/assets/15043878/4be505c9-b306-429a-99f9-ea228d7adf49

 
 # How to test 

- set up vault and use boundary controller
- create a vault credential library
- you can add/update/delete credential mapping overrides

## Checklist:


ui/admin/app/components/form/credential-library/vault-generic/index.js
```diff
@@ -30,4 +30,23 @@ export default class FormCredentialLibraryVaultGenericComponent extends Componen
...
+
+  /**
+   * Clear the previously selected key value pair when toggling between credential types on a new form
+   *
+   */
+  @action
+  selectCredentialType({ target: { value } }) {
+    if (
+      this.args.model.isNew &&
+      Object.keys(this.args.model.credential_mapping_overrides || {}).length
```
ZedLi, line 46: Personal taste but I prefer `??` instead these days for clarity

addons/api/addon/serializers/credential-library.js
```diff
@@ -31,9 +32,39 @@ export default class CredentialLibrarySerializer extends ApplicationSerializer {
...
+    const { credential_type, type, credential_mapping_overrides, isNew } =
+      snapshot.record;
+    if (
+      type === TYPE_CREDENTIAL_LIBRARY_VAULT_GENERIC &&
+      credential_mapping_overrides &&
+      credential_type &&
+      !isNew
+    ) {
+      //API expects to send null to fields if it is undefined or deleted
+      const remainingAttrs = Object.entries(
```
ZedLi, line 46: Why is this needed?

ui/admin/app/components/form/field/list-wrapper/select-text/index.hbs
```diff
@@ -0,0 +1,111 @@
...
+            <Hds::Form::Select::Base
+              @value={{this.newOptionKey}}
+              @width={{@width}}
+              disabled={{@disabled}}
+              aria-labelledby={{keyHeaderID}}
+              {{on 'change' (set-from-event this 'newOptionKey')}}
+              as |F|
+            >
+              <F.Options>
+                <option value='' />
```
ZedLi, line 77: Shouldn't this be not selectable?

addons/api/addon/serializers/credential-library.js
DhariniJeeva: When updating the select list, the API explicitly wants us to send null to fields we don't need. Otherwise, they will return the unmodified value from previous versions. 

For ex, If I have a username attr and want to change that to a `password` attr, I will need to send both the fields and make the username null. Otherwise, they will send values for both fields.

ui/admin/app/components/form/field/list-wrapper/select-text/index.js
```diff
@@ -0,0 +1,123 @@
...
+   */
+  get options() {
+    return this.args?.options || this.args?.model?.[this.args.name];
+  }
+
+  /**
+   * Determines if we need to show an empty row to the users to enter more key/value pairs
+   * @type {object}
+   */
+  get showNewRow() {
```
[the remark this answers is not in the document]
DhariniJeeva: actually I had that thought and that would definitely eliminate the need for an extra function but wasn't sure if that was a better option. I can update it

addons/api/addon/serializers/credential-library.js
ZedLi: I had a setup where I changed the key and it wasn't setting null, just not sending the field which was working fine, is this expected?

https://github.com/hashicorp/boundary-ui/assets/5783847/e53f8d24-fca3-48b3-8268-6d07eebd7b38

addons/api/addon/serializers/credential-library.js
```diff
@@ -31,9 +32,39 @@ export default class CredentialLibrarySerializer extends ApplicationSerializer {
...
+    const value = super.serializeAttribute(...arguments);
+    const { credential_type, type, credential_mapping_overrides, isNew } =
+      snapshot.record;
+    if (
+      type === TYPE_CREDENTIAL_LIBRARY_VAULT_GENERIC &&
+      credential_mapping_overrides &&
+      credential_type &&
+      !isNew
+    ) {
+      //API expects to send null to fields if it is undefined or deleted
```
ZedLi, line 45:
```suggestion
      // API expects to send null to fields if it is undefined or deleted
```

Can we add a space between the slashes for comments 😅

DhariniJeeva: oops I somehow always forget to add spaces there 😿

addons/api/addon/serializers/credential-library.js
DhariniJeeva: It is set in the serializer, so I don't think you would see it here?   I can see the null being sent to the API in the network call

ZedLi: I think I understand now, I don't see it still but it doesn't matter from the API perspective. This was just ember re-using old values bc it assumes they're still relevant if we don't tell it otherwise

addons/api/addon/serializers/credential-library.js
```diff
@@ -31,9 +32,39 @@ export default class CredentialLibrarySerializer extends ApplicationSerializer {
...
+            obj,
+            credential_mapping_overrides[key] ? {} : { [key]: null },
+          ),
+        {},
+      );
+
+      json.credential_mapping_overrides = {
+        ...credential_mapping_overrides,
+        ...remainingAttrs,
+      };
```
ZedLi, line 60: Few comments regarding this, I don't think you needed to use `Object.entries` since you only used the keys anyways. And couldn't we have included the actual values of the fields as part of the reduce rather than keeping them separate and doing something like this:

```suggestion
      json.credential_mapping_overrides = Object.keys(
        options.mapping_overrides[credential_type],
      ).reduce((obj, key) => {
        if (credential_mapping_overrides[key]) {
          obj[key] = credential_mapping_overrides[key];
        } else {
          obj[key] = null;
        }
        return obj;
      }, {});
```

ui/admin/app/components/form/credential-library/vault-generic/index.js
```diff
@@ -30,4 +30,23 @@ export default class FormCredentialLibraryVaultGenericComponent extends Componen
...
+    return options.mapping_overrides[this.args.model.credential_type];
+  }
+
+  /**
+   * Clear the previously selected key value pair when toggling between credential types on a new form
+   *
+   */
+  @action
+  selectCredentialType({ target: { value } }) {
+    if (
```
ZedLi, line 44: Actually do you need this if check? Couldn't you just set it regardless if they're switching types? It should always be an empty object anyways right?

DhariniJeeva: good call, yes, I don't need it, since a user can't update this type once its created, I will remove

ui/admin/app/components/form/field/list-wrapper/select-text/index.js
```diff
@@ -0,0 +1,123 @@
...
+      const field = this.args.name;
+
+      const newObj = Object.keys(this.args.model[field] || {}).reduce(
+        (obj, key) => {
+          key !== selectedKey && (obj[key] = this.args.model[field][key]);
+          return obj;
+        },
+        {},
+      );
+      set(this.args.model, field, newObj);
```
ZedLi, line 120: This seems unnecessary to reduce over the object when you really just want to delete a key (I also want to mention this component assumes an object now instead of an array, which is unexpected).

You could have done something like: 
```suggestion
      const { [selectedKey]: _, ...newObj } = this.args.model[field];
      set(this.args.model, field, newObj);
```
or

```suggestion
      const newObj = { ...this.args.model[field] };
      delete newObj[selectedKey];
      set(this.args.model, field, newObj);

```

ui/admin/app/components/form/field/list-wrapper/select-text/index.js
```diff
@@ -0,0 +1,123 @@
...
+        });
+      }
+    } else {
+      const field = this.args.name;
+      const newRow = {};
+      newRow[this.newOptionKey] = this.newOptionValue;
+      this.args.model[field] = {
+        ...this.args.model[field],
+        ...newRow,
+      };
```
ZedLi, line 94: This could also be simplified a bit:

```suggestion
      this.args.model[field] = {
        ...this.args.model[field],
        [this.newOptionKey]: this.newOptionValue,
      };
```

ui/admin/app/components/form/field/list-wrapper/select-text/index.hbs
DhariniJeeva: I added the text, but if we make this disabled, the select option isn't getting selected when there's one item in the list anyway. The user won't be able to select this empty option and click `add`, since we only enable `add` when there a valid option is selected from the dropdown. I updated the PR description to show this behavior.

ui/admin/app/components/form/field/list-wrapper/select-text/index.js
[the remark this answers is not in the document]
DhariniJeeva: I added a new arg `removeDuplicates` which allows the consumer to decide if they want to use that feature, let me know your thoughts

ui/admin/app/components/form/field/list-wrapper/select-text/index.js
DhariniJeeva: simplified

ui/admin/app/components/form/field/list-wrapper/select-text/index.js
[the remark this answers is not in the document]
DhariniJeeva: I thought about that, but this form is essentially a key/value that doesn't allow duplicates and I liked that we didn't have to use our custom transform object-as-array for this..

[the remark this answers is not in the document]
ZedLi: We can probably talk about this component after we get it in but there's no reason this form _shouldn't_ allow duplicates, that happens to be an implementation detail for specifically the vault credential overrides mapping which we've now baked and coupled into this component

ui/admin/app/components/form/field/list-wrapper/select-text/index.hbs
ZedLi: Oh but i meant the user should never be able to select this option right? It's not like `credential_type` where the user might want to select it as that "removes" the choice. In this case, the user would just delete the row

ui/admin/app/components/form/credential-library/vault-generic/index.hbs
```diff
@@ -140,6 +140,46 @@
...
+        {{/if}}
+      </:fieldset>
+
+      <:field as |F|>
+        <F.SelectText
+          @name='credential_mapping_overrides'
+          @selectOptions={{this.mappingOverrides}}
+          @model={{@model}}
+          @disabled={{form.disabled}}
+          @removeDuplicates='true'
```
ZedLi, line 176: Shouldn't this be a boolean and not a string?

ui/admin/app/components/form/field/list-wrapper/select-text/index.hbs
DhariniJeeva: ty, done

PR #2158, opened 2024-02-23
test(e2e): Add enterprise tests for creating TCP/SSH targets

moduli:
This PR adds 2 new tests to the Admin UI e2e test suite. These tests are specific to the enterprise version due to how the enterprise version has a selector for the target type.
- Creates a TCP target and connects to it
- Creates an SSH target and connects to it

This PR also includes a few adjustments
- Fixes some `await has no effect` warnings by updating the `return` description on some methods
- Fixes the tags on the `Delete Resources` test to be AWS only (it creates an AWS Dynamic Host catalog, which is only available in an AWS scenario)


ui/admin/tests/e2e/helpers/boundary-ui.js
```diff
@@ -312,6 +438,53 @@ exports.addBrokeredCredentialsToTarget = async (
   await expect(page.getByRole('link', { name: credentialName })).toBeVisible();
 };
 
+/**
+ * Uses the UI to navigate to the specified Target and add the Brokered Credentials to it.
+ * @param {Page} page Playwright page object
+ * @param {string} targetName Name of the target associated with the session
+ * @param {string} credentialName Name of the credentials to be added to the target
+ */
+exports.addInjectedCredentialsToTarget = async (
```
stasryzhov, line 447:
```suggestion
 * Uses the UI to navigate to the specified Target and add the Injected Credentials to it.
 * @param {Page} page Playwright page object
 * @param {string} targetName Name of the target associated with the session
 * @param {string} credentialName Name of the credentials to be added to the target
 */
exports.addInjectedCredentialsToTarget = async (
```

stasryzhov (approved):

PR #2161, opened 2024-02-26
Fix credential library mocks

DhariniJeeva:
## Description

This PR fixes the mock data when a vault credential library is updated with mapping_overrides

## How to Test

1. go to credential-library->vault
2. edit an existing vault credential library and include mapping override attributes
3. click save
4. the form should reflect only the attributes you added in the above steps 

## Checklist:

### DIFF
--- a/addons/api/mirage/config.js
+++ b/addons/api/mirage/config.js
@@ -632,7 +632,26 @@ function routes() {
       { queryParams: { credential_store_id: credentialStoreId } },
     ) => credentialLibraries.where({ credentialStoreId }),
   );
-  this.get('/credential-libraries/:id');
+  this.get(
+    '/credential-libraries/:id',
+    function ({ credentialLibraries }, { params: { id } }) {
+      const library = credentialLibraries.find(id);
+      // BE API only updates these fields if there's an update mask, otherwise they remain unmodified
+      // but we don't need this flow in our mocks, so we are deleting fields with null value
+      const { credentialMappingOverrides } = library;
+      if (
+        credentialMappingOverrides &&
+        Object.keys(credentialMappingOverrides).length
+      ) {
+        Object.keys(credentialMappingOverrides).forEach((key) => {
+          if (credentialMappingOverrides[key] === null) {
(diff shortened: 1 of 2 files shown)


cameronperera (approved):

addons/api/mirage/config.js
```diff
@@ -632,7 +632,23 @@ function routes() {
       { queryParams: { credential_store_id: credentialStoreId } },
     ) => credentialLibraries.where({ credentialStoreId }),
   );
-  this.get('/credential-libraries/:id');
+  this.get(
+    '/credential-libraries/:id',
+    function ({ credentialLibraries }, { params: { id } }) {
+      const library = credentialLibraries.find(id);
+      // BE API only updates these fields if there's an update mask, otherwise they remain unmodified
```
ZedLi, line 639: Sorry I just saw this PR but shouldn't this be handled in the mirage serializers?

DhariniJeeva: I tried to mimic the API. We send these in the PATCH request, but we don't get these back from the GET req. If I were to add these in the serializer, wouldn't that change the flow of how this is done in the API?

ZedLi: Are you referring to our our application serializers or mirage serializers? The mirage serializers would just handle how the data shows up in mirage so it shouldn't affect anything to do with the API?

DhariniJeeva: I was referring to the Mirage serializer and trying to align it with how it works when a real API is used. I don't know if we usually add these in Mirage serializers. I can move it there, but my reasoning was that we don't serialize this when a real API is involved. So, I did this in the GET request.

ZedLi: I'm not sure I understand, don't we normally do these type of of things normally in the mirage serializers? To get it to match how the real API looks like?

The mirage configs I thought were mainly for changing what the actual data is, not how it should look like

DhariniJeeva: Had offline discussion, and looks like that's where this processing should exist - addressed it here

PR #2164, opened 2024-02-27
test(e2e): Add LDAP Admin UI e2e test

moduli:
This PR adds an Admin UI e2e test that walks through setting up an LDAP auth method.


ui/admin/tests/e2e/helpers/boundary-ui.js
```diff
@@ -668,9 +668,10 @@ exports.addMemberToGroup = async (page, userName) => {
 /**
  * Uses the UI to create a new role. Assumes you have selected the desired scope.
  * @param {Page} page Playwright page object
- * @param {string} roleName Name of the new role
  */
-exports.createNewRole = async (page, roleName) => {
+exports.createRole = async (page) => {
```
stasryzhov, line 672: 👍

ui/admin/tests/e2e/tests/auth-method-ldap.spec.js
```diff
@@ -0,0 +1,141 @@
...
+    await page.getByLabel('Description').fill('LDAP Auth Method');
+    await page.getByLabel('Address').fill(process.env.E2E_LDAP_ADDR);
+    await page.getByLabel('Bind DN').fill(process.env.E2E_LDAP_ADMIN_DN);
+    await page
+      .getByLabel('Bind Password')
+      .fill(process.env.E2E_LDAP_ADMIN_PASSWORD);
+    await page.getByRole('switch', { name: 'Discover DN' }).click();
+    await page.getByLabel('User DN').fill(process.env.E2E_LDAP_DOMAIN_DN);
+    await page.getByLabel('User Attribute').fill('uid');
+    await page.getByLabel('Group DN').fill(process.env.E2E_LDAP_DOMAIN_DN);
```
stasryzhov, line 70: Just to double check, are the User DN and Group DN supposed to be the same?

moduli: Yup. I believe it describes what part of the "LDAP tree" users and groups live. In our case, I've set up the LDAP server for users and groups to live in the same part of the tree.

DhariniJeeva (approved):
Thanks for getting this in! looks great

ui/admin/tests/e2e/tests/auth-method-ldap.spec.js
```diff
@@ -0,0 +1,141 @@
...
+    'E2E_LDAP_GROUP_NAME',
+  ]);
+
+  await checkBoundaryCli();
+});
+
+test('Set up LDAP auth method @ce @ent @docker', async ({ page }) => {
+  await page.goto('/');
+  let org;
+  let connect;
```
DhariniJeeva, line 42: not a blocker- I think these two can be `const`, I don't see them being reassigned anywhere

moduli: I believe those need to be `let`. If I change to `const`, I get this warning
```
'const' declarations must be initialized.ts (1155)
```

ui/admin/tests/e2e/helpers/boundary-ui.js
```diff
@@ -668,9 +668,10 @@ exports.addMemberToGroup = async (page, userName) => {
...
+  const roleName = 'Credential ' + nanoid();
```
DhariniJeeva, line 673: not a blocker totally a nitpick- should this be called `Role`?

moduli: Yup! Will fix that.

PR #2176, opened 2024-03-08
refactor: 💡 remove route-action usage for groups

lisbet-alvarez:
✅ Closes: https://hashicorp.atlassian.net/browse/ICU-12605

:tickets: [Jira ticket](https://hashicorp.atlassian.net/browse/ICU-12605)

## Description

Remove route-action usage for groups. NOTE: The add-members route still uses two route-actions due to filtering. However this will be replaced by future filtering.

## How to Test

1. Start boundary dev instance
2. Test that saving changes to a group and new group still works.
3. Test that editing a group and saving or canceling still works.
4. Test that deleting a group still works.
5. Test that adding and removing members to a group still works.

## Checklist:

### DIFF
--- a/ui/admin/app/controllers/scopes/scope/groups/group/index.js
+++ b/ui/admin/app/controllers/scopes/scope/groups/group/index.js
@@ -1,0 +1,5 @@
+import Controller, { inject as controller } from '@ember/controller';
+
+export default class ScopesScopeGroupsGroupIndexController extends Controller {
+  @controller('scopes/scope/groups/index') groups;
+}
--- a/ui/admin/tests/unit/controllers/scopes/scope/auth-methods/auth-method/accounts/account/index-test.js
+++ b/ui/admin/tests/unit/controllers/scopes/scope/auth-methods/auth-method/accounts/account/index-test.js
@@ -1,3 +1,8 @@
+/**
+ * Copyright (c) HashiCorp, Inc.
+ * SPDX-License-Identifier: MPL-2.0
+ */
+
 import { module, test } from 'qunit';
 import { setupTest } from 'admin/tests/helpers';
 
--- a/ui/admin/tests/unit/controllers/scopes/scope/auth-methods/auth-method/accounts/account/index-test.js
+++ b/ui/admin/tests/unit/controllers/scopes/scope/auth-methods/auth-method/accounts/account/index-test.js
@@ -1,5 +1,5 @@
 import { module, test } from 'qunit';
-import { setupTest } from 'admin/tests/helpers';
+import { setupTest } from 'ember-qunit';
 
 module(
   'Unit | Controller | scopes/scope/auth-methods/auth-method/accounts/account/index',
--- a/ui/admin/tests/unit/controllers/scopes/scope/auth-methods/auth-method/accounts/account/index-test.js
+++ b/ui/admin/tests/unit/controllers/scopes/scope/auth-methods/auth-method/accounts/account/index-test.js
@@ -1,5 +1,5 @@
 import { module, test } from 'qunit';
-import { setupTest } from 'ember-qunit';
+import { setupTest } from 'admin/tests/helpers';
 
 module(
   'Unit | Controller | scopes/scope/auth-methods/auth-method/accounts/account/index',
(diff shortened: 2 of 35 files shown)


ui/admin/app/controllers/scopes/scope/groups/group/index.js
```diff
@@ -0,0 +1,5 @@
+import Controller, { inject as controller } from '@ember/controller';
+
+export default class ScopesScopeGroupsGroupIndexController extends Controller {
+  @controller('scopes/scope/groups/index') groups;
```
calcaide, line 4: None blocking question: I do understand this line injects the groups contoller (`ui/admin/app/controllers/scopes/scope/groups/index.js` is that correct?

ui/admin/tests/unit/controllers/scopes/scope/auth-methods/auth-method/accounts/account/index-test.js
```diff
@@ -1,5 +1,5 @@
 import { module, test } from 'qunit';
-import { setupTest } from 'admin/tests/helpers';
+import { setupTest } from 'ember-qunit';
```
calcaide, line 2: None blocking comment and for learninng purposes: I was not aware we need to perform this changes, should we consider a refactor or eliminate the helper we were using before? [link to code](https://github.com/hashicorp/boundary-ui/blob/main/ui/admin/tests/helpers/index.js#L41-L45).

I would like to learn a bit more about this change, do you have some context, perhaps link to documentation, on this change? Thank you!!

ui/admin/tests/unit/controllers/scopes/scope/groups/group/index-test.js
```diff
@@ -0,0 +1,13 @@
+import { module, test } from 'qunit';
+import { setupTest } from 'ember-qunit';
+
+module('Unit | Controller | scopes/scope/groups/group/index', function (hooks) {
+  setupTest(hooks);
+
+  test('it exists', function (assert) {
```
calcaide, line 7: None blocking comment: I just realize that we do not have any real controller test in our codebase. We are just testing that the controller exists. I will bring this topic to tech time since I think is a big one and needs team members' opinion on it.

calcaide (approved):
Great work on this refactor, followed the test procedure successfully with no issues at all!

Thanks for this and keep the good work 💪

ui/admin/tests/unit/controllers/scopes/scope/auth-methods/auth-method/accounts/account/index-test.js
lisbet-alvarez: ahh good catch! i didn't realize there was a helper there, when i auto-generated this test file it automatically imported from `admin/tests/helpers` and i got the message "Could not find a declaration file for module" so I changed it to what i saw another controller file was importing from. Now that you point this out I am noticing that tests pass either way, so I'll reverse this change.

ui/admin/tests/unit/controllers/scopes/scope/groups/group/index-test.js
lisbet-alvarez: good point, and thank you!

cameronperera: Agreed that we should be testing our controllers, especially now that we are moving our actions from the routes to the controllers. We have some example [tests for controllers](https://github.com/hashicorp/boundary-ui/blob/main/ui/admin/tests/unit/controllers/scopes/scope/session-recordings/session-recording/channels-by-connection/index-test.js) but they seem to be mostly for getters which are a bit easier to test.

cameronperera (approved):

ui/admin/tests/unit/controllers/scopes/scope/groups/group/index-test.js
lisbet-alvarez: For actions added to the controller in this PR, I'll make tickets to add tests. For future work, ill add the tests in the same PR.

calcaide: @lisbet-alvarez Amazing!! yeah, creating tickets to bring tests later is ok. In this case we do not have patterns to follow so all good providing separate PR's for those so we can focus on them and bring discussions.

ui/admin/tests/unit/controllers/scopes/scope/auth-methods/auth-method/accounts/account/index-test.js
calcaide: @lisbet-alvarez this is interesting, should we bring this topic to tech time and discuss the need for `admin/test/helpers` so far, what I understand of such helper is that is a wrapper but aside from wrapping we are doing nothing custom, so maybe we should rely on the default one until we have the need for customize?

PR #2179, opened 2024-03-13
fix: 🐛 update destination path for autoUpdater workflow

cameronperera:
✅ Closes: https://hashicorp.atlassian.net/browse/ICU-12878

:tickets: [Jira ticket](https://hashicorp.atlassian.net/browse/ICU-12878)

## Description

Our update process in the desktop client was running into conflicts with our ASAR changes we made for the `1.7.0` release. The autoUpdater workflow was trying to store update specific data in a directory that no longer exist. We changed the destination for the update specific data and the workflow is no longer erroring out. We also moved the destination logic inside the try catch so we can catch the error if one is thrown.

## How to Test

To properly test this, you will need a signed build of the desktop client. AutoUpdater will not work without it. Please use a signed artifact from this [build](https://github.com/hashicorp/boundary-ui-releases/actions/runs/8255809808) to test.

## Checklist:

### DIFF
--- a/ui/desktop/electron-app/src/helpers/app-updater.js
+++ b/ui/desktop/electron-app/src/helpers/app-updater.js
@@ -62,10 +62,10 @@ const createAppUpdaterConfig = (url, version, destination) => {
 };
 
 const downloadAndInstallUpdate = async (version, url) => {
-  const destination = path.resolve(__dirname, '..', 'nextVersion');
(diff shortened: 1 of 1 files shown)


ZedLi (approved):

calcaide (approved):
Let's goo!!! Thanks for this fix!!

lisbet-alvarez (approved):
Awesome! 🥳

PR #2180, opened 2024-03-14
Refactor key value comp

DhariniJeeva:
This PR refactors the keyValue component by making it dynamic. Now, consumers can pass `key` and `value` to this component. This removes the need to create specific `key` and `value` input types every time when there's a new input design requirement.

https://github.com/hashicorp/boundary-ui/assets/15043878/d53255ee-5cd1-4b88-842f-71f7a4ee9bbe

## How to Test

## Checklist:


ui/admin/app/components/form/field/list-wrapper/key-value/index.js
```diff
@@ -29,6 +29,38 @@ export default class MappingListComponent extends Component {
...
+
+  /**
+   * Determines if we need to show an empty row to the users to enter more key/value pairs based on removeDuplicates arg,
+   * by default it is true
+   * @type {object}
+   */
+  get showNewRow() {
+    if (this.args.removeDuplicates) {
+      return (
+        (this.options || []).length !== (this.args.selectOptions || []).length
```
ZedLi, line 57:
```suggestion
        (this.options ?? []).length !== (this.args.selectOptions ?? []).length
```

Does `this.options?.length` !== `this.args.selectOptions?.length` not work?

ZedLi (commented):
Looks nice!

ui/admin/app/components/form/field/list-wrapper/key-value/index.js
DhariniJeeva: no that totally works, I'll simplify this

addons/api/addon/serializers/credential-library.js
```diff
@@ -36,23 +36,30 @@ export default class CredentialLibrarySerializer extends ApplicationSerializer {
...
-        if (credential_mapping_overrides[key]) {
-          obj[key] = credential_mapping_overrides[key];
-        } else {
-          obj[key] = null;
+      options.mapping_overrides[credential_type].forEach((item) => {
+        const exisiting = credential_mapping_overrides.find(
+          (obj) => obj.key === item,
+        );
+        if (!exisiting) {
+          credential_mapping_overrides.push({ key: item, value: null });
```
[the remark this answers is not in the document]
DhariniJeeva: The issue is with receiving the snapshot record data in array format (`credential_mapping_overrides`) before the `transformation` process. However, the JSON data must be in object format, so I need to do the transformation after setting the unused attributes to `null`. 

open to suggestions if there is a better approach to handle this

addons/api/addon/transforms/object-as-array.js
```diff
@@ -4,10 +4,11 @@
  */
 
 import Transform from '@ember-data/serializer/transform';
+import { TrackedObject } from 'tracked-built-ins';
 
 export default class ObjectAsArrayTransform extends Transform {
   deserialize(serialized) {
-    const obj = serialized || {};
+    const obj = new TrackedObject(serialized || {});
```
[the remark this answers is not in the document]
DhariniJeeva: hmm, you are right, it is not being used, I removed it.

addons/api/addon/serializers/credential-library.js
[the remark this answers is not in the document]
ZedLi: Out of curiosity, does leaving it just as an array not work? I thought what would happen was it would go through this serializer and then the transformer on its way out, is that not how it works?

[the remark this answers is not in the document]
DhariniJeeva: It seems like the `serializeAttribute` method did not undergo transformation once we assigned it to the JSON. I moved this entire processing to the vault specific `serialize` method, which we already had. Since we are updating the serialized attribute, we didn't have to perform any additional transformation. Let me know your thoughts on that.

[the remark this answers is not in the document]
ZedLi: Yeah that makes sense!

ui/admin/app/components/form/field/list-wrapper/key-value/index.hbs
```diff
@@ -3,39 +3,76 @@
...
+                  disabled=@disabled
+                  value=option.key
+                  ariaLabel=keyHeaderID
+                  setContext=(set-from-event option 'key')
+                )
+                select=(component
+                  'form/field/list-wrapper/key-value/select'
+                  disabled=@disabled
+                  value=option.key
+                  selectOptions=@selectOptions
```
ZedLi, line 44: Shouldn't `selectOptions` be passed directly to the component rather than from the key value component? If I'm reading this right we're passing in `selectOptions` from KeyValue which would mean we can't use two selects?

DhariniJeeva: Good point! With the current design, it is not possible to have two dropdowns. Passing directly to the component does not get yielded(maybe I'm doing something wrong here?). What do you think about adding separate key and value select options to the keyValue instead of having a single `selectoptions`?

ZedLi: Hmm I was able to get it to yield it but I had to remove `showNewRow` and `selectOptions` since they were using `selectOptions`, is that what you were seeing?

DhariniJeeva: yup, it works without those, but I can't access the consumer's nested component args in the `keyvalue` component. This means we won't be able to have unique rows which are necessary for the credential library.

ZedLi: I was thinking we could let the consumer handle that logic instead of baking it into the component, what do you think? That means we'd have to pass in the necessary logic for that one consumer

ui/admin/app/components/form/credential-library/vault-generic/index.hbs
```diff
@@ -169,14 +169,21 @@
...
 
       <:field as |F|>
-        <F.SelectText
+        <F.KeyValue
           @name='credential_mapping_overrides'
           @selectOptions={{this.mappingOverrides}}
           @model={{@model}}
           @disabled={{form.disabled}}
           @removeDuplicates={{true}}
           @width='400px'
```
calcaide, line 178: None blocking question: I know the value was set already, but I think this value should be `100%`.

Screenshot at `400px`:
Screenshot 2024-03-28 at 8 30 35 AM

Screenshot at `100%`:
Screenshot 2024-03-28 at 8 31 13 AM

addons/api/addon/transforms/account-value-map-array.js
```diff
@@ -9,13 +9,13 @@ export default class AccountValueMapArrayTransform extends Transform {
   deserialize(serialized) {
     const accountValues = serialized || [];
     return accountValues.map((accountValueString) => ({
-      from: accountValueString.split('=')[0],
-      to: accountValueString.split('=')[1],
+      key: accountValueString.split('=')[0],
+      value: accountValueString.split('=')[1],
```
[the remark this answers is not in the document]
calcaide: 100% agreed with this! Great call!

calcaide (approved):
I left a couple of coments but not blocking. AMAZING work with this component refactor and the screens that use it, thank you so much for this!! 🙌

ui/admin/app/components/form/field/list-wrapper/key-value/index.hbs
DhariniJeeva: The issue is we have two separate new and existing row blocks in the component and a single `selectOptions` arg from the consumer. If we were to handle this logic(filter out duplicates) in the consumer, the consumer wouldn't know if they are modifying the new or the existing row.

ui/admin/app/components/form/credential-library/vault-generic/index.hbs
DhariniJeeva: thanks for the catch, fixed

ui/admin/app/components/form/field/list-wrapper/key-value/index.hbs
DhariniJeeva: moved to consumer as discussed

ui/admin/app/components/form/field/list-wrapper/key-value/select.hbs
```diff
@@ -0,0 +1,19 @@
+<Hds::Form::Select::Base
+  @value={{@value}}
+  @width={{@width}}
+  disabled={{@disabled}}
+  aria-label={{t 'titles.value'}}
```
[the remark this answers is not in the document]
DhariniJeeva: Thanks for the catch, updated this!

ui/admin/app/components/form/field/list-wrapper/key-value/index.hbs
```diff
@@ -2,40 +2,78 @@
...
```
ZedLi, line 45: Should these `selectOptions` still be here? Shouldn't they be passed in directly now instead of from `KeyValue`?

ui/admin/app/components/form/auth-method/ldap/index.hbs
```diff
@@ -382,124 +382,46 @@
...
+      {{/if}}
+    </:fieldset>
+    <:field as |F|>
+      <F.KeyValue
+        @name='account_attribute_maps'
+        @options={{@model.account_attribute_maps}}
+        @model={{@model}}
+        @keyLabel={{t 'form.from_claim.label'}}
+        @valueLabel={{t 'form.to_claim.label'}}
+        @selectOptions={{this.toAttributes}}
```
ZedLi, line 413: Should this be on the `V.select` instead?

ui/admin/app/components/form/auth-method/oidc/index.hbs
```diff
@@ -247,138 +247,46 @@
...
-  </Rose::Form::Fieldset>
+    </:fieldset>
+    <:field as |F|>
+      <F.KeyValue
+        @name='account_claim_maps'
+        @options={{@model.account_claim_maps}}
+        @model={{@model}}
+        @keyLabel={{t 'form.from_claim.label'}}
+        @valueLabel={{t 'form.to_claim.label'}}
+        @selectOptions={{this.toClaims}}
```
ZedLi, line 278: Same here, shouldn't this be on the `V.select`?

ui/admin/app/components/form/field/list-wrapper/key-value/index.js
```diff
@@ -29,6 +29,20 @@ export default class MappingListComponent extends Component {
...
   }
+
+  /**
+   * Determines if we need to show an empty row to the users to enter more key/value pairs based on removeDuplicates arg,
+   * by default it is true
+   * @type {object}
+   */
+  get showNewRow() {
+    if (this.args.removeDuplicates) {
+      return this.options?.length !== this.args.allowedEntries;
```
ZedLi, line 40: Shouldn't we pass in the entire function `showNewRow` and let the user handle this logic rather than passing `removeDuplicates` and `allowedEntries`? We can check if `this.args.showNewRow` is a func and then call it if it is.

DhariniJeeva: good call, I moved it to consumer and passed the func to component

ui/admin/app/components/form/field/list-wrapper/key-value/index.hbs
```diff
@@ -2,40 +2,76 @@
...
-              aria-labelledby={{keyHeaderID}}
-              {{on 'input' (set-from-event option 'key')}}
-            />
+            {{yield
+              (hash
+                text=(component
+                  'form/field/list-wrapper/key-value/text'
+                  disabled=@disabled
+                  value=option.key
+                  ariaLabel=keyHeaderID
```
ZedLi, line 38: I don't think the `ariaLabel`'s can be this generated GUID, it should be human readable otherwise the screen reader would be reading a GUID out haha

Do we not have a guaranteed label anymore? Should it have still been the `aria-labelledby`?

DhariniJeeva: oops yes, the other ones have `aria-labelledby`

cameronperera (approved):

PR #2181, opened 2024-03-14
feat: 🎸 mirage support and initial route for aliases

DhariniJeeva:
Jira - https://hashicorp.atlassian.net/browse/ICU-12932, https://hashicorp.atlassian.net/browse/ICU-12934

## Description

This PR adds an initial list route and mirage support for alias resources. 

Untitled

## How to Test

- Using mirage, go to /scopes/global/aliases and look at the dev console to see the generated mirage data

## Checklist:

### DIFF
--- a/addons/api/addon/generated/models/alias.js
+++ b/addons/api/addon/generated/models/alias.js
@@ -50,6 +50,7 @@ export default class GeneratedAliasModel extends BaseModel {
   destination_id;
 
   @attr('object', {
+    isNestedAttribute: true,
     description:
       'This contains a host_id which is the id of the host that the session will be authorized for.',
   })
--- a/addons/api/addon/generated/models/alias.js
+++ b/addons/api/addon/generated/models/alias.js
@@ -1,0 +1,57 @@
+/**
+ * Copyright (c) HashiCorp, Inc.
+ * SPDX-License-Identifier: MPL-2.0
+ */
+
+import BaseModel from '../../models/base';
+import { attr } from '@ember-data/model';
+
+/**
+ * Alias contains all fields related to an Alias resource
+ */
+export default class GeneratedAliasModel extends BaseModel {
+  // =attributes
+
+  @attr('string', {
+    description: 'The type of the resource, to help differentiate schemas',
+  })
+  type;
+
+  @attr('string', {
+    description: 'Optional name for identification purposes.',
+  })
+  name;
+
+  @attr('string', {
+    description: 'Optional user-set description for identification purposes.',
+  })
+  description;
+
+  @attr('date', {
+    description: 'The time this resource was created\nOutput only.',
+    readOnly: true,
+  })
+  created_time;
+
+  @attr('date', {
+    description: 'The time this resource was last updated\nOutput only.',
+    readOnly: true,
+  })
+  updated_time;
+
+  @attr('number', {
+    description: 'Current version number of this resource.',
+  })
+  version;
+
+  @attr('string', {
+    description: 'This is the id of the resource that this Alias points to.',
+  })
+  destination_id;
+
+  @attr('object', {
+    description:
+      'This contains a host_id which is the id of the host that the session will be authorized for.',
+  })
+  authorize_session_arguments;
+}
--- a/addons/api/addon/models/alias.js
+++ b/addons/api/addon/models/alias.js
@@ -5,5 +5,6 @@
 
 import GeneratedAliasModel from '../generated/models/alias';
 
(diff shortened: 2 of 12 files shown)


addons/api/addon/generated/models/alias.js
```diff
@@ -0,0 +1,57 @@
...
+  @attr('string', {
+    description: 'This is the id of the resource that this Alias points to.',
+  })
+  destination_id;
+
+  @attr('object', {
+    description:
+      'This contains a host_id which is the id of the host that the session will be authorized for.',
+  })
+  authorize_session_arguments;
```
lisbet-alvarez, line 56: i think this may be missing `isNestedAttribute: true`

addons/api/addon/models/alias.js
```diff
@@ -0,0 +1,9 @@
+/**
+ * Copyright (c) HashiCorp, Inc.
+ * SPDX-License-Identifier: MPL-2.0
+ */
+
+import GeneratedAliasModel from '../generated/models/alias';
+
+export const TYPE_ALIAS = 'target';
```
lisbet-alvarez, line 8: non-blocking: from my understanding,  think we are only going to support aliases for targets for now. Since in the future we may want to support more types for many resources we should also add `export const TYPES_ALIAS = Object.freeze([...]);` . And perhaps rename `TYPE_ALIAS` to `TYPE_ALIAS_TARGET`

addons/api/mirage/generated/factories/alias.js
```diff
@@ -0,0 +1,20 @@
...
+
+import { Factory } from 'miragejs';
+import { faker } from '@faker-js/faker';
+
+/**
+ * GeneratedAliasModelFactory
+ */
+export default Factory.extend({
+  name: () => faker.word.words(),
+  type: 'target',
```
lisbet-alvarez, line 14: small suggestion: for `type` perhaps we should import it from the model

addons/api/addon/generated/models/alias.js
DhariniJeeva: ah good catch, will add this

addons/api/addon/models/alias.js
DhariniJeeva: yeah that makes sense

addons/api/mirage/factories/scope.js
```diff
@@ -36,6 +36,7 @@ export default factory.extend({
     // Worker permissions only available on the global scope
     if (this.type === 'global') {
       collectionActions.workers = ['create:worker-led', 'list'];
+      collectionActions.aliases = ['create', 'list'];
```
ZedLi, line 39: Is it global only? Also the comment should change if we add this here

DhariniJeeva: yeah it is global only for now, in the future we might extend it to other scopes

lisbet-alvarez (approved):
🥳 LGTM 🙌

addons/api/mirage/factories/scope.js
DhariniJeeva: added

addons/api/mirage/factories/alias.js
```diff
@@ -0,0 +1,20 @@
...
+ * Copyright (c) HashiCorp, Inc.
+ * SPDX-License-Identifier: MPL-2.0
+ */
+
+import factory from '../generated/factories/alias';
+import permissions from '../helpers/permissions';
+import generateId from '../helpers/id';
+
+export default factory.extend({
+  id: () => generateId('alt_'),
```
cameronperera, line 11: nonblocking question for learning: Is the `alt_` prefix specific to the target type similar to `ttcp_` or `tssh_` for TCP and SSH targets?

DhariniJeeva: Yes, it is a prefix for alias IDs that the backend uses.

addons/api/mirage/generated/factories/alias.js
```diff
@@ -0,0 +1,21 @@
...
+ */
+
+import { Factory } from 'miragejs';
+import { faker } from '@faker-js/faker';
+import { TYPE_ALIAS_TARGET } from 'api/models/alias';
+
+/**
+ * GeneratedAliasModelFactory
+ */
+export default Factory.extend({
```
calcaide, line 13: I think we are missing `destination_id` field?

DhariniJeeva: added!

PR #2184, opened 2024-03-14
test(e2e): Add test for worker

moduli:
This PR attempts to add an Admin UI e2e test for workers.

### DIFF
--- a/ui/admin/tests/e2e/tests/worker.spec.js
+++ b/ui/admin/tests/e2e/tests/worker.spec.js
@@ -1,0 +1,68 @@
+/**
+ * Copyright (c) HashiCorp, Inc.
+ * SPDX-License-Identifier: MPL-2.0
+ */
+
+/* eslint-disable no-undef */
+const { expect } = require('@playwright/test');
+const { test } = require('@playwright/test');
+const { checkEnv, authenticatedState } = require('../helpers/general');
+const { checkBoundaryCli } = require('../helpers/boundary-cli');
+
+test.use({ storageState: authenticatedState });
+
+test.beforeAll(async () => {
+  await checkEnv(['E2E_WORKER_TOKEN']);
+
+  await checkBoundaryCli();
+});
+
+test('Create a worker @ce @ent @docker', async ({ page }) => {
+  await page.goto('/');
+  await page
+    .getByRole('navigation', { name: 'General' })
+    .getByRole('link', { name: 'Workers' })
+    .click();
+  await page.getByRole('link', { name: 'New', exact: true }).click();
+
+  // Populate config fields
+  await page.getByLabel('Worker public address').fill('worker1.example.com');
+  await page.getByLabel('Config file path').fill('/home/ubuntu/boundary');
+  await page.getByLabel('Initial upstreams').fill('10.0.0.1, 10.0.0.2');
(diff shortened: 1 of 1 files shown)


ui/admin/tests/e2e/tests/worker.spec.js
```diff
@@ -0,0 +1,68 @@
...
+
+  const text = await page.getByRole('alert').innerText();
+  if (text.includes('Success')) {
+    await page.getByRole('button', { name: 'Dismiss' }).click();
+    await page.getByRole('button', { name: 'Done' }).click();
+  } else if (text.includes('Error')) {
+    await page.getByRole('button', { name: 'Dismiss' }).click();
+  } else {
+    throw new Error('Unexpected error');
+  }
```
moduli, line 67: I'm not sure how I feel about this. With the way things are set up, the worker token that we pass to the test suite is only available for one use. Subsequent attempts to run this test (say, when this gets run on multiple browser types) will result in an error when trying to register that same token.

I'm not sure what direction to go in, though... any other ideas?

calcaide: Be aware I am missing quite a bit of context here, but for the sake of throwing ideas to the air:

By token we pass to the test, we mean `E2E_WORKER_TOKEN`is that correct?

How do we get that token in first place? Do we run a command on the CLI? if we get it through the CLI, perhaps we can create a boundary-cli helper that retrieve it and we call such method at the beginning of each test run?
Could be this an option or I am missing too much context?

moduli: Good questions.

To get this token, we need to start a worker instance and extract the token from its logs. With the way things are structured right now, this occurs once during the infra setup stage and gets passed to the test suite via `E2E_WORKER_TOKEN`, like you noted.

I was wondering if we could add some helper methods to start another worker instance within this test suite, but I'm a little concerned about adding a bunch of stuff within this test suite to support that. I might experiment to see what it looks like.

To do this, we'd have to execute this in the test
```
# We'd need to provide a worker-config.hcl file and run this command
> boundary server -config worker-config.hcl

# We would need to extract the worker token from the logs
# We can pass that token to worker form to register

# For cleanup, we'll need to end the `boundary server` command

```

calcaide: Gotcha, I see your concerns about adding a bunch of stuff for supporting things approach. Seems like we are going to cross the line of SOC if we go that path.

Another possible idea... If we run this test, and at the end, after adding the worker successfully, we delete such worker from Admin UI, would the deletion allow us to re-use the Token to create a new one? If that is possible, I think it will make more sense to add the step of deletion at the end of the creation test, so we can repeat the operation.

This will bring the side effect of not running this specific test in parallel tho.

And thanks for the explanation 😃

moduli: Good idea, but I believe it's a one-time use token.

We aren't planning on making these tests run in parallel as some tests might affect state that affects other tests.

calcaide (approved):
Thanks for adding this test!!

moduli:
Closing this for now. Will want to spend some time to think about the approach.

PR #2185, opened 2024-03-15
chore: Add routes for aliases

DhariniJeeva:
:tickets: [Jira ticket](https://hashicorp.atlassian.net/browse/ICU-12936)

## Description

This PR adds new routes for aliases, also controllers to support new route action pattern

alias

## How to Test

## Checklist:

### DIFF
--- a/ui/admin/app/routes/scopes/scope/aliases/alias.js
+++ b/ui/admin/app/routes/scopes/scope/aliases/alias.js
@@ -1,0 +1,26 @@
+/**
+ * Copyright (c) HashiCorp, Inc.
+ * SPDX-License-Identifier: MPL-2.0
+ */
+
+import Route from '@ember/routing/route';
+import { inject as service } from '@ember/service';
+
+export default class ScopesScopeAliasesAliasRoute extends Route {
+  @service store;
+  @service can;
+
+  // =methods
+
+  /**
+   * Load an alias in current scope.
+   * @param {object} params
+   * @param {string} params.alias_id
+   * @return {Promise{AliasModel}}
+   */
+  async model({ alias_id }) {
+    return this.store.findRecord('alias', alias_id, {
+      reload: true,
+    });
+  }
+}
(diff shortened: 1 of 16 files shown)


ui/admin/app/routes/scopes/scope/aliases/alias.js
```diff
@@ -0,0 +1,26 @@
...
+ * Copyright (c) HashiCorp, Inc.
+ * SPDX-License-Identifier: MPL-2.0
+ */
+
+import Route from '@ember/routing/route';
+import { inject as service } from '@ember/service';
+
+export default class ScopesScopeAliasesAliasRoute extends Route {
+  @service store;
+  @service can;
```
calcaide, line 11: None blocking comment: Despite we are not using right now the service `can`, I do understand we add it because later we will add permissions and therefore service `can` is necessary?

DhariniJeeva: Actually, the 'can' may not be necessary at all. I'm looking at other similar routes where we didn't need it either. I'm surprised the linter did not catch this, but it will be needed in the '/new' routes. I'll remove this, ty for catching this

ZedLi (approved):

ui/admin/app/routes/scopes/scope/aliases/alias.js
calcaide: You are right, the linter should help us with this.

Since this is an specific ember thing the "default" eslint will not catch it as something we declared and we did not use. But looking at the `eslint-plugin-ember` I found this rule [no-unused-services](https://github.com/ember-cli/eslint-plugin-ember/blob/master/docs/rules/no-unused-services.md) it is not part of the default rules so if we want it we have to activate it. I will bring this topic to tech time and we can discuss as a team 😉

DhariniJeeva: yup, that's a good idea! removed it

calcaide (approved):
Awesome work!! 💪 thanks for this

cameronperera (approved):

PR #2186, opened 2024-03-15
refact(e2e): Move env outside of helper methods

moduli:
This PR does some refactoring in the Admin UI e2e test suite to move any direct use of environment variables outside of helper methods to make it consistent and explicit where they are used within tests.


stasryzhov (approved):
🚀

PR #2194, opened 2024-03-20
Icu 12587 add filtering pagination support for sessions

cameronperera:
:tickets: [Jira ticket](https://hashicorp.atlassian.net/browse/ICU-12587)

Also closes Jira ticket for route-action migration for sessions route in [Admin UI](https://hashicorp.atlassian.net/browse/ICU-12609)

Initial screen:
Screenshot 2024-03-21 at 10 37 36 AM
Searching by session Id:
Screenshot 2024-03-21 at 10 38 31 AM
Filtering by target:
Screenshot 2024-03-21 at 10 38 42 AM

## How to Test

Start a local `boundary dev` instance and start a few sessions. Navigate to the sessions tab in the project view. You should be able to search and filter the sessions along with paginate.

:technologist: [Admin preview](https://boundary-cbw9ykpbq-hashicorp.vercel.app/scopes/s_1sheljmjgm/sessions)

## Checklist:

### DIFF
--- a/addons/api/addon/utils/indexed-db-query.js
+++ b/addons/api/addon/utils/indexed-db-query.js
@@ -225,8 +225,7 @@ const getKey = (key) => {
 
 const cleanValues = (values) => {
   return values.reduce((result, obj) => {
-    // eslint-disable-next-line no-unused-vars
-    const [key, value] = Object.entries(obj)[0];
+    const [, value] = Object.entries(obj)[0];
     if (value) result.push(obj);
     return result;
   }, []);
--- a/addons/api/addon/utils/indexed-db-query.js
+++ b/addons/api/addon/utils/indexed-db-query.js
@@ -223,14 +223,6 @@ const getKey = (key) => {
   }
 };
 
-const cleanValues = (values) => {
-  return values.reduce((result, obj) => {
-    const [, value] = Object.entries(obj)[0];
-    if (value) result.push(obj);
-    return result;
-  }, []);
-};
-
 /**
  * Cleans the object to remove any keys that have a value of an empty array. This makes
  * it easier so we don't have to worry about empty arrays getting queried in indexed db.
@@ -243,10 +235,9 @@ const cleanObject = (obj) => {
   return Object.entries(obj).reduce((result, [key, value]) => {
     if (Array.isArray(value)) {
       if (value.length > 0) {
-        result[key] = cleanValues(value);
+        result[key] = value;
       }
     } else if (value.values.length > 0) {
-      value.values = cleanValues(value.values);
       result[key] = value;
     }
 
--- a/addons/api/addon/utils/indexed-db-query.js
+++ b/addons/api/addon/utils/indexed-db-query.js
@@ -223,6 +223,15 @@ const getKey = (key) => {
   }
 };
 
+const cleanValues = (values) => {
+  return values.reduce((result, obj) => {
+    // eslint-disable-next-line no-unused-vars
+    const [key, value] = Object.entries(obj)[0];
+    if (value) result.push(obj);
+    return result;
+  }, []);
+};
+
 /**
  * Cleans the object to remove any keys that have a value of an empty array. This makes
  * it easier so we don't have to worry about empty arrays getting queried in indexed db.
@@ -235,9 +244,10 @@ const cleanObject = (obj) => {
   return Object.entries(obj).reduce((result, [key, value]) => {
     if (Array.isArray(value)) {
       if (value.length > 0) {
-        result[key] = value;
+        result[key] = cleanValues(value);
       }
     } else if (value.values.length > 0) {
+      value.values = cleanValues(value.values);
       result[key] = value;
     }
 
--- a/addons/rose/addon/styles/hds/overrides.scss
+++ b/addons/rose/addon/styles/hds/overrides.scss
@@ -31,6 +31,13 @@
     color: var(--token-form-indicator-optional-color);
   }
 
+  // Override for sortable HDS table columns that ignore selected theme in app
+  // and use the system default theme causing color mismatch if selected theme
+  // is not the system default theme
+  .hds-table__th-sort button {
+    color: var(--token-color-foreground-strong);
+  }
+
   // TODO: This is a temporary hack to fix an issue causing weird whitespace
   //   and scrollbars from an accessible class in HDS. We have to use !important
   //   as they already forced it with !important and we can't override it otherwise
--- a/ui/admin/app/controllers/scopes/scope/sessions/index.js
+++ b/ui/admin/app/controllers/scopes/scope/sessions/index.js
@@ -9,6 +9,7 @@ import {
 } from 'api/models/session';
 import { action } from '@ember/object';
 import { debounce } from 'core/decorators/debounce';
+import { notifySuccess, notifyError } from 'core/decorators/notify';
 
 export default class ScopesScopeSessionsIndexController extends Controller {
   // =services
@@ -93,8 +94,22 @@ export default class ScopesScopeSessionsIndexController extends Controller {
     this.page = 1;
   }
 
+  /**
+   * Refreshes the all data for the current page
+   */
   @action
-  async refresh() {
-    await this.send('refreshAll');
+  refresh() {
+    this.send('refreshAll');
+  }
+
+  /**
+   * Cancels the specified session and notifies the user of success or error
+   * @param {SessionModel} session
+   */
+  @action
+  @notifyError(({ message }) => message, { catch: true })
+  @notifySuccess('notifications.canceled-success')
+  cancelSession(session) {
+    session.cancelSession();
   }
 }
--- a/ui/admin/app/controllers/scopes/scope/sessions/index.js
+++ b/ui/admin/app/controllers/scopes/scope/sessions/index.js
@@ -38,6 +38,10 @@ export default class ScopesScopeSessionsIndexController extends Controller {
   @tracked page = 1;
   @tracked pageSize = 10;
 
+  /**
+   * Returns object of filters to be used for displaying selected filters
+   * @returns {object}
+   */
   get filters() {
     return {
       allFilters: {
--- a/ui/admin/app/controllers/scopes/scope/sessions/index.js
+++ b/ui/admin/app/controllers/scopes/scope/sessions/index.js
@@ -1,0 +1,91 @@
+import Controller from '@ember/controller';
+import { inject as service } from '@ember/service';
+import { tracked } from '@glimmer/tracking';
+import {
+  STATUS_SESSION_ACTIVE,
+  STATUS_SESSION_PENDING,
+  STATUS_SESSION_CANCELING,
+  statusTypes,
+} from 'api/models/session';
+import { action } from '@ember/object';
+import { debounce } from 'core/decorators/debounce';
+
+export default class ScopesScopeSessionsIndexController extends Controller {
+  // =services
+
+  @service store;
+  @service intl;
+
+  // =attributes
+
+  queryParams = [
+    'search',
+    { users: { type: 'array' } },
+    { targets: { type: 'array' } },
+    { status: { type: 'array' } },
+    'page',
+    'pageSize',
+  ];
+
+  @tracked search = '';
+  @tracked users = [];
+  @tracked targets = [];
+  @tracked status = [
+    STATUS_SESSION_ACTIVE,
+    STATUS_SESSION_PENDING,
+    STATUS_SESSION_CANCELING,
+  ];
+  @tracked page = 1;
+  @tracked pageSize = 10;
+
+  get filters() {
+    return {
+      allFilters: {
+        users: this.model.allUsers,
+        targets: this.model.allTargets,
+        status: this.sessionStatusOptions,
+      },
+      selectedFilters: {
+        users: this.users,
+        targets: this.targets,
+        status: this.status,
+      },
+    };
+  }
+
+  /**
+   * Returns all status types for sessions
+   * @returns {[object]}
+   */
+  get sessionStatusOptions() {
+    return statusTypes.map((status) => ({
+      id: status,
+      name: this.intl.t(`resources.session.status.${status}`),
+    }));
+  }
+
+  // =actions
+
+  /**
+   * Handles input on each keystroke and the search queryParam
+   * @param {object} event
+   */
+  @action
+  @debounce(250)
+  handleSearchInput(event) {
+    const { value } = event.target;
+    this.search = value;
+    this.page = 1;
+  }
+
+  /**
+   * Sets the scopes query param to value of selectedScopes
+   * to trigger a query and closes the dropdown
+   * @param {function} close
+   */
+  @action
+  applyFilter(paramKey, selectedItems) {
+    this[paramKey] = [...selectedItems];
+    this.page = 1;
+  }
+}
--- a/ui/admin/app/controllers/scopes/scope/sessions/index.js
+++ b/ui/admin/app/controllers/scopes/scope/sessions/index.js
@@ -109,7 +109,7 @@ export default class ScopesScopeSessionsIndexController extends Controller {
   @action
   @notifyError(({ message }) => message, { catch: true })
   @notifySuccess('notifications.canceled-success')
-  cancelSession(session) {
-    session.cancelSession();
+  async cancelSession(session) {
+    await session.cancelSession();
   }
 }
--- a/ui/admin/app/controllers/scopes/scope/sessions/index.js
+++ b/ui/admin/app/controllers/scopes/scope/sessions/index.js
@@ -92,4 +92,9 @@ export default class ScopesScopeSessionsIndexController extends Controller {
     this[paramKey] = [...selectedItems];
     this.page = 1;
   }
+
+  @action
+  async refresh() {
+    await this.send('refreshAll');
+  }
 }
--- a/ui/admin/app/controllers/scopes/scope/sessions/index.js
+++ b/ui/admin/app/controllers/scopes/scope/sessions/index.js
@@ -83,9 +83,9 @@ export default class ScopesScopeSessionsIndexController extends Controller {
   }
 
   /**
-   * Sets the scopes query param to value of selectedScopes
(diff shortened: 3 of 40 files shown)


addons/api/addon/utils/indexed-db-query.js
```diff
@@ -223,6 +223,15 @@ const getKey = (key) => {
   }
 };
 
+const cleanValues = (values) => {
+  return values.reduce((result, obj) => {
+    // eslint-disable-next-line no-unused-vars
+    const [key, value] = Object.entries(obj)[0];
```
lisbet-alvarez, line 229: we can use an underscore to replace `key` so that we don't have to disable lint rule
```suggestion
    const [_, value] = Object.entries(obj)[0];
```

ui/admin/app/controllers/scopes/scope/sessions/index.js
```diff
@@ -0,0 +1,95 @@
...
+  handleSearchInput(event) {
+    const { value } = event.target;
+    this.search = value;
+    this.page = 1;
+  }
+
+  /**
+   * Sets the scopes query param to value of selectedScopes
+   * to trigger a query and closes the dropdown
+   * @param {function} close
```
lisbet-alvarez, line 88: im realizing the comment for this action needs to be updated to have two of `@param`.

ui/admin/app/routes/scopes/scope/sessions/index.js
```diff
@@ -4,5 +4,200 @@
...
+    },
+  };
+
+  allUsers;
+  allSessions;
+
+  @runEvery(POLL_TIMEOUT_SECONDS * 1000)
+  poller() {
+    return this.refresh();
+  }
```
lisbet-alvarez, line 54: nice that this works even with the new changes!!

ui/admin/app/controllers/scopes/scope/sessions/index.js
cameronperera: Oh yeah, looks like I never updated it from when I first created the action and it only handled one queryParam

ui/admin/app/routes/scopes/scope/sessions/index.js
```diff
@@ -4,5 +4,200 @@
...
+    sessions.forEach(async (session) => {
+      if (session.user_id) {
+        session.user = await this.store.findRecord('user', session.user_id);
+      }
+      if (session.target_id) {
+        await this.store.findRecord('target', session.target_id);
+      }
+    });
+
+    // Query all sessions, users, and targets for defining filtering values if entering route for the first time
```
lisbet-alvarez, line 99: I think this causes issues with displaying filters because even after polling refreshes the page, not all targets are displayed in the dropdown. The route is not being entered for the first time, so I'm assuming this gets skipped. 

Another thing, I think that retrieving allSessions might not be necessary since it's not needed for the filters.

addons/api/addon/utils/indexed-db-query.js
```diff
@@ -223,6 +223,14 @@ const getKey = (key) => {
   }
 };
 
+const cleanValues = (values) => {
```
DhariniJeeva, line 226: Can you add a comment explaining why we need this?

cameronperera: Yes. We were getting `null` values from the model (related to the peekRecord and findRecord issue). So the indexed-db-query was getting input that looked like this: `{ equals: null }`. So this `cleanValues` method will remove those `null` or `undefined` values if passed in.

ui/admin/app/routes/scopes/scope/sessions/index.js
```diff
@@ -4,5 +4,196 @@
...
+      query: { search, filters },
+      page,
+      pageSize,
+      force_refresh: true,
+    };
+
+    const sessions = await this.store.query('session', queryOptions);
+    const totalItems = sessions.meta?.totalItems;
+
+    sessions.forEach(async (session) => {
```
lisbet-alvarez, line 81: perhaps we should replace `forEach` due to it [not awaiting](https://gist.github.com/joeytwiddle/37d2085425c049629b80956d3c618971) for promises to resolve.

addons/api/addon/utils/indexed-db-query.js
ZedLi: I don't understand, null is a valid value isn't it? Why are we cleaning it here?

addons/api/addon/utils/indexed-db-query.js
```diff
@@ -223,6 +223,14 @@ const getKey = (key) => {
...
+  return values.reduce((result, obj) => {
```
ZedLi, line 227: Isn't `values` an array, why are we reducing it instead of filtering?

addons/api/addon/utils/indexed-db-query.js
```diff
@@ -223,6 +223,14 @@ const getKey = (key) => {
...
+    const [, value] = Object.entries(obj)[0];
```
ZedLi, line 228: Why use `Object.entries` and then throw away the key value?

addons/api/addon/utils/indexed-db-query.js
```diff
@@ -223,6 +223,14 @@ const getKey = (key) => {
...
+    if (value) result.push(obj);
```
ZedLi, line 229: What if value was `0`?

ui/admin/app/templates/scopes/scope/sessions/index.hbs
```diff
@@ -17,134 +17,143 @@
...
+              </B.Td>
+              <B.Td>
+                <SessionStatus @model={{B.data}} />
+              </B.Td>
+              <B.Td>
+                {{#if (can 'cancel session' B.data)}}
+                  <Hds::Button
+                    @color='secondary'
+                    @text={{t 'actions.cancel'}}
+                    {{on 'click' (route-action 'cancelSession' B.data)}}
```
ZedLi, line 119: Are we still using `route-action`s?

ui/admin/app/routes/scopes/scope/sessions/index.js
```diff
@@ -4,5 +4,196 @@
...
+      filters.status.push({ equals: item });
+    });
+
+    const queryOptions = {
+      scope_id,
+      recursive: true,
+      query: { search, filters },
+      page,
+      pageSize,
+      force_refresh: true,
```
ZedLi, line 75: What's this `force_refresh` for?

ui/admin/app/routes/scopes/scope/sessions/index.js
```diff
@@ -4,5 +4,196 @@
...
+    const sessions = await this.store.query('session', queryOptions);
+    const totalItems = sessions.meta?.totalItems;
+
+    sessions.forEach(async (session) => {
+      if (session.user_id) {
+        session.user = await this.store.findRecord('user', session.user_id);
+      }
+      if (session.target_id) {
+        await this.store.findRecord('target', session.target_id);
+      }
```
ZedLi, line 87: Why do we need to `findRecord` when we've already queried for all the users and targets? This seems like unnecessary extra API requests when we could just load what we need into the store?

ui/admin/app/routes/scopes/scope/sessions/index.js
```diff
@@ -4,5 +4,196 @@
...
+  /**
+   * Get all the targets but only load them once when entering the route.
+   * @param scope_id
+   * @returns {Promise<void>}
+   */
+  async getAllTargets(scope_id) {
+    const uniqueSessionTargetIds = new Set(
+      this.allSessions.map((session) => session.target_id),
+    );
+    const filters = { id: { logicalOperator: 'or', values: [] } };
```
ZedLi, line 158: Isn't the default `or`?

ui/admin/app/routes/scopes/scope/sessions/index.js
cameronperera: Sorry, copy pasta. Removed.

addons/api/addon/utils/indexed-db-query.js
cameronperera: While manually testing we were getting a `null` value from `session.target_id`. Happy to revisit this after I address the other feedback.

ui/admin/app/routes/scopes/scope/sessions/index.js
cameronperera: I actually removed the whole thing based on feedback from another PR comment.

ui/admin/app/routes/scopes/scope/sessions/index.js
cameronperera: refactored to not need this anymore.

ui/admin/app/routes/scopes/scope/sessions/index.js
cameronperera: Removed.

ui/admin/app/templates/scopes/scope/sessions/index.hbs
cameronperera: I was trying to keep the scope of this PR limited to just the pagination and filtering but I can update this now if you'd prefer.

cameronperera: `cancelSession` is an easy one so I went ahead and moved it from the router to the controller.

addons/api/addon/utils/indexed-db-query.js
cameronperera: After discussion, we are removing this for now and going to address this in a later PR

ui/admin/app/routes/scopes/scope/sessions/index.js
```diff
@@ -4,5 +4,167 @@
...
 import Route from '@ember/routing/route';
+import { inject as service } from '@ember/service';
+import { action } from '@ember/object';
 
-export default class ScopesScopeSessionsIndexRoute extends Route {}
+export default class ScopesScopeSessionsIndexRoute extends Route {
+  // =services
+
+  @service store;
+  @service resourceFilterStore;
```
lisbet-alvarez, line 14: this service can be removed

ui/admin/app/routes/scopes/scope/sessions/index.js
```diff
@@ -4,5 +4,167 @@
...
+      refreshModel: true,
+    },
+    pageSize: {
+      refreshModel: true,
+    },
+  };
+
+  allUsers;
+  allSessions;
+
```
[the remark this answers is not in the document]
cameronperera: Thanks. I know I added that at one point so unsure how it disappeared

lisbet-alvarez (approved):
🙌

ui/admin/app/routes/scopes/scope/sessions/index.js
```diff
@@ -4,5 +4,167 @@
...
+    const filters = { id: { values: [] } };
+    uniqueSessionUserIds.forEach((userId) => {
+      filters.id.values.push({ equals: userId });
+    });
+    const allUsersQuery = {
+      scope_id: 'global',
+      recursive: true,
+      query: { filters },
+    };
+    this.allUsers = await this.store.query('user', allUsersQuery);
```
ZedLi, line 131: Did we stress test this? I wasn't suggesting we push all of these resources to the store, just that we have it in indexeddb so we should be taking what we need out of it and pushing it to the store

ui/admin/app/routes/scopes/scope/sessions/index.js
```diff
@@ -4,5 +4,171 @@
...
+
+  /**
+   * Get all the targets but only load them once when entering the route.
+   * @param scope_id
+   * @returns {Promise<void>}
+   */
+  async getAllTargets(scope_id) {
+    const uniqueSessionTargetIds = new Set(
+      this.allSessions
+        .filter((session) => session.user_id)
```
ZedLi, line 144:
```suggestion
        .filter((session) => session.target_id)
```

ui/admin/app/templates/scopes/scope/sessions/index.hbs
```diff
@@ -17,134 +17,143 @@
...
+          <A.Header @title={{t 'titles.no-results-found'}} />
+          <A.Body
+            @text={{t
+              (if
+                this.search
+                'descriptions.no-search-results'
+                'descriptions.no-filter-results'
+              )
+              query=this.search
+              resource='sessions'
```
ZedLi, line 142: Should be translated string here

ui/admin/app/routes/scopes/scope/sessions/index.js
```diff
@@ -4,5 +4,171 @@
...
+    this.allSessions = await this.store.query('session', allSessionsQuery, {
+      pushToStore: false,
+    });
+  }
+
+  /**
+   * Get all the users but only load them once when entering the route.
+   * @returns {Promise<void>}
+   */
+  async getAllUsers() {
```
ZedLi, line 118: I would make these names a bit more descriptive since I do see that they're only pulling in users and targets that are on a session, I thought these were originally pulling in _all_ users and targets based on the function names

ZedLi (approved):
Looks great!!

ui/admin/app/routes/scopes/scope/sessions/index.js
cameronperera: Good catch. Fixed

ui/admin/app/routes/scopes/scope/sessions/index.js
cameronperera: Great point. I will update now.

addons/rose/addon/styles/hds/overrides.scss
```diff
@@ -31,6 +31,13 @@
     color: var(--token-form-indicator-optional-color);
   }
 
+  // Override for sortable HDS table columns that ignore selected theme in app
+  // and use the system default theme causing color mismatch if selected theme
+  // is not the system default theme
+  .hds-table__th-sort button {
```
calcaide, line 37: None blocking comment: Did we ask/mention this to design system team (HDS) ?

calcaide:
I would suggest to add to the test procedure, what fields are we suppose to perform search?

Performing manual test:
- When using the search bar, I could just search by ID and status. i.e: `s_z9aN3Am2Wt` will show me just the sessions with this id.
- When using filters, with and without\ search, behaviour was as expected.

calcaide:
None blocking comment:
The initial state of the column status shows an ascendant arrow, I understand this is because is sorting content ascendant? isn't the expected behaviour to have NO order applied at first load until user wants to change the sorting order? in other words, it is expected to show the arrow with the sorting direction even if the user did not click to sort?
(attached screenshot for better understanding)

Screenshot 2024-04-02 at 10 54 45 AM

ui/admin/app/routes/scopes/scope/sessions/index.js
cameronperera: Updated.

ui/admin/app/templates/scopes/scope/sessions/index.hbs
cameronperera: Fixed.

addons/rose/addon/styles/hds/overrides.scss
cameronperera: Not yet. We are a bit behind on our HDS version so we need to update first before bringing this to the team.

calcaide: @cameronperera good call 😉

ui/admin/app/templates/scopes/scope/sessions/index.hbs
```diff
@@ -17,134 +17,143 @@
...
+          <S.TextInput
+            @value={{this.search}}
+            @type='search'
+            placeholder={{t 'actions.search'}}
+            aria-label={{t 'actions.search'}}
+            {{on 'input' this.handleSearchInput}}
+          />
+          <S.Generic>
+            <Dropdown
+              @name={{t 'resources.user.title'}}
```
calcaide, line 33: none blocking comment: should we use translated string for `@name`? have in mind i18n strings will change the value and are Uppercase, depending of the usage you want for this attribute I will consider hardcoding the value.

calcaide (approved):
Left a none blocking comment, great job!! 😉

ui/admin/app/routes/scopes/scope/sessions/index.js
```diff
@@ -4,5 +4,174 @@
...
+  associatedTargets;
+
+  /**
+   * Loads all sessions under the current scope and encapsulates them into
+   * an array of objects containing their associated users and targets.
+   * @return {Promise{[{sessions: [SessionModel], allSessions: [SessionModel], associatedUsers: [UserModel], associatedTargets: [TargetModel], totalItems: number}]}}
+   */
+  async model({ search, users, targets, status, page, pageSize }) {
+    const { id: scope_id } = this.modelFor('scopes.scope');
+    const filters = {
```
ZedLi, line 53: Shouldn't the scope id filter be added here?

cameronperera: Yes they should. Thanks for catching that.

ui/admin/app/routes/scopes/scope/sessions/index.js
```diff
@@ -4,5 +4,172 @@
...
+    targets.forEach((target) => {
+      filters.target_id.push({ equals: target });
+    });
+    status.forEach((item) => {
+      filters.status.push({ equals: item });
+    });
+
+    const queryOptions = {
+      scope_id,
+      recursive: true,
```
ZedLi, line 71: I think we should also remove these scope id and recursive parameters from the store queries in this file since they're not doing anything right now. We can revisit not having it hard coded later

cameronperera: Yep. I noticed that after the change in the groups work. Removing now.

ui/admin/app/templates/scopes/scope/sessions/index.hbs
cameronperera: That is a good point. We are using translations for this in a few places. I will add this as a tech time topic to discuss.

PR #2197, opened 2024-03-22
chore: 🤖 aliases list view

DhariniJeeva:
:tickets: [Jira ticket](https://hashicorp.atlassian.net/browse/ICU-12925) and https://hashicorp.atlassian.net/browse/ICU-12927

## Description

This PR covers list view, edit, and delete functionality(note: delete doesn't have the modal popup yet, which means we can't clear the aliases yet)

Untitled

## How to Test

- get the latest main from BE, run make install and start the controller
- create an alias by providing `target id host id, and a value`
- After successfully creation, you should now see the alias on the list view
- create another alias, but this time, don't provide the optional fields- `target id` and `hostid`
- After successfully creation, you should now see the alias on the list view, and the target column would show as `not set`

## Checklist:

### DIFF
--- a/addons/api/mirage/scenarios/default.js
+++ b/addons/api/mirage/scenarios/default.js
@@ -125,12 +125,9 @@ export default function (server) {
   });
 
   // Aliases
-  const aliases = server.createList('alias', 3, { scope: globalScope });
+  const destination_id = server.schema.targets.all().models[0].id;
+  server.createList('alias', 3, { scope: globalScope, destination_id });
 
-  aliases.forEach((alias) => {
-    const destination_id = server.schema.targets.all().models[0].id;
-    alias.update({ destination_id });
-  });
   // Workers
   server.createList('worker', 3, { scope: globalScope });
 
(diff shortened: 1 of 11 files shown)


cameronperera (approved):

addons/api/mirage/scenarios/default.js
```diff
@@ -127,6 +124,13 @@ export default function (server) {
     server.createList('role', 3, { scope });
   });
 
+  // Aliases
+  const aliases = server.createList('alias', 3, { scope: globalScope });
+
+  aliases.forEach((alias) => {
```
calcaide, line 130: I think this is not need.

`destination_id` is a constant that always return the same value and we are going through each alias to override an attribute that we know the value.

I will use the same attribute overrides, [link to docs](https://miragejs.com/docs/main-concepts/factories/#attribute-overrides), we are already using with `scope.`

for example:
```jacascript
const destination_id = server.schema.targets.all().models[0].id;
server.createList('alias', 3, { scope: globalScope, destination_id });
```

DhariniJeeva: ah yes good call, updated

calcaide (approved):
LGTM!! Nice work! thanks 🙌

PR #2198, opened 2024-03-22
test(e2e): Add check for playing back session recording

moduli:
:tickets: [Jira ticket](https://hashicorp.atlassian.net/browse/ICU-13091)

This PR updates an Admin UI e2e test to add a check for playing back a session recording.

### DIFF
--- a/ui/admin/tests/e2e/tests/session-recording-ent.spec.js
+++ b/ui/admin/tests/e2e/tests/session-recording-ent.spec.js
@@ -75,7 +75,7 @@ test('Verify session recording can be deleted @ent @aws', async ({ page }) => {
     const project = projects.items.filter((obj) => obj.name == projectName)[0];
     const projectId = project.id;
 
-    // Create Storage Bucket
+    // Create storage bucket
     await page.getByRole('link', { name: 'Orgs', exact: true }).click();
     const storageBucketName = await createStorageBucket(
       page,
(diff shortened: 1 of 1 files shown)


ui/admin/tests/e2e/tests/session-recording-ent.spec.js
```diff
@@ -140,11 +136,30 @@ test('Verify session recording can be deleted @ent @aws', async ({ page }) => {
...
+      .getByRole('row', { name: targetName })
+      .getByRole('link', { name: 'View' })
+      .click();
+    await page
+      .getByRole('cell', { name: 'Channel 1' })
+      .locator('..')
+      .getByRole('link', { name: 'Play' })
+      .click();
+    await page.locator('div.session-recording-player').hover();
+    await page.locator('.ap-playback-button').click();
```
moduli, line 157: Definitely not the best way to access these elements, but they are the only ones available. I *think* we technically could add like a `test-id` to `session-recording-player`, but wasn't sure if it was worth it. I *think* the playback button is out of our control.

stasryzhov (approved):

ui/admin/tests/e2e/tests/session-recording-ent.spec.js
moduli: Leaving this for now.

Issue #2192, opened 2024-03-19
`recursive` parameter in `scopesCheck` breaks with strict rbac


macmiranda:
Another issue has been created in the main Boundary repo that I think is related to the `recursive` parameter in `scopesCheck` below
https://github.com/hashicorp/boundary-ui/blob/0902e42cb05ecbab9bd6549eba1f705c17541541/ui/desktop/app/routes/scopes.js#L26

When `recursive` is `true`, _at least one_ of your orgs must allow unauthenticated scopes list, otherwise the Boundary API will respond with a 500 status code, which makes the check fail and Boundary Desktop show the modal (thinking that it's actually talking to an older version of the API).

_Originally posted by @macmiranda in https://github.com/hashicorp/boundary/issues/4370#issuecomment-1999332273_

cameronperera:
Hi @macmiranda,

Thanks for letting us know about this issue. I have been trying to replicate this in relation to the other [issue](https://github.com/hashicorp/boundary/issues/4370#issuecomment-1999332273) you mentioned.

To properly replicate this issue, would you be able to share the grant string for your scopes that was causing the Boundary API to response with a 500?

macmiranda:
Hi @cameronperera 

To reproduce:

1. Start Boundary server in dev mode
2. Log in as admin
3. Modify the Global role Login Grants, Edit form, _make sure **Global** is selected_ as the Scope Id

<img src="https://github.com/hashicorp/boundary-ui/assets/17923899/63d7cbd3-0344-4784-8385-19a6aa009d91"  height="500" />

(this is intentional. I don't want unauthenticated users to be able to see which scopes exist under the Global one. Since my auth method is global, users are able to authenticate themselves and then have access to other scopes)

4. On the CLI
```
export BOUNDARY_ADDR=http://localhost:9200
```
```
boundary scopes list -recursive
Error from controller when performing list on scopes

Error information:
  Kind:                Internal
  Message:             output fields not found when building scope proto
  Status:              500
  context:             Error from controller when performing list on scopes
```
but without `-recursive`
```
boundary scopes list

Scope information:
  ID:                    o_1234567890
    Name:                Generated org scope
    Description:         Provides an initial org scope in Boundary
    Authorized Actions:
      no-op
```

PR #2199, opened 2024-03-22
fix: 🐛 remove recursive param from list scopes endpoint

cameronperera:
✅ Closes: https://hashicorp.atlassian.net/browse/ICU-12543

:tickets: [Jira ticket](https://hashicorp.atlassian.net/browse/ICU-12543)

## Description

A user has configured their boundary cluster to not allow listing all scopes when not authenticated. When checking to see if the controller is a supported version for the desktop client, we were using the `recursive: true` queryParam which was causing the API to return a `500` response which causes the desktop client to assume the controller is unsupported. Simply removing the `recursive: true` queryParam and only list the `global` scope, we remove that issue and properly determine if the controller is supported.

## How to Test

Follow steps mentioned [here](https://github.com/hashicorp/boundary-ui/issues/2192#issuecomment-2014765882). 

## Checklist:

### DIFF
--- a/ui/desktop/app/routes/scopes.js
+++ b/ui/desktop/app/routes/scopes.js
@@ -23,7 +23,6 @@ export default class ScopesRoute extends Route {
     try {
       const scopesCheck = await adapter.query(this.store, scopeSchema, {
         page_size: 1,
-        recursive: true,
       });
       // we use the response_type to determine if pagination is supported
       // and this will still pass even if there are no scopes returned


DhariniJeeva (approved):

lisbet-alvarez (approved):
🙌

calcaide (approved):
Good job!!

ui/desktop/app/routes/scopes.js
```diff
@@ -23,7 +23,6 @@ export default class ScopesRoute extends Route {
     try {
       const scopesCheck = await adapter.query(this.store, scopeSchema, {
         page_size: 1,
```
calcaide, line 25: Totally for learning purposes: I was checking the `adapter.query` [object documentation](https://api.emberjs.com/ember-data/5.3/classes/%3CInterface%3E%20Adapter/methods/query?anchor=query). I do understand `page_size` and `recursive` are params that belong to the`query`, is that correct?

cameronperera: Correct. In this instance, I am simply removing recursive so it only fetches the scopes at the global level. Which is this case, will always be the `global` scope. I had added `recursive: true` prior because that is simply what we typically do when fetching all scopes but due to permission configurations, the user could make that fetch fail by not allowing all scopes to be visible to unauthenticated users (which is completely valid for them to do).

PR #2203, opened 2024-03-25
feat: 🎸 authenticate to ferry on authentication

johanbrandhorst:
Replace the 'boundary daemon add-token' command with a simple 'boundary auth-tokens read -id <token_id>', relying on the boundary CLI to do the work of sending the token to the client and Ferry daemons.

:tickets: [Jira ticket](https://hashicorp.atlassian.net/browse/ICU-13142)

## Description

Authenticate the Ferry DNS daemon at the same time as the client daemon.

## How to Test

Run Ferry, login to the Desktop client, run `boundary ferry status` which should confirm that the user is authenticated.

## Checklist:

  - I couldn't find where to add tests for this feature

### DIFF
--- a/ui/desktop/electron-app/src/services/client-daemon-manager.js
+++ b/ui/desktop/electron-app/src/services/client-daemon-manager.js
@@ -109,23 +109,41 @@ class ClientDaemonManager {
 }
 
 const addTokenCliCommand = (token) => {
-  const addTokenCommand = [
-    'daemon',
-    'add-token',
+  const sanitizedToken = sanitizer.base62EscapeAndValidate(token);
+  // The format of Boundary tokens is
+  //   token-id_token-secret
+  // where token-id has the format
+  //   at_1234567890,
+  // so we extract the token id by finding the last '_' and taking everything before it.
+  lastUnderscoreIndex = sanitizedToken.lastIndexOf('_');
+  if (lastUnderscoreIndex === -1) {
+    return Promise.reject({
+      statusCode: 400,
+      message: 'Invalid token format',
+    });
+  }
+  const tokenId = sanitizedToken.substring(0, lastUnderscoreIndex);
+  // Successfully calling any Boundary CLI command with a token
+  // will add the token both to the Client daemon and the Ferry DNS daemon,
+  // so we just do a simple read on the input token and let the CLI do the rest.
+  const readTokenCommand = [
+    'auth-tokens',
+    'read',
+    '-id',
+    tokenId,
     `-addr=${runtimeSettings.clusterUrl}`,
     '-format=json',
     '-token=env://BOUNDARY_TOKEN',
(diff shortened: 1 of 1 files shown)


ui/desktop/electron-app/src/services/client-daemon-manager.js
```diff
@@ -109,23 +109,41 @@ class ClientDaemonManager {
...
+  if (lastUnderscoreIndex === -1) {
+    return Promise.reject({
+      statusCode: 400,
+      message: 'Invalid token format',
+    });
+  }
+  const tokenId = sanitizedToken.substring(0, lastUnderscoreIndex);
+  // Successfully calling any Boundary CLI command with a token
+  // will add the token both to the Client daemon and the Ferry DNS daemon,
+  // so we just do a simple read on the input token and let the CLI do the rest.
```
ZedLi, line 128: I didn't realize calling any CLI command with a token automatically handles adding tokens to daemons, was this a recent change?

Also, is the recommendation to just go through the CLI for adding tokens at this point since it handles adding multiple tokens? I had a [TODO](https://github.com/hashicorp/boundary-ui/pull/2203/files#diff-3a31bb3a8a9063bb4dc5b0c1dba204044a1bc4543a264758d7cbd0962f195207R76) regarding whether we should just always go through the CLI for adding tokens and it sounds like it might be better at this point? This will only add the token right now for windows since it's the only one that calls the CLI directly (mac/linux adds the token to the client daemon directly right now).

If you prefer, I can update this PR if that's the recommended way.

johanbrandhorst: Automatically adding the token to the client daemon on any command was shipped with search and filtering - we definitely could've communicated that better, sorry! On Boundary `main` right now it will also add the token to the ferry daemon. I'd say going through the CLI is probably the simplest thing for all platforms, yeah.

One caveat I do want to mention is that ferry can be configured to serve on a different port than the default, and in these cases a separate flag will need to be set to let the CLI know where to direct its request (`--ferry-port`). I think we're OK with not supporting that until such a time as we have a way for the user to configure that in their Desktop client, but it's something to keep in mind.

If you'd like to commandeer this PR to make those changes, that'd be great! We only really care about it for Mac for now, since we're not shipping transparent sessions for Linux right now, but it can't hurt :).

ZedLi: Sure I can do that, really appreciate you opening this PR! And thanks for the info, this is all good to know. 

Just for my edification, does the ferry service also expose its API over a unix socket like the client daemon?

calcaide (approved):
Terrific work here!! simpler and cleaner with the new approach 🙌

lisbet-alvarez (approved):
🙌

cameronperera (approved):

PR #2206, opened 2024-03-27
configure Desktop logging

DhariniJeeva:

## Description

This PR installs `electron log` to add logs to desktop client.  

## How to Test

- run binary locally and perform search on DC 
- you can see the log file being generated in the below mentioned location based on your OS 

on Linux: ~/.config/{app name}/logs/main.log
on macOS: ~/Library/Logs/{app name}/main.log
on Windows: %USERPROFILE%\AppData\Roaming\{app name}\logs\main.log

## Checklist:


addons/api/package.json
```diff
@@ -30,7 +30,6 @@
     "ember-cli-mirage": "^2.4.0",
     "ember-data": "~4.12.0",
     "ember-fetch": "^8.1.1",
-    "ember-auto-import": "^2.6.3",
```
calcaide, line 33: Are we deleting `ember-auto-import`?

ui/desktop/electron-app/src/services/client-daemon-manager.js
```diff
@@ -154,6 +156,8 @@ const searchCliCommand = (requestData) => {
   }
   const sanitizedToken = sanitizer.base62EscapeAndValidate(requestData.token);
 
+  log.info('logging search command:', searchCommand);
```
calcaide, line 159: I think, before we log `searchCommand` we should delete the `token` field from it, we can not log sensitive data such as tokens.

calcaide:
I will suggest to add a log when we detect the clientDaemon is failing and we fall back to the controller, [not sure should be here](https://github.com/hashicorp/boundary-ui/blob/main/addons/api/addon/handlers/client-daemon-handler.js#L119-L139), but somewhere around there.
The idea of my suggestion is to know:

- When and how often the daemon fails and we need to fallback to the controller.
- To know if we are fetching from daemon or controller.

addons/api/package.json
calcaide: Good catch!

ui/desktop/electron-app/src/services/client-daemon-manager.js
calcaide: Taking as example a line of the generated log:
```
{
  method: 'GET',
  path: 'http://internal.boundary.local/v1/search?force_refresh=true&query=&auth_token_id=at_JhnjID5a06&resource=targets',
  socketPath: '/Users/ccorvo/.boundary/socket/daemon.sock'
}
```

Inside the path, I see the `auth_token_id` is this something safe to log?

calcaide:
Awesome!! this is looking great @DhariniJeeva !!

ZedLi:
Is this going to get merged into main? What was the plan for this?

calcaide:
Ups!! good catch @ZedLi , definetly this should be in a feature or LL branch, so we can keep working on it, build experimental binaries if need until we decide to bring it to `main`

DhariniJeeva:
I wasn't sure when we would merge this or what the scope of this work is, but for now, as Carlos suggested, I will create an LLB for this work.

ZedLi:
Closed and handled in https://github.com/hashicorp/boundary-ui/pull/2234

PR #2214, opened 2024-04-03
Icu 12933 updates to target model to support aliasing

DhariniJeeva:

## Description

This PR adds support to the target model to allow the creation of aliases within targets. 

NOTE: After you create it, you cannot see the aliases in the UI. This will be done in a separate PR. 

https://github.com/hashicorp/boundary-ui/assets/15043878/6cf737d2-d431-466a-a036-670d6d4e77b5

target

## How to Test

1. Get the latest boundary main, `make install` and  start the controller 
2. go to targets `/targets` within a project
3. you should only be able to see the option to create aliases with the target creation form if the user has access to create aliases 
4. Input name, select TCP target, input default port, and enter 0 or more aliases. 
5. Hit submit and save successfully.

**Note**: the above steps are for TCP targets; if you want to test SSH targets, you should use enterprise binary.  

## Checklist:


addons/api/addon/models/role.js
```diff
@@ -27,7 +27,7 @@ export default class RoleModel extends GeneratedRoleModel {
    * see obvious.  Instead, the application layer is expected to load referenced
    * users and groups as needed.
    */
-  @attr('principal-array', {
```
DhariniJeeva, line 30: I refactored this to a generic array which can be used in other resources as well.

ui/admin/app/components/form/role/add-principals/index.js
```diff
@@ -35,7 +35,7 @@ export default class FormRoleAddPrincipalsIndexComponent extends Component {
   get filteredPrincipals() {
     // Retrieve principal IDs assigned to current role
     const currentPrincipalIDs = this.args.model.principals.map(
-      (principal) => principal.principal_id,
+      (principal) => principal.id,
```
DhariniJeeva, line 38: I tested the role/principals workflow to ensure this refactor didn't cause any regression

addons/api/addon/serializers/target.js
```diff
@@ -44,6 +44,14 @@ export default class TargetSerializer extends ApplicationSerializer {
         injectedApplicationCredentialSourceIDs,
       );
 
+    if (serialized?.with_aliases) {
+      // API expects scope id along with every alias value
+      serialized.with_aliases = serialized.with_aliases.map((item) => ({
+        ...item,
+        scope_id: 'global',
```
calcaide, line 51: None blocking comment: In the future, I guess we should pass a dynamic value as `scope_id`?

addons/api/tests/unit/serializers/target-test.js
```diff
@@ -36,6 +36,7 @@ module('Unit | Serializer | target', function (hooks) {
       address: '0.0.0.0',
       enable_session_recording: false,
       storage_bucket_id: null,
+      with_aliases: [{ value: 'a', scope_id: 'global' }],
```
calcaide, line 39: If the serializer adds `scope_id` within `with_aliases` shouldn't attempt to create the record without passing `scope_id`? so we validate the serializer is correctly adding the `scope_id: 'global'`?

addons/api/tests/unit/serializers/target-test.js
```diff
@@ -85,6 +87,7 @@ module('Unit | Serializer | target', function (hooks) {
       enable_session_recording: false,
       storage_bucket_id: null,
       address: '0.0.0.0',
+      with_aliases: [],
```
calcaide, line 90: Why are we passing `with_aliases` empty?

addons/api/tests/unit/transforms/array-test.js
```diff
@@ -0,0 +1,51 @@
...
+ * SPDX-License-Identifier: MPL-2.0
+ */
+
+import { module, test } from 'qunit';
+import { setupTest } from 'ember-qunit';
+
+module('Unit | Transform | array', function (hooks) {
+  setupTest(hooks);
+
+  // Replace this with your real tests.
```
calcaide, line 12: I will suggest deleting this first automated test.

ui/admin/app/components/form/target/details/index.hbs
```diff
@@ -325,6 +325,42 @@
...
+        {{#if @model.errors.with_aliases}}
+          <F.Error as |E|>
+            {{#each @model.errors.with_aliases as |error|}}
+              <E.Message>{{error.message}}</E.Message>
+            {{/each}}
+          </F.Error>
+        {{/if}}
+      </:fieldset>
+
+      <:field as |F|>
```
calcaide, line 354: None blocking comment: I am seeing when adding aliases to a target, every input has the same `aria-label=Value`. I know this is not specific to this PR and more to the component that we use, but what do you think? should we try to get a more descriptive `aria-label` or value is good enough?

ui/admin/tests/acceptance/targets/create-test.js
```diff
@@ -323,6 +325,33 @@ module('Acceptance | targets | create', function (hooks) {
     assert.strictEqual(getTCPTargetCount(), tcpTargetCount);
   });
 
+  test('can add aliases during target creation', async function (assert) {
```
calcaide, line 328: None blocking comment: Wondering, should we add another test case were we attempt to add  2 or 3 aliases? thoughts? Not necessary on this PR, since this is an initiative WIP 😉

calcaide (approved):
I tested manually the creation of a target with alias for TCP. The creation was successful, and as you mention on the PR description I could NOT see on the UI after creation the alias value within the target detail screen, but inspecting the network request I can see it. Also, I can see the alias information inspecting ember data store for target.

Great job!!! left some comments 😉 terrific work!! 💃 💪

addons/api/tests/unit/serializers/target-test.js
DhariniJeeva: Ooh that is a good point, I'll update accordingly

addons/api/tests/unit/serializers/target-test.js
DhariniJeeva: since I made a test with non empty array in previous test case, I didn't do here and other places, but since this is SSH targets, it makes sense to do a non empty test case? I'll pass a non empty arr

addons/api/addon/serializers/target.js
DhariniJeeva: Yup, when we have more than a global scope, we should allow the user to select the scope that they want to create in the UI, and we can pass in dynamically then. Right now, it is only global scope, hence I thought additional calls weren't necessary

ui/admin/app/components/form/target/details/index.hbs
DhariniJeeva: Good point- I agree, we should use a descriptive label. I'll think more about this and do a refactor separately to update the component

PR #2220, opened 2024-04-05
chore: 🤖 Model & Mirage support for MinIO

calcaide:

## Description

First iteration of model & mirage support for MinIO.

After discussion on Tech time (Thursday April the 4th) I decided to keep this as simple as possible until we progress a bit more on the initiative: add routes, a way to change providers and such. BUT happy to change approach if reviewers suggest so.

## How to Test

Be aware this PR do NOT need a boundary instance since is changing models/mirage, on coming PR's, the usage of a boundary instance will be mandatory. [LINK to Vercel deployment](https://boundary-ui-git-chore-13335-model-support-minio-hashicorp.vercel.app/scopes/global/storage-buckets).

### Mirage at listing
Open storage buckets list, validate we generate some storage buckets with minio as provider.

### Mirage at create
- Open storage bucket, click new storage bucket.
- Open the browser console.
- Fill or not (not mandatory) and click Save.
- Inspect on the browser console the POST to `/v1/storage-buckets`. Be aware the param plugin_name is incorrect at the moment (will fix this on coming PR).
- The body request should contain `endpoint_url`.

### Mirage at storage bucket read
- Open storage buckets list.
- Open your browser console.
- Click on a Storage bucket with minio as provider.
- Inspect on the browser console the GET to `/v1/storage-buckets/sb_id`.
- The body request should contain:
    - `endpoint_url` within attributes.
    - plugin name should be minio.

## Checklist:

~- [ ] I have added before and after screenshots for UI changes~
~- [ ] I have added JSON response output for API changes~
~- [ ] I have commented on my code, particularly in hard-to-understand areas~
~- [ ] My changes generate no new warnings~
~- [ ] I have added tests that prove my fix is effective or that my feature works~

### DIFF
--- a/addons/api/addon/models/storage-bucket.js
+++ b/addons/api/addon/models/storage-bucket.js
@@ -81,7 +81,7 @@ export default class StorageBucketModel extends GeneratedStorageBucketModel {
    * True if the storage bucket plugin type is MinIO
    * @type {boolean}
    */
-  get isMinio() {
+  get isMinIO() {
     return this.compositeType === TYPE_STORAGE_BUCKET_PLUGIN_MINIO;
   }
 
--- a/addons/api/addon/models/storage-bucket.js
+++ b/addons/api/addon/models/storage-bucket.js
@@ -9,8 +9,10 @@ export const TYPE_STORAGE_BUCKET_PLUGIN = 'plugin';
 export const TYPES_STORAGE_BUCKET = Object.freeze([TYPE_STORAGE_BUCKET_PLUGIN]);
 
 export const TYPE_STORAGE_BUCKET_PLUGIN_AWS_S3 = 'aws';
+export const TYPE_STORAGE_BUCKET_PLUGIN_MINIO = 'minio';
 export const TYPES_STORAGE_BUCKET_PLUGIN = Object.freeze([
   TYPE_STORAGE_BUCKET_PLUGIN_AWS_S3,
+  TYPE_STORAGE_BUCKET_PLUGIN_MINIO,
 ]);
 
 export const TYPE_CREDENTIAL_STATIC = 'static';
@@ -73,6 +75,14 @@ export default class StorageBucketModel extends GeneratedStorageBucketModel {
    */
   get isAWS() {
     return this.compositeType === TYPE_STORAGE_BUCKET_PLUGIN_AWS_S3;
+  }
+
+  /**
+   * True if the storage bucket plugin type is MinIO
+   * @type {boolean}
+   */
+  get isMinio() {
+    return this.compositeType === TYPE_STORAGE_BUCKET_PLUGIN_MINIO;
   }
 
   /**
--- a/addons/api/addon/serializers/storage-bucket.js
+++ b/addons/api/addon/serializers/storage-bucket.js
@@ -4,7 +4,10 @@
  */
 
 import ApplicationSerializer from './application';
-import { TYPE_CREDENTIAL_DYNAMIC } from '../models/storage-bucket';
+import {
+  TYPE_CREDENTIAL_DYNAMIC,
+  TYPE_STORAGE_BUCKET_PLUGIN_AWS_S3,
+} from '../models/storage-bucket';
 
 const fieldsByCredentialType = {
   static: [
@@ -25,6 +28,12 @@ const fieldsByCredentialType = {
   ],
 };
 export default class StorageBucketSerializer extends ApplicationSerializer {
+  /**
+   * Serializes storage-bucket. ATM just AWS needs specific serialization,
+   * for the other plugin types use default.
+   * @param {object} snapshot
+   * @returns
+   */
   serialize(snapshot) {
     const { credentialType } = snapshot.record;
     const pluginType = snapshot.record.plugin?.name;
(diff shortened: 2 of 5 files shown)


addons/api/addon/serializers/storage-bucket.js
```diff
@@ -12,27 +12,49 @@ const fieldsByCredentialType = {
...
+    'disable_credential_rotation',
+    'endpoint_url',
   ],
 };
 export default class StorageBucketSerializer extends ApplicationSerializer {
   serialize(snapshot) {
-    const serialized = super.serialize(...arguments);
     const { credentialType } = snapshot.record;
+    const pluginType = snapshot.record.plugin?.name;
+    const serialized = super.serialize(...arguments);
```
calcaide, line 31: I am trying to follow [credential-store](https://github.com/hashicorp/boundary-ui/blob/main/addons/api/addon/serializers/credential-store.js) approach, but since in this case there are some common serialization I decided to serialize here, and pass to specific type methods the already serialized object. Let me know if I should follow a different approach.

addons/api/tests/unit/serializers/storage-bucket-test.js
```diff
@@ -222,6 +255,62 @@ module('Unit | Serializer | storage bucket', function (hooks) {
     assert.deepEqual(serializedRecord, expectedResult);
   });
 
+  test('it serializes when updateing a MinIO (static credential by default) storage bucket correctly', async function (assert) {
```
calcaide, line 258: I am thinking on adding a test were we update an AWS storage bucket to be MinIO, but not sure that even makes sense, I will check with stakeholders and if does sense I will add such test.

addons/api/addon/models/storage-bucket.js
```diff
@@ -75,6 +77,14 @@ export default class StorageBucketModel extends GeneratedStorageBucketModel {
     return this.compositeType === TYPE_STORAGE_BUCKET_PLUGIN_AWS_S3;
   }
 
+  /**
+   * True if the storage bucket plugin type is MinIO
+   * @type {boolean}
+   */
+  get isMinio() {
```
DhariniJeeva, line 84: this is a nitpick - should we consolidate the usage here? how about MinIO?

addons/api/addon/serializers/storage-bucket.js
DhariniJeeva: I think we should refactor and simplify this serializer if we introduce `for` in the generated model, take a look at `authmethod` [model](https://github.com/hashicorp/boundary-ui/blob/main/addons/api/addon/generated/models/auth-method.js#L214).

addons/api/addon/serializers/storage-bucket.js
```diff
@@ -12,27 +12,49 @@ const fieldsByCredentialType = {
...
     if (!snapshot.record.isNew) {
       delete serialized.bucket_name;
       delete serialized.bucket_prefix;
     }
 
-    if (credentialType === TYPE_CREDENTIAL_DYNAMIC) {
-      serialized.attributes.disable_credential_rotation = true;
+    switch (pluginType) {
+      case 'minio':
+        return this.serializeMinio(serialized);
```
DhariniJeeva, line 41: I think inversing this might be a better approach. We can make MinIO the default option since it doesn't have any serialization, and we don't need a separate case for this. Thoughts?

addons/api/addon/models/storage-bucket.js
calcaide: Sure! Nice catch! 😉

addons/api/addon/serializers/storage-bucket.js
calcaide: @DhariniJeeva 
After a conversation on slack, the storage bucket can not be serialized using the `for` to map with a type. The type in storage bucket is only plugin, so we can not use it to differentiate between plugin types, therefor the usage of `for` at serializing will not work. 
In the storage-bucket model, we are using `compositeType` to identify the plugin type.

addons/api/addon/serializers/storage-bucket.js
calcaide: @DhariniJeeva 
I think you are totally right, ATM there is no need for specific MinIO serialization. So will perform a specific one just for AWS and return a default one valid for MinIO.
Thanks for the catch!

addons/api/tests/unit/serializers/storage-bucket-test.js
```diff
@@ -126,6 +159,7 @@ module('Unit | Serializer | storage bucket', function (hooks) {
           bucket_name: 'bucketname',
           bucket_prefix: 'bucketprefix',
           worker_filter: 'workerfilter',
+          plugin: { name: 'aws' },
```
calcaide, line 162: Why the need to add `plugin` in this test?

Before this PR, the serializer didn't check `pluginType` to perform a different serialization. Now we do and when `pluginType` is unknown, returns default serialization when we expect an aws `pluginType` serialization. Therefore, I add the plugin name so when checking `pluginType` it returns aws rather than unknown.

addons/api/tests/unit/serializers/storage-bucket-test.js
calcaide: After checking with BE, the plugin type MUST be set at creation time ONLY. Can not be edited.

DhariniJeeva (approved):
Looks good to me! 🎉

cameronperera (approved):

PR #2223, opened 2024-04-05
feat: 🎸 Update how we store tokens in indexed-db

ZedLi:
## Description
Added logic to make the list tokens stored in indexed DB more granular. I get a hash of the query parameters that will get sent to the controller and use that as the id for the token. This also means extra logic for when a token is invalid as we will need to delete all tokens of the same resource type.

## How to Test
Go through all the resources that will used indexed DB and confirm that tokens are being correctly added. Delete or add new scopes to test invalid list tokens for scope resources. 

## Checklist:


lisbet-alvarez (approved):
🙌 Thx for all this work & the detailed comments!

PR #2225, opened 2024-04-05
Target aliases list view

DhariniJeeva:
:tickets: [Jira ticket](https://hashicorp.atlassian.net/browse/ICU-12955)

## Description

This PR adds a list of associated aliases in the side bar

**When there is no alias associated:**
without alias association
**list view:**
alias list view
**flyout:**
flyoyt

https://github.com/hashicorp/boundary-ui/assets/15043878/a93f6c54-14ee-4e2d-9eb5-96d9c8455b40

## How to Test

1. checkout latest from the boundary main and run the controller
2. In the UI, create a TCP target with 0 or more aliases 
3. After successful creation, you should be able to view them on the right sidebar
4. If there are no aliases, you should see an option to add a new one. Currently it is just a button and does not have a route attached. 

## Checklist:

### DIFF
--- a/addons/api/mirage/scenarios/default.js
+++ b/addons/api/mirage/scenarios/default.js
@@ -125,15 +125,17 @@ export default function (server) {
   });
 
   // Aliases
-  const targets = server.schema.targets.all().models[0];
-  const destination_id = server.schema.targets.all().models[0].id;
-  const aliases = server.createList('alias', 3, {
+  const aliasDestinationTarget = server.schema.targets.all().models[0];
+  const { id: destination_id } = aliasDestinationTarget;
+  const aliases = server.createList('alias', 4, {
     scope: globalScope,
     destination_id,
   });
-  targets.update({
-    aliases: [{ id: aliases[0].id, value: aliases[0].value }],
+
+  aliasDestinationTarget.update({
+    aliases: aliases.map((alias) => ({ id: alias.id, value: alias.value })),
   });
+
   // Workers
   server.createList('worker', 3, { scope: globalScope });
 
--- a/addons/api/mirage/scenarios/default.js
+++ b/addons/api/mirage/scenarios/default.js
@@ -125,9 +125,15 @@ export default function (server) {
   });
 
   // Aliases
+  const targets = server.schema.targets.all().models[0];
   const destination_id = server.schema.targets.all().models[0].id;
-  server.createList('alias', 3, { scope: globalScope, destination_id });
-
+  const aliases = server.createList('alias', 3, {
+    scope: globalScope,
+    destination_id,
+  });
+  targets.update({
+    aliases: [{ id: aliases[0].id, value: aliases[0].value }],
+  });
   // Workers
   server.createList('worker', 3, { scope: globalScope });
 
--- a/ui/admin/app/styles/app.scss
+++ b/ui/admin/app/styles/app.scss
@@ -623,6 +623,24 @@
   }
 }
 
+.target-sidebar-aliases {
+  margin-bottom: 3rem;
+
+  .description {
+    color: var(--token-form-helper-text-color);
+    font-weight: var(--token-typography-font-weight-normal);
+    font-size: var(--token-typography-body-200-font-size);
+  }
+
+  .hds-link-standalone {
+    margin-top: 1rem;
+  }
+
+  .link-list-item {
+    margin-bottom: 0.5rem;
+  }
+}
+
 .link-list-item {
   display: flex;
   align-items: center;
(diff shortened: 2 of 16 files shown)


ui/admin/tests/acceptance/targets/enable-session-recording/index-test.js
```diff
@@ -12,215 +12,218 @@ import { authenticateSession } from 'ember-simple-auth/test-support';
...
-
-  hooks.beforeEach(function () {
-    featuresService = this.owner.lookup('service:features');
-    // Generate resources
-    instances.scopes.global = this.server.create('scope', { id: 'global' });
-    instances.scopes.org = this.server.create('scope', {
-      type: 'org',
-      scope: { id: 'global', type: 'global' },
+module(
+  'Acceptance | targets | enable session recording | index',
```
DhariniJeeva, line 16: I only updated the module name here, as there is another file with the same module name

ui/admin/app/styles/app.scss
```diff
@@ -623,6 +623,24 @@
   }
 }
 
+.target-sidebar-aliases {
```
ZedLi, line 626: I don't think this is related to this PR (but I don't know if it existed since our last release) but I noticed the migration button for me is...wonky

https://github.com/hashicorp/boundary-ui/assets/5783847/0766fdc6-5be0-4f50-aca0-9803052e6937

ui/admin/app/templates/scopes/scope/targets/target/index.hbs
```diff
@@ -30,9 +30,74 @@
...
+              </Hds::Flyout>
+            {{/if}}
+          {{else}}
+            <Hds::Button
+              @icon='plus'
+              @text={{t 'resources.target.actions.add-an-alias'}}
+              @color='secondary'
+              {{! @route='' }}
+            />
+          {{/if}}
```
ZedLi, line 97: Can a target can have more than one alias? If that's correct doesn't this else check prevent you from adding more if an alias is added?

addons/api/mirage/scenarios/default.js
```diff
@@ -125,9 +125,15 @@ export default function (server) {
...
+  const targets = server.schema.targets.all().models[0];
   const destination_id = server.schema.targets.all().models[0].id;
-  server.createList('alias', 3, { scope: globalScope, destination_id });
-
+  const aliases = server.createList('alias', 3, {
+    scope: globalScope,
+    destination_id,
+  });
+  targets.update({
+    aliases: [{ id: aliases[0].id, value: aliases[0].value }],
```
ZedLi, line 135: Did you only want one alias specifically instead of including all the aliases?
Could we also include a target with multiple aliases?

```suggestion
    aliases: aliases.map((alias) => ({ id: alias.id, value: alias.value })),
```

ui/admin/app/templates/scopes/scope/targets/target/index.hbs
```diff
@@ -30,9 +30,74 @@
...
+          </p>
+          {{#if @model.aliases}}
+
+            {{#each this.aliases as |alias|}}
+              <LinkListPanel as |P|>
+                <P.Item
+                  @icon='at-sign'
+                  @color='boundary'
+                  @text={{alias.value}}
+                  {{! @route=' ' }}
```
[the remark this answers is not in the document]
DhariniJeeva: Yes, this will be updated in the following PR, which is currently in a draft state

ui/admin/app/templates/scopes/scope/targets/target/index.hbs
DhariniJeeva: A target can have multiple aliases only at creation time in the target model. To add an alias within the alias domain/model, we use this button, but in the alias model we are allowed to add only one alias at a time.

addons/api/mirage/scenarios/default.js
DhariniJeeva: good call, updated

ui/admin/app/styles/app.scss
DhariniJeeva: I'm guessing this has always been there due to the container wrapper that we have inside `F.control` . I moved the button within the container wrapper, and it seemed to fix the issue. Can you check if you are still seeing this?

ui/admin/app/templates/scopes/scope/targets/target/index.hbs
DhariniJeeva: yes, it does. the target route will have this update

calcaide (approved):
Terrific job!! Aliases getting closer, looks very nice 😃

PR #2226, opened 2024-04-08
Icu 13374 add permission checks for filtering resources in sessions and targets

cameronperera:
:tickets: [Jira ticket](https://hashicorp.atlassian.net/browse/ICU-13374)

## Description

We add new resources being queried in the `/sessions` and `/targets` routes. Those resources have their own permissions that need to be checked before fetching. This adds those checks and modifies the respective templates to hide the filters if needed.

### Targets route
before:
Screenshot 2024-04-08 at 12 03 45 PM
after:
Screenshot 2024-04-08 at 12 03 23 PM

### Sessions route
before:
Screenshot 2024-04-08 at 5 17 13 PM
after:
Screenshot 2024-04-08 at 5 17 30 PM

## How to Test

To test the targets route, remove the `list` permission for sessions.

To test the sessions route, remove the 'list' permission for targets and users.

## Checklist:

### DIFF
--- a/ui/admin/app/routes/scopes/scope/sessions/index.js
+++ b/ui/admin/app/routes/scopes/scope/sessions/index.js
@@ -10,6 +10,7 @@ import { action } from '@ember/object';
 export default class ScopesScopeSessionsIndexRoute extends Route {
   // =services
 
+  @service can;
   @service store;
 
   // =attributes
@@ -40,8 +41,8 @@ export default class ScopesScopeSessionsIndexRoute extends Route {
   };
 
   allSessions;
-  associatedUsers;
-  associatedTargets;
+  associatedUsers = [];
+  associatedTargets = [];
 
   /**
    * Loads all sessions under the current scope and encapsulates them into
@@ -49,7 +50,12 @@ export default class ScopesScopeSessionsIndexRoute extends Route {
    * @return {Promise{[{sessions: [SessionModel], allSessions: [SessionModel], associatedUsers: [UserModel], associatedTargets: [TargetModel], totalItems: number}]}}
    */
   async model({ search, users, targets, status, page, pageSize }) {
-    const { id: scope_id } = this.modelFor('scopes.scope');
+    const scope = this.modelFor('scopes.scope');
+    // Fetch the global and org scope to check user permissions
+    const orgScope = await this.store.findRecord('scope', scope.scope.id);
+    const globalScope = await this.store.findRecord('scope', 'global');
+
+    const { id: scope_id } = scope;
     const filters = {
       scope_id: [{ equals: scope_id }],
       status: [],
@@ -80,10 +86,17 @@ export default class ScopesScopeSessionsIndexRoute extends Route {
     if (!this.allSessions) {
       await this.getAllSessions(scope_id);
     }
-    if (!this.associatedUsers) {
+    if (
+      this.can.can('list model', globalScope, { collection: 'users' }) &&
+      this.can.can('list model', orgScope, { collection: 'users' }) &&
+      this.associatedUsers.length === 0
+    ) {
       await this.getAssociatedUsers();
     }
-    if (!this.associatedTargets) {
+    if (
+      this.can.can('list model', scope, { collection: 'targets' }) &&
+      this.associatedTargets.length === 0
+    ) {
       await this.getAssociatedTargets();
     }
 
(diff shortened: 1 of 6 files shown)


ui/admin/app/routes/scopes/scope/targets/index.js
```diff
@@ -60,20 +60,21 @@ export default class ScopesScopeTargetsIndexRoute extends Route {
...
 
-    const sessions = await this.store.query('session', {
-      query: {
-        filters: {
-          scope_id: [{ equals: scope_id }],
-          status: [
-            { equals: STATUS_SESSION_ACTIVE },
-            { equals: STATUS_SESSION_PENDING },
-          ],
+    let sessions = [];
```
lisbet-alvarez, line 63: small nitpick: is it necessary to initialize sessions outside the if statement if its only used to be passed into the addActiveSessionFilters function

ui/admin/tests/acceptance/targets/list-test.js
```diff
@@ -239,6 +239,18 @@ module('Acceptance | targets | list', function (hooks) {
     assert.dom(`[href="${urls.tcpTarget}"]`).doesNotExist();
   });
 
+  test('active sessions filter is hidden if user does not have permission to list sessiosn', async function (assert) {
```
lisbet-alvarez, line 242:
```suggestion
  test('active sessions filter is hidden if user does not have permission to list sessions, async function (assert) {
```

ui/admin/tests/acceptance/sessions/list-test.js
```diff
@@ -229,6 +229,21 @@ module('Acceptance | sessions | list', function (hooks) {
     assert.dom('tbody tr').exists({ count: 1 });
   });
 
+  test('users filter is hidden if no users', async function (assert) {
```
lisbet-alvarez, line 232: small suggestion: perhaps these two test case titles can be updated to be similar to the one in the targets folder below

lisbet-alvarez (approved):
🙌 Looks great !

ui/admin/app/routes/scopes/scope/sessions/index.js
```diff
@@ -81,10 +87,17 @@ export default class ScopesScopeSessionsIndexRoute extends Route {
     if (!this.allSessions) {
       await this.getAllSessions(scope_id);
     }
-    if (!this.associatedUsers) {
+    if (
+      this.can.can('list model', globalScope, { collection: 'users' }) &&
+      this.can.can('list model', orgScope, { collection: 'users' }) &&
```
ZedLi, line 92: Can these permission checks be moved into the helper methods since they're used multiple times?

ui/admin/app/routes/scopes/scope/sessions/index.js
```diff
@@ -169,12 +182,27 @@ export default class ScopesScopeSessionsIndexRoute extends Route {
...
+      await this.getAssociatedTargets();
+    }
 
     return super.refresh(...arguments);
   }
+
+  setupController(controller) {
+    const scope = this.modelFor('scopes.scope');
+    super.setupController(...arguments);
+    controller.setProperties({ scope });
```
ZedLi, line 206: Is this used in the sessions template?

cameronperera: Nope. Thanks for catching this. I thought I needed it in the template but ended up not and didn't remove this.

ui/admin/app/routes/scopes/scope/sessions/index.js
cameronperera: Yep.

ui/admin/app/routes/scopes/scope/sessions/index.js
```diff
@@ -162,18 +171,32 @@ export default class ScopesScopeSessionsIndexRoute extends Route {
     );
   }
 
+  async canListUsers(orgId) {
+    const orgScope = await this.store.findRecord('scope', orgId);
+    const globalScope = await this.store.findRecord('scope', 'global');
```
ZedLi, line 176: These two should be passed in rather than fetched so we don't call multiple times. I think using a `canListUsers` is fine but you also could have just wrapped the permissions check as part of `getAssociatedUsers`.

ZedLi (approved):
Looks good! Just had one last comment

ui/admin/app/routes/scopes/scope/sessions/index.js
cameronperera: Right. That would be much better. I will update that now.

PR #2227, opened 2024-04-09
fix: 🐛 modal bubbling up when user has not updated model

lisbet-alvarez:
✅ Closes: https://hashicorp.atlassian.net/browse/ICU-13106

:tickets: [Jira ticket](https://hashicorp.atlassian.net/browse/ICU-13106)

## Description

Fixes the "Discard unsaved changes?" modal from popping up when the user made no changes to the model. Bug was seen in the global settings page when trying to navigate away.

Before:

https://github.com/hashicorp/boundary-ui/assets/107949262/ac34d865-f60c-4959-946a-8e0b8283b3c7

After:

https://github.com/hashicorp/boundary-ui/assets/107949262/f9ad6e4b-c007-4ca8-8d91-dfbd226bd2e9

## How to Test

1. start boundary dev instance.
2. login in
3. navigate to "Global settings"
4. Attempt to navigate to a different route, and validate that "Discard unsaved changes?" modal does not pop up.
( look at slack thread in jira ticket for additional details if necessary )

## Checklist:

### DIFF
--- a/ui/admin/app/routes/scopes/scope.js
+++ b/ui/admin/app/routes/scopes/scope.js
@@ -33,17 +33,13 @@ export default class ScopesScopeRoute extends Route {
    * @param {string} params.scope_id
    * @return {Promise{ScopeModel}}
    */
-  model({ scope_id: id }) {
+  async model({ scope_id: id }) {
     // Since only global and org scopes are authenticatable, we can infer type
     // from ID because global has a fixed ID.
     const type = id === 'global' ? 'global' : 'org';
     return this.store.findRecord('scope', id).catch(() => {
       const maybeExistingScope = this.store.peekRecord('scope', id);
       const scopeOptions = { id, type };
-      /* istanbul ignore else */
-      if (type === 'global') {
-        scopeOptions.name = this.intl.t('titles.global');
-      }
       return (
         maybeExistingScope || this.store.createRecord('scope', scopeOptions)
       );
--- a/ui/admin/app/components/header-nav/index.hbs
+++ b/ui/admin/app/components/header-nav/index.hbs
@@ -4,7 +4,11 @@
 }}
 
 <Rose::Dropdown
-  @text={{this.scope.org.displayName}}
+  @text={{if
+    this.scope.org.isGlobal
+    (t 'titles.global')
+    this.scope.org.displayName
+  }}
   @icon={{if
     this.scope.org.isGlobal
     'flight-icons/svg/globe-16'


ui/admin/app/routes/scopes/scope.js
```diff
@@ -33,17 +33,13 @@ export default class ScopesScopeRoute extends Route {
    * @param {string} params.scope_id
    * @return {Promise{ScopeModel}}
    */
-  model({ scope_id: id }) {
+  async model({ scope_id: id }) {
```
cameronperera, line 36: Does this need to be `async` since we are not `awaiting` anything?

lisbet-alvarez: Marked it as async since it returns a promise

ui/admin/app/routes/scopes/scope.js
```diff
@@ -33,17 +33,13 @@ export default class ScopesScopeRoute extends Route {
...
     // Since only global and org scopes are authenticatable, we can infer type
     // from ID because global has a fixed ID.
     const type = id === 'global' ? 'global' : 'org';
     return this.store.findRecord('scope', id).catch(() => {
       const maybeExistingScope = this.store.peekRecord('scope', id);
       const scopeOptions = { id, type };
-      /* istanbul ignore else */
-      if (type === 'global') {
-        scopeOptions.name = this.intl.t('titles.global');
```
ZedLi, line 45: Was this originally just to have the capitalized name in the other file you updated? 😕

ZedLi (approved):
Great catch! Thanks for the fix

ui/admin/app/routes/scopes/scope.js
lisbet-alvarez: I'm assuming that could have been the reason (so that the displayname shows the capitalized version), or since this was added a while ago maybe in the first iterations the global scope didn't have a name already defined

ui/admin/app/routes/scopes/scope.js
cameronperera: I think we only need `async` if we are explicitly using `await`. Otherwise, ember will handle the promise for us implicitly. For example, we if did `const response = await this.store.findRecord('scope', id)` then checked if the response contains an error, we would need to explicitly add `async`. I could be wrong and fully ok with that. Just making sure I know why `async` is needed. This [link](https://guides.emberjs.com/release/routing/specifying-a-routes-model/#toc_fetch-example) is what I am basing my assumption on.

lisbet-alvarez: Yup ur correct, it's only explicitly needed when using await. For context, my thought process was that it was better practice to label functions that return a promise as async for clarity/readability, and so that the return value is always a promise but after looking at the link u posted and reading this [article](https://dev.to/somedood/best-practices-for-es2017-asynchronous-functions-async-await-39ji) I'm leaning towards removing it to avoid unnecessary wrapping.

ZedLi: I think we discussed this before but you don't _need_ async if you're just returning a promise but I think you should. I'm of the opinion that if you're returning a promise, you should always mark the function as async for readability and to make it obvious the function is asynchronous. It also guarantees any return is a promise so you don't accidentally return a normal value and a promise in the same function. 

There shouldn't be any downside, there isn't any unnecessary wrapping since you're just returning a promise regardless, the unnecessary wrapping would occur if you returned an awaited promise (i.e. `return await asyncFunction()`)

cameronperera: That makes since for documentation purposes. In this case, I think we can lean on the doc comment `@return {Promise{ScopeModel}}` to make it clear for future engineers.

cameronperera (approved):

PR #2229, opened 2024-04-10
Split storage-bucket form at creation time

calcaide:
:tickets: [Jira ticket](https://hashicorp.atlassian.net/browse/ICU-12610)
:tickets: [Jira ticket](https://hashicorp.atlassian.net/browse/ICU-13348)
:tickets: [Jira ticket](https://hashicorp.atlassian.net/browse/ICU-13441)

## Description

IMPORTANT: 
- This PR has NOT BEEN tested (yet) with a real minio deployment (working on it). But I think is mature enough to review.
- NO CONTROLLERS tests: I will provide those in a coming PR.

Split storage-bucket form in two types: aws and minion. Add support for minio at storage-bucket creation.

## Screenshots:

Before:
Screenshot 2024-04-12 at 11 03 36 AM

After:
Screenshot 2024-04-16 at 10 44 37 AM
Screenshot 2024-04-16 at 10 44 53 AM

### Mirage
- Creation of a storage-bucket using both plugin types works.

### Boundary instance + MinIO infra
- We are working on how to spin a MinIO infra so we can trully test it.

## Checklist:


addons/core/translations/resources/en-us.yaml
```diff
@@ -435,6 +435,9 @@ storage-bucket:
     dynamic:
       title: Dynamic
       description: Specify an Assume Role used to generate credentials to use with this AWS bucket
+  plugin-types:
+    aws: Aws
```
DhariniJeeva, line 439: I believe this should be `AWS`

DhariniJeeva: Actually, figma shows `Amazon S3`

ui/admin/app/components/form/storage-bucket/index.hbs
```diff
@@ -2,455 +2,23 @@
...
+    updateScope=this.updateScope
+    scopeFieldDescription=this.scopeFieldDescription
+    scopes=@scopes
+    pluginTypes=this.pluginTypes
+    changePluginType=@changePluginType
+    changeCredentialType=@changeCredentialType
+    credentials=this.credentials
+    selectedCredentialType=this.selectedCredentialType
+    updateTypeSelection=this.updateTypeSelection
+    showDynamicCredentials=this.showDynamicCredentials
```
DhariniJeeva, line 20: Is everything shared by both aws and minIO?

addons/core/translations/resources/en-us.yaml
calcaide: @DhariniJeeva 
You are totally right, changing it! thanks for the catch! 😉

ui/admin/app/components/form/storage-bucket/index.hbs
calcaide: @DhariniJeeva 
You are right, the next one, are just need by AWS.

```
credentials=this.credentials
selectedCredentialType=this.selectedCredentialType
updateTypeSelection=this.updateTypeSelection
showDynamicCredentials=this.showDynamicCredentials
```

calcaide: @DhariniJeeva 
TBH I do believe the storage-bucket form itself needs a refactor, I think [this could be simplified](https://github.com/hashicorp/boundary-ui/blob/main/ui/admin/app/components/form/storage-bucket/index.js) and I am ok creating a jira ticket for it.

BUT if you have a solution to not pass these 4 attributes mentioned above that belongs JUST to AWS I will apply such solution now, and leavet the refactor of the component for later. 

Let me know your thoughts 😉

DhariniJeeva: I was thinking we could split the JS file and have the AWS-specific actions/getters in the AWS folder since we now have separate folders for each type. Any additional refactoring can be postponed later due to time constraints.

ui/admin/tests/acceptance/storage-buckets/create-test.js
```diff
@@ -86,7 +90,7 @@ module('Acceptance | storage-buckets | create', function (hooks) {
     assert.strictEqual(getStorageBucketCount(), storageBucketCount + 1);
   });
 
-  test('users can create a new storage bucket with org scope', async function (assert) {
+  test('users can create a new storage bucket aws plugin type with org scope', async function (assert) {
```
DhariniJeeva, line 93: Do we need a similar test case for minIO?

ui/admin/app/components/form/storage-bucket/minio/index.hbs
```diff
@@ -0,0 +1,341 @@
...
+    @value={{@model.bucket_name}}
+    @isInvalid={{@model.errors.bucket_name}}
+    disabled={{form.disabled}}
+    readOnly={{not @model.isNew}}
+    {{on 'input' (set-from-event @model 'bucket_name')}}
+    as |F|
+  >
+    <F.Label>{{t 'resources.storage-bucket.form.bucket_name.label'}}</F.Label>
+    <F.HelperText>{{t
+        'resources.storage-bucket.form.bucket_name.help'
```
ZedLi, line 154: Should these helper texts say something else since they reference AWS in the helper text?

ui/admin/app/components/form/storage-bucket/minio/index.hbs
```diff
@@ -0,0 +1,341 @@
...
+    {{/each}}
+  </Hds::Form::RadioCard::Group>
+
+  {{! endpoint_url }}
+  <Hds::Form::TextInput::Field
+    name='endpoint_url'
+    @isRequired={{true}}
+    @value={{@model.endpoint_url}}
+    @isInvalid={{@model.errors.endpoint_url}}
+    disabled={{form.disabled}}
```
ZedLi, line 124: The helper text says this field can't be updated later, should this field also be readonly like the name and region?

ui/admin/app/components/form/storage-bucket/minio/index.hbs
```diff
@@ -0,0 +1,341 @@
...
+        {{#each @model.errors.description as |error|}}
+          <E.Message>{{error.message}}</E.Message>
+        {{/each}}
+      </F.Error>
+    {{/if}}
+  </Hds::Form::Textarea::Field>
+
+  {{! scope }}
+  {{#if @model.isNew}}
+    <Hds::Form::Select::Field
```
ZedLi, line 52: It seems like the scope dropdown gets reset when you switch between the aws and minio type in the form but it's still there in the data model

ui/admin/tests/acceptance/storage-buckets/create-test.js
ZedLi, line 1: Do we have any update tests?

ui/admin/app/components/form/storage-bucket/minio/index.hbs
calcaide: @ZedLi 
Thanks for the catch. Yeah I am trying to review this part, I do believe, as per current implementation is a bit of a mess for such simple field: tracked properties, action that calls another internal method... I do believe this could be based on model value solely without need for those mentioned methods (`updateScopeFieldDescription`, `updateScope` and `scopeFieldDescription`) I will keep you posted when I push the updates.

calcaide: @ZedLi 
Once again, thanks for the catch.

I changed the behaviour of how we set the initial scope and how this component behaves.

**Initial scope:**
Previous implementation: sets scope to global at the new route file. When changing provider the route gets refreshed because we use queryParams, therefore scope is set to global overriding user previously selected scope.

Current implementation: we do not set the scope until the user specified.

**Component behaviour**
With the initial scope changes, now we know if @model.scope is not set, means the user did not choose yet. So we do not need a tracked property to choose what helper text to show (scopeFieldDescription) neither a method to update the description (updateScopeFieldDescription).

ui/admin/app/components/form/storage-bucket/minio/index.hbs
calcaide: @ZedLi 
Good catch! I am triple checking with BE. In case can't not be updated later I will make it readonly like name and region, otherwise I will change the description.

ui/admin/app/components/form/storage-bucket/minio/index.hbs
calcaide: @ZedLi 
Good catch! I think you are totally right, alsog the field region, access key id and secret access key mention AWS within the helper, I will check with design about these.

calcaide: @ZedLi 
All the helper texts that have any reference to AWS have been updated.

ui/admin/tests/acceptance/storage-buckets/create-test.js
calcaide: @ZedLi 
You are right, we have not update tests. As per this slack message and the sake of the release, I will provide these tests within a separate PR.

ui/admin/tests/acceptance/storage-buckets/create-test.js
calcaide: @DhariniJeeva 
You are right. I will add such test. As we discussed within stand up, and posted on the boundary-ui channel, [here message](https://hashicorp.slack.com/archives/C012GTH1C4E/p1713291713830339) for the sake of the release I will provide this test on a separate PR.

ui/admin/app/components/form/storage-bucket/minio/index.hbs
calcaide: @ZedLi 
After [checking with BE](https://hashicorp.slack.com/archives/C04N3MLQUBY/p1713224380077909) `endpoint_url` is indeed `readonly`. This change has been applied and now `endpoint_url` is `readonly` field

DhariniJeeva (approved):
Looks good!

PR #2233, opened 2024-04-11
Fix tests from rebase

ZedLi:
## Description
There were some broken tests after rebasing onto main. This mainly fixes the issues. There was one interesting issue that arose from this [PR](https://github.com/hashicorp/boundary-ui/pull/2214). The model was changed to now use a tracked array instead of a normal array. I had originally made the assumption we wouldn't have any ember-like arrays in our models but this is now obviously not a valid assumption so we have to handle cleaning those up before sending it to indexed db.

Most likely we will encounter a similar type of error with tracked objects in the future but none of our current models use those yet.

### DIFF
--- a/addons/api/addon/services/indexed-db.js
+++ b/addons/api/addon/services/indexed-db.js
@@ -28,17 +28,20 @@ export const formatDbName = (userId, clusterUrl) =>
   `boundary-${userId}-${clusterUrl}`;
 
 /**
- * Recursively copy over the object and convert any booleans to "true" or "false" so they can be indexed
+ * Recursively copy over the object and convert any booleans to "true" or "false" so they can be indexed.
+ * Also converts any ember-like arrays back to a standard array.
  * @param result
  * @param object
  * @returns {*}
  */
-const convertBooleansInObject = (result, object) => {
+const convertFields = (result, object) => {
   Object.entries(object).forEach(([key, value]) => {
     if (typeOf(value) === 'object') {
-      result[key] = convertBooleansInObject({}, value);
+      result[key] = convertFields({}, value);
     } else if (typeOf(value) === 'boolean') {
       result[key] = value ? 'true' : 'false';
+    } else if (typeOf(value) === 'array') {
+      result[key] = [...value];
     } else {
       result[key] = value;
     }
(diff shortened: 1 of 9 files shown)


addons/api/addon/services/indexed-db.js
```diff
@@ -48,17 +51,31 @@ const convertBooleansInObject = (result, object) => {
...
+      const attribute = schema.attributes.get(key);
+
+      // If attribute type is array, we transform it back to a tracked array
+      if (attribute?.type === 'array') {
+        const transformer = serializer.transformFor(attribute.type);
+        result[key] = transformer.deserialize(value, attribute.options);
+        return;
+      }
+
+      result[key] = value;
```
ZedLi, line 78: I could have also just made _any_ array into a tracked array which shouldn't hurt anything but is also not the exact data we would be getting back after a normal normalization.

I also didn't want to just apply a transform deserialization for any array-like object since we have a few fields that are standard arrays but just do simple transforms (e.g `string-array`, `object-array` etc.) and I would be then deserializing twice.

So unfortunately the safest way of handling this is for me to explicitly check for the new `array` type which is the only ember-like array we transform currently.

cameronperera: I think this is a good approach to this issue.

cameronperera (approved):

addons/api/addon/services/indexed-db.js
lisbet-alvarez: Agreed, I also think this is the best approach.

ui/admin/app/controllers/scopes/scope/credential-stores/index.js
```diff
@@ -90,7 +90,7 @@ export default class ScopesScopeCredentialStoresIndexController extends Controll
   @notifySuccess('notifications.delete-success')
   async delete(credentialStore) {
     await credentialStore.destroyRecord();
-    await this.router.replaceWith('scopes.scope.credential-stores');
+    this.router.replaceWith('scopes.scope.credential-stores');
```
lisbet-alvarez, line 93: hmm thx for fixing this, I guess best path forward is to not await transitions unless strictly necessary

lisbet-alvarez (approved):
🙌 Looks great!

ui/admin/app/controllers/scopes/scope/credential-stores/index.js
ZedLi: I think this was actually a mistake I had made in rebasing, I might have added back some awaits 🤔

PR #2238, opened 2024-04-12
Icu 13218 update indexed db parser unit tests to test for filtering on boolean fields

lisbet-alvarez:
:tickets: [Jira ticket](https://hashicorp.atlassian.net/browse/ICU-13218)

## Description
Silence deprecation warnings and add test cases for bool fields.

before:

image

after:

image

## How to Test

Run unit tests and make sure no deprecation warnings are shown in dev tools console

## Checklist:

### DIFF
--- a/addons/api/tests/unit/utils/indexed-db-query-test.js
+++ b/addons/api/tests/unit/utils/indexed-db-query-test.js
@@ -4,10 +4,12 @@ import { setupTest } from 'ember-qunit';
 import setupMirage from 'ember-cli-mirage/test-support/setup-mirage';
 import { setupIndexedDb } from 'api/test-support/helpers/indexed-db';
 import { pluralize } from 'ember-inflector';
+import { camelize } from '@ember/string';
 import { TYPE_TARGET_TCP, TYPE_TARGET_SSH } from 'api/models/target';
 
 const seedIndexDb = async (resource, store, indexedDb, server) => {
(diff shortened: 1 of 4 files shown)


cameronperera (approved):
LGTM

addons/api/tests/unit/utils/indexed-db-query-test.js
```diff
@@ -494,4 +496,32 @@ module('Unit | Utility | indexed-db-query', function (hooks) {
...
+
+    assert.strictEqual(result.length, 1);
+  });
+
+  test('it filters on boolean type fields with falsy value', async function (assert) {
+    const authMethod = this.server.create('auth-method', {
+      name: 'Generated_auth_method',
+      scopeId: scope1.id,
+    });
+    scope1.update({ primaryAuthMethodId: authMethod.id });
```
ZedLi, line 519: Does this actually set the `is_primary` for the auth method itself? I don't think we have anything in the config that would update the auth method model when we patch this

ZedLi (approved):
Thanks for adding more tests!

addons/api/tests/unit/utils/indexed-db-query-test.js
lisbet-alvarez: yeah, not in the config, only in the mirage [auth-method serializer ](https://github.com/hashicorp/boundary-ui/blob/main/addons/api/mirage/serializers/auth-method.js#L20)

PR #2239, opened 2024-04-12
Alias clear modal

DhariniJeeva:
:tickets: [Jira ticket](https://hashicorp.atlassian.net/browse/ICU-12928)

## Description

This PR adds a delete modal with an extra option `clear` so users can clear aliases instead of deleting it

https://github.com/hashicorp/boundary-ui/assets/15043878/8df848a7-4cc3-4d0b-952f-3c5210025f84

## How to Test

1. using boundary controller, create a few aliases 
2. After successful creation, you can see the aliases pop up in /aliases view 
3. click on the `...` icon on the list table, you should see `edit` and `delete` options
4. click on delete. The hds modal should show three options(delete, clear and cancel) if there's a target associated with the resource
5. If there's no target associated, you should only see two options - delete and cancel 

## Checklist:

### DIFF
--- a/addons/api/addon/initializers/deprecations.js
+++ b/addons/api/addon/initializers/deprecations.js
@@ -1,0 +1,27 @@
+/**
+ * Copyright (c) HashiCorp, Inc.
+ * SPDX-License-Identifier: MPL-2.0
+ */
+
+import { registerDeprecationHandler } from '@ember/debug';
+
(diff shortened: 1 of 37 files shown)


calcaide (approved):
Awesome!!! tested manually with boundary instance and everything works smooth and as expected 🥳 great work 💪 🙌

PR #2241, opened 2024-04-12
feat: 🎸 Add migration check to delete index DB when updating

ZedLi:
✅ Closes: https://hashicorp.atlassian.net/browse/ICU-13223

## Description
This PR aims to prevent any migration that might be needed in the future for indexed DB by destroying the previous DB if a version is detected. 

@hashicc was able to help figure out the original issue and it came down to us actually doing two transitions which mean we were hitting the index before indexed db was setup. 

## How to Test
Increment the version of index DB and confirm it correctly deletes previous tables and sets up a fresh instance.

## Checklist:


addons/api/addon/services/indexed-db.js
```diff
@@ -88,23 +88,46 @@ const unconvertFields = (result, object, schema, serializer) => {
...
   // =attributes
   #db;
+  #version = 1;
+
   get db() {
     return this.#db;
   }
 
-  setup(dbName) {
+  async setup(dbName) {
```
ZedLi, line 105: Before:
image

After:
image

It seems on initial login (not from an authentication restoration) some requests are firing off before index db is properly setup.

I was suspicious that the session service I override [here](https://github.com/hashicorp/boundary-ui/blob/9f11e8a25de5cb89e27b143803d8b5f91755b6da/ui/admin/app/services/session.js#L13) can't actually be async since it seems the [source code](https://ember-simple-auth.com/api/services_session.js.html) doesn't take into account if the overridable method needs to be async. 

The setup call in the service _should_ be awaited but it didn't seem to make a difference. I tried setting up indexed DB in our [authenticate method](https://github.com/hashicorp/boundary-ui/blob/bcf004b8a46af963016a382c738072479dc57d97/ui/admin/app/routes/scopes/scope/authenticate/method.js#L64) before any transition but it seems like for some reason ember still fires off requests for scopes before index DB is ready. 

I'm not sure what the solution is as it seems the moment the `setup` method is async, some of the requests for the ember routes aren't properly awaiting now.

hashicc:
Some notes related to [this comment in the pr](https://github.com/hashicorp/boundary-ui/pull/2241/files#r1563385039) where there's a race condition between getting an indexeddb setup _and then_ making calls that are use the db for caches. 

The goal is to deterministically `await` an indexeddb setup _before_ making calls with `IndexedDbHandler#request`. From reading the code and @ZedLi's comment, I think the issue specifically arises [here](https://github.com/hashicorp/boundary-ui/blob/9f11e8a25de5cb89e27b143803d8b5f91755b6da/ui/admin/app/services/session.js#L19) since as @ZedLi already mentioned even if `async` was added to `handleAuthentication` and the setup was awaited, there's [nothing in ember-simple-auth awaiting the `handleAuthentication` hook itself to finish](https://github.com/mainmatter/ember-simple-auth/blob/748e46fb67564d2e4e82bdfb5724a9fdc02f551b/packages/ember-simple-auth/src/services/session.ts#L149-L151). Although, [their docs](https://github.com/mainmatter/ember-simple-auth/blob/748e46fb67564d2e4e82bdfb5724a9fdc02f551b/guides/managing-current-user.md?plain=1#L168-L176) do use an async `handleAuthentication` example (I think they just leave that promise hanging? 🤔). 

Here are some possible solutions toe ensure that the `setup` is awaited by the time `IndexedDbHandler#request` is called. I think the first example is probably the way to go.

## 1. Instead of routing ourselves, route within `handleAuthentication`
[Described in this comment on the ember-simple-auth repo](https://github.com/mainmatter/ember-simple-auth/pull/2285#issuecomment-807876500). Also removing `this.router.transitionTo('index');` [here](ui/admin/app/controllers/scopes/scope/authenticate/method/index.js) and letting that be taken care of in the hook.

Within the hook we could either not call super and transition ourselves or call super with the route we want to go to, it does some extra work with `routeAfterAuthentication` [here](https://github.com/mainmatter/ember-simple-auth/blob/748e46fb67564d2e4e82bdfb5724a9fdc02f551b/packages/ember-simple-auth/src/-internals/routing.js#L33-L50). It seems like the intention of `routeAfterAuthentication` is to default to using the value from configuration but I can't find current documentation on the way that should be done so using the super call explicitly might be more clear.

## 2. Setup promise state stashed on IndexedDb
One way this could be solved is having the IndexedDb service have some sort of promise stashed on it that is awaited before making calls in the handler
// in IndexedDbHandler#request
await this.indexedDb.isSetup // is a promise that resolves when `setup` is finished
const { db: indexedDb } = this.indexedDb ?? {};
That's kind of ugly since it forces the await in the request. 

## 3. Chase all root `transitionTo` and add `await indexeddb.setup()` before
The other thing is that the setup could be pushed to right after this.session.authenticate succeeds but before this.router.transitionTo('index');. I'm not sure about the OIDC case though, when it just gets redirected, is the session authenticated before the indexeddb would setup is called in the application route? Sprinkling the setup calls (maybe only in a few spots) before the transitions is not great because it's a bit hard to track.

## 4. Chase all root model hooks and include a `beforeModel` with  `await indexeddb.setup()`
Find all routes that would make a call to `IndexedDbHandler#request` in routes and making sure that `await indexeddb.setup()` is called first. This ends up being boilerplate that will probably be forgotten about. 

## 5. `authenticated` or `protected` "invisible" route
* "Invisible" because it is not part of the "url" but *is* part of the route call order
* This acts as route guard to ensure setup for any child routes in a `beforeModel`, similar to what is above but is enforced by ember
```js
// router.js
// top-level route layer that "hosts" all authenticated child routes
this.route('authenticated', { resetNamespace: true }, function () {
  // any routes that need "authenticated" `beforeModel` checks and are expected to be authenticated
  // move all existing child routes into here
});
```

From a URL level nothing changes but from a route name change this can be annoying, all migrated routes that are referenced by route name now need to be added with an `authenticated`: `transitionTo("account.change-password")` becomes `transitionTo("authenticated.account-change-password")`

```js
class AuthenticatedRoute {
	async beforeModel() {
		// Using the `AuthenticatedRoute` removes the need to have this
		// sprinkled across the application since it's a pre-req for
		// entering this layer of routes and its child routes

		// enforce that this is setup before entering child routes
		await indexeddb.setup();
	}
}
```
I believe once you're in this layer it's hooks don't get called again unless something like `ApplicationRoute.refresh()` gets called and recalls its hooks and all child route's hooks recursively.

hashicc:
Here's a [branch](https://github.com/hashicorp/boundary-ui/compare/hashicc/migration-check-delete-db-when-updating?expand=1) based on suggestion 1 above

addons/api/addon/services/indexed-db.js
ZedLi: It seems the main issue was now that the indexedDb setup was async, we were actually hitting this [transition](https://github.com/hashicorp/boundary-ui/blob/0af81d9bc1354f632b6199ef660232d57d0f19ab/ui/admin/app/controllers/scopes/scope/authenticate/method/index.js#L50) before it finished.

In fact, as it turns out, the `handleAuthentication` event provided by ember-simple-auth actually bakes in [transitioning for you](https://github.com/mainmatter/ember-simple-auth/blob/483d4786ff3aedfda19234246da4bfed63f25368/packages/ember-simple-auth/src/-internals/routing.js#L48) so we were doing it twice. 

OIDC was in fact already depending on this transition so it makes sense to have it just be handled in `handleAuthentication`.

ui/admin/app/services/session.js
```diff
@@ -15,13 +15,15 @@ export default class SessionService extends BaseSessionService {
...
+  async handleAuthentication() {
     const userId = this.data?.authenticated?.user_id;
     const hostUrl = this.window.location.host;
     if (userId && hostUrl) {
-      this.indexedDb.setup(formatDbName(userId, hostUrl));
+      await this.indexedDb.setup(formatDbName(userId, hostUrl));
     }
+
+    // We let ember-simple-auth handle transitioning back to the index after authentication.
+    // This route can be configured in our environment config.
```
ZedLi, line 26: I didn't find much documentation about this but you can configure where it will redirect by setting an object in our environment config like this (the [default](https://github.com/mainmatter/ember-simple-auth/blob/483d4786ff3aedfda19234246da4bfed63f25368/packages/ember-simple-auth/src/configuration.js#L3) is `index`):

```
let ENV = {
    'ember-simple-auth': {
      routeAfterAuthentication: 'example.route',
    }
}
```

hashicc: Thanks for chasing that down

hashicc (approved):
One minor refactor suggestion! Otherwise everything looks great. Tested the indexed-db code locally with no db, existing db, and db upgrade, and everything worked 🚀

cameronperera (approved):
This looks great!

addons/api/addon/services/indexed-db.js
```diff
@@ -97,23 +97,45 @@ const unconvertFields = (result, object, schema, serializer) => {
...
     // Don't run setup again if we already have one or if we didn't get a name
     if (this.#db || !dbName) {
       return;
     }
 
+    // If the database doesn't exist, just create it and skip checking version
```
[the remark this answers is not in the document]
ZedLi: Great suggestion, looks much cleaner!

cameronperera (approved):

PR #2245, opened 2024-04-15
fix: 🐛 Fix windows being unable to search client daemon

ZedLi:
✅ Closes: https://hashicorp.atlassian.net/browse/ICU-13466

## Description
When using the client daemon to search on windows, we use the CLI instead of the unix socket for our requests. Unfortunately that means we spawn a separate process to retrive the results and one idiosyncracy of this is that `spawnSync` buffers its outputs which means we were hitting the default max buffer size. `spawn` doesn't have this issue because it streams its outputs and we didn't previously use `spawnSync` for any command that produced large amounts of output.

The node default is 1MB and ~1000 targets will end up being around there from my quick testing. Because we weren't getting the full result set, `jsonify` would silently error out trying to parse the output and we would fall back to always calling the controller.

## How to Test
Setup boundary dev with a couple thousand targets. Either remove the if check to only use the CLI for windows or make a windows build. Confirm that we're not making any extra network requests or that `jsonify` isn't erroring out. 

## Checklist:


ui/desktop/electron-app/src/helpers/spawn-promise.js
```diff
@@ -78,6 +78,12 @@ module.exports = {
    * @returns {{stdout: string | undefined, stderr: string | undefined}}   */
   spawnSync(args, envVars = {}) {
     const childProcess = spawnSync(path(), args, {
+      // Some of our outputs (namely client daemon searching) can be very large.
+      // This an undocumented hack to allow for an unlimited buffer size which
+      // could change at any time. If it does, we should just set an arbitrarily
+      // large buffer size or switch to spawn and stream the output and handle
+      // killing the process ourselves.
+      maxBuffer: undefined,
```
ZedLi, line 86: Do people prefer I just set an arbitrarily large buffer size instead rather than use an undocumented behavior? If so, what should I set it to?

ui/desktop/electron-app/src/services/client-daemon-manager.js
```diff
@@ -56,12 +55,18 @@ class ClientDaemonManager {
...
     }
 
-    const stopDaemonCommand = ['daemon', 'stop'];
-    spawnSync(stopDaemonCommand);
+    // Kill the process if it's still running
+    if (this.#clientDaemonProcess && !this.#clientDaemonProcess.killed) {
+      isWindows()
+        ? treeKill(this.#clientDaemonProcess.pid)
+        : this.#clientDaemonProcess.kill();
+    }
```
ZedLi, line 69: I realized we likely weren't properly killing the process when starting up the client daemon 😓

lisbet-alvarez (approved):
Great catch! Tested locally, and I am no longer seeing any extra network requests!

ui/desktop/electron-app/src/helpers/spawn-promise.js
lisbet-alvarez: I think that for now, this hack is good but perhaps we can revisit this topic during techtime?

PR #2248, opened 2024-04-16
MinIO support for Admin UI

calcaide:

## Description

Add MinIO support for CRUD operation within Admin UI.

IMPORTANT: The creation has NOT been tested with a real Boundary controller (WIP).

Be aware this PR is missing:
- Acceptance tests for Update operation.
- Accepntace test for creation with MinIO under Org scope (we have just for global).

### Before:
before-list

before-dynamic

before-static

### After:

after-list

after-static

after-dynamic

after-minio

## How to Test

## Checklist:


ui/admin/app/routes/scopes/scope/storage-buckets/new.js
```diff
@@ -71,7 +79,6 @@ export default class ScopesScopeStorageBucketsNewRoute extends Route {
       region,
       worker_filter,
     });
-    record.scopeModel = scopeModel;
```
lisbet-alvarez, line 74: non-blocking comment: since this is removed, now if we save without choosing a scope and navigate to storage-buckets table list view, the page errors out (using mirage). I'm assuming this won't be an issue when running against a boundary instance since API will probably not let user save if they haven't chosen a scope.

lisbet-alvarez (approved):
🥳 Awesome work!!!

DhariniJeeva (approved):
🎉  looks amazing!! great job

ui/admin/app/routes/scopes/scope/storage-buckets/new.js
calcaide: @lisbet-alvarez 

Very good point. You are totally correct, under a real usage case, the user should NOT be able to create a storage-bucket without selecting a scope.

But under mirage, nothing stops you from creating a storage-bucket without scope because there is no validation.

I think this is worth to discussion and I will bring it for the next tech time.

PR #2250, opened 2024-04-17
fix: 🐛 transition aborted error when deleting group

cameronperera:
## Description

fix: 🐛 transition aborted error when deleting group

Before:
Screenshot 2024-04-17 at 10 36 34 AM

After:
https://github.com/hashicorp/boundary-ui/assets/29386339/fd3240fd-62a1-49d5-81b8-bf3b41dbc193

## How to Test

Start a boundary instance, create a group, then delete. You should see no errors.

## Checklist:

### DIFF
--- a/ui/admin/app/controllers/scopes/scope/groups/index.js
+++ b/ui/admin/app/controllers/scopes/scope/groups/index.js
@@ -69,7 +69,7 @@ export default class ScopesScopeGroupsIndexController extends Controller {
   @notifySuccess('notifications.delete-success')
   async delete(group) {
     await group.destroyRecord();
-    await this.router.replaceWith('scopes.scope.groups');
+    this.router.replaceWith('scopes.scope.groups');
     await this.router.refresh();
   }
 


ui/admin/app/controllers/scopes/scope/groups/index.js
```diff
@@ -69,7 +69,7 @@ export default class ScopesScopeGroupsIndexController extends Controller {
   @notifySuccess('notifications.delete-success')
   async delete(group) {
     await group.destroyRecord();
-    await this.router.replaceWith('scopes.scope.groups');
+    this.router.replaceWith('scopes.scope.groups');
```
moduli, line 72: Disclaimer: I know nothing about this stuff

Looking at other instances of this `async delete(...)`, I'm not seeing a consistent pattern. Some have an `await` here, some don't have a refresh, etc. Are they supposed to be consistent or are there intricacies depending on the resource?

cameronperera: I have done some testing and I am baffled as well. We have had this pop up in a few places but like you said, it is inconsistent. I thought it might be the combo of `this.router.replaceWith()` and `this.router.refresh()` but that doesn't seem to be true either. It is something we are keeping our eyes on. Thank you for finding this one.

moduli (approved):

PR #2253, opened 2024-04-17
Add minio missing unit tests

calcaide:
## Description

This PR will add missing tests to the [recently merged MinIO support](https://github.com/hashicorp/boundary-ui/pull/2248). 

Test this PR provides:
- Acceptance test to create a new storage bucket MinIO plugin type with org scope. [Mentioned in this comment](https://github.com/hashicorp/boundary-ui/pull/2229#discussion_r1566255677)
- Update tests for MinIO plugin type. [Mentioned in this comment](https://github.com/hashicorp/boundary-ui/pull/2229#discussion_r1566320930)

## How to Test

Run the test suite with `yarn test` within `boundary-ui/ui/admin` directory.

## Checklist:


ui/admin/app/components/form/storage-bucket/minio/index.hbs
```diff
@@ -166,6 +166,7 @@
   {{! bucket_name }}
   <Hds::Form::TextInput::Field
     name='bucket_name'
+    @id='bucket_name'
```
calcaide, line 169: The usage of ID is to could use standard selectors also for error messages.

Once we assign attribute `@id`, such id will be used within the `<input id='bucket_name'>` but also it will be added to the error message as follows: `error-bucket_name`.  This pattern allows using a standard and descriptive selector to select a specific field and its error message.

cameronperera: nice find! That should make tests against error states for our inputs much easier.

cameronperera (approved):

ui/admin/app/components/form/storage-bucket/minio/index.hbs
ZedLi: Should we move away from using IDs after the discussion in https://github.com/hashicorp/boundary-ui/pull/2278?

calcaide: @ZedLi 
Ups!! totally forgot in this PR I was using those too!!
Sure!! as per discussion on #2278 I will proceed to rely on `data-test-*` for this PR, I will require a review from you and @cameronperera as soon I got that in 😉 
Thanks for the feedback and sorry for the back/forth.

ZedLi: No problem!! Just wanted to make sure we were on the same page

ui/admin/app/components/form/storage-bucket/aws/index.hbs
```diff
@@ -140,28 +140,30 @@
...
-      }}</F.HelperText>
-    {{#if @model.errors.bucket_name}}
-      <F.Error as |E|>
-        {{#each @model.errors.bucket_name as |error|}}
-          <E.Message>{{error.message}}</E.Message>
-        {{/each}}
-      </F.Error>
-    {{/if}}
-  </Hds::Form::TextInput::Field>
+  <div data-test-bucket-name>
```
calcaide, line 143: @ZedLi @cameronperera 

I had to add this wrapper. If you add the `data-test-bucket-name` as attribute of `<Hds::Form::TextInput::Field>`, when it renders it just adds the `data-test-bucket-name` attribute within the input, making impossible to use the `data-test` to select the error message or other elements within the `Hds::Form::TextInput::Field`.

Let me know your thoughts

ui/admin/app/components/form/storage-bucket/minio/index.hbs
calcaide: @ZedLi 
I applied changes 😉

ui/admin/app/components/form/storage-bucket/aws/index.hbs
cameronperera: Interesting. I think this is a fine approach since we only have to do it for inputs that we need to test error states. Other inputs, we can continue to rely on the `name` attribute to select.

cameronperera (approved):

ui/admin/app/components/form/storage-bucket/aws/index.hbs
ZedLi: If we're just trying to select the error messages, is there a reason we couldn't have added the data test attribute on the `F.error`?

cameronperera: That is a great point. Just add it to the error which keeps it closer to the thing we want to test. I retract my previous statement.

calcaide: @ZedLi @cameronperera 

For the uses cases I have seen within the PoC of sharing selectors, the idea behind this wrapper is not just to make the error message selectable, but any element within the component that wraps.

That said, maybe I was trying to solve a problem that we do not have right here. So I will proceed with the changes of adding the `data-test-` directly to the error message, and when we will see how we solve future problems.

cameronperera (approved):

PR #2254, opened 2024-04-17
chore(e2e): Update playwright to 1.43

moduli:
When running the Admin UI e2e tests against 0.16.0, there were some issues with trying to click on a link on the new pagination tables (links were not seen as visible). For some reason, updating the playwright version fixed this.

Steps to upgrade
- Updated `ui/admin/package.json` and `/ui/desktop/package.json` to the new version
- In the root directory, ran `yarn install`


ui/desktop/package.json
```diff
@@ -104,7 +104,7 @@
     "eslint-plugin-qunit": "^8.0.1",
     "js-bexpr": "hashicorp/js-bexpr#9b4a4b54d85eba67fdfc0990133d1518d890b1e1",
     "loader.js": "^4.7.0",
-    "playwright": "^1.30.0",
+    "playwright": "^1.43.0",
```
moduli, line 107: Reading about this, I'm wondering if this package is unnecessary?

https://playwright.dev/docs/library#:~:text=Introduction%E2%80%8B,end%20Test%20Runner%20and%20experience.

calcaide: This is a very good point.

Within ui/desktop e2e testing is a bit different than on ui/admin. Desktop relies on electron to wrap the UI as app. Therefore, the API's need to interact with the different electron internals are specific and you need to actually import `electron` from Playwright.

I am pretty sure this is a leftover of the update we gave playwright when you introduced to Admin. When we introduced Playwright to the project was for Desktop e2e and since electron support is experimental for Playwright they stated on their documentation to use `playwright` and not `playwright/test`, [you can see it here](https://playwright.dev/docs/api/class-electron).

BUT when you introduced Admin e2e I remember we had an issue with naming collisions when importing Playwright. We changed the Desktop e2e test to use `Playwright` instead of `Playwright/test`, so we can delete Playwright and keep `Playwright/test`.

Feel free to add such change on this PR or let me know and I will take care of the deletion of `playwright`.

Note: the desktop e2e test have not been run in ages, plus the only tests there were a PoC, so Admin UI has all the priority.

moduli: Chatted with Carlos. We're going to merge as-is and do any cleanup when we do work on the desktop client.

PR #2256, opened 2024-04-18
feat: 🎸 Add aliases to list view

ZedLi:
✅ Closes: https://hashicorp.atlassian.net/browse/ICU-13487

## Description
Add aliases to list view. 

This needs to also be tested on an older version of the CLI where aliases aren't in the client daemon to make sure it's working correctly.

image

## Checklist:

### DIFF
--- a/addons/api/addon/models/target.js
+++ b/addons/api/addon/models/target.js
@@ -135,6 +135,12 @@ export default class TargetModel extends GeneratedTargetModel {
     return Boolean(pendingOrActiveSessions.length);
   }
 
+  get aliases() {
+    return this.store
+      .peekAll('alias')
+      .filter((alias) => alias?.destination_id === this.id);
+  }
+
   // =methods
 
   /**
--- a/addons/api/addon/models/target.js
+++ b/addons/api/addon/models/target.js
@@ -135,7 +135,7 @@ export default class TargetModel extends GeneratedTargetModel {
     return Boolean(pendingOrActiveSessions.length);
   }
 
-  get aliases() {
+  get associatedAliases() {
     return this.store
       .peekAll('alias')
       .filter((alias) => alias?.destination_id === this.id);
--- a/ui/desktop/app/routes/scopes/scope/projects/targets/index.js
+++ b/ui/desktop/app/routes/scopes/scope/projects/targets/index.js
@@ -104,7 +104,10 @@ export default class ScopesScopeProjectsTargetsIndexRoute extends Route {
     const allTargetsPromise = !this.allTargets
       ? this.makeAllTargetsQuery(orgScope, orgFilter)
       : Promise.resolve();
-    const sessionsPromise = this.makeSessionQuery(orgScope, scopes, orgFilter);
+
+    const sessions = await this.makeSessionQuery(orgScope, scopes, orgFilter);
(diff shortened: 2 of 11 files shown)


ui/desktop/app/routes/scopes/scope/projects/targets/index.js
```diff
@@ -134,14 +116,30 @@ export default class ScopesScopeProjectsTargetsIndexRoute extends Route {
...
-    let targets = await this.store.query('target', query);
-    const totalItems = targets.meta?.totalItems;
+    const targetsPromise = this.store.query('target', query);
+
+    let [targets, sessions, allTargets] = await Promise.all([
+      targetsPromise,
+      sessionsPromise,
+      allTargetsPromise,
+      aliasPromise,
+    ]);
```
ZedLi, line 126: I collected up all the promises to execute them in parallel rather than us doing them sequentially, it seemed like we were starting to have quite a few requests which adds up:

This is the time it takes for an already cached client daemon for a load with an all targets request and one without:
Screenshot 2024-04-19 at 11 24 47 AM

And this is the same setup but with the promises running in parallel:
Screenshot 2024-04-19 at 11 22 50 AM

addons/api/addon/models/target.js
```diff
@@ -135,6 +135,12 @@ export default class TargetModel extends GeneratedTargetModel {
     return Boolean(pendingOrActiveSessions.length);
   }
 
+  get associatedAliases() {
+    return this.store
+      .peekAll('alias')
```
DhariniJeeva, line 140: Does this name here refer to the assignment we did in L30, or is it the actual resource name in the store?

ZedLi: It's the name of the resources in the stores, so `alias` models

DhariniJeeva (approved):
Looks great! thanks for the work

calcaide:
None blocking comment: Could we add some labels to the targets created in mirage? at the moment, form the aproximetly 25 targets we create with mirage just 1 has labels assigned. Would be nice to have a bit more, so when using mirage you see some labels.

calcaide (approved):
Awesome!! thanks for the work!!

ZedLi:
> None blocking comment: Could we add some labels to the targets created in mirage? at the moment, form the aproximetly 25 targets we create with mirage just 1 has labels assigned. Would be nice to have a bit more, so when using mirage you see some labels.

What are you referring to when you say "labels" here?

calcaide:
> > None blocking comment: Could we add some labels to the targets created in mirage? at the moment, form the aproximetly 25 targets we create with mirage just 1 has labels assigned. Would be nice to have a bit more, so when using mirage you see some labels.
> 
> What are you referring to when you say "labels" here?

I apologize, I wanted to say Aliases, but my brain keeps taking "Labels" 😅

lisbet-alvarez (approved):
Tested locally, everything looks good! 🙌

ZedLi:
> I apologize, I wanted to say Aliases, but my brain keeps taking "Labels" 😅

Ahh I see that makes sense, yeah I think we can do that. I don't normally use mirage with DC so didn't think about that but I think mirage can be adjusted in a separate PR.

PR #2257, opened 2024-04-18
fix: 🐛 aliasing bugs

DhariniJeeva:
:tickets: [Jira ticket](https://hashicorp.atlassian.net/browse/ICU-13525)

## Description

This pr fixes a few things - 
1. Hides alias sidebar container when there are no associated aliases in DC 
2. Hides target ID from alias form creation/update within a target
3. Adds confirmation messages for delete and clear aliases 

 read-only in new form
t1

editable in edit state
t2

## How to Test

## Checklist:

### DIFF
--- a/addons/core/translations/en-us.yaml
+++ b/addons/core/translations/en-us.yaml
@@ -60,6 +60,7 @@ questions:
   remove-confirm: Are you sure you want to remove this association?
   abandon-confirm: Are you sure you want to navigate away?  Your unsaved changes will be lost.
   retry-confirm: Would you like to retry?
+  clear-confirm: Are you sure you want to clear the resource id?
 themes:
   system: System Default
   light: Light
--- a/ui/admin/app/controllers/scopes/scope/aliases/index.js
+++ b/ui/admin/app/controllers/scopes/scope/aliases/index.js
@@ -18,7 +18,6 @@ export default class ScopesScopeAliasesIndexController extends Controller {
   // =attributes
   @tracked showModal = false;
   @tracked selectedAlias;
-  @tracked hasDestinationId;
   // =actions
 
   /**
@@ -57,9 +56,6 @@ export default class ScopesScopeAliasesIndexController extends Controller {
    */
   @action
   toggleModal(alias) {
-    if (alias?.destination_id) {
-      this.hasDestinationId = true;
-    }
     this.showModal = !this.showModal;
     this.selectedAlias = alias;
   }
@@ -73,11 +69,10 @@ export default class ScopesScopeAliasesIndexController extends Controller {
   @notifyError(({ message }) => message, { catch: true })
   @notifySuccess('notifications.clear-success')
   async clearAlias(alias) {
+    this.showModal = false;
     alias.destination_id = '';
-    this.hasDestinationId = false;
     await alias.save();
     await this.router.refresh();
-    this.showModal = false;
   }
 
   /**
@@ -91,6 +86,7 @@ export default class ScopesScopeAliasesIndexController extends Controller {
   async deleteAlias(alias) {
     await alias.destroyRecord();
     this.showModal = false;
+    await this.router.replaceWith('scopes.scope.aliases');
     await this.router.refresh();
   }
 }
--- a/ui/admin/app/controllers/scopes/scope/aliases/index.js
+++ b/ui/admin/app/controllers/scopes/scope/aliases/index.js
@@ -8,16 +8,13 @@ import { inject as service } from '@ember/service';
 import { action } from '@ember/object';
 import { loading } from 'ember-loading';
 import { notifySuccess, notifyError } from 'core/decorators/notify';
-import { tracked } from '@glimmer/tracking';
+import { confirm } from 'core/decorators/confirm';
 
 export default class ScopesScopeAliasesIndexController extends Controller {
   // =services
   @service can;
   @service router;
 
-  // =attributes
-  @tracked showModal = false;
-  @tracked selectedAlias;
   // =actions
 
   /**
@@ -52,24 +49,15 @@ export default class ScopesScopeAliasesIndexController extends Controller {
   }
 
   /**
-   * Handles modal visibility and sets the destination_id to a tracked variable so we can handle the clear button visibility on the modal
-   */
-  @action
-  toggleModal(alias) {
-    this.showModal = !this.showModal;
-    this.selectedAlias = alias;
-  }
-
-  /**
    * Remove destaination_id from alias
    * @param {AliasModel} alias
    */
   @action
   @loading
+  @confirm('questions.clear-confirm')
   @notifyError(({ message }) => message, { catch: true })
   @notifySuccess('notifications.clear-success')
   async clearAlias(alias) {
-    this.showModal = false;
     alias.destination_id = '';
     await alias.save();
     await this.router.refresh();
@@ -81,11 +69,11 @@ export default class ScopesScopeAliasesIndexController extends Controller {
    */
   @action
   @loading
+  @confirm('resources.alias.messages.delete')
   @notifyError(({ message }) => message, { catch: true })
   @notifySuccess('notifications.delete-success')
   async deleteAlias(alias) {
     await alias.destroyRecord();
-    this.showModal = false;
     await this.router.replaceWith('scopes.scope.aliases');
     await this.router.refresh();
   }
--- a/ui/admin/app/templates/scopes/scope/aliases/alias/index.hbs
+++ b/ui/admin/app/templates/scopes/scope/aliases/alias/index.hbs
@@ -23,12 +23,76 @@
       </nav.link>
     </Rose::Nav::Tabs>
   </page.navigation>
-
+  {{#if (can 'delete model' @model)}}
+    <page.actions>
+      <Hds::Dropdown as |dd|>
+        <dd.ToggleButton
+          data-test-manage-dropdown
+          @text={{t 'actions.manage'}}
+          @color='secondary'
+        />
+        <dd.Interactive
+          data-test-manage-alias-dropdown-delete
+          @color='critical'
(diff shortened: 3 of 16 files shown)


ui/admin/app/templates/scopes/scope/targets/target/manage-alias.hbs
```diff
@@ -48,6 +48,7 @@
   <page.body>
     <Form::Alias
       @model={{@model}}
+      @hideTargetId='true'
```
DhariniJeeva, line 51: waiting on feedback for this, not sure if we should still hide the target id in edit state

DhariniJeeva: this has been changed to readonly in new and editable in edit view, see [thread](https://hashicorp.slack.com/archives/C012GTH1C4E/p1713474813278209) here

addons/core/translations/en-us.yaml
```diff
@@ -60,6 +60,7 @@ questions:
   remove-confirm: Are you sure you want to remove this association?
   abandon-confirm: Are you sure you want to navigate away?  Your unsaved changes will be lost.
   retry-confirm: Would you like to retry?
+  clear-confirm: Are you sure you want to clear the resource id?
```
ZedLi, line 63: I get a second modal that pops up for a fraction of a second when I hit clear:


https://github.com/hashicorp/boundary-ui/assets/5783847/b13d3cb7-6a1a-4d69-8af6-8934e8b26ee2

ZedLi (changes requested):
Was it also intentional to not have a manage dropdown in the alias details page? I can only delete from the list view

DhariniJeeva:
> Was it also intentional to not have a manage dropdown in the alias details page? I can only delete from the list view

ah I missed that, will add

addons/core/translations/en-us.yaml
DhariniJeeva: thanks! updated this

ui/admin/app/templates/scopes/scope/aliases/alias/index.hbs
```diff
@@ -23,12 +23,76 @@
...
+        <dd.ToggleButton
+          data-test-manage-dropdown
+          @text={{t 'actions.manage'}}
+          @color='secondary'
+        />
+        <dd.Interactive
+          data-test-manage-alias-dropdown-delete
+          @color='critical'
+          @text={{t 'actions.delete'}}
+          {{on 'click' (fn this.aliases.toggleModal @model)}}
```
ZedLi, line 38: Hmm the one from managed targets has two dropdown options for both clear and delete, how come these ones are different and both options are in a modal when you "delete"?

ui/admin/app/templates/scopes/scope/aliases/alias/index.hbs
```diff
@@ -23,12 +23,76 @@
       </nav.link>
     </Rose::Nav::Tabs>
   </page.navigation>
-
+  {{#if (can 'delete model' @model)}}
```
ZedLi, line 26: Do you need delete permissions if you just want to clear?

ui/admin/app/controllers/scopes/scope/aliases/index.js
```diff
@@ -73,11 +69,10 @@ export default class ScopesScopeAliasesIndexController extends Controller {
   @notifyError(({ message }) => message, { catch: true })
   @notifySuccess('notifications.clear-success')
   async clearAlias(alias) {
```
ZedLi, line 71: I notice we have a route action `clearAlias` and `deleteAlias` as well as controller ones? Did we need multiple versions that do almost the same thing?

DhariniJeeva: yes we need as the other ones are for the aliases within target route

ZedLi: We couldn't just share it? Since you can inject controllers?

DhariniJeeva: Can we share controllers from one route to another? 🤔

ZedLi: I think technically yes but we can talk about this offline in tech time, this isn't blocking

ui/admin/app/templates/scopes/scope/aliases/alias/index.hbs
DhariniJeeva: uh... I thought we did, checking with BE

calcaide:
I tested deleting an alias from Aliases list and from the targets > target detail > Alias side bar > manage > Alias detail> manage > delete and the confirmation modals are different, the design and the content. Is this expected?

Modal from Aliases domain:
Screenshot 2024-04-19 at 2 04 50 PM

Modal from Targets domain:
Screenshot 2024-04-19 at 2 04 57 PM

DhariniJeeva:
> I tested deleting an alias from Aliases domain (list and alias domain) and from the targets domain, targets list > target detail > Alias side bar > manage > Alias detail> manage > delete and the confirmation modals are different, the design and the content. Is this expected?
> 
> Modal from Aliases domain: Screenshot 2024-04-19 at 2 04 50 PM
> 
> Modal from Targets domain: Screenshot 2024-04-19 at 2 04 57 PM

yes we will be getting rid of the modal due to permission issues, see this thread [here](https://hashicorp.slack.com/archives/C06NK4E647J/p1713547552640849) for more details.

ui/admin/app/templates/scopes/scope/aliases/alias/index.hbs
```diff
@@ -23,12 +23,76 @@
...
+    <page.actions>
+      <Hds::Dropdown as |dd|>
+        <dd.ToggleButton
+          data-test-manage-dropdown
+          @text={{t 'actions.manage'}}
+          @color='secondary'
+        />
+        <dd.Interactive
+          data-test-manage-alias-dropdown-delete
```
calcaide, line 35: None blocking comment: are we planning on using these `data-test-*` selectors for tests?

DhariniJeeva: thanks, we don't need this, removed

ui/admin/app/templates/scopes/scope/aliases/alias/index.hbs
DhariniJeeva: discussed offline, we don't need del permissions to clear, we removed the modal and listed the actions now

ui/admin/app/templates/scopes/scope/aliases/index.hbs
```diff
@@ -48,11 +48,11 @@
                   @route='scopes.scope.aliases.alias'
                   @model={{B.data.alias.id}}
                 >
-                  {{B.data.alias.displayName}}
+                  {{B.data.alias.value}}
```
DhariniJeeva, line 51: this should be value instead of name

DhariniJeeva: this is based on design/product [feedback](https://hashicorp.slack.com/archives/C012GTH1C4E/p1713560858991069). Value is a required field and we show that in the table as it makes more sense than ID for aliases

cameronperera (approved):
LGTM

ui/admin/app/templates/scopes/scope/aliases/alias/index.hbs
```diff
@@ -23,7 +23,36 @@
...
+      canRead=(can 'read model' @model) canDelete=(can 'delete model' @model)
+    )
+    as |perms|
+  }}
+    <page.actions>
 
+      {{#if (or perms.canRead perms.canDelete)}}
+        <Hds::Dropdown as |dd|>
+          <dd.ToggleButton @text={{t 'actions.manage'}} @color='secondary' />
+          {{#if (and perms.canRead @model.destination_id)}}
```
ZedLi, line 37: Shouldn't we be checking update permissions?

calcaide (approved):
LGTM, great bugs fixing 🧰 🙌

ui/admin/app/templates/scopes/scope/aliases/alias/index.hbs
DhariniJeeva: actually I wasn't sure, our `canUpdate` is bundled with the `canSave` in base [model](https://github.com/hashicorp/boundary-ui/blob/f65f483611105ac2735fda075a4184a77cda293d/addons/api/addon/abilities/model.js#L46) ability and we normally check for `read` before letting the user click on a resource to edit.  so I checked for canRead.  But in this case, the user wouldn't hit save so we should check for `canUpdate`?

ZedLi: Right, normally we'd use `canSave` from edit view but in this case you could actually update it without read permissions straight from the list view, you could just have edit permissions for whatever reason and just clear it without even reading the alias. We should obviously test this but I would expect us to be just checking `canUpdate`

ui/admin/tests/acceptance/aliases/update-test.js
```diff
@@ -78,4 +93,21 @@ module('Acceptance | aliases | update', function (hooks) {
     assert.dom(NAME_FIELD_SELECTOR).hasValue(`${name}`);
     assert.strictEqual(instances.alias.name, name);
   });
+
+  test('users have the option to clear an alias', async function (assert) {
```
ZedLi, line 97: It probably would be good to have a quick test that the user can't clear without update permissions but this is not blocking

ZedLi (approved):
Looks great! Thanks for the fixes

ui/admin/tests/acceptance/aliases/update-test.js
DhariniJeeva: agree, added!

PR #2258, opened 2024-04-19
test(e2e): Fixes for 0.16.0

moduli:
This PR makes some adjustments to the Admin UI e2e test suite based on changes for 0.16.0.
- The "New Storage Bucket" page no longer defaults to a `global` scope, so we now have to specify a scope. The test now creates the bucket within the created org scope, and the respective changes 
- Multiple tests were modified to create a target with an address associated with it to reduce the number of actions needed. We still have one test that creates a target and associated a host source to it.
- The Sessions page no longer shows the Refresh button in its empty state. Needed to modify the logic of `waitForSessionToBeVisible` accordingly
- Some link locators were modified (removed a `.getByRole('article')`)

### DIFF
--- a/ui/admin/tests/e2e/helpers/boundary-cli.js
+++ b/ui/admin/tests/e2e/helpers/boundary-cli.js
@@ -626,7 +626,7 @@ exports.waitForSessionRecordingCli = async (storageBucketId) => {
   do {
     i = i + 1;
     const sessionRecordings = JSON.parse(
-      execSync('boundary session-recordings list -format json'),
+      execSync('boundary session-recordings list --recursive -format json'),
     );
     filteredSessionRecording = sessionRecordings.items.filter(
       (obj) =>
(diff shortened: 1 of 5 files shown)


stasryzhov (approved):

calcaide:
Nice work!!

I have to apologize for this:

`The "New Storage Bucket" page no longer defaults to a global scope, so we now have to specify a scope. The test now creates the bucket within the created org scope, and the respective changes`

Is an implementation change we took recently and I could give a heads up 🙏

calcaide (approved):

moduli:
> Nice work!!
> 
> I have to apologize for this:
> 
> `The "New Storage Bucket" page no longer defaults to a global scope, so we now have to specify a scope. The test now creates the bucket within the created org scope, and the respective changes`
> 
> Is an implementation change we took recently and I could give a heads up 🙏

No need to apologize. Since we don't have this hooked up in any automated way, yet, we're going to run across things like this.

PR #2260, opened 2024-04-19
chore: 🤖 add doc links for aliases

DhariniJeeva:
## Description

Adds doc links for aliases

see [PR](https://github.com/hashicorp/web/pull/1452) which adds the redirects

sc
scr

## How to Test

## Checklist:

- ~[ ] I have added tests that prove my fix is effective or that my feature works

### DIFF
--- a/ui/admin/app/templates/scopes/scope/aliases/index.hbs
+++ b/ui/admin/app/templates/scopes/scope/aliases/index.hbs
@@ -11,10 +11,11 @@
   <page.header>
     <h2>
       {{t 'resources.alias.title_plural'}}
-      {{! TODO: Add doc link for alias }}
+      <DocLink @doc='aliases' @iconSize='large' />
     </h2>
     <p>{{t 'resources.alias.description'}}</p>
   </page.header>
+
   <page.actions>
     {{#if (can 'create model' this.scope collection='aliases')}}
       {{#if @model}}
--- a/ui/admin/app/templates/scopes/scope/aliases/new.hbs
+++ b/ui/admin/app/templates/scopes/scope/aliases/new.hbs
@@ -17,10 +17,9 @@
   <page.header>
     <h2>
       {{t 'resources.alias.titles.new'}}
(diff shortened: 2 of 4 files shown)


ui/admin/app/templates/scopes/scope/aliases/index.hbs
```diff
@@ -11,10 +11,11 @@
   <page.header>
     <h2>
       {{t 'resources.alias.title_plural'}}
-      {{! TODO: Add doc link for alias }}
+      <DocLink @doc='aliases' @iconSize='large' />
```
calcaide, line 14: None blocking comment: if possible, I will suggest to use HDS link instead of our custom `<DocLink>`. [Here there is an example](https://github.com/hashicorp/boundary-ui/pull/2037/files#diff-d0edf5ba3ad67a92b84cd9ef362c648a9b07cb0e25a434c5b9f425a90d726fe2R13-R15) of how to use it. 
Let me know your thoughts 😉

calcaide (approved):

ui/admin/app/templates/scopes/scope/aliases/index.hbs
DhariniJeeva: So I think we can't use this for doc links in headers that require only icons? I quickly tested it, and it did not have an icon-only option. Let me know if I missed something

calcaide: You are right. On the PoC it was done this way because Joshua suggested going with the "Learn more" text rather than the icon. 
That makes me think that for visual consistency with the rest, let's keep aliases as you implemented, later on, if we refactor the headers we can do it all at once, to have consistency with the design. 😉

DhariniJeeva: ah, that makes sense, for consistency, we will keep the doc link and once we have the POC merged, we can replace with HDS 👍

ZedLi (approved):
Looks good!

PR #2273, opened 2024-04-25
fix: 🐛 Fix error when deleting auth method

ZedLi:
✅ Closes: https://hashicorp.atlassian.net/browse/ICU-13624

## Description
There was an issue with deleting an auth method where we tried to access a model field after deleting it. Weirdly enough ember only throws the error when we make the auth method primary and then removing it as primary.

image

## How to Test
* Create a password auth method in an org
* Mark it as primary
* Remove it as primary
* Delete it
* Should not error 

## Checklist:

### DIFF
--- a/ui/admin/app/controllers/scopes/scope/auth-methods/index.js
+++ b/ui/admin/app/controllers/scopes/scope/auth-methods/index.js
@@ -142,11 +142,11 @@ export default class ScopesScopeAuthMethodsIndexController extends Controller {
   @notifyError(({ message }) => message, { catch: true })
   @notifySuccess('notifications.delete-success')
   async delete(authMethod) {
+    const scopeID = authMethod.scopeID;
     await authMethod.destroyRecord();
     // Reload the scope, since this is where the primary_auth_method_id is
     // stored.  An auth method deletion could affect this field.
-    const scope = this.store.peekRecord('scope', authMethod.scopeID);
-    await scope.reload();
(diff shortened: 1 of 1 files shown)


cameronperera (approved):

lisbet-alvarez (approved):
Great catch, and thanks for fixing this 🙌

ui/admin/app/controllers/scopes/scope/auth-methods/index.js
```diff
@@ -142,11 +142,11 @@ export default class ScopesScopeAuthMethodsIndexController extends Controller {
   @notifyError(({ message }) => message, { catch: true })
   @notifySuccess('notifications.delete-success')
   async delete(authMethod) {
+    const scopeID = authMethod.scopeID;
     await authMethod.destroyRecord();
     // Reload the scope, since this is where the primary_auth_method_id is
     // stored.  An auth method deletion could affect this field.
-    const scope = this.store.peekRecord('scope', authMethod.scopeID);
-    await scope.reload();
+    this.store.findRecord('scope', scopeID, { reload: true });
```
lisbet-alvarez, line 149: one small thing: i think this needs `await`

ZedLi: Good catch, I updated it

DhariniJeeva (approved):
thanks for the fix!

calcaide (approved):

PR #2149, opened 2024-02-20
chore: 🤖 Migrate `form.input` to `Hds:TExtInput::Field`

dvonanderson:
✅ Closes: https://hashicorp.atlassian.net/browse/ICU-12457
✅ Closes: https://hashicorp.atlassian.net/browse/ICU-12458

:tickets: [Jira ticket](https://hashicorp.atlassian.net/browse/ICU-12457)
:tickets: [Jira ticket](https://hashicorp.atlassian.net/browse/ICU-12458)

## Description
- Migrate `form/account/oidc` and `form/account/password` forms to use `Hds::TextInput::Field` instead of `form.input`
- Fix corresponding test
- Added `data-test-*` attributes to error messaging test
- Used `InfoField` component for readonly attributes for `OIDC`

## How to Test

:technologist: [Admin preview](PATH_TO_FEATURE)

:desktop_computer: [Desktop preview](PATH_TO_FEATURE)

## Checklist:

~~- [ ] I have added before and after screenshots for UI changes~~
~~- [ ] I have added JSON response output for API changes~~
~~- [ ] I have added steps to reproduce and test for bug fixes in the description~~
~~- [ ] I have commented on my code, particularly in hard-to-understand areas~~
- [x] My changes generate no new warnings
~~- [ ] I have added tests that prove my fix is effective or that my feature works`~~

PR #2275, opened 2024-04-26
fix: 🐛 Fix account password not being properly set

ZedLi:
## Description
There was a bug introduced in https://github.com/hashicorp/boundary-ui/pull/2149. We are not properly using the right field for account passwords.

## How to Test
Make a new account in a password auth and confirm a password is sent as part of the request. Create a new user with that account and login with it and it should work. 

## Checklist:


DhariniJeeva (approved):
thanks for the fix!

lisbet-alvarez (approved):
Great catch! Thanks for the fix!

PR #2282, opened 2024-05-01
Icu 11936 support node 20 stop support node 16

cameronperera:
:tickets: [Jira ticket](https://hashicorp.atlassian.net/browse/ICU-11936)

## Description

support node 20 stop support node 16

## How to Test

Update local node to version 20.*. You should be able to run test, lint, format, build, and start the app locally with no issues.
node v20.* will show the below error message due to our ember version. So far there have been no issues and the Github runners will continue to use node v18 for now.
`WARNING: Ember CLI v4.12.2 is not tested against Node v20.12.2. See "https://github.com/ember-cli/ember-cli/blob/master/docs/node-support.md" to find out which version of Node is best to use.`

## Checklist:


.github/workflows/build-admin-ui.yaml
```diff
@@ -21,7 +21,7 @@ jobs:
             node-modules-
       - uses: actions/setup-node@60edb5dd545a775178f52524783378180af0d1f8 # v4.0.2
         with:
-          node-version: 16
+          node-version: 18
```
ZedLi, line 24: Should we try 20?

cameronperera: We could try that but I was keeping this in line with what [ember suggest](https://github.com/ember-cli/ember-cli/blob/master/docs/node-support.md) for `v4.12.2`. I know we still plan to get to ember v5 and that would push us into node v20.
This also accurately reflects which version we use in [boundary-ui-releases workflow](https://github.com/hashicorp/boundary-ui-releases/blob/main/.github/workflows/build-packages.yml).

cameronperera: To argue against my own statement, we could test updating our workflows to node v20. Then we would have one less hurdle to getting ember to v5.

ZedLi: Yup, I was going to suggest we should be trying to push our CI and workflows to 20 since we'd still be behind if we just go to 18.

calcaide: Indeed in this comment, 20 should be the one we aim for. Nice work!!

calcaide (changes requested):
As mentioned on a comment, for the github workflow we should try to aim node 20.

Very good job!! this is much need to get rid of warnings and to could support necessary coming changes within `boundary-ui-releases` appreciate the work!

calcaide (approved):

PR #2283, opened 2024-05-02
chore: Update to use Node20 version of Actions

moduli:
This PR updates some GitHub Actions to use a Node20 version of the Action in order to avoid deprecation warnings.

In this case, `atlassian/gajira-*` actions are no longer maintained, but they are using node16, which is soon to be deprecated. This change removes the use of those actions. (This borrows changes from https://github.com/hashicorp/vault-workflows-common/pull/64)

### DIFF
--- a/.github/workflows/jira.yml
+++ b/.github/workflows/jira.yml
@@ -29,13 +29,6 @@ jobs:
       env:
         GITHUB_TOKEN: ${{ secrets.JIRA_SYNC_GITHUB_TOKEN }}
 
-    - name: Login
-      uses: atlassian/gajira-login@45fd029b9f1d6d8926c6f04175aa80c0e42c9026 # v3.0.1
-      env:
-        JIRA_BASE_URL: ${{ secrets.JIRA_BASE_URL }}
-        JIRA_USER_EMAIL: ${{ secrets.JIRA_USER_EMAIL }}
-        JIRA_API_TOKEN: ${{ secrets.JIRA_API_TOKEN }}
-
     - name: Set ticket type
       if: github.event.action == 'opened' && !steps.boundary-team-role.outputs.role
       id: set-ticket-type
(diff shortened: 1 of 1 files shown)


moduli:
Trying to figure out a way to test this...
- One way would be to just merge this and test in this repo
- An alternative is to find some other internal repo and create a workflow that points to the one in this branch. However, I'm not finding any good examples that contain the needed secrets...

.github/workflows/jira.yml
```diff
@@ -62,27 +59,49 @@ jobs:
...
+        JIRA_BASE_URL: ${{ secrets.JIRA_BASE_URL }}
+        JIRA_USER_EMAIL: ${{ secrets.JIRA_USER_EMAIL }}
+        JIRA_API_TOKEN: ${{ secrets.JIRA_API_TOKEN }}
       with:
         # cf[10089] is Issue Link custom field
         jql: 'issuetype = "GH Issue" and cf[10089]="${{ github.event.issue.html_url || github.event.pull_request.html_url }}"'
 
     - name: Sync comment
       if: github.event.action == 'created' && steps.search.outputs.issue
       uses: atlassian/gajira-comment@164913891625fe50e9836957902e0bf7d9ef99a8 # v3.0.1
```
lisbet-alvarez, line 72: Thank you for updating this workflow!! Looks like we were using the deprecated `atlassian/gajira-comment` action here, luckily there is `tomhjp/gh-action-jira-comment@6eb6b9ead70221916b6badd118c24535ed220bd9 # v0.2.0` that can be used instead.

lisbet-alvarez:
I'm onboard with the first option of just merging then testing.

lisbet-alvarez (approved):
Looks great, thank you!

moduli:
Tested with
- https://github.com/hashicorp/boundary-ui/issues/2288
- https://hashicorp.atlassian.net/browse/ICU-13742

moduli:
Looks like there's an issue when reopening a ticket. Investigating...

PR #2284, opened 2024-05-03
refactor: 💡 remove route-action usage for host-catalogs

lisbet-alvarez:
:tickets: [Jira ticket](https://hashicorp.atlassian.net/browse/ICU-12606)

## Description
Remove route-action usage for host-catalogs, hosts, and host-sets.

## How to Test
1. start boundary dev instance
2. test crudl for host-catalogs
3. test crudl for hosts
4. test crudl for host-sets
**verify no strange warnings or errors appear in console**

## Checklist:


ui/admin/tests/unit/controllers/scopes/scope/host-catalogs/index-test.js
```diff
@@ -32,5 +75,55 @@ module(
       assert.strictEqual(controller.page, 1);
       assert.strictEqual(controller.search, searchValue);
     });
+
+    test('cancel action rolls-back changes on the specified model', async function (assert) {
```
calcaide, line 79: I would like to constructively discuss a bit this approach.

I think this approach is valid, but at the same time I think we should perform assertions within the same store record.

My thinking is next (I can be wrong):
- We retrieve the record.
- We perform changes within the record.
- We assert those changes have been performed.
- We Cancel changes (rollback)
- We assert the previous changes have been rollback within the same record.

In code I am thinking something like this:

```
const hostCatalog = await store.findRecord(
        'host-catalog',
        instances.hostCatalog.id,
);

hostCatalog.name = 'test';
assert.strictEqual(hostCatalog.name, 'test');
await controller.cancel(hostCatalog);
assert.notEqual(hostCatalog.name, 'test');
```

ui/admin/app/templates/scopes/scope/host-catalogs/host-catalog/host-sets/host-set/add-hosts.hbs
```diff
@@ -22,8 +22,8 @@
     <Form::HostSet::AddHosts
       @model={{@model.hostSet}}
       @hosts={{@model.hosts}}
-      @submit={{route-action 'addHosts' @model.hostSet}}
-      @cancel={{route-action 'cancel'}}
+      @submit={{fn this.addHosts @model.hostSet}}
+      @cancel={{this.cancel}}
```
calcaide, line 26: None blocking question: why this one does not use the `fn`?

ui/admin/tests/unit/controllers/scopes/scope/host-catalogs/host-catalog/host-sets/host-set/add-hosts-test.js
```diff
@@ -0,0 +1,89 @@
...
+      await visit(urls.addHosts);
+      const hostSetBefore = await store.findRecord(
+        'host-set',
+        instances.hostSet.id,
+      );
+
+      assert.deepEqual(hostSetBefore.host_ids, []);
+
+      await controller.addHosts(hostSetBefore, [instances.host.id]);
+      const hostSetAfter = await store.findRecord(
```
calcaide, line 81: Do we really need to use hostSetAfter?

I do believe we could keep using the same store record for this test.

My thinking is:
- We retrieve hostSet record.
- We assert it has no host_ids.
- We call controller.addHosts. Therefore a host_id is add to the record
- We check the record has the desired host_id.

Thoughts?

lisbet-alvarez: My thought process for calling findRecord again was to assert that the changes performed on that model instance persisted in the store. After thinking about this, thanks for your suggestion, perhaps that's not necessary since in both cases we are getting back a reference to the same instance. I'll update this.

ui/admin/tests/unit/controllers/scopes/scope/host-catalogs/index-test.js
lisbet-alvarez: Thanks for the feedback, as mentioned above, ill also refactor these tests, and existing tests to use the same instance for both asserting changes.

cameronperera (approved):
Looks great. Thanks for tackling so much for this refactor.

calcaide (approved):
Awesome work @lisbet-alvarez 🙌

PR #2304, opened 2024-05-22
feat: 🎸 Add support for polling transparent sessions

ZedLi:
✅ Closes: https://hashicorp.atlassian.net/browse/ICU-13831

# Description
Added functionality to poll the new `boundary ferry sessions` API. Still trying to figure out how to test this.

image

## How to Test
* Setup ferry and boundary according to the docs [here](https://github.com/hashicorp/ferry?tab=readme-ov-file#ferry).
* Copy over the new CLI to our cli folder and use `BYPASS_CLI_SETUP=true`
* Add a brokered credential to the `Generated localhost ssh target with an alias` target.
* SSH to the default `ssh.boundary.dev` alias and you should get a notification (though you might need to also enable it on your OS). 
* Clicking the notification should open up the correct session with credentials displayed.

## Checklist

### DIFF
--- a/ui/desktop/app/routes/scopes/scope/projects.js
+++ b/ui/desktop/app/routes/scopes/scope/projects.js
@@ -76,7 +76,6 @@ export default class ScopesScopeProjectsRoute extends Route {
 
         try {
           const aliases = await this.store.query('alias', {
-            recursive: true,
             scope_id: 'global',
             force_refresh: true,
           });
@@ -92,7 +91,7 @@ export default class ScopesScopeProjectsRoute extends Route {
         }
 
         window.location.href = `serve://boundary/#/scopes/${orgScope}/projects/sessions/${session.session_authorization.session_id}`;
-        this.ipc.invoke('focusWindow');
+        await this.ipc.invoke('focusWindow');
       };
     });
   }
--- a/ui/desktop/app/routes/scopes/scope/projects.js
+++ b/ui/desktop/app/routes/scopes/scope/projects.js
@@ -27,6 +27,7 @@ export default class ScopesScopeProjectsRoute extends Route {
   @service session;
   @service resourceFilterStore;
   @service router;
+  @service store;
   @service intl;
   @service ipc;
   @service clientAgentSessions;
@@ -54,6 +55,9 @@ export default class ScopesScopeProjectsRoute extends Route {
     return projects;
   }
 
+  /**
+   * Poll for new sessions with credentials. Sends a notification for each new session that has a credential.
+   */
   @runEvery(2000)
   async poller() {
     const sessions =
@@ -67,8 +71,27 @@ export default class ScopesScopeProjectsRoute extends Route {
         {
           body: this.intl.t('notifications.connected-to-target.description'),
         },
-      ).onclick = () => {
-        window.location.href = `serve://boundary/#/scopes/global/projects/sessions/${session.session_authorization.session_id}`;
+      ).onclick = async () => {
+        let orgScope = 'global';
+
+        try {
+          const aliases = await this.store.query('alias', {
+            recursive: true,
+            scope_id: 'global',
+            force_refresh: true,
+          });
+          const alias = aliases.find((alias) => alias.value === session.alias);
+
+          const target = await this.store.findRecord(
+            'target',
+            alias.destination_id,
+          );
+          orgScope = target.scope.parent_scope_id;
+        } catch (e) {
+          // Do nothing and default scope to global if an error occurs
+        }
+
+        window.location.href = `serve://boundary/#/scopes/${orgScope}/projects/sessions/${session.session_authorization.session_id}`;
         this.ipc.invoke('focusWindow');
       };
     });
--- a/ui/desktop/app/routes/scopes/scope/projects.js
+++ b/ui/desktop/app/routes/scopes/scope/projects.js
@@ -5,7 +5,6 @@
 
 import Route from '@ember/routing/route';
 import { inject as service } from '@ember/service';
-import runEvery from 'ember-pollster/decorators/route/run-every';
 
 /**
  * TODO:  This route is now vestigial.  Desktop does not provide project-based
@@ -22,6 +21,10 @@ import runEvery from 'ember-pollster/decorators/route/run-every';
  * It no longer makes sense as a dedicated route.
  */
 export default class ScopesScopeProjectsRoute extends Route {
+  // =attributes
+
+  job;
+
   // =services
 
   @service session;
@@ -31,6 +34,8 @@ export default class ScopesScopeProjectsRoute extends Route {
   @service intl;
   @service ipc;
   @service clientAgentSessions;
+  @service pollster;
+  @service flashMessages;
 
   // =methods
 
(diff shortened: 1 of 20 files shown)


ui/desktop/electron-app/src/services/client-agent-daemon-manager.js
```diff
@@ -0,0 +1,43 @@
...
+
+const { spawnSync, spawn } = require('../helpers/spawn-promise');
+
+class ClientAgentDaemonManager {
+  /**
+   * Checks the status of the client agent daemon.
+   * @returns {string}
+   */
+  status() {
+    const clientAgentStatusCommand = ['ferry', 'status', '-format=json'];
```
johanbrandhorst, line 14: FYI the `boundary ferry` command has been renamed `boundary client-agent`. Flags and environment variables are also renamed accordingly. `boundary ferry` will not work with `0.16.1`.

ZedLi: Yup, I'm planning on renaming those in a separate PR to handle the client daemon to `cache` as well. My current build still uses `ferry` so I kept it in, I can add a TODO to make it more explicit

ui/desktop/app/routes/scopes/scope/projects.js
```diff
@@ -49,4 +53,24 @@ export default class ScopesScopeProjectsRoute extends Route {
...
+    sessions.forEach((session) => {
+      new window.Notification(
+        this.intl.t('notifications.connected-to-target.title', {
+          target: session.alias,
+        }),
+        {
+          body: this.intl.t('notifications.connected-to-target.description'),
+        },
+      ).onclick = () => {
+        window.location.href = `serve://boundary/#/scopes/global/projects/sessions/${session.session_authorization.session_id}`;
```
ZedLi, line 71: I wasn't sure if hardcoding the scope to `global` would be an issue but it turns out as long as we can list the scopes, it should be fine. It seems like it still works even if we don't have access to the global scope (and get a 403) but it'll still lead there correctly. 

I'm not sure where else to go, I could just use the default scope of wherever they're currently in but an alias could lead to a different scope.

After typing all this out, I realize I could also look through the resolvable aliases, find the target, find the scope that target is in and use that, which might be better. I'll try that

ui/desktop/electron-app/src/services/client-agent-daemon-manager.js
johanbrandhorst: No worries, just wanted to make sure we'd communicated it! As you were :)

ui/desktop/app/routes/scopes/scope/projects.js
```diff
@@ -49,4 +53,24 @@ export default class ScopesScopeProjectsRoute extends Route {
     );
     return projects;
   }
+
+  @runEvery(2000)
```
ZedLi, line 57: Not sure if I should still be using `ember-pollster`, did we ever have any plans to get off of this?

ui/desktop/electron-app/src/services/client-agent-daemon-manager.js
```diff
@@ -0,0 +1,45 @@
...
+    const { stdout } = spawnSync(clientAgentStatusCommand);
+
+    return JSON.parse(stdout);
```
ZedLi, line 19: Note that I started using `JSON.parse` rather than `jsonify` here. `jsonify` was swallowing any errors that occurred during parsing which I was not a fan of. I'd rather we bubble up the error and we can let the client choose what to do with the error (most likely just log it).

ui/desktop/electron-app/src/services/client-agent-daemon-manager.js
calcaide: I had no idea about the rename, so good to know 😄

ui/desktop/app/routes/scopes/scope/projects.js
calcaide: I will prefer to get off `ember-pollster`. But TBH I do think that could be done in a separate PR/context since is used in our codebase in other places. For now, I do think is totally fine to keep using it and later on we can consider a refactor?

As long as I know there were no plans for that, maybe something we can ask on tech-frontend how other teams are dealing with this and bring the topic to tech time.

ui/desktop/app/routes/scopes/scope/projects.js
```diff
@@ -49,4 +54,46 @@ export default class ScopesScopeProjectsRoute extends Route {
...
+        }),
+        {
+          body: this.intl.t('notifications.connected-to-target.description'),
+        },
+      ).onclick = async () => {
+        let orgScope = 'global';
+
+        try {
+          const aliases = await this.store.query('alias', {
+            recursive: true,
```
DhariniJeeva, line 79: do we need this recursive, since aliases are only available at the global level?

DhariniJeeva: Also, should we query list-resolvable aliases or this? When would we use the other endpoint?

ZedLi: Yeah recursive isn't needed here, I'll update that. The trick is this is actually querying for resolvable aliases haha, see [here](https://github.com/hashicorp/boundary-ui/blob/79a932bebb57f9759589bf1384f7e7dc8356fbe5/addons/api/addon/handlers/client-daemon-handler.js#L30)

ui/desktop/electron-app/src/ipc/handlers.js
```diff
@@ -114,6 +115,13 @@ handle('toggleFullscreenWindow', () => {
  */
 handle('closeWindow', () => app.quit());
 
+/**
+ * Focus the window
+ */
+handle('focusWindow', async () => {
```
DhariniJeeva, line 121: Is async needed here?

ui/desktop/app/routes/scopes/scope/projects.js
```diff
@@ -49,4 +54,46 @@ export default class ScopesScopeProjectsRoute extends Route {
...
+   */
+  @runEvery(2000)
+  async poller() {
+    const sessions =
+      await this.clientAgentSessions.getNewSessionsWithCredentials();
+
+    sessions.forEach((session) => {
+      new window.Notification(
+        this.intl.t('notifications.connected-to-target.title', {
+          target: session.alias,
```
DhariniJeeva, line 69: Why are we passing this target info? Do we need this for the notification?

ZedLi: This is so we have a name for the what we're connecting to
image

ui/desktop/electron-app/src/ipc/handlers.js
ZedLi: Nope, removed

ui/desktop/app/routes/scopes/scope/projects.js
DhariniJeeva: My bad. I didn't realize that was the session's name :D

DhariniJeeva (approved):
looks great!

calcaide:
After following the instructions within the description, I am able to run this in mac but I am NOT getting notifications. I will ask you for help on reviewing this. Most likely something on my env regarding notifications.

calcaide (approved):
After following [new instructions to run Boundary + client agent](https://github.com/hashicorp/boundary-client-agent?tab=readme-ov-file#fetch-and-install-boundary-enterprise) I was able to run successfully an ssh connection with client agent and get notifications through the desktop client.

ui/desktop/electron-app/src/services/client-agent-daemon-manager.js
ZedLi: I went back to jsonify and decided I'll log the error when we add logging since I need to parse both stdout and stderr if stdout doesn't have parse-able output

DhariniJeeva (approved):

PR #2305, opened 2024-05-23
fix: 🐛 target patch with_Aliases field

DhariniJeeva:
# Description
This PR fixes target patch when there is with_aliases field

## Screenshots (

https://github.com/hashicorp/boundary-ui/assets/15043878/f3594ede-871a-4047-b93e-ec877123bced

if appropriate)

## How to Test

- create. a target with aliases 
- after successful submit, edit the form and update any of the fields 
- it should patch successfully and the network payload should not include `with_aliases` attr
## Checklist


ZedLi (approved):
Looks good, thanks for fixing this!

PR #2308, opened 2024-05-24
feat: 🎸 add grant_scope_ids to role model

cameronperera:
✅ Closes: https://hashicorp.atlassian.net/browse/ICU-12709

# Description

Adds `grant_scope_ids` to the role model and adds `setGrantScopes` and `removeGrantScopes` methods to make use of the `:set-grant-scopes` and `:remove-grant-scopes` actions in the API.

Screenshot 2024-05-24 at 3 11 42 PM

## Checklist

### DIFF
--- a/addons/api/addon/generated/models/role.js
+++ b/addons/api/addon/generated/models/role.js
@@ -28,6 +28,12 @@ export default class GeneratedRoleModel extends BaseModel {
   })
   grant_scope_id;
 
+  @attr('array', {
+    description:
+      "The scopes to which this role applies.  Defaults to the role's scope.",
+  })
+  grant_scope_ids;
+
   @attr('date', {
     description: 'The time this resource was created\nOutput only.',
     readOnly: true,
--- a/addons/api/addon/models/role.js
+++ b/addons/api/addon/models/role.js
@@ -89,20 +89,6 @@ export default class RoleModel extends GeneratedRoleModel {
   }
 
   /**
-   * Removes grant scope ids on the role via the `remove-grant-scopes` method.
-   * @param {[string]} grantScopeIDs
-   * @return {Promise}
-   */
-  removeGrantScopes(grantScopeIDs) {
-    return this.save({
-      adapterOptions: {
-        method: 'remove-grant-scopes',
-        grantScopeIDs,
-      },
-    });
-  }
-
-  /**
    * Saves grant strings on the role via the `set-grants` method.
    * See serializer and adapter for more information.
    * @param {[string]} grantStrings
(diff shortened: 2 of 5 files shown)


addons/api/addon/generated/models/role.js
```diff
@@ -28,6 +28,12 @@ export default class GeneratedRoleModel extends BaseModel {
   })
   grant_scope_id;
```
DhariniJeeva, line 29: Will we still use this `grant_scope_id` or will that be replaced with `grant_scope_ids`?

DhariniJeeva (approved):

addons/api/addon/generated/models/role.js
cameronperera: `grant_scope_id` but that will be done later when we remove all references to that attribute.

addons/api/addon/serializers/role.js
```diff
@@ -15,23 +15,32 @@ export default class RoleSerializer extends ApplicationSerializer {
...
+    const grantStrings = get(snapshot, 'adapterOptions.grantStrings');
     if (grantStrings)
       serialized = this.serializeWithGrantStrings(snapshot, grantStrings);
+
     const principalIDs = snapshot?.adapterOptions?.principalIDs;
     if (principalIDs)
       serialized = this.serializeWithPrincipals(snapshot, principalIDs);
+
+    const grantScopeIDs = snapshot?.adapterOptions?.grantScopeIDs;
+    if (grantScopeIDs)
```
ZedLi, line 36: I know this was how it was previously but can we use braces for if statements that aren't on the same line?

addons/api/addon/models/role.js
```diff
@@ -74,6 +74,34 @@ export default class RoleModel extends GeneratedRoleModel {
...
+      },
+    });
+  }
+
+  /**
+   * Removes grant scope ids on the role via the `remove-grant-scopes` method.
+   * @param {[string]} grantScopeIDs
+   * @return {Promise}
+   */
+  removeGrantScopes(grantScopeIDs) {
```
ZedLi, line 96: When would we call `removeGrantScopes` instead of calling `setGrantScopes`?

addons/api/addon/serializers/role.js
cameronperera: Yep that is right. I will fix these.

addons/api/addon/models/role.js
cameronperera: Looking at our designs, we won't need a dedicated `removeGrantScopes` anymore. I will remove this.

ZedLi (approved):
Looks great!

PR #2310, opened 2024-05-28
test(e2e): 0.16.1 updates

moduli:
This PR updates the Admin UI e2e test suite based on results from testing Boundary 0.16.1. There are two changes here
1. Updates the test suite to utilize the AWS region specified in the environment. The end-to-end test suite can now run against different AWS regions, and this change uses the information passed from infra creation
2. Updates a locator for a label that was used to attach a storage bucket to a target.


calcaide (approved):
Nice one!! thanks for this!

PR #2312, opened 2024-05-30
test(e2e): Add test for minio storage bucket

moduli:
This PR is a follow-up to https://github.com/hashicorp/boundary-enterprise/pull/1112.

This PR adds an Admin UI e2e test to validate session recording using a MinIO storage bucket. This test is fairly similar to the existing session recording test using an AWS storage bucket outside of the creation of a storage bucket using a different provider.

https://hashicorp.atlassian.net/browse/ICU-13930


calcaide (approved):
Terrific work!!! thank you for these tests!!

PR #2313, opened 2024-05-30
test(e2e): Update password test to log in with new account

moduli:
This PR updates an Admin UI e2e test to log in using a new password account. The existing test created a new account, but we want to ensure we can log in with that account. We attempt to log in after initial account creation AND after setting a new password.

There was a bug in a previous release with an issue on initial password creation: https://github.com/hashicorp/boundary-ui/pull/2275 

https://hashicorp.atlassian.net/browse/ICU-13940


ui/admin/tests/e2e/tests/auth-method-ldap.spec.js
```diff
@@ -135,6 +125,14 @@ test('Set up LDAP auth method @ce @ent @docker', async ({ page }) => {
     await createRole(page);
     await addPrincipalToRole(page, ldapManagedGroupName);
   } finally {
+    await authenticateBoundaryCli(
+      process.env.BOUNDARY_ADDR,
+      process.env.E2E_PASSWORD_AUTH_METHOD_ID,
+      process.env.E2E_PASSWORD_ADMIN_LOGIN_NAME,
+      process.env.E2E_PASSWORD_ADMIN_PASSWORD,
+    );
+    let orgs = JSON.parse(execSync('boundary scopes list -format json'));
```
calcaide, line 134: None blocking comment: these 2 vars `orgs` and `org` can be `const`, because they are not redefined.

ui/admin/tests/e2e/tests/auth-method-password.spec.js
```diff
@@ -32,24 +29,120 @@ test('Verify new auth-method can be created and assigned to users @ce @ent @aws
...
-    org = orgs.items.filter((obj) => obj.name == orgName)[0];
-    await createPasswordAuthMethod(page);
-    await createPasswordAccount(page, 'test-user', 'password');
-    await setPasswordToAccount(page, 'password2');
-    await makeAuthMethodPrimary(page);
-    await createUser(page);
-    await addAccountToUser(page);
-  } finally {
-    await deleteOrgCli(org.id);
+    let orgs = JSON.parse(execSync('boundary scopes list -format json'));
```
calcaide, line 142: None blocking comment: these two vars can be `const`.

moduli: 👍 I'll update

ui/admin/tests/e2e/tests/auth-method-ldap.spec.js
moduli: I'll update

calcaide (approved):
Awesome work!!! thanks for these tests, much need on the UI!! 💪

PR #2314, opened 2024-05-30
test(e2e): Update ldap test to log in with ldap account

moduli:
Similar to https://github.com/hashicorp/boundary-ui/pull/2313, this PR updates an Admin UI e2e test to log in using an LDAP account that was created.

https://hashicorp.atlassian.net/browse/ICU-13941


calcaide (approved):
Amazing work!! Thanks for this!

PR #2318, opened 2024-06-03
Icu 13451 desktop client logging

ZedLi:
# Description
This adds desktop client logging on all environments with an initial pass of things to log. 

By default, the logs are written to:

### Linux: 

    ~/.config/Boundary/logs/desktop-client.log

### macOS: 

    ~/Library/Logs/Boundary/desktop-client.log

### Windows:

    %USERPROFILE%\AppData\Roaming\Boundary\logs\desktop-client.log

[desktop-client.log](https://github.com/user-attachments/files/15539786/desktop-client.log)

## How to Test
Run the desktop client and experiment with generating errors. You'll have to manually do this usually by throwing manual errors. You can `tail -f` the log file to watch for new log entries.

To test log rotation, you can set the [maxSize](https://github.com/megahertz/electron-log/blob/master/docs/transports/file.md#maxsize-number) to something smaller.

## Checklist

### DIFF
--- a/addons/api/.eslintrc.js
+++ b/addons/api/.eslintrc.js
@@ -25,6 +25,7 @@ module.exports = {
   env: {
     browser: true,
   },
+  globals: { __electronLog: 'true' },
   rules: {
     'ember/no-get': 'off',
     'ember/no-get-with-default': 'off',
--- a/addons/api/.eslintrc.js
+++ b/addons/api/.eslintrc.js
@@ -25,7 +25,6 @@ module.exports = {
   env: {
     browser: true,
   },
-  globals: { __electronLog: 'true' },
   rules: {
     'ember/no-get': 'off',
     'ember/no-get-with-default': 'off',
--- a/addons/api/addon/handlers/client-daemon-handler.js
+++ b/addons/api/addon/handlers/client-daemon-handler.js
@@ -120,8 +120,9 @@ export default class ClientDaemonHandler {
                 'searchClientDaemon',
                 remainingQuery,
               );
-            } catch {
+            } catch (err) {
               // If it fails again just fall back to fetching controller data
+              __electronLog?.info('Failed to add token to client daemon', err);
             }
           }
 
--- a/addons/api/addon/handlers/client-daemon-handler.js
+++ b/addons/api/addon/handlers/client-daemon-handler.js
@@ -122,11 +122,16 @@ export default class ClientDaemonHandler {
               );
             } catch (err) {
               // If it fails again just fall back to fetching controller data
-              __electronLog?.info('Failed to add token to client daemon', err);
+              __electronLog?.error('Failed to add token to daemons', err);
             }
-          }
+          } else {
+            __electronLog?.error(
+              'Failed to search cache daemon; falling back to search controller',
+              e,
+            );
 
-          return fetchControllerData(context, next);
+            return fetchControllerData(context, next);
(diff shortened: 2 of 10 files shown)


ui/desktop/electron-app/src/index.js
```diff
@@ -54,6 +56,33 @@ protocol.registerSchemesAsPrivileged([
...
+log.transports.console.level = false;
+log.transports.file.format =
+  '[{y}-{m}-{d} {h}:{i}:{s}.{ms}{z}] [{level}] {text}';
+log.transports.file.fileName = 'desktop-client.log';
+log.transports.file.archiveLogFn = (file) => {
+  file = file.toString();
+  const info = path.parse(file);
+
+  // This renames old files to the format of `desktop-client-(iso string date).log`
+  // TODO: Should we compress or delete old log files after a certain time?
```
ZedLi, line 70: Not exactly sure how we should handle log rotation currently. The default logic is it rotates to a `file.old.log` so it will constantly delete old files by just renaming them (i.e. there's a `main.log` and a `main.old.log` and with default log size of 1MB, it basically keeps at most 2MB of the most recent logs).

I didn't want it to lose older logs but now I'm also not deleting any old logs, do we think it's necessary to have some sort of cleanup before we release? Or for now would it be better to keep the default logic and just only have it keep the two most recent log files?

addons/api/.eslintrc.js
```diff
@@ -25,6 +25,7 @@ module.exports = {
   env: {
     browser: true,
   },
+  globals: { __electronLog: 'true' },
```
ZedLi, line 28: It's unfortunate we need this but it should only be used in the `client-daemon-handler`...but this is a very leaky implementation detail now. 

Open to other thoughts, I could always make a call to the ipc handler to make log statements in the main process instead of renderer as well.

ui/desktop/electron-app/src/index.js
DhariniJeeva: Since we removed the rotation logic, will we now delete the old logs and only show the two recent logs that fit into our maxsize?

DhariniJeeva (approved):
LGTM

ui/desktop/electron-app/src/index.js
ZedLi: Yup exactly

addons/api/addon/handlers/client-daemon-handler.js
```diff
@@ -3,6 +3,7 @@
  * SPDX-License-Identifier: BUSL-1.1
  */
 
+/* global __electronLog */
```
calcaide, line 6: What is the purpose of this comment?

ui/desktop/electron-app/src/index.js
ZedLi: It'd be the same log file

addons/api/addon/handlers/client-daemon-handler.js
ZedLi: It's to let eslint know that we're using a global variable

ui/desktop/electron-app/src/index.js
calcaide: Thanks for clarification

calcaide (approved):
Manually tested on Mac OS and the output is as expected.

Awesome work!! now we have logs 🥳

PR #2322, opened 2024-06-07
Icu 12708 add scopes tab route in role route

cameronperera:
# Description

This adds the "Scopes" tab in the role screen and list any grant scopes associated with the role.

UPDATED NOTE: (see attached image below for context)
In our designs, if the grant scope of "this" is present, we want to use the current scope and allow the user to navigate to that scope. This will be added in a follow up PR so right now it displays the name of the current scope.
Screenshot 2024-06-12 at 8 54 57 AM

Screenshot 2024-06-06 at 11 10 52 AM

## How to Test

Best to test in a dev instance. Create a role via the UI or CLI then run this command in your terminal:
```shell
curl --location --request POST 'http://localhost:9200/v1/roles/<YOUR_ROLE_ID_HERE>:set-grant-scopes' \
--header 'Authorization: Bearer <YOUR_TOKEN_HERE>' \
--header 'Content-Type: application/json' \
--data-raw '{
    "grant_scope_ids": [
        "descendants",
        "o_1234567890",
        "p_1234567890",
        "this"
    ],
    "version": 1
}'
``` 
Make sure you add your bearer token and roleID.  You will need to update your `version` if you make subsequent request.
After successful response, navigate to `/scopes/global/roles/<YOUR_ROLE_ID>/scopes` to view the new table.

## Checklist

### DIFF
--- a/ui/admin/app/routes/scopes/scope/roles/role/scopes.js
+++ b/ui/admin/app/routes/scopes/scope/roles/role/scopes.js
@@ -9,9 +9,7 @@ export default class ScopesScopeRolesRoleScopesRoute extends Route {
 
   async model() {
     const role = this.modelFor('scopes.scope.roles.role');
-    console.log('role: ', role);
     if (!GRANT_SCOPE_KEYWORDS.some((id) => role.grant_scope_ids.includes(id))) {
-      console.log('ENTERED');
       await this.store.query('scope', {
         scope_id: role.scope.id,
         recursive: true,


addons/api/addon/models/role.js
```diff
@@ -42,6 +51,61 @@ export default class RoleModel extends GeneratedRoleModel {
       : null;
   }
 
+  /**
+   * Convenience for looking up the grant scopes, if loaded.
+   */
+  get grantScopes() {
```
cameronperera, line 57: I just realized I forgot tests for this. I will add these asap.

cameronperera: Test have been added.

DhariniJeeva:
regarding the screenshot - should the `global` title in the parent scope also be capitalized similarly to the other scope?

addons/api/addon/models/role.js
```diff
@@ -42,6 +51,61 @@ export default class RoleModel extends GeneratedRoleModel {
...
+      // spreading into sortedScopeIDs if globalID is undefined
+      const globalID = this.grant_scope_ids.filter((id) => {
+        return id === 'global';
+      });
+      const orgIDs = this.grant_scope_ids.filter((id) => {
+        return id.startsWith('o_');
+      });
+      const projectIDs = this.grant_scope_ids.filter((id) => {
+        return id.startsWith('p_');
+      });
```
DhariniJeeva, line 75: I don't have much context here- do we want to sort the project and org here? In some resources, we display global first and display the rest of the scopes after global and don't sort them so wondering if this is needed

addons/api/addon/models/role.js
```diff
@@ -42,6 +51,61 @@ export default class RoleModel extends GeneratedRoleModel {
...
+   * Convenience for looking up the grant scopes, if loaded.
+   */
+  get grantScopes() {
+    // filter out scopes that start with an o_ or p_
+    // then sort the keywords
+    const grantScopes = [];
+    if (this.grant_scope_ids) {
+      const keywordIDs = this.grant_scope_ids.filter((id) => {
+        return !id.startsWith('o_') && !id.startsWith('p_') && id !== 'global';
+      });
```
ZedLi, line 64: Why not just use the keywords from `GRANT_SCOPE_KEYWORDS` to find them rather than doing the inverse?

addons/api/addon/models/role.js
```diff
@@ -42,6 +51,61 @@ export default class RoleModel extends GeneratedRoleModel {
...
+      const orgIDs = this.grant_scope_ids.filter((id) => {
+        return id.startsWith('o_');
+      });
+      const projectIDs = this.grant_scope_ids.filter((id) => {
+        return id.startsWith('p_');
+      });
+
+      const sortedKeywords = keywordIDs.sort((a, b) => {
+        return (
+          GRANT_SCOPE_KEYWORDS.indexOf(a) - GRANT_SCOPE_KEYWORDS.indexOf(b)
```
ZedLi, line 79: Maybe make a comment on the array that the order is important and affects sorting if we're doing this

cameronperera:
> regarding the screenshot - should the `global` title in the parent scope also be capitalized similarly to the other scope?

Great question. In most cases we add custom logic for displaying Global scope info. In the parent scope, I am just showing the name (or id). I can update this to display Global instead.

addons/api/addon/models/role.js
cameronperera: In the original designs we didn't need this but when reading all of the grant scopes on a role, we wanted a way to view the keywords (this, children, descendants), orgs, and projects in the same table. With that, we decided to have the keywords first, global second, orgs third, and projects last. We wanted global and orgs to be at the top and not mixed in among the projects.

addons/api/addon/models/role.js
```diff
@@ -42,6 +51,61 @@ export default class RoleModel extends GeneratedRoleModel {
...
+        );
+      });
+
+      const sortedScopeIDs = [
+        ...sortedKeywords,
+        ...globalID,
+        ...orgIDs,
+        ...projectIDs,
+      ];
```
ZedLi, line 88: If the goal was for a custom sort order, why not have just written a custom sort function for all of these in the first place instead of just for the `sortedKeywords`?

ui/admin/app/controllers/scopes/scope/roles/role/scopes.js
```diff
@@ -0,0 +1,31 @@
...
+import {
+  GRANT_SCOPE_THIS,
+  GRANT_SCOPE_CHILDREN,
+  GRANT_SCOPE_DESCENDANTS,
+} from 'api/models/role';
+
+export default class ScopesScopeRolesRoleScopesController extends Controller {
+  @controller('scopes/scope/roles/index') roles;
+
+  @action
```
ZedLi, line 17: Should these be actions? It seems like these are just local helper functions.

addons/core/translations/resources/en-us.yaml
```diff
@@ -560,6 +560,22 @@ role:
     description: Grants are permissions which allow roles to take actions and access resources.
     actions:
       create: New Grant
+  scope:
+    title: Scope
+    title_plural: Scopes
+    description: Role grant sets can be applied to multiple descendants underneath this scope.
```
ZedLi, line 566: It looks like we're using this description for any role page, but this isn't applicable if it's org (children only) or project (no descendants or children) scopes right?

addons/api/addon/models/role.js
cameronperera: Because I got lost in the woods refactoring this a few times. Thanks for the good suggestion. Updated.

addons/api/addon/models/role.js
cameronperera: added

addons/api/addon/models/role.js
cameronperera: I actually did have a custom sort order originally but I felt it was overkill just to sort the keywords in the correct order along with orgs and scope. I decided to go with filtering it down. I don't feel like this will be a performance issue but happy to talk some more about it.

ui/admin/app/controllers/scopes/scope/roles/role/scopes.js
cameronperera: No they shouldn't be actions. Removed.

addons/core/translations/resources/en-us.yaml
cameronperera: Oh that is a good point. I can ask if we should use more generic language here or we could change the language depending on the scope. Good catch.

addons/api/addon/models/role.js
ZedLi: Why is it overkill when you're just doing the same thing in a more manual way? Doesn't it simplify the logic greatly?

addons/api/addon/models/role.js
```diff
@@ -33,6 +42,66 @@ export default class RoleModel extends GeneratedRoleModel {
    */
   @attr({ readOnly: true, emptyArrayIfMissing: true }) grant_strings;
 
+  /**
+   * Convenience for looking up the grant scopes, if loaded.
+   */
+  get grantScopes() {
+    const grantScopes = [];
+    if (this.grant_scope_ids) {
+      const sortedScopeIDs = this.grant_scope_ids.slice().sort((a, b) => {
```
ZedLi, line 51: You might be able to use the new [toSorted](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/toSorted) function for this.

Otherwise I think doing 
```suggestion
      const sortedScopeIDs = [...this.grant_scope_ids].sort((a, b) => {
```
is a bit more obvious that you're copying the array first

addons/api/addon/models/role.js
```diff
@@ -33,6 +42,66 @@ export default class RoleModel extends GeneratedRoleModel {
...
+
+        // both a and b are projects - keep original order
+        if (a.startsWith('p_') && b.startsWith('p_')) return 0;
+        // only a is a project
+        if (a.startsWith('p_')) return -1;
+        // only b is a project
+        if (b.startsWith('p_')) return 1;
+
+        // default - keep original order
+        return 0;
```
ZedLi, line 83: I see what you mean by this being overkill, but I think the logic can be simplified a bit more by using your original sorting idea with the `GRANT_SCOPE_KEYWORDS`.

If you have an array or object like this:
```
{ 
  "this": 0,
  "children": 1,
  "descendants": 2,
  "global": 3,
  "o": 4,
  "p": 5
}
```

You can still do the sort function logic with `sortOrder.indexOf(a) - sortOrder.indexOf(b)` or `sortOrder[a] - sortOrder[b]` depending if you use an array or object (I don't think there's a big performance difference for iterating over such a small array).

The only gotcha is you have to do a little processing first if the ids are orgs or project ones since you have to slice off the extra characters after the underscore, but that's relatively straightforward since you could just do something like `a.split('_')[0]` applied to all your ids.

addons/api/mirage/generated/factories/scope.js
```diff
@@ -10,7 +10,6 @@ import { faker } from '@faker-js/faker';
  * GeneratedScopeModelFactory
  */
 export default Factory.extend({
-  name: () => faker.word.words(),
```
DhariniJeeva, line 13: is this not needed anymore?

DhariniJeeva: oops just noticed we get the name from the file above

ui/admin/app/controllers/scopes/scope/roles/role/scopes.js
```diff
@@ -0,0 +1,27 @@
...
+import {
+  GRANT_SCOPE_THIS,
+  GRANT_SCOPE_CHILDREN,
+  GRANT_SCOPE_DESCENDANTS,
+} from 'api/models/role';
+
+export default class ScopesScopeRolesRoleScopesController extends Controller {
+  @controller('scopes/scope/roles/index') roles;
+
+  isKeywordThis(id) {
```
DhariniJeeva, line 16: Do we need separate methods? Since we are passing in the ID from the template, can we just check if the ID exists in the array of grants such as [GRANT_SCOPE_THIS, GRANT_SCOPE_CHILDREN,
  GRANT_SCOPE_DESCENDANTS]?

addons/api/addon/models/role.js
cameronperera: Resolving this comment and will continue in new comment.

addons/api/addon/models/role.js
cameronperera: I see what you are getting at. I will update the sort now.

addons/core/translations/resources/en-us.yaml
cameronperera: This has been updated.

addons/api/addon/models/role.js
cameronperera: Updated. Thanks for the suggestion.

ui/admin/app/controllers/scopes/scope/roles/role/scopes.js
cameronperera: Actually looking this over I think I can refactor this to one specific for this as we want to use the current scope of the role if they have "this". Along with a second method that checks if it is "children" or "descendants" and just use the id in my translations to get the correct text. I will refactor this now. Thanks!

cameronperera: Done.

ZedLi (approved):

ui/admin/app/routes/scopes/scope.js
```diff
@@ -53,6 +53,12 @@ export default class ScopesScopeRoute extends Route {
    * a project.  These are used for scope navigation.
    */
   async afterModel(model) {
+    // fetch all scopes for multi-scope grants
+    // TODO: address if existing logic needs corrected
+    await this.store.query('scope', {
+      scope_id: model.scopeID,
+      recursive: true,
+    });
```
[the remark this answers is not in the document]
cameronperera: Updated.

DhariniJeeva (approved):
looks great!

ui/admin/app/routes/scopes/scope/roles/role/scopes.js
```diff
@@ -0,0 +1,22 @@
...
+export default class ScopesScopeRolesRoleScopesRoute extends Route {
+  // =services
+
+  @service store;
+
+  async model() {
+    const role = this.modelFor('scopes.scope.roles.role');
+    console.log('role: ', role);
+    if (!GRANT_SCOPE_KEYWORDS.some((id) => role.grant_scope_ids.includes(id))) {
+      console.log('ENTERED');
```
lisbet-alvarez, line 14: heads up line 12 and line 14 have leftover console logs

cameronperera: Thanks. Removed

ui/admin/app/routes/scopes/scope/roles/role/scopes.js
```diff
@@ -0,0 +1,20 @@
...
+  // =services
+
+  @service store;
+
+  async model() {
+    const role = this.modelFor('scopes.scope.roles.role');
+    if (!GRANT_SCOPE_KEYWORDS.some((id) => role.grant_scope_ids.includes(id))) {
+      await this.store.query('scope', {
+        scope_id: role.scope.id,
+        recursive: true,
```
lisbet-alvarez, line 15: should this have `pushToStore: false` in case role scope is global and it recursively loads a lot of scopes?

PR #2323, opened 2024-06-07
test(e2e): Add test for global settings

moduli:
This PR adds an Admin UI e2e test that utilizes the Global Settings page. This also have to cover the following issue: https://github.com/hashicorp/boundary-ui/pull/2227

https://hashicorp.atlassian.net/browse/ICU-14003


calcaide (approved):
Tested manually and all good!! thanks for this work!

moduli:
> Just wondering, who (i.e. with what grants) can update the name and description on the global settings page?

I believe those are fixed and cannot be changed

PR #2327, opened 2024-06-10
fix: 🐛 appUpdater support ARM64

calcaide:
# Description
Add ARM64 support for the appUpdater

:tickets: [Jira ticket](https://hashicorp.atlassian.net/browse/ICU-13666)

### Test 1:
- Within dev environment, start the app.
- When the autoUpdater launches, say YES (so it downloads the latest version.
- Make sure on the terminal the donwloaded files is the latet published DC
Screenshot 2024-06-10 at 2 45 04 PM

### Test 2: (to be run ONLY on ARM chips)
- Grab ARM64 binary [from this test build](https://github.com/hashicorp/boundary-ui-releases/actions/runs/9667402379).
- Install the test binary within your system (drag & drop to apps folder).
- Run the test binary.
- The test build MUST trigger the autoupdater, since the version set on the test build `1.2.3` is lower than the latest publish version `2.0.3`.
- Say YES to update.
- Check the version has been updated to ARM64 DC `2.0.3`.

### Test 3: (to be run ONLY on ARM chips)
This test checks a very important step, if the user has the AMD64 but its arch is ARM64, we must swap from AMD to ARM

- Grab AMD64 (x64) binary [from this test build](https://github.com/hashicorp/boundary-ui-releases/actions/runs/9667402379).
- Install the test binary within your system (drag & drop to apps folder).
- Run the test binary.
- The test build MUST trigger the autoupdater, since the version set on the test build 1.2.3 is lower than the latest publish version 2.0.3.
- Say YES to update.
- Check the version has been updated to ARM64 DC 2.0.3.

### Test 4: (to be run ONLY on INTEL chips)
- Grab AMD64 (x64) binary [from this test build](https://github.com/hashicorp/boundary-ui-releases/actions/runs/9667402379).
- Install the test binary within your system (drag & drop to apps folder).
- Run the test binary.
- The test build MUST trigger the autoupdater, since the version set on the test build `1.2.3` is lower than the latest publish version `2.0.3`.
- Say YES to update.
- Check the version has been updated to AMD64 DC `2.0.3`.


lisbet-alvarez (approved):
Awesome work! I performed Test 1 and Test 2 and everything works as expected!

ui/desktop/electron-app/src/helpers/app-updater.js
```diff
@@ -19,6 +20,33 @@ if (debug && process.env.APP_UPDATER_CURRENT_VERSION) {
...
 
+const returnArchitectureToUpdate = () => {
+  const nodeArchitecture = process.arch;
+
+  if (nodeArchitecture === 'arm64') {
+    return nodeArchitecture;
+  }
+
+  if (nodeArchitecture === 'amd64') {
+    const runsOnRosetta = exec(
```
ZedLi, line 31: We should use `execSync` or `spawnSync` rather than `exec`, that simplifies it a bit more.

You're returning the child process and using that as the check instead of the output. You would have to check what the output is returning (and then based on the output determine if we're running in rosetta).

Keep in mind that outputs are buffers by default FYI

ui/desktop/electron-app/src/helpers/app-updater.js
```diff
@@ -19,6 +20,33 @@ if (debug && process.env.APP_UPDATER_CURRENT_VERSION) {
...
 }
 
+const returnArchitectureToUpdate = () => {
+  const nodeArchitecture = process.arch;
+
+  if (nodeArchitecture === 'arm64') {
+    return nodeArchitecture;
+  }
+
+  if (nodeArchitecture === 'amd64') {
```
ZedLi, line 30: I don't think you need this if check since if it's not `arm64`, we can assume the process is now amd64

ZedLi (changes requested):

ui/desktop/electron-app/src/helpers/app-updater.js
```diff
@@ -19,6 +20,31 @@ if (debug && process.env.APP_UPDATER_CURRENT_VERSION) {
...
+    if (runsOnRosetta) {
+      return 'arm64';
+    } else {
+      return nodeArchitecture; // amd64
+    }
+  } catch (err) {
+    console.error(`returnArchitectureToUpdate error: ${err}`);
+  }
+
+  return 'arm64';
```
ZedLi, line 45: This only gets here if we get an error right? Should we just return back `nodeArchitecture` to be safe and remove line 39 since we don't need that else then?

ui/desktop/electron-app/src/helpers/app-updater.js
```diff
@@ -19,6 +20,31 @@ if (debug && process.env.APP_UPDATER_CURRENT_VERSION) {
...
+  if (nodeArchitecture === 'arm64') {
+    return nodeArchitecture;
+  }
+
+  try {
+    const runsOnRosetta = execSync(
+      'sysctl -in sysctl.proc_translated',
+    ).toString();
+
+    if (runsOnRosetta) {
```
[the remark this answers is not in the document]
calcaide: You are right, change it to `parseInt`

ui/desktop/electron-app/src/helpers/app-updater.js
```diff
@@ -19,6 +21,29 @@ if (debug && process.env.APP_UPDATER_CURRENT_VERSION) {
...
+
+  try {
+    const runsOnRosetta = parseInt(
+      execSync('sysctl -in sysctl.proc_translated'),
+    );
+
+    if (runsOnRosetta) {
+      return 'arm64';
+    } else {
+      return nodeArchitecture; // amd64
```
ZedLi, line 40: Can we move this to after the try catch so it always gets returned whether we error out or not? If it's arm64 it'll hit the normal return statement

ui/desktop/electron-app/src/helpers/app-updater.js
```diff
@@ -19,6 +21,28 @@ if (debug && process.env.APP_UPDATER_CURRENT_VERSION) {
...
+  try {
+    const runsOnRosetta = parseInt(
+      execSync('sysctl -in sysctl.proc_translated'),
+    );
+
+    if (runsOnRosetta) {
+      return 'arm64';
+    }
+  } catch (err) {
+    log.error(`returnArchitectureToUpdate error: ${err}`);
```
ZedLi, line 41:
```suggestion
    log.error(`returnArchitectureToUpdate: ${err}`);
```

The error in the message is redundant since we log it as an error

ZedLi (approved):
Looks great! Thanks for this change, tested on M1 mac and works as expected

ui/desktop/electron-app/src/helpers/app-updater.js
calcaide: Sure

ZedLi:
I just discovered this helper [method](https://www.electronjs.org/docs/latest/api/app#apprunningunderarm64translation-readonly-macos-windows) from electron which simplifies things quite a bit and also has the advantage of also working on windows. It looks like it does the [same thing](https://github.com/electron/electron/blob/540a42fb96c8e954e668f8c77e8cb059975ac72f/shell/browser/api/electron_api_app_mac.mm#L63) under the hood for mac at least.

Note that this is available on electron 28, it was previously just `app.runningUnderRosettaTranslation` and was [removed](https://www.electronjs.org/docs/latest/breaking-changes#removed-apprunningunderrosettatranslation)

calcaide:
> I just discovered this helper [method](https://www.electronjs.org/docs/latest/api/app#apprunningunderarm64translation-readonly-macos-windows) from electron which simplifies things quite a bit and also has the advantage of also working on windows. It looks like it does the [same thing](https://github.com/electron/electron/blob/540a42fb96c8e954e668f8c77e8cb059975ac72f/shell/browser/api/electron_api_app_mac.mm#L63) under the hood for mac at least.
> 
> Note that this is available on electron 28, it was previously just `app.runningUnderRosettaTranslation` and was [removed](https://www.electronjs.org/docs/latest/breaking-changes#removed-apprunningunderrosettatranslation)

@ZedLi Nice finding!! I will definitely apply this implementation because is way better. Way better = the method is mantained by electron, no need for CLI command + compatible with mac and windows.

ZedLi (approved):

PR #2331, opened 2024-06-12
test(e2e): Add check for multiple host sources

moduli:
This PR updates some Admin UI e2e tests to add/remove additional host sources to a target. 
- The existing test (CE only) was updated to add and remove a second host source to a target
- A second test (ENT only) mirrors the existing test, but uses SSH targets

## Testing
There are two tests modified in this PR
- One is a CE test that can be run on either Docker or AWS infra
- One is an ENT test that can be run on either Docker or AWS infra
```
# CE
yarn run e2e:ce:docker 

# Enterprise
yarn run e2e:ent:docker
```

https://hashicorp.atlassian.net/browse/ICU-14009


calcaide (approved):
Tested CE edition with no problems.

Tested Enterprise edition with initial issues, but you fixed very quickly!! now running no problems.

Thanks for adding these tests

PR #2332, opened 2024-06-12
chore: 🤖 upgrade electron

cameronperera:
✅ Closes: https://hashicorp.atlassian.net/browse/ICU-13697

# Description

I have updated electron to v31. I've gone through the electron breaking [changes](https://www.electronjs.org/docs/latest/breaking-changes) and fixed one breaking change that affected us.  

## How to Test
[Here's](https://github.com/hashicorp/boundary-ui-releases/actions/runs/9570586017) a build in our CI of this electron upgrade.

Confirm that the app runs. Note that this includes the client-daemon to cache-daemon changes which the currently CLI hasn't updated yet so you'll see errors for that.

## Checklist

### DIFF
--- a/ui/desktop/electron-app/package.json
+++ b/ui/desktop/electron-app/package.json
@@ -41,7 +41,7 @@
     "@electron-forge/maker-dmg": "^7.2.0",
     "@electron-forge/maker-squirrel": "^7.2.0",
     "@electron-forge/maker-zip": "^7.2.0",
-    "electron": "26.2.4"
+    "electron": "31.0.1"
   },
   "resolutions": {
     "highlight.js": "^10.4.1"
--- a/ui/desktop/electron-app/src/index.js
+++ b/ui/desktop/electron-app/src/index.js
@@ -125,7 +125,7 @@ const createWindow = (partition, closeWindowCB) => {
     browserWindow.loadURL(emberAppURL);
   });
 
-  browserWindow.webContents.on('crashed', () => {
+  browserWindow.webContents.on('render-process-gone', () => {
     console.log(
       'Your Ember app (or other code) in the main window has crashed.',
     );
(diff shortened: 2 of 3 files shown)


ui/desktop/electron-app/package.json
```diff
@@ -40,10 +40,11 @@
...
     "@electron-forge/maker-squirrel": "^7.2.0",
     "@electron-forge/maker-zip": "^7.2.0",
-    "electron": "26.2.4",
+    "electron": "28.3.3",
     "unzipper": "^0.11.5"
   },
   "resolutions": {
-    "highlight.js": "^10.4.1"
+    "highlight.js": "^10.4.1",
+    "node-abi": "3.64.0"
```
ZedLi, line 48: You shouldn't need a resolution here it looks like. Our `node-abi` versions can be brought up manually to 3.64.0 in the `yarn.lock` file without a resolution.

cameronperera: Right we have chatted about this before a while back. I will update this.

cameronperera: Updated. Let me know if I updated the correct parts.

ZedLi (approved):

ZedLi:
> When trying to upgrade to v29 and v30, I started running into node-gyp build issues.

I took a look into this, if you also upgrade `nan`, it should also fix the issue.

ZedLi (commented):
The `crashed` event was [removed](https://www.electronjs.org/docs/latest/breaking-changes#removed-crashed-event-on-webcontents-and-webview) which we'll have to update

ui/desktop/electron-app/src/index.js
```diff
@@ -125,7 +125,7 @@ const createWindow = (partition, closeWindowCB) => {
     browserWindow.loadURL(emberAppURL);
   });
 
-  browserWindow.webContents.on('crashed', () => {
+  browserWindow.webContents.on('render-process-gone', () => {
```
cameronperera, line 128: thanks for catching this change

calcaide (approved):

PR #2333, opened 2024-06-12
test(e2e): Add check for multiple credential sources

moduli:
This PR updates some Admin UI e2e tests to add/remove additional credential sources to a target. 
- The existing test (CE only) was updated to add and remove a second brokered credential source to a target
- A second test (ENT only) mirrors the existing test, but uses SSH targets and adds both brokered and injected credentials to a target
- 
## Testing
There are two tests modified in this PR
- One is a CE test that can be run on either Docker or AWS infra
- One is an ENT test that can be run on either Docker or AWS infra
```
# CE
yarn run e2e:ce:docker 

# Enterprise
yarn run e2e:ent:docker
```

https://hashicorp.atlassian.net/browse/ICU-14010

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -4,11 +4,41 @@ License text copyright (c) 2020 MariaDB Corporation Ab, All Rights Reserved.
 Parameters
 
 Licensor:             HashiCorp, Inc.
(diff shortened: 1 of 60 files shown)


calcaide (approved):
I tested CE & Enterprise with no issues.

Thanks for the amazing tests!!!

PR #2338, opened 2024-06-12
feat: 🎸 add No scopes add to roles/role/scopes route

cameronperera:
✅ Closes: https://hashicorp.atlassian.net/browse/ICU-13878

# Description

If a role has no grant scopes, we will display a simple message with an action to direct the user to add scopes. I will update the route once we have a route for it to navigate to.

Screenshot 2024-06-12 at 3 11 26 PM

## How to Test

Start a `boundary dev` instance and remove all grant scope ids via the CLI. Now you should see the "no scopes added" message.

## Checklist

### DIFF
--- a/addons/api/mirage/serializers/role.js
+++ b/addons/api/mirage/serializers/role.js
@@ -38,11 +38,6 @@ export default ApplicationSerializer.extend({
 
     if (!json.principals.length) delete json.principals;
 
-    // default grant scopes
-    if (json.grant_scope_ids.length === 0) {
-      json.grant_scope_ids = ['this'];
-    }
-
     return json;
   },
 });
--- a/ui/admin/app/templates/scopes/scope/roles/role/scopes.hbs
+++ b/ui/admin/app/templates/scopes/scope/roles/role/scopes.hbs
@@ -151,7 +151,7 @@
       <Hds::ApplicationState as |A|>
         <A.Header @title={{t 'resources.role.scope.messages.title'}} />
         <A.Body @text={{t 'resources.role.scope.messages.description'}} />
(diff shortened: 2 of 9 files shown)


addons/api/mirage/serializers/role.js
```diff
@@ -38,11 +38,6 @@ export default ApplicationSerializer.extend({
 
     if (!json.principals.length) delete json.principals;
 
-    // default grant scopes
```
cameronperera, line 41: Now that we can simply define the `grant_scope_ids` as `['this']` we don't need this serializer code anymore.

ui/admin/app/templates/scopes/scope/roles/role/scopes.hbs
```diff
@@ -147,6 +147,21 @@
           </B.Tr>
         </:body>
       </Hds::Table>
+    {{else}}
+      <Hds::ApplicationState as |A|>
+        <A.Header @title={{t 'resources.role.scope.messages.title'}} />
+        <A.Body @text={{t 'resources.role.scope.messages.description'}} />
+        {{#if (can 'create model' this.scope collection='roles')}}
```
lisbet-alvarez, line 154: I'm thinking that if the link is going to direct users to the manage-scopes route then maybe the permission we should check is can "set-grant-scopes" on a role?

cameronperera: Absolutely. I forgot to update the exact ability. Thanks, I will update now.

cameronperera: Fixed now.

lisbet-alvarez (approved):
Looks great!!

PR #2345, opened 2024-06-14
workflow action for copyright headers

DhariniJeeva:
# Description
This PR proposes an automated process to validate copyright headers. The job will throw an error when headers are missing, and the user must fix those for the pull request to be valid. 

**note**: I intentionally made the job fail for us to view how this would look like. 

screenshot

## Checklist


.github/workflows/copyright-validate.yaml
```diff
@@ -0,0 +1,19 @@
+name: Check Copyright Headers
+
+on:
+  # - pull_request [should we include for all types]
+  push:
```
DhariniJeeva, line 5: Right now, this is targeted at this specific branch push. But the idea is to run this job for every pull request

.github/workflows/copyright-validate.yaml
```diff
@@ -0,0 +1,19 @@
...
+      - gh-copyright-workflow
+jobs:
+  copywrite:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@a5ac7e51b41094c92402da3b24376905380afc29 # v4.1.6
+      - uses: hashicorp/setup-copywrite@32638da2d4e81d56a0764aa1547882fc4d209636 # v1.1.3
+        name: Setup Copywrite
+        with:
+          version: v0.16.4
```
ZedLi, line 16: Any reason we pin to this version?

DhariniJeeva: I noticed this is the version used in other products but we have `version: v0.18.0` mentioned in setup-copywrite [repo](https://github.com/hashicorp/setup-copywrite), I can change to this version, not sure if there's a way to get latest always instead of pinning?

ZedLi: I think it will pull the latest if you just don't include the version

DhariniJeeva: I removed the version and the checksum verification [step](https://github.com/hashicorp/setup-copywrite?tab=readme-ov-file#faq), which didn't seem necessary in our workflow

cameronperera (commented):
Looks good to me. Thanks for digging into this for us! Happy to approve once this is ready for review.

.github/workflows/copyright-validate.yaml
```diff
@@ -0,0 +1,19 @@
...
+    branches:
+      - gh-copyright-workflow
+jobs:
+  copywrite:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout Repo
+        uses: actions/checkout@a5ac7e51b41094c92402da3b24376905380afc29 # v4.1.6
+
+      - name: Install copywrite
```
calcaide, line 15: None blocking comment:
I will suggest just to define `uses` for checkout + copywrite, as [we have in other jobs](https://github.com/hashicorp/boundary-ui/blob/main/.github/workflows/monorepo-validate.yaml#L27-L28), and just use `name` for check header compliance.
I do suggest this approach because we do not perform operations within checkout action or install copywrite, we define the action.

In example:
```
- uses: actions/checkout@a5ac7e51b41094c92402da3b24376905380afc29 # v4.1.6
- uses: hashicorp/setup-copywrite@32638da2d4e81d56a0764aa1547882fc4d209636 # v1.1.3
- name: Check Header Compliance
        run: copywrite headers --plan

calcaide (approved):
I left a none blocking comment.

Happy to approve when this becomes a none draft PR. Thanks for taking care of this.

.github/workflows/copyright-validate.yaml
DhariniJeeva: yeah that's fair, I'll remove the labels

ZedLi (approved):

cameronperera (approved):

PR #2349, opened 2024-06-17
feat: 🎸 create get allTags method that includes tag type

cameronperera:
✅ Closes: https://hashicorp.atlassian.net/browse/ICU-13891

# Description

This is some of the initial api/model changes for worker tag improvements. We have a story that will need data organized in a specific way from how it is returned from the API. The `getAllTags` method will combine the config and api tags into a single list that also denotes their tag type. Here is the [design](https://www.figma.com/design/ZzeL7kAVjNENx6kzZutua3/Worker-Tags-Improvements?node-id=803-24752&t=EH3Wc78KnpSAImBa-0) of how this data will be used in a table.

Screenshot 2024-06-14 at 11 32 24 AM

## Checklist

### DIFF
--- a/addons/api/addon/models/worker.js
+++ b/addons/api/addon/models/worker.js
@@ -44,7 +44,7 @@ export default class WorkerModel extends GeneratedWorkerModel {
    * Returns the config tags as an array of key/value pair objects.
    * @type {[object]}
    */
-  getConfigTagList() {
+  get configTagList() {
     if (!this.config_tags) {
       return null;
     }
@@ -58,7 +58,7 @@ export default class WorkerModel extends GeneratedWorkerModel {
    * Returns the api tags as an array of key/value pair objects.
    * @type {[object]}
    */
-  getApiTagList() {
+  get apiTagList() {
     if (!this.api_tags) {
       return null;
     }
@@ -72,11 +72,8 @@ export default class WorkerModel extends GeneratedWorkerModel {
    * Returns all tags as an array of key/value pair objects with tag type.
    * @type {[object]}
    */
-  getAllTags() {
-    return [
-      ...(this.getConfigTagList() ?? []),
-      ...(this.getApiTagList() ?? []),
-    ];
+  get allTags() {
+    return [...(this.configTagList ?? []), ...(this.apiTagList ?? [])];
   }
 
   @attr({
--- a/addons/api/addon/models/worker.js
+++ b/addons/api/addon/models/worker.js
@@ -31,14 +31,24 @@ export default class WorkerModel extends GeneratedWorkerModel {
    * @type {number}
    */
   get tagCount() {
-    if (!this.config_tags) {
+    if (!this.config_tags && !this.api_tags) {
       return 0;
     }
 
-    return Object.values(this.config_tags).reduce(
-      (previousCount, currentTags) => previousCount + currentTags.length,
-      0,
-    );
+    const configTagCount = this.config_tags
+      ? Object.values(this.config_tags).reduce(
+          (previousCount, currentTags) => previousCount + currentTags.length,
+          0,
+        )
+      : 0;
+    const apiTagCount = this.api_tags
+      ? Object.values(this.api_tags).reduce(
+          (previousCount, currentTags) => previousCount + currentTags.length,
+          0,
+        )
+      : 0;
+
+    return configTagCount + apiTagCount;
   }
 
   /**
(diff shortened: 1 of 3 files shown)


addons/api/addon/models/worker.js
```diff
@@ -55,6 +65,34 @@ export default class WorkerModel extends GeneratedWorkerModel {
...
+   * @type {[object]}
+   */
+  getAllTags() {
+    return [
+      ...(this.getConfigTagList() || []).map((tag) => ({
+        ...tag,
+        type: 'config',
+      })),
+      ...(this.getApiTagList() || []).map((tag) => ({ ...tag, type: 'api' })),
+    ];
```
ZedLi, line 93: Is this needed? Can we just look at `canonical_tags` instead? That should be a union of unique config tags and api tags

addons/api/addon/models/worker.js
```diff
@@ -31,14 +31,24 @@ export default class WorkerModel extends GeneratedWorkerModel {
    * @type {number}
    */
   get tagCount() {
-    if (!this.config_tags) {
+    if (!this.config_tags && !this.api_tags) {
```
ZedLi, line 34: Same here, should we just have been looking at `canonical_tags` instead rather than adding them up ourselves?

addons/api/addon/models/worker.js
cameronperera: Just relying on `canonical_tags` doesn't give us insight into what kind of tag it is. Only api tags can be created or removed and we need the type to determine what actions the user can take for each tag. In the below API response, canonical_tags doesn't give me all the info I need.
```json
{
    "canonical_tags": {
        "boundary.cloud.hashicorp.com:managed": [
            "true"
        ],
        "key1": [
            "dev"
        ],
        "key2": [
            "prod",
            "test"
        ]
    },
    "config_tags": {
        "boundary.cloud.hashicorp.com:managed": [
            "true"
        ]
    },
    "api_tags": {
        "key1": [
            "dev"
        ],
        "key2": [
            "prod",
            "test"
        ]
    },
}
```

addons/api/addon/models/worker.js
cameronperera: For sure here we could use `canonical_tags`. I will update.

addons/api/addon/models/worker.js
ZedLi: I see, makes sense then.

addons/api/addon/models/worker.js
```diff
@@ -55,6 +65,34 @@ export default class WorkerModel extends GeneratedWorkerModel {
...
+   * Returns all tags as an array of key/value pair objects with tag type.
+   * @type {[object]}
+   */
+  getAllTags() {
+    return [
+      ...(this.getConfigTagList() || []).map((tag) => ({
+        ...tag,
+        type: 'config',
+      })),
+      ...(this.getApiTagList() || []).map((tag) => ({ ...tag, type: 'api' })),
```
ZedLi, line 81: Could the type have just been added in the `getApiTagList` and `getConfigTagList` instead of here and then you'd just only merge the two?

cameronperera: Originally I didn't want to modify the output of getConfigTagList but just adding `type` doesn't hurt anything. Thanks for the suggestion, I will update now.

addons/api/addon/models/worker.js
```diff
@@ -51,10 +50,35 @@ export default class WorkerModel extends GeneratedWorkerModel {
...
-      value.map((tag) => ({ key, value: tag })),
+      value.map((tag) => ({ key, value: tag, type: 'config' })),
+    );
+  }
+
+  /**
+   * Returns the api tags as an array of key/value pair objects.
+   * @type {[object]}
+   */
+  getApiTagList() {
```
ZedLi, line 61: Unrelated to your PR but looking at this I'm pretty sure I should have made `getConfigTagList` into a getter instead of a function lol

addons/api/addon/models/worker.js
```diff
@@ -27,15 +27,14 @@ export default class WorkerModel extends GeneratedWorkerModel {
   }
 
   /**
-   * Returns the number of config tags present on the worker.
+   * Returns the number of canonical tags present on the worker.
    * @type {number}
    */
   get tagCount() {
-    if (!this.config_tags) {
+    if (!this.canonical_tags) {
```
ZedLi, line 34: The only gotcha is I'm not sure if `canonical_tags` dedupes the same tag so I'm not sure if that will match ours if we display all api and config tags separately which will show the same tags

ZedLi (approved):

addons/api/addon/models/worker.js
cameronperera: It would only dedupe a tag if it was set as a config and api tag right? I did some testing locally and I cannot set an api tag that is already a config tag. This may not be an issue.

addons/api/addon/models/worker.js
cameronperera: Well I can fix that now, no worries

lisbet-alvarez (approved):
🙌

PR #2352, opened 2024-06-17
test(e2e): Add check to detach bucket from target

moduli:
This PR updates some Admin UI e2e tests to add a check to detach a storage bucket from a target.

## Testing
The tests that are affected by this include...
- Session Recording (AWS)
- Session Recording (MinIO)

To be able to run all of the tests, you'll need to use the boundary-enterprise and run tests against 2 different scenarios.
```
yarn run e2e:ent:docker

enos scenario launch e2e_ui_aws_ent builder:local
yarn run e2e:ent:aws
```

https://hashicorp.atlassian.net/browse/ICU-14076


calcaide (approved):
Run both `ent:docker` and `ent:aws` with no issues!! Thanks for this

PR #2355, opened 2024-06-18
Icu 13889 update tags column to action that opens hds flyout and add link to tags tab if more than 10 tags

cameronperera:
# Description

When viewing the workers table, you can now quickly view the tags on a worker with an Hds Flyout

https://github.com/hashicorp/boundary-ui/assets/29386339/d8534048-50ce-46e0-b763-94ca83620a99

Example of a key that is longer than 16 characters
Screenshot 2024-06-18 at 3 23 20 PM

## How to Test

Visit vercel build and view workers. You should be able to click on the tags count in a table row and view the tags for that specific worker. If there are 10 or more tags, it should only show 10 and navigate the user to the worker route.

## Checklist

### DIFF
--- a/ui/admin/app/controllers/scopes/scope/workers/index.js
+++ b/ui/admin/app/controllers/scopes/scope/workers/index.js
@@ -31,8 +31,7 @@ export default class ScopesScopeWorkersIndexController extends Controller {
   }
 
   /**
-   * Build the display name for a worker tag and will
-   * truncate the key if it's too long.
+   * Build the display name for a worker tag
    * @param {object} tag
    * @returns {string}
    */
--- a/ui/admin/app/controllers/scopes/scope/workers/index.js
+++ b/ui/admin/app/controllers/scopes/scope/workers/index.js
@@ -7,6 +7,7 @@ import Controller from '@ember/controller';
 import { inject as service } from '@ember/service';
 import { action } from '@ember/object';
 import { loading } from 'ember-loading';
+import { tracked } from '@glimmer/tracking';
 import { confirm } from 'core/decorators/confirm';
 import { notifySuccess, notifyError } from 'core/decorators/notify';
 
@@ -16,7 +17,41 @@ export default class ScopesScopeWorkersIndexController extends Controller {
   @service router;
   @service can;
 
+  // =attributes
+
+  @tracked selectedWorker;
+
+  /**
+   * Get the first 10 tags of the selected worker.
+   * @type {object[]}
+   */
+  get selectedWorkerTags() {
+    // only show the first 10 tags if there are more than 10
+    return this.selectedWorker.allTags.slice(0, 10);
+  }
+
+  /**
+   * Build the display name for a worker tag and will
+   * truncate the key if it's too long.
+   * @param {object} tag
+   * @returns {string}
+   */
+  tagDisplayName(tag) {
+    const key =
+      tag.key.length > 16 ? tag.key.substring(0, 16) + '...' : tag.key;
+    return `${key} = ${tag.value}`;
+  }
+
   // =actions
+
+  /**
+   * Toggle the tags flyout to display or hide the tags of a worker.
+   * @param {object} selectedWorker
+   */
+  @action
+  selectWorker(selectedWorker) {
+    this.selectedWorker = selectedWorker;
+  }
 
   /**
    * Rollback changes on workers.
--- a/ui/admin/app/controllers/scopes/scope/workers/index.js
+++ b/ui/admin/app/controllers/scopes/scope/workers/index.js
@@ -37,9 +37,7 @@ export default class ScopesScopeWorkersIndexController extends Controller {
    * @returns {string}
    */
   tagDisplayName(tag) {
-    const key =
-      tag.key.length > 16 ? tag.key.substring(0, 16) + '...' : tag.key;
-    return `${key} = ${tag.value}`;
+    return `${tag.key} = ${tag.value}`;
   }
 
   // =actions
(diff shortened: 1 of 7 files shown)


ui/admin/app/templates/scopes/scope/workers/index.hbs
```diff
@@ -145,6 +149,56 @@
...
+                      {{this.tagDisplayName B.data}}
+                    </Hds::Text::Code>
+                  </B.Td>
+                  <B.Td>
+                    <Hds::Badge @text={{B.data.type}} />
+                  </B.Td>
+                </B.Tr>
+              </:body>
+            </Hds::Table>
+            {{#if (eq this.selectedWorkerTags.length 10)}}
```
lisbet-alvarez, line 187: I think this should be slightly updated. The "View more tags" link should be displayed if there are more than 10 tags. So in the case that there are only 10 tags, the link will still be displayed.

ui/admin/app/controllers/scopes/scope/workers/index.js
```diff
@@ -16,8 +17,42 @@ export default class ScopesScopeWorkersIndexController extends Controller {
...
+    return this.selectedWorker.allTags.slice(0, 10);
+  }
+
+  /**
+   * Build the display name for a worker tag and will
+   * truncate the key if it's too long.
+   * @param {object} tag
+   * @returns {string}
+   */
+  tagDisplayName(tag) {
```
ZedLi, line 39: I don't think we should truncate the actual text. We should accomplish this with `text-overflow: ellipsis` since if you try and copy and paste the key now, it's the truncated text but if you do it with CSS, you still get the full text.

ui/admin/app/templates/scopes/scope/workers/index.hbs
cameronperera: Great point. Updated and will push up shortly.

ui/admin/app/controllers/scopes/scope/workers/index.js
cameronperera: Right, `text-overflow` is what we have done in the past. Thanks.

Yep that is a good point. With that in mind, should we wrap the whole key/value pair with `text-overflow` or the key and value separately?

ZedLi: It depends what we're going for, doing it separately means you can control how big the key and value are and independently overflow. I'm not sure if key or value might be more important to a user so I'm not sure what's a better UX but I think either could make sense

cameronperera: Ok. I will play around with wrapping each separately. I think there is value in always trying to show the key and value.

cameronperera: Updated. Let know what you think. Here is an example of what it looks like for the different cases: long key, long key and value, and long value.
Screenshot 2024-06-21 at 3 24 14 PM

lisbet-alvarez (approved):
Great work!!

ui/admin/app/controllers/scopes/scope/workers/index.js
```diff
@@ -16,8 +17,40 @@ export default class ScopesScopeWorkersIndexController extends Controller {
...
+   * @type {object[]}
+   */
+  get selectedWorkerTags() {
+    // only show the first 10 tags if there are more than 10
+    return this.selectedWorker.allTags.slice(0, 10);
+  }
+
+  /**
+   * Build the display name for a worker tag and will
+   * truncate the key if it's too long.
```
ZedLi, line 35: I think this comment needs to be updated now after the truncation logic was removed

ui/admin/app/styles/app.scss
```diff
@@ -1036,3 +1036,18 @@
...
+.view-more-tags {
+  margin-top: 1.5rem;
+}
+
+.tag-container {
+  display: flex;
+  max-width: 310px;
+}
+
+.tag-shortener {
```
ZedLi, line 1050: Shouldn't this have used `tag-container` or at least been nested inside it?

ui/admin/app/templates/scopes/scope/workers/index.hbs
```diff
@@ -145,6 +149,58 @@
...
+                (hash label=(t 'resources.worker.flyout.table.tag'))
+                (hash label=(t 'resources.worker.flyout.table.type'))
+              }}
+              @valign='middle'
+            >
+              <:body as |B|>
+                <B.Tr>
+                  <B.Td>
+                    <Hds::Text::Code class='tag-container'>
+                      <div class='tag-shortener'>
```
ZedLi, line 178: Was this div actually necessary? Couldn't we have just had the text in the `Hds::Text::Code`?

ui/admin/app/controllers/scopes/scope/workers/index.js
cameronperera: Thanks. Updated now.

ui/admin/app/templates/scopes/scope/workers/index.hbs
cameronperera: Nope not needed. I originally used a different approach that needed the div but changed it and didn't go back and remove it. Updated.

ui/admin/app/controllers/scopes/scope/workers/index.js
cameronperera: Screenshot 2024-06-24 at 3 42 02 PM
Sharing again to show this is still displaying how we want after style change.

ZedLi (approved):
Looks good!

PR #2356, opened 2024-06-18
fix: 🐛 Fix unzipping changing cli permissions

ZedLi:
# Description
This fixes the issue @calcaide reported in this [thread](https://hashicorp.slack.com/archives/C02CGB68P8S/p1717535140624269).

It turns out, our new unzipping library doesn't properly preserve permissions from the zip archive and removes the executable file permissions on the CLI. This is a known issue for this [library](https://github.com/ZJONSSON/node-unzipper/issues/216).

I swapped to `adm-zip` which can properly handle the original file permissions and also has no dependencies while still being maintained. 

Before:
image

After:
image

## How to Test
Delete your current CLI and let the script download a new one either through building or starting dev. Confirm that the CLI permissions are correct. 

## Checklist

### DIFF
--- a/ui/desktop/electron-app/config/cli.js
+++ b/ui/desktop/electron-app/config/cli.js
@@ -7,7 +7,7 @@ const fs = require('fs');
 const os = require('os');
 const path = require('path');
 const https = require('https');
-const unzipper = require('unzipper');
+const admZip = require('adm-zip');
 const { isMac, isWindows, isLinux } = require('../src/helpers/platform.js');
 
 const artifactDestination = path.resolve(__dirname, '..', 'cli');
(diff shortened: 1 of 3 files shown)


ui/desktop/electron-app/config/cli.js
```diff
@@ -69,12 +69,10 @@ const downloadArtifact = (version) => {
...
-  return fs
-    .createReadStream(artifactPath)
-    .pipe(unzipper.Extract({ path: destination }))
-    .on('close', () => {
-      console.log('Extract artifact to: ', destination);
-    });
+  const zip = new admZip(artifactPath);
+  // The first boolean flag is to overwrite files, the second is to preserve original
+  // file permissions which doesn't seem to be documented.
+  return zip.extractAllTo(destination, true, true);
```
ZedLi, line 75: It's not documented but [this method](https://github.com/cthackers/adm-zip/blob/05c25b123284cd0d3c8930995914ed09dbb8cf43/adm-zip.js#L598) has a boolean flag that lets us preserve the original permissions 🤷

calcaide (approved):
Thank you for taking care of this.

I did tested with the changes on main and within this PR and I can confirm permissions are fixed.

Before:
Screenshot 2024-06-18 at 1 28 11 PM

After:
Screenshot 2024-06-19 at 1 51 13 PM

PR #2360, opened 2024-06-21
chore: 🤖 Fix unzip permissions part 2: electric boogaloo

ZedLi:
# Description
The previous PR https://github.com/hashicorp/boundary-ui/pull/2356 fixed the permissions issue but had a memory overflow issue again in our build pipeline. Unfortunately it seems like that's a much harder issue to solve for so for now I just fixed the permissions issue from unzipping manually and reverted the other change. 

Here's a [workflow run](https://github.com/hashicorp/boundary-ui-releases/actions/runs/9620711437) to test with and confirm it builds correctly (keep in mind the CLI will still error out when calling the cache daemon currently due to the rename)

image

## How to Test
Delete your current CLI and let the script download a new one either through building or starting dev. Confirm that the CLI permissions are correct.

## Checklist

### DIFF
--- a/ui/desktop/electron-app/config/cli.js
+++ b/ui/desktop/electron-app/config/cli.js
@@ -7,7 +7,7 @@ const fs = require('fs');
 const os = require('os');
 const path = require('path');
 const https = require('https');
-const admZip = require('adm-zip');
+const unzipper = require('unzipper');
 const { isMac, isWindows, isLinux } = require('../src/helpers/platform.js');
 
 const artifactDestination = path.resolve(__dirname, '..', 'cli');
@@ -69,10 +69,12 @@ const downloadArtifact = (version) => {
 const extract = (artifactPath, destination) => {
   if (!fs.existsSync(destination)) fs.mkdirSync(destination);
 
-  const zip = new admZip(artifactPath);
-  // The first boolean flag is to overwrite files, the second is to preserve original
-  // file permissions which doesn't seem to be documented.
-  return zip.extractAllTo(destination, true, true);
+  return fs
+    .createReadStream(artifactPath)
+    .pipe(unzipper.Extract({ path: destination }))
+    .on('close', () => {
+      console.log('Extract artifact to: ', destination);
+    });
 };
 
 module.exports = {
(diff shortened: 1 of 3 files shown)


DhariniJeeva (approved):

calcaide (approved):

calcaide:
Tested manually.

I tested deleting the CLI folder before starting the desktop, and tested without deleting the CLI folder. Both worked as expected and no permission issues are present.

Thanks for the fix.

ui/desktop/electron-app/config/cli.js
```diff
@@ -66,13 +66,31 @@ const downloadArtifact = (version) => {
...
+          .stream()
+          .pipe(
+            fs.createWriteStream(path.join(destination, file.path), {
+              // Creating a new file stream sets the permission bits on the file
+              // to node's default. We need to set the permission bits ourselves.
+              // We could just force the mode to 0o755, but we can also just convert it.
+              // This should also work on windows as the permission bits will still be set
+              // correctly but they just won't have any effect.
+              // Taken from here: https://github.com/thejoshwolfe/yauzl/issues/101#issuecomment-448073570
+              mode: file.externalFileAttributes >>> 16,
```
lisbet-alvarez, line 86: clarifying question: so if we shift the `externalFileAttributes` field by 16 the result will always be the binary version of `0o755`?

ZedLi: Not necessarily, it depends on the original externalFileAttributes. The issue was the `externalFileAttributes` wasn't in a format that could be passed into mode directly (I think it's encoded?) to set the permissions, and this was to mainly convert it.

The original file permissions for the boundary executable was `755` which it ended up converting into but if you notice in the screenshot, the license file ended up as `644` (I'm not sure if it ends up as `644` bc it was invalid or if that was the original permissions but we really only care about the CLI).

I could have also alternatively only extracted out the CLI but I would then have to get the right name depending if we're on windows or not but that didn't seem necessary

lisbet-alvarez: Thank you for the explanation, that helped a lot!

PR #2362, opened 2024-06-24
Icu 13892 create tags tab route and view

cameronperera:
# Description

This PR adds the new `/workers/worker/tags` route and table view for displaying worker tags.

NOTE: This does not include removing tags from the details view of a worker yet. That will come in a later PR.

https://github.com/hashicorp/boundary-ui/assets/29386339/5d6c388a-5226-4139-bc35-dc4ba17a60e0

## How to Test

Using the vercel deployment:
Navigate to a worker that has tags and click on the new "Tags" tab of a worker.

## Checklist

### DIFF
--- a/addons/core/translations/resources/en-us.yaml
+++ b/addons/core/translations/resources/en-us.yaml
@@ -914,6 +914,7 @@ worker:
     tag_count: '{tagCount, plural, =1 {# tag} other {# tags}}'
   tags:
     title_plural: Tags
+    description: Tags are used to define the type of worker and to make it distinguishable amongst other workers.
   form:
     cluster_id:
       label: Boundary Cluster ID
--- a/addons/core/translations/resources/en-us.yaml
+++ b/addons/core/translations/resources/en-us.yaml
@@ -912,6 +912,9 @@ worker:
     ip_address: IP Address
     release_version: Release Version
     tag_count: '{tagCount, plural, =1 {# tag} other {# tags}}'
+    key: Key
+    value: Value
+    type: Type
   tags:
     title_plural: Tags
     description: Tags are used to define the type of worker and to make it distinguishable amongst other workers.
--- a/addons/core/translations/resources/en-us.yaml
+++ b/addons/core/translations/resources/en-us.yaml
@@ -912,6 +912,8 @@ worker:
     ip_address: IP Address
     release_version: Release Version
     tag_count: '{tagCount, plural, =1 {# tag} other {# tags}}'
+  tags:
+    title_plural: Tags
   form:
     cluster_id:
       label: Boundary Cluster ID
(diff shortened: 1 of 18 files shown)


addons/core/translations/resources/en-us.yaml
```diff
@@ -912,6 +912,12 @@ worker:
     ip_address: IP Address
     release_version: Release Version
     tag_count: '{tagCount, plural, =1 {# tag} other {# tags}}'
+    key: Key
```
cameronperera, line 915: I am realizing now I have some duplication of translations between this PR and my other [PR](https://github.com/hashicorp/boundary-ui/pull/2355). I will correct this once one of them is merged.

ui/admin/app/components/workers/worker/actions/index.hbs
```diff
@@ -0,0 +1,20 @@
+{{!
+  Copyright (c) HashiCorp, Inc.
+  SPDX-License-Identifier: BUSL-1.1
+}}
+
+<Rose::Dropdown
```
DhariniJeeva, line 6: Not a blocker- Can we replace this with HDS?

DhariniJeeva (approved):
Looks good

ui/admin/app/components/workers/worker/actions/index.hbs
cameronperera: I was thinking of making it an HDS dropdown but I don't think any of our other "manage" dropdown uses have been updated to HDS either. I opted to not change it so it doesn't stand out. Might be an easy change across the board though because I don't think we have that many "manage" dropdowns.

cameronperera: I see that we are updating these in other places like multi-scope grants. I will update it here as well.

lisbet-alvarez (approved):
Great work!! 🙌

ui/admin/app/templates/scopes/scope/workers/worker/tags.hbs
```diff
@@ -0,0 +1,62 @@
...
+            <B.Td>
+              <Hds::Text::Code @tag='pre'>{{B.data.value}}</Hds::Text::Code>
+            </B.Td>
+            <B.Td>
+              <Hds::Badge @text={{B.data.type}} />
+            </B.Td>
+          </B.Tr>
+        </:body>
+      </Hds::Table>
+    {{/if}}
```
lisbet-alvarez, line 58: non-blocking: are we going to add the empty app state for this tab in a later PR?

cameronperera: Absolutely. I was going to add that when I created the "create new tags" route and view.

PR #2366, opened 2024-06-25
test(e2e): Add test for search and pagination

moduli:
This PR adds an e2e test for search and pagination. The test creates 15 targets and interacts with a variety of elements on the targets list page to search and paginate through the list of targets.

Additionally, support for HTTP requests to Boundary was added. The new test makes use of HTTP requests to quickly generate the 15 targets. 

# Testing
One test was added that works for both CE and Enterprise
```
# CE
yarn run e2e:ce:docker
```

https://hashicorp.atlassian.net/browse/ICU-14123


calcaide (approved):
Let's gooo!!!! I run these e2e without issue!! awesome coverage addition, thanks for this 🥳

PR #2368, opened 2024-06-28
Icu 13895 create create new tags route and view

cameronperera:
# Description

This PR adds the worker tag creation workflow.

~The test, `cannot visit worker create tags without proper permissions`, is still failing and I am not sure why. I am opening this PR for review while I dig into this more. I will push up a fix asap.~
Test has been fixed!

NOTE: I have not added the "no tags" view yet. That will be a later PR.

https://github.com/hashicorp/boundary-ui/assets/29386339/1ecdaf23-e20f-41d1-87ff-a34f0ee5b740

Payload sent to `POST /v1/workers/<workerID>:set-worker-tags`
Screenshot 2024-07-02 at 2 24 51 PM

## How to Test

Visit a worker in the vercel deployment. From here you can create new tags via the "Manage" dropdown and create new tags.

## Checklist

### DIFF
--- a/addons/api/addon/models/worker.js
+++ b/addons/api/addon/models/worker.js
@@ -93,7 +93,6 @@ export default class WorkerModel extends GeneratedWorkerModel {
 
   @attr({
     description: 'The api tags set for the worker.\nOutput only.',
-    readOnly: true,
   })
   api_tags;
 
--- a/addons/api/addon/models/worker.js
+++ b/addons/api/addon/models/worker.js
@@ -6,6 +6,8 @@
 import GeneratedWorkerModel from '../generated/models/worker';
 import { attr } from '@ember-data/model';
 
+export const TYPE_WORKER_PKI = 'pki';
+
 export const TAG_TYPE_CONFIG = 'config';
 export const TAG_TYPE_API = 'api';
 
@@ -26,7 +28,7 @@ export default class WorkerModel extends GeneratedWorkerModel {
    * @type {boolean}
    */
   get isPki() {
-    return this.type === 'pki';
+    return this.type === TYPE_WORKER_PKI;
   }
 
   /**
--- a/addons/api/addon/models/worker.js
+++ b/addons/api/addon/models/worker.js
@@ -117,4 +117,15 @@ export default class WorkerModel extends GeneratedWorkerModel {
       },
     });
   }
+
+  /**
+   * Method to set api tags on the worker.
+   * @param {object} apiTags
+   * @return {Promise}
+   */
+  setApiTags(apiTags) {
+    return this.save({
+      adapterOptions: { method: 'set-worker-tags', apiTags },
+    });
+  }
 }
--- a/addons/api/mirage/generated/factories/worker.js
+++ b/addons/api/mirage/generated/factories/worker.js
@@ -5,6 +5,7 @@
 
 import { Factory } from 'miragejs';
 import { faker } from '@faker-js/faker';
+import { TYPE_WORKER_PKI } from 'api/models/worker';
 
 /**
  * GeneratedWorkerModelFactory
@@ -13,7 +14,7 @@ import { faker } from '@faker-js/faker';
 export default Factory.extend({
   name: () => faker.word.words(),
   description: () => faker.word.words(),
-  type: 'pki',
+  type: TYPE_WORKER_PKI,
   created_time: () => faker.date.recent(),
   updated_time: () => faker.date.recent(),
   version: () => faker.number.int(),
--- a/addons/api/mirage/generated/factories/worker.js
+++ b/addons/api/mirage/generated/factories/worker.js
@@ -13,6 +13,7 @@ import { faker } from '@faker-js/faker';
 export default Factory.extend({
   name: () => faker.word.words(),
   description: () => faker.word.words(),
+  type: 'pki',
   created_time: () => faker.date.recent(),
   updated_time: () => faker.date.recent(),
   version: () => faker.number.int(),
--- a/addons/core/translations/resources/en-us.yaml
+++ b/addons/core/translations/resources/en-us.yaml
@@ -904,6 +904,7 @@ worker:
   actions:
     register: Register Worker
     delete: Remove Worker
+    create_tags: Create new tags
   messages:
     welcome:
       title: Welcome to Workers
--- a/addons/core/translations/resources/en-us.yaml
+++ b/addons/core/translations/resources/en-us.yaml
@@ -919,6 +919,10 @@ worker:
     type: Type
   tags:
     title_plural: Tags
+    create:
+      title: Create new tags
+      description: Create new api worker tags. You can only add API tags in the admin UI. Config tags are added in the worker config file.
+      success: Tag(s) successfully created
     description: Tags are used to define the type of worker and to make it distinguishable amongst other workers.
   flyout:
     title: Tags in {workerName}
--- a/addons/core/translations/resources/en-us.yaml
+++ b/addons/core/translations/resources/en-us.yaml
@@ -922,7 +922,7 @@ worker:
     title_plural: Tags
     create:
       title: Create new tags
-      description: Create new api worker tags. You can only add API tags in the admin UI. Config tags are added in the worker config file.
+      description: Create new api worker tags. You can only add API tags in the admin UI or CLI. Config tags are added in the worker config file.
       success: Tag(s) successfully created
     description: Tags are used to define the type of worker and to make it distinguishable amongst other workers.
   flyout:
--- a/ui/admin/app/components/form/worker/create-tags/index.js
+++ b/ui/admin/app/components/form/worker/create-tags/index.js
@@ -15,60 +15,6 @@ export default class FormWorkerCreateTagsIndexComponent extends Component {
   @service confirm;
   @service intl;
 
-  // =attributes
-
-  hasRemovedExitHandler = false;
-
-  constructor() {
-    super(...arguments);
-    // this is added to prevent the user from accidentally navigating away
-    // from the page when they have unsaved changes
-    this.addExitHandler();
-  }
-
-  // adds an exit handler to prevent the user from accidentally navigating away
-  addExitHandler() {
-    this.router.on('routeWillChange', this, this.confirmTransition);
-  }
-
-  // removes the exit handler
-  removeExitHandler() {
-    if (!this.hasRemovedExitHandler) {
-      this.router.off('routeWillChange', this, this.confirmTransition);
-      this.hasRemovedExitHandler = true;
-    }
-  }
-
-  // checks if the transition is aborted and if there are any tags
-  confirmTransition(transition) {
-    if (transition.isAborted) return;
-
-    if (this.args.apiTags.length) {
-      this.confirmExit(transition);
-    }
-  }
-
-  // displays a confirmation dialog to the user
-  async confirmExit(transition) {
-    transition.abort();
-    try {
-      await this.confirm.confirm(this.intl.t('questions.abandon-confirm'), {
-        title: 'titles.abandon-confirm',
-        confirm: 'actions.discard',
-      });
-      this.args.apiTags = this.args.apiTags.splice(0, this.args.apiTags.length);
-      transition.retry();
-    } catch (e) {
-      // if user denies, do nothing
-    }
-  }
-
-  // removes the exit handler when the component is destroyed
-  willDestroy() {
-    super.willDestroy(...arguments);
-    this.removeExitHandler();
-  }
-
   // =actions
 
   /**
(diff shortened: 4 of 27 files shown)


ui/admin/tests/acceptance/workers/create-tags-test.js
```diff
@@ -0,0 +1,144 @@
...
+    worker: null,
+    tags: null,
+    createTags: null,
+  };
+
+  hooks.beforeEach(function () {
+    instances.scopes.global = this.server.create('scope', { id: 'global' });
+    const scope = instances.scopes.global;
+    instances.worker = this.server.create('worker', {
+      scopeId: scope.id,
```
ZedLi, line 50: This needs to be the entire scope model passed in, not just the ID

```suggestion
      scope,
```

cameronperera: Thanks for spotting that. I updated it in a few other places as well.

ui/admin/tests/unit/controllers/scopes/scope/workers/new-test.js
```diff
@@ -57,14 +57,4 @@ module('Unit | Controller | scopes/scope/workers/new', function (hooks) {
 
     assert.notEqual(worker.name, 'test');
   });
-
-  test('createWorkerLed action saves changes on the specified model', async function (assert) {
```
cameronperera, line 61: This test was removed because it is just a pass through for the model method `createWorkerLed` and did not do anything else.

addons/api/mirage/generated/factories/worker.js
```diff
@@ -13,6 +13,7 @@ import { faker } from '@faker-js/faker';
 export default Factory.extend({
   name: () => faker.word.words(),
   description: () => faker.word.words(),
+  type: 'pki',
```
lisbet-alvarez, line 16: Non-blocking: For this type value should we import it from the model instead?

ui/admin/app/components/form/worker/create-tags/index.js
```diff
@@ -0,0 +1,114 @@
...
+      } else {
+        existingApiTags[key] = Array.from(
+          new Set([...existingApiTags[key], ...values]),
+        );
+      }
+    });
+
+    this.args.submit(existingApiTags);
+  }
+}
```
lisbet-alvarez, line 114: i think we usually try to add comments to methods and actions. can we do that in this file as well.

ui/admin/app/components/form/worker/create-tags/index.js
```diff
@@ -0,0 +1,114 @@
...
+  }
+
+  async confirmExit(transition) {
+    transition.abort();
+    try {
+      await this.confirm.confirm(this.intl.t('questions.abandon-confirm'), {
+        title: 'titles.abandon-confirm',
+        confirm: 'actions.discard',
+      });
+      this.apiTags = A([]);
```
lisbet-alvarez, line 65: question for understanding: why do we need to set back to empty array before retrying the transition ?

ui/admin/tests/unit/controllers/scopes/scope/workers/worker/create-tags-test.js
```diff
@@ -0,0 +1,21 @@
...
+    setupTest(hooks);
+
+    test('it exists', function (assert) {
+      let controller = this.owner.lookup(
+        'controller:scopes/scope/workers/worker/create-tags',
+      );
+      assert.ok(controller);
+    });
+  },
+);
```
lisbet-alvarez, line 21: i think the test cases for the actions are missing here, only if possible to test.

addons/api/mirage/generated/factories/worker.js
cameronperera: Great point. I will update that.

ui/admin/app/components/form/worker/create-tags/index.js
```diff
@@ -0,0 +1,114 @@
...
+export default class FormWorkerCreateTagsIndexComponent extends Component {
+  // =services
+
+  @service router;
+  @service confirm;
+  @service intl;
+
+  // =attributes
+
+  @tracked apiTags = A([]);
```
ZedLi, line 31: We shouldn't be using `A` anymore as we've switched to `tracked-built-ins`

ui/admin/app/components/form/worker/create-tags/index.js
```diff
@@ -0,0 +1,114 @@
...
+
+class Tag {
+  @tracked key;
+  @tracked value;
+
+  constructor(key, value) {
+    this.key = key;
+    this.value = value;
+  }
+}
```
ZedLi, line 20: This looks the same as a previous usage in worker, shouldn't it be pulled out to be shared?

ui/admin/app/components/form/worker/create-tags/index.js
ZedLi, line 1: I'm not sure I understand why this logic is living in this `create-tags` form component. Why didn't this logic happen in the route this form exists on since it seems like this component shouldn't need to know about transition logic?

Couldn't you have used the `willTransition` event and just done the logic there?

ui/admin/app/components/form/worker/create-tags/index.js
```diff
@@ -0,0 +1,114 @@
...
+  @action
+  save() {
+    this.removeExitHandler();
+
+    if (this.apiTags.isEmpty) {
+      this.router.transitionTo('scopes.scope.workers.worker.tags');
+      return;
+    }
+
+    const existingApiTags = this.args.model.api_tags || {};
```
ZedLi, line 98:
```suggestion
    const existingApiTags = this.args.model.api_tags ?? {};
```

ui/admin/app/components/form/worker/create-tags/index.js
```diff
@@ -0,0 +1,114 @@
...
+    this.apiTags.forEach((tag) => {
+      let key = tag.key;
+      let values = tag.value.split(',');
+
+      if (!existingApiTags[key]) {
+        existingApiTags[key] = values;
+      } else {
+        existingApiTags[key] = Array.from(
+          new Set([...existingApiTags[key], ...values]),
+        );
```
ZedLi, line 108:
```suggestion
        existingApiTags[key] = [...new Set([...existingApiTags[key], ...values])];
```

I think you can also spread `Set`

cameronperera: Updated.

ui/admin/app/components/form/worker/create-tags/index.js
cameronperera: Updated.

ui/admin/app/components/form/worker/create-tags/index.js
cameronperera: Updated. I also changed it slightly to not trigger the dialog when the user clicks cancel. That was the same UX I saw in other places where the dialog would only appear when trying to navigate away by ways other than clicking cancel.

ui/admin/app/components/form/worker/create-tags/index.js
```diff
@@ -0,0 +1,64 @@
...
+import { inject as service } from '@ember/service';
+import { action } from '@ember/object';
+import Tag from '../tag';
+
+export default class FormWorkerCreateTagsIndexComponent extends Component {
+  // =services
+
+  @service router;
+  @service confirm;
+  @service intl;
```
lisbet-alvarez, line 16: i think confirm and intl services can be removed now.

ui/admin/tests/unit/controllers/scopes/scope/workers/worker/create-tags-test.js
cameronperera: Thanks. Added.

addons/api/addon/models/worker.js
```diff
@@ -96,7 +98,6 @@ export default class WorkerModel extends GeneratedWorkerModel {
 
   @attr({
     description: 'The api tags set for the worker.\nOutput only.',
-    readOnly: true,
```
lisbet-alvarez, line 99: non-blocking: im wondering why remove `readOnly: true` if this field can only be saved via the setApiTags method ?

lisbet-alvarez (approved):
🙌 👍

addons/api/addon/models/worker.js
cameronperera: I removed it since it can be updated via the API. However, I am not sure if the way we update it changes that. Happy to change it back and/or continue discussing this.

addons/core/translations/resources/en-us.yaml
```diff
@@ -919,6 +920,10 @@ worker:
     type: Type
   tags:
     title_plural: Tags
+    create:
+      title: Create new tags
+      description: Create new api worker tags. You can only add API tags in the admin UI. Config tags are added in the worker config file.
```
ZedLi, line 925: Is it right to say `you can only add API tags in the admin UI` since you can add it through the CLI too?

ui/admin/app/components/form/worker/create-worker-led/index.js
```diff
@@ -8,16 +8,7 @@ import { tracked } from '@glimmer/tracking';
 import { inject as service } from '@ember/service';
 import { action } from '@ember/object';
 import { A } from '@ember/array';
-
-class Tag {
-  @tracked key;
-  @tracked value;
```
ZedLi, line 14: The new tag class doesn't use tracked fields anymore which changed the functionality in the create worker form, the tags don't update immediately anymore

ui/admin/app/routes/scopes/scope/workers/worker/create-tags.js
```diff
@@ -0,0 +1,55 @@
...
+  constructor() {
+    super(...arguments);
+    /**
+     * If user attempts to navigate away from unsaved changes, the user is
+     * asked to confirm that they would like to discard the changes.  If the
+     * user chooses discard, changes apiTags are cleared and the
+     * transition is retried.  If the user cancels discard, the transition is
+     * aborted.
+     */
+    this.router.on('routeWillChange', async (transition) => {
```
ZedLi, line 26: Is there a reason you need to use the `router` instead of `willTransition` when you're already in the route?

ui/admin/app/routes/scopes/scope/workers/worker/create-tags.js
```diff
@@ -0,0 +1,55 @@
...
+      const fromName = transition?.from?.name;
+      const toName = transition?.to?.name;
+      // This will prevent the confirmation dialog from showing when the user
+      // saves or cancels the form.
+      if (toName === 'scopes.scope.workers.worker.tags') {
```
ZedLi, line 31: What happens if the user enters the create tags form from the tags page and then tries to go back? It seems like this will prevent the check from happening for unsaved data.

ZedLi (changes requested):

addons/core/translations/resources/en-us.yaml
cameronperera: Good catch. I will update this.

ui/admin/app/components/form/worker/create-worker-led/index.js
cameronperera: Oops. Thanks for catching that. Tested and working.

ui/admin/app/routes/scopes/scope/workers/worker/create-tags.js
cameronperera: As Hoobastank would say: "The reason is me".

Updating now.

ui/admin/app/controllers/scopes/scope/workers/worker/create-tags.js
```diff
@@ -41,8 +41,6 @@ export default class ScopesScopeWorkersWorkerCreateTagsController extends Contro
    */
   @action
   cancel() {
-    // Clear the apiTags array when canceling
-    this.apiTags = new TrackedArray([]);
```
ZedLi: Isn't this behavior different? Don't we normally just rollback the changes and let the user cancel instead of a popup appearing when trying to cancel?

ui/admin/app/routes/scopes/scope/workers/worker/create-tags.js
```diff
@@ -0,0 +1,34 @@
...
+
+export default class ScopesScopeWorkersWorkerCreateTagsRoute extends Route {
+  // =services
+
+  @service confirm;
+  @service intl;
+
+  @action
+  async willTransition(transition) {
+    if (this.controller.get('apiTags').length) {
```
ZedLi, line 20: I think if we access the controller in a route, we're supposed to still call `controllerFor` according to the [rule](https://github.com/ember-cli/eslint-plugin-ember/blob/master/docs/rules/no-controller-access-in-routes.md)

cameronperera: Sorry, not sure why I changed that. I used controllerFor in my first interation.

ui/admin/app/controllers/scopes/scope/workers/worker/create-tags.js
cameronperera: Yep that is right. Changing back.

ZedLi (approved):
Looks great!

lisbet-alvarez (approved):

PR #2371, opened 2024-07-01
feat: 🎸 add search and filtering to scopes tab in a role

lisbet-alvarez:
✅ Closes: https://hashicorp.atlassian.net/browse/ICU-13886

# Description

https://github.com/hashicorp/boundary-ui/assets/107949262/99366dd2-d41a-400c-b174-07f11cb6f3cc

## Checklist

### DIFF
--- a/ui/admin/app/controllers/scopes/scope/roles/role/manage-scopes/index.js
+++ b/ui/admin/app/controllers/scopes/scope/roles/role/manage-scopes/index.js
@@ -14,6 +14,12 @@ export default class ScopesScopeRolesRoleManageScopesIndexController extends Con
 
   @service router;
 
+  // =attributes
+
+  queryParams = ['showCheckIcon'];
+
+  showCheckIcon = false;
+
   // =actions
 
   /**
@@ -27,14 +33,14 @@ export default class ScopesScopeRolesRoleManageScopesIndexController extends Con
   @notifySuccess('notifications.add-success')
   async setGrantScopes(role, grantScopeIDs) {
     await role.setGrantScopes(grantScopeIDs);
-    await this.router.replaceWith('scopes.scope.roles.role.scopes');
+    this.router.replaceWith('scopes.scope.roles.role.scopes');
   }
 
   /**
    * Redirect to role scopes as if nothing ever happened.
    */
   @action
-  async cancel() {
-    await this.router.replaceWith('scopes.scope.roles.role.scopes');
+  cancel() {
+    this.router.replaceWith('scopes.scope.roles.role.scopes');
   }
 }
--- a/ui/admin/app/controllers/scopes/scope/roles/role/manage-scopes/index.js
+++ b/ui/admin/app/controllers/scopes/scope/roles/role/manage-scopes/index.js
@@ -16,8 +16,6 @@ export default class ScopesScopeRolesRoleManageScopesIndexController extends Con
 
   // =attributes
 
-  queryParams = ['showCheckIcon'];
-
   showCheckIcon = false;
 
   // =actions
--- a/ui/admin/app/templates/scopes/scope/roles/role/scopes.hbs
+++ b/ui/admin/app/templates/scopes/scope/roles/role/scopes.hbs
@@ -21,137 +21,192 @@
       <DocLink @doc='role' @iconSize='large' />
     </h2>
     <p>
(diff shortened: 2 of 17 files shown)


ui/admin/app/templates/scopes/scope/roles/role/scopes.hbs
```diff
@@ -21,137 +21,192 @@
...
+              </B.Td>
+              <B.Td>
+                {{#if B.data.scope.isGlobal}}
+                  <Hds::Link::Inline
+                    @route='scopes.scope'
+                    @model={{B.data.scope.id}}
+                    @icon='globe'
+                    @iconPosition='leading'
+                  >
+                    {{~if
```
lisbet-alvarez, line 154: the `~` inside the curly braces is used to control whitespace in hbs. It removed the extra space caused by the newline.

ui/admin/app/templates/scopes/scope/roles/role/scopes.hbs
```diff
@@ -21,137 +21,192 @@
...
+                  <Hds::Link::Inline
+                    @route='scopes.scope'
+                    @model={{B.data.scope.id}}
+                    @icon='org'
+                    @iconPosition='leading'
+                  >
+                    {{~if B.data.scope.name B.data.scope.name B.data.scope.id}}
+                  </Hds::Link::Inline>
+                {{else}}
+                  {{if B.data.scope.name B.data.scope.name B.data.scope.id}}
```
cameronperera, line 170: non-blocking: Why do we need the `~` on line 167 but not here?

cameronperera (approved):
LGTM

ui/admin/app/templates/scopes/scope/roles/role/scopes.hbs
cameronperera: TIL

DhariniJeeva (approved):

ui/admin/app/templates/scopes/scope/roles/role/scopes.hbs
ZedLi: Is there CSS or something else that is making the whitespace in the HTML not collapse?

lisbet-alvarez: hmm good point, im not sure, ill investigate

ui/admin/app/templates/scopes/scope/roles/role/scopes.hbs
lisbet-alvarez: good point, tho now that i think about it i don't think this else is necessary at all since in every case the parent scope will either be org or global

ui/admin/app/templates/scopes/scope/roles/role/scopes.hbs
lisbet-alvarez: looked into it and it seems to be an ember [hbs bug](https://github.com/emberjs/ember.js/issues/19603#issuecomment-1121406604) with rendering extra whitespace. And there wasn't any css (that i could find) that would cause this whitespace issue.

ZedLi: Wow interesting, good find! Thanks for the info

ui/admin/app/controllers/scopes/scope/roles/role/manage-scopes/index.js
```diff
@@ -14,6 +14,12 @@ export default class ScopesScopeRolesRoleManageScopesIndexController extends Con
 
   @service router;
 
+  // =attributes
+
+  queryParams = ['showCheckIcon'];
```
ZedLi, line 19: I'm wondering if you need this queryParam in the first place. It feels kind of out of place since it's being used to just pass a flag to know if something should be displayed after saving. (Not sure how you feel about it, I could be convinced)

It looks like you only set it in the controllers on transition, is it possible to just set a property on the parent controller instead by using `@controller` and resetting when necessary?

lisbet-alvarez: Great point, I updated to be just a property instead!

ZedLi (approved):
Looks great!

PR #2376, opened 2024-07-09
Add a postMake arch conversion from x64 to amd64

calcaide:
# Description

This PR introduces changes to the DC binary built process. These changes should be merged along with this boundary-ui-releases PR. Otherwise, the pipeline runs will fail.

Changes:
When the `arch` is x64 (electron-forge default) change it to amd64.
When the `platform` is win32 (electron-forge default) change it to `windows`. Otherwise, if win32 is present in the artifact name, the signature process fails because it just accepts windows.

## Checklist

### DIFF
--- a/ui/desktop/electron-app/config/forge.config.js
+++ b/ui/desktop/electron-app/config/forge.config.js
@@ -103,6 +103,9 @@ module.exports = {
           fs.mkdirSync(destination, { recursive: true });
         // Copy artifacts
         artifacts.forEach(async (artifact) => {
+          if (arch === 'x64') {
+            arch = 'amd64';
+          }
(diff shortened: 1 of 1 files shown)


ui/desktop/electron-app/config/forge.config.js
```diff
@@ -103,6 +103,9 @@ module.exports = {
           fs.mkdirSync(destination, { recursive: true });
         // Copy artifacts
         artifacts.forEach(async (artifact) => {
+          if (arch === 'x64') {
```
calcaide, line 106: I add this arch "rewrite" at the `postMake` hook, so shouldn't affect any other part of the build process.

ZedLi (approved):

cameronperera (approved):

PR #2377, opened 2024-07-09
Icu 13897 create remove tag modal

cameronperera:
# Description

Add a modal that removes a single worker tag

https://github.com/user-attachments/assets/801311a3-2ab6-45b2-98bb-7f85b242714a

Screenshot 2024-07-09 at 5 35 20 PM

## How to Test

Navigate to the tags tab of a worker that has api tags. Click on the action at the end of that row and select remove tag. Type `REMOVE` to enable the Remove button. Tag should be removed from the list.

## Checklist

### DIFF
--- a/addons/api/addon/serializers/worker.js
+++ b/addons/api/addon/serializers/worker.js
@@ -60,18 +60,4 @@ export default class WorkerSerializer extends ApplicationSerializer {
       api_tags: apiTags,
     };
   }
-
-  normalize(typeClass, hash, ...rest) {
-    const normalizedHash = structuredClone(hash);
-    const normalized = super.normalize(typeClass, normalizedHash, ...rest);
-
-    // api_tags is not returned by the API if it is null so we need to set it
-    // to an empty object if it is not present to prevent not updating the
-    // worker correctly if the api_tags are removed.
-    if (!normalized.data.attributes.api_tags) {
-      normalized.data.attributes.api_tags = {};
-    }
-
-    return normalized;
-  }
 }
--- a/addons/api/addon/serializers/worker.js
+++ b/addons/api/addon/serializers/worker.js
@@ -60,4 +60,18 @@ export default class WorkerSerializer extends ApplicationSerializer {
       api_tags: apiTags,
     };
   }
+
+  normalize(typeClass, hash, ...rest) {
+    const normalizedHash = structuredClone(hash);
+    const normalized = super.normalize(typeClass, normalizedHash, ...rest);
+
+    // api_tags is not returned by the API if it is null so we need to set it
+    // to an empty object if it is not present to prevent not updating the
+    // worker correctly if the api_tags are removed.
+    if (!normalized.data.attributes.api_tags) {
+      normalized.data.attributes.api_tags = {};
+    }
+
+    return normalized;
+  }
 }
--- a/addons/api/addon/serializers/worker.js
+++ b/addons/api/addon/serializers/worker.js
@@ -19,11 +19,19 @@ export default class WorkerSerializer extends ApplicationSerializer {
     let serialized = super.serialize(...arguments);
     const workerGeneratedAuthToken =
       snapshot?.adapterOptions?.workerGeneratedAuthToken;
-    if (workerGeneratedAuthToken)
+    if (workerGeneratedAuthToken) {
       serialized = this.serializeWithGeneratedToken(
         snapshot,
         workerGeneratedAuthToken,
       );
+    }
+    const apiTags = snapshot?.adapterOptions?.apiTags;
+    if (apiTags) {
+      serialized = this.serializeWithApiTags(snapshot, apiTags);
+    } else {
+      // remove api_tags from payload if not set
+      delete serialized.api_tags;
+    }
     return serialized;
   }
 
@@ -37,6 +45,19 @@ export default class WorkerSerializer extends ApplicationSerializer {
     return {
       scope_id: snapshot.attr('scope').scope_id,
       worker_generated_auth_token: workerGeneratedAuthToken,
+    };
+  }
+
+  /**
+   * Returns a payload with api_tags and version serialized.
+   * @param {Snapshot} snapshot
+   * @param {object} apiTags
+   * @return {object}
+   */
+  serializeWithApiTags(snapshot, apiTags) {
+    return {
+      version: snapshot.attr('version'),
+      api_tags: apiTags,
     };
   }
 }
--- a/addons/api/mirage/config.js
+++ b/addons/api/mirage/config.js
@@ -681,13 +681,20 @@ function routes() {
       const id = idMethod.split(':')[0];
       const method = idMethod.split(':')[1];
       const worker = workers.find(id);
-      let updatedAttrs = {};
+      let updatedAttrs = {
+        version: attrs.version,
+        apiTags: worker.api_tags,
+      };
 
       if (method === 'remove-worker-tags') {
-        updatedAttrs = {
-          version: attrs.version,
-          api_tags: attrs.api_tags,
-        };
+        const key = Object.keys(attrs.apiTags)[0];
+        const value = attrs.apiTags[key][0];
+        updatedAttrs.apiTags[key] = updatedAttrs.apiTags[key].filter((tag) => {
+          return tag !== value;
+        });
+        if (updatedAttrs.apiTags[key].length === 0) {
+          delete updatedAttrs.apiTags[key];
+        }
       }
 
       return worker.update(updatedAttrs);
--- a/addons/api/mirage/config.js
+++ b/addons/api/mirage/config.js
@@ -693,9 +693,6 @@ function routes() {
         updatedAttrs.apiTags[key] = updatedAttrs.apiTags[key].filter((tag) => {
           return tag !== value;
         });
-        if (updatedAttrs.apiTags[key].length === 0) {
-          delete updatedAttrs.apiTags[key];
-        }
       }
 
       if (method === 'set-worker-tags') {
--- a/addons/api/mirage/config.js
+++ b/addons/api/mirage/config.js
@@ -674,6 +674,25 @@ function routes() {
     });
     return workers.create(newWorker.attrs);
   });
+  this.post(
+    '/workers/:idMethod',
+    function ({ workers }, { params: { idMethod } }) {
+      const attrs = this.normalizedRequestAttrs();
+      const id = idMethod.split(':')[0];
+      const method = idMethod.split(':')[1];
+      const worker = workers.find(id);
+      let updatedAttrs = {};
+
+      if (method === 'remove-worker-tags') {
+        updatedAttrs = {
+          version: attrs.version,
+          api_tags: attrs.api_tags,
+        };
+      }
+
+      return worker.update(updatedAttrs);
+    },
+  );
 
   // storage-buckets
   this.get(
--- a/addons/api/mirage/config.js
+++ b/addons/api/mirage/config.js
@@ -699,7 +699,7 @@ function routes() {
       }
 
       if (method === 'set-worker-tags') {
-        updatedAttrs.apiTags = attrs.apitTags
+        updatedAttrs.apiTags = attrs.apiTags;
       }
 
       return worker.update(updatedAttrs);
--- a/addons/api/mirage/config.js
+++ b/addons/api/mirage/config.js
@@ -684,19 +684,21 @@ function routes() {
       const worker = workers.find(id);
       let updatedAttrs = {
         version: attrs.version,
-        apiTags: worker.api_tags,
+        api_tags: worker.api_tags,
       };
 
       if (method === 'remove-worker-tags') {
         const key = Object.keys(attrs.apiTags)[0];
         const value = attrs.apiTags[key][0];
-        updatedAttrs.apiTags[key] = updatedAttrs.apiTags[key].filter((tag) => {
-          return tag !== value;
-        });
+        updatedAttrs.api_tags[key] = updatedAttrs.api_tags[key].filter(
+          (tag) => {
+            return tag !== value;
+          },
+        );
       }
 
       if (method === 'set-worker-tags') {
-        updatedAttrs.apiTags = attrs.apiTags;
+        updatedAttrs.api_tags = attrs.apiTags;
       }
 
       return worker.update(updatedAttrs);
--- a/addons/api/mirage/config.js
+++ b/addons/api/mirage/config.js
@@ -16,6 +16,7 @@ import initializeMockIPC from './scenarios/ipc';
 import makeBooleanFilter from './helpers/bexpr-filter';
 import { faker } from '@faker-js/faker';
 import { asciicasts } from './data/asciicasts';
+import { TYPE_WORKER_PKI } from 'api/models/worker';
 
 const isTesting = environmentConfig.environment === 'test';
 
@@ -669,7 +670,7 @@ function routes() {
 
     // This POST only takes in a token so we need to generate a random worker to return
     const newWorker = this.create('worker', {
-      type: 'pki',
+      type: TYPE_WORKER_PKI,
       scope: globalScope,
     });
     return workers.create(newWorker.attrs);
@@ -695,6 +696,10 @@ function routes() {
         if (updatedAttrs.apiTags[key].length === 0) {
           delete updatedAttrs.apiTags[key];
         }
+      }
+
+      if (method === 'set-worker-tags') {
+        updatedAttrs.apiTags = attrs.apitTags
       }
 
       return worker.update(updatedAttrs);
--- a/addons/api/mirage/serializers/worker.js
+++ b/addons/api/mirage/serializers/worker.js
@@ -16,7 +16,7 @@ export default ApplicationSerializer.extend({
     if (json.api_tags) {
       const keys = Object.keys(model.api_tags);
       keys.forEach((key) => {
-        if (json.api_tags[key].length === 0) {
+        if (json.api_tags[key]?.length === 0) {
           delete json.api_tags[key];
         }
       });
--- a/addons/api/mirage/serializers/worker.js
+++ b/addons/api/mirage/serializers/worker.js
@@ -7,4 +7,21 @@ import ApplicationSerializer from './application';
 
 export default ApplicationSerializer.extend({
   modelName: 'worker',
+
+  _hashForModel(model) {
+    const json = ApplicationSerializer.prototype._hashForModel.apply(
+      this,
+      arguments,
+    );
+    if (json.api_tags) {
+      const keys = Object.keys(model.api_tags);
+      keys.forEach((key) => {
+        if (json.api_tags[key].length === 0) {
+          delete json.api_tags[key];
+        }
+      });
+    }
+
+    return json;
+  },
 });
--- a/addons/api/tests/unit/models/worker-test.js
+++ b/addons/api/tests/unit/models/worker-test.js
@@ -6,6 +6,7 @@
 import { module, test } from 'qunit';
 import { setupTest } from 'ember-qunit';
 import setupMirage from 'ember-cli-mirage/test-support/setup-mirage';
+import { TYPE_WORKER_PKI } from 'api/models/worker';
 
 module('Unit | Model | worker', function (hooks) {
   setupTest(hooks);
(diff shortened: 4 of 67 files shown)


addons/api/mirage/config.js
```diff
@@ -682,13 +682,24 @@ function routes() {
...
+      };
+
+      if (method === 'remove-worker-tags') {
+        const key = Object.keys(attrs.apiTags)[0];
+        const value = attrs.apiTags[key][0];
+        updatedAttrs.apiTags[key] = updatedAttrs.apiTags[key].filter((tag) => {
+          return tag !== value;
+        });
+        if (updatedAttrs.apiTags[key].length === 0) {
+          delete updatedAttrs.apiTags[key];
```
DhariniJeeva, line 697: should this be handled in the serializer?

addons/api/mirage/config.js
```diff
@@ -682,13 +682,24 @@ function routes() {
...
+        }
+      }
 
       if (method === 'set-worker-tags') {
-        updatedAttrs = {
-          version: attrs.version,
-          apiTags: attrs.apiTags,
-        };
+        updatedAttrs.apiTags = attrs.apitTags
```
DhariniJeeva, line 702: do we not need version? 🤔

addons/core/translations/resources/en-us.yaml
```diff
@@ -925,6 +925,15 @@ worker:
       description: Create new api worker tags. You can only add API tags in the admin UI or CLI. Config tags are added in the worker config file.
       success: Tag(s) successfully created
     description: Tags are used to define the type of worker and to make it distinguishable amongst other workers.
+    remove:
+      title: Remove worker tag?
+      success: Tag successfully removed
+      warning: will be removed but removing these may affect certain targets
+      confirm:
+        label: Confirm removal
+        help: 'Enter “REMOVE” to confirm — this is case-sensitive.'
```
DhariniJeeva, line 934: just curious- why are we doing this pattern over showing a confirmation modal with just delete or cancel as that's what HDS [follows](https://helios.hashicorp.design/components/modal#critical).

ui/admin/app/styles/app.scss
```diff
@@ -1054,3 +1054,14 @@
   overflow: hidden;
   text-overflow: ellipsis;
 }
+
+#remove-tag-modal {
```
DhariniJeeva, line 1058: not opposed to this - but I don't think we use IDs for styles in our repo, curious why ID over classnames

ui/admin/app/controllers/scopes/scope/workers/worker/tags.js
```diff
@@ -4,10 +4,75 @@
...
+   * Opens the remove tag modal.
+   * @param {string} action
+   * @param {object} apiTag
+   */
+  @action
+  toggleModal(action, apiTag = null) {
+    if (action === 'remove') {
+      this.modals.remove = !this.modals.remove;
+      this.apiTagToRemove = apiTag;
+      this.removalConfirmation = null;
```
DhariniJeeva, line 75: do we need to set `removalConfirmation`  to null in this action?  we set this to `null` on page load and we only reassign on input change, so I wonder if this is needed here

addons/api/mirage/config.js
cameronperera: Thanks for the suggestion. I forgot we had mirage serializers. I have updated this. Please let me know if what I added is correct.

ui/admin/app/styles/app.scss
cameronperera: Absolutely right. Updated.

addons/api/mirage/config.js
cameronperera: Version is being set on line 686 above.

ui/admin/app/controllers/scopes/scope/workers/worker/tags.js
cameronperera: I added it here because the user can open the remove modal, type something in then cancel or close the modal. Now if they open the modal again the same input will still be there.

addons/core/translations/resources/en-us.yaml
cameronperera: That is a great question that I do not have an answer for. I can talk with design about this.

ZedLi:
I think you need to refresh after removing tags since it's not showing up as being removed

https://github.com/hashicorp/boundary-ui/assets/5783847/f729c956-5d78-46d9-9aa5-1fd022240ff4

cameronperera:
> I think you need to refresh after removing tags since it's not showing up as being removed
> 
>  Screen.Recording.2024-07-10.at.6.05.37.PM.mov

Thanks for catching that. I will add a normalizer to set `api_tags` to an empty object if not present in the API response.

cameronperera:
Updated to handle no api_tags in the API response
https://github.com/hashicorp/boundary-ui/assets/29386339/b9090612-3f58-4682-b6e9-0b82093f6393

addons/api/tests/unit/models/worker-test.js
```diff
@@ -71,6 +71,38 @@ module('Unit | Model | worker', function (hooks) {
...
     await model.setApiTags(tags);
   });
+  
+  test('it has a `removeApiTags` method that targets a specific POST API', async function (assert) {
+    assert.expect(1);
+    const workerId = 'w_123';
+    const tags = {
+      tag1: ['value1'],
+    };
+    this.server.post(
```
calcaide, line 81: What is the purpose of mocking a `post` request within this test?

addons/api/tests/unit/models/worker-test.js
```diff
@@ -71,6 +71,38 @@ module('Unit | Model | worker', function (hooks) {
     const model = store.peekRecord('worker', workerId);
     await model.setApiTags(tags);
   });
+  
+  test('it has a `removeApiTags` method that targets a specific POST API', async function (assert) {
```
calcaide, line 75: I am a bit confuse by this test, is a unit test but we are mocking quite a bit of stuff and it has no assertions? I am sure I am missing context, but can you ellaborate a bit the reasoning for this test?

ui/admin/app/styles/app.scss
```diff
@@ -1047,3 +1047,14 @@
   overflow: hidden;
   text-overflow: ellipsis;
 }
+
+.remove-tag-modal {
```
calcaide, line 1051: What is the purpose of adding styles this specific? 

Let me explain my question, I might have assume things. 
This modal looks like a confirmation modal, is that correct? if so, should we do it generic? or use one of the confirmation modals we already have?

I am fine if we have a new confirmation modal,but I would do it generic and not specific to `remove-tag`.

addons/api/tests/unit/models/worker-test.js
cameronperera: We are mocking the `post` so we can assert that the correct body is being added to the payload when we call `/workers/WORKER_ID:remove-worker-tags`.

ui/admin/app/templates/scopes/scope/workers/worker/tags.hbs
```diff
@@ -61,11 +62,80 @@
...
+        </M.Header>
+        <M.Body>
+          <Hds::Text::Body @color='primary'>
+            {{this.tagDisplayName}}
+            {{t 'resources.worker.tags.remove.warning'}}
+          </Hds::Text::Body>
+          {{! template-lint-disable no-autofocus-attribute }}
+          <Hds::Form::TextInput::Field
+            name='removal-confirmation'
+            autofocus={{true}}
```
calcaide, line 106: What is the purpose of the autofocus here? as per your video in the PR description, once the dialog opens, the field is NOT autofocusing (you have to click on it).

addons/api/tests/unit/models/worker-test.js
cameronperera: The assertion is on line 85. We are mocking it here so we can assert the correct payload is added to the `post` request. Does this help explain what is going on here?

ui/admin/app/templates/scopes/scope/workers/worker/tags.hbs
cameronperera: Great question. I forgot to update the video. I was looking at some [HDS documentation for modals](https://helios.hashicorp.design/components/modal?tab=code#form-within-a-modal-dialog) and they recommend adding the `autofocus` attribute if there is a form in the modal. So I added it after the original video was made. Sorry for the confusion.

cameronperera: Video updated!

addons/api/tests/unit/models/worker-test.js
calcaide: I answered this one on the comment above, my bad mixxing this 2 comments but they are related.

ui/admin/app/templates/scopes/scope/workers/worker/tags.hbs
calcaide: Awesome, now makes sense 😃  Thank you @cameronperera

ui/admin/app/templates/scopes/scope/workers/worker/tags.hbs
```diff
@@ -61,11 +62,80 @@
...
+              }}</F.Label>
+            <F.HelperText>{{t
+                'resources.worker.tags.remove.confirm.help'
+              }}</F.HelperText>
+          </Hds::Form::TextInput::Field>
+        </M.Body>
+        <M.Footer as |F|>
+          <Hds::ButtonSet>
+            <Hds::Button
+              type='button'
```
lisbet-alvarez, line 122: is type necessary to be specified ?

ui/admin/app/templates/scopes/scope/workers/worker/tags.hbs
```diff
@@ -61,11 +62,80 @@
...
+              disabled={{(not this.isRemovalConfirmed)}}
```
lisbet-alvarez, line 123:
```suggestion
              disabled={{not this.isRemovalConfirmed}}
```

DhariniJeeva (approved):
looks good!

ui/admin/app/controllers/scopes/scope/workers/worker/tags.js
```diff
@@ -4,10 +4,73 @@
...
   tagTypeConfig = TAG_TYPE_CONFIG;
+  tagTypeApi = TAG_TYPE_API;
+
+  @tracked removeModal = false;
+  @tracked removalConfirmation = null;
+  @tracked modalTag = null;
+
+  /**
+   * Build the display name for a worker tag
+   * @returns {string}
```
lisbet-alvarez, line 31: For getters we usually use `@type` instead of `@returns`
```suggestion
   * @type {string}
```

ui/admin/app/templates/scopes/scope/workers/worker/tags.hbs
```diff
@@ -38,6 +38,7 @@
           (hash label=(t 'resources.worker.table.key'))
           (hash label=(t 'resources.worker.table.value'))
           (hash label=(t 'resources.worker.table.type'))
+          (hash label=(t 'titles.actions') isVisuallyHidden=true width='60px')
```
lisbet-alvarez, line 41: ah great catch on adding the `isVisuallyHidden` option!

addons/core/translations/resources/en-us.yaml
cameronperera: Talked with design and product. Since removing a worker tag can also affect a target that might be using that tag.

ui/admin/app/controllers/scopes/scope/workers/worker/tags.js
cameronperera: Right, thank you

ui/admin/app/templates/scopes/scope/workers/worker/tags.hbs
cameronperera: Not at all. I originally had this as a submit but changed it back. However, `button` is the default, so it has been removed.

ui/admin/app/styles/app.scss
cameronperera: I see what you mean here. I do think this is our only instance of a modal like this (needing to confirm before allowing the user to remove). Regardless, I see your point that it doesn't need to be specific to tags. I will update the class name.

addons/api/tests/unit/models/worker-test.js
calcaide: @cameronperera 

For the sake of this PR. I will be ok as it is, there are previous tests within this same file following same approach. But definitely I think we should bring this topic to discussion in tech time.

addons/api/addon/serializers/worker.js
```diff
@@ -60,4 +60,18 @@ export default class WorkerSerializer extends ApplicationSerializer {
...
   }
+
+  normalize(typeClass, hash, ...rest) {
+    const normalizedHash = structuredClone(hash);
+    const normalized = super.normalize(typeClass, normalizedHash, ...rest);
+
+    // api_tags is not returned by the API if it is null so we need to set it
+    // to an empty object if it is not present to prevent not updating the
+    // worker correctly if the api_tags are removed.
+    if (!normalized.data.attributes.api_tags) {
```
ZedLi, line 71: Should we have this for the other tag fields?

Also, we have a `emptyArrayIfMissing` option for array fields on models to handle this, should we have something similar to `emptyObjectIfMissing` as well?

addons/api/mirage/serializers/worker.js
```diff
@@ -7,4 +7,21 @@ import ApplicationSerializer from './application';
...
 export default ApplicationSerializer.extend({
   modelName: 'worker',
+
+  _hashForModel(model) {
+    const json = ApplicationSerializer.prototype._hashForModel.apply(
+      this,
+      arguments,
+    );
+    if (json.api_tags) {
+      const keys = Object.keys(model.api_tags);
```
ZedLi, line 17: I get an error here in mirage:

image

ui/admin/app/controllers/scopes/scope/workers/worker/tags.js
```diff
@@ -4,10 +4,73 @@
...
+  }
+
+  /**
+   * Opens the remove tag modal.
+   * @param {string} action
+   * @param {object} apiTag
+   */
+  @action
+  toggleModal(action, apiTag = null) {
+    if (action === 'remove') {
```
ZedLi, line 70: What's the purpose of this `action`? Are you trying to share this with edit? Why not just use two separate functions instead?

ui/admin/app/controllers/scopes/scope/workers/worker/tags.js
```diff
@@ -4,10 +4,73 @@
...
 
+  // =services
+  @service intl;
+
+  // =attributes
+
   tagTypeConfig = TAG_TYPE_CONFIG;
+  tagTypeApi = TAG_TYPE_API;
+
+  @tracked removeModal = false;
```
ZedLi, line 25: Is this needed? Can't you just key off of `modalTag` to know whether or not the modal should be shown?

ui/admin/app/templates/scopes/scope/workers/worker/tags.hbs
```diff
@@ -61,11 +62,78 @@
...
+                'resources.worker.tags.remove.confirm.help'
+              }}</F.HelperText>
+          </Hds::Form::TextInput::Field>
+        </M.Body>
+        <M.Footer as |F|>
+          <Hds::ButtonSet>
+            <Hds::Button
+              disabled={{not this.isRemovalConfirmed}}
+              @color='critical'
+              @text='Remove'
```
calcaide, line 124: I think this should be a translated string?

ui/admin/app/templates/scopes/scope/workers/worker/tags.hbs
```diff
@@ -61,11 +62,78 @@
...
+              {{on 'click' this.removeApiTag}}
+            />
+            <Hds::Button
+              @color='secondary'
+              @text='Cancel'
```
calcaide, line 129: I think this should be a translated string?

addons/api/mirage/serializers/worker.js
cameronperera: What exactly triggered this error?

ui/admin/app/templates/scopes/scope/workers/worker/tags.hbs
```diff
@@ -61,11 +62,78 @@
                 <Hds::Badge @text={{B.data.type}} />
               {{/if}}
             </B.Td>
+            <B.Td>
+              {{#if (eq B.data.type this.tagTypeApi)}}
+                <Hds::Dropdown as |dd|>
+                  <dd.ToggleIcon
+                    @icon='more-horizontal'
+                    @text='Overflow Options'
```
calcaide, line 70: This should be a translated string

ui/admin/app/controllers/scopes/scope/workers/worker/tags.js
cameronperera: Yes the plan was to use `toggleModal` for both as they do the same thing (except one toggles remove and one toggles edit) but happy to split them up.

ui/admin/app/controllers/scopes/scope/workers/worker/tags.js
cameronperera: If we just had one modal that would be possible but since we will have 2 modals on this screen, we need to specify which one to open.

ui/admin/app/controllers/scopes/scope/workers/worker/tags.js
calcaide: Agreed with Zed, if this is to share `remove` and `edit` actions, I will create two separate methods for those, that deal with the edit or removal state of the tags, and rely on `toggleModal` to deal with the modal state.

ui/admin/app/templates/scopes/scope/workers/worker/tags.hbs
cameronperera: This one actually isn't seen by the user and is only there because it is required. Here is [Hds documentation](https://helios.hashicorp.design/components/dropdown?tab=code#toggleicon-as-overflow-menu) on its use. Happy to create a translation for this but decided not to since it isn't seen.

ui/admin/app/templates/scopes/scope/workers/worker/tags.hbs
cameronperera: Fixed.

ui/admin/app/templates/scopes/scope/workers/worker/tags.hbs
cameronperera: Fixed.

ui/admin/app/templates/scopes/scope/workers/worker/tags.hbs
ZedLi: It says it's still used for aria-label which means screen reader users will still be using it no?

addons/api/mirage/serializers/worker.js
ZedLi: trying to save an api_tag when there are no api_tags

ui/admin/app/templates/scopes/scope/workers/worker/tags.hbs
calcaide: I would vote, unless is an internal or machine attribute (such as `data-test-*`), everything else that we send for rendering I will suggest being under intl.

cameronperera: Translation added.

ui/admin/app/controllers/scopes/scope/workers/worker/tags.js
cameronperera: Updated so it is specific to just the remove modal.

ui/admin/app/controllers/scopes/scope/workers/worker/tags.js
```diff
@@ -4,10 +4,70 @@
...
+    this.modalTag = null;
+    this.removalConfirmation = null;
+  }
+
+  /**
+   * Opens the remove tag modal.
+   * @param {object} apiTag
+   */
+  @action
+  toggleRemoveModal(apiTag = null) {
```
ZedLi, line 68: Is there a reason you need to default to `null`?

ZedLi (approved):
Looks good!

ui/admin/app/controllers/scopes/scope/workers/worker/tags.js
cameronperera: I set it to null by default for when closing but if we are always setting it to whatever the new tag is, this actually isn't needed. I will remove it.

lisbet-alvarez (approved):

PR #2384, opened 2024-07-12
Icu 13894 create edit tags modal

cameronperera:
# Description

Add the ability to edit worker tags and show an empty state message and action when there are no tags

Edit tags:
https://github.com/user-attachments/assets/345074fb-c357-4e17-b911-5c6fc1db61c8

No tags on a worker:
Screenshot 2024-07-12 at 2 33 59 PM

## Checklist

### DIFF
--- a/ui/admin/app/controllers/scopes/scope/workers/worker/tags.js
+++ b/ui/admin/app/controllers/scopes/scope/workers/worker/tags.js
@@ -24,7 +24,10 @@ export default class ScopesScopeWorkersWorkerTagsController extends Controller {
 
   @tracked removeModal = false;
   @tracked removalConfirmation = null;
+  @tracked editModal = false;
   @tracked modalTag = null;
+  @tracked editKey = null;
+  @tracked editValue = null;
 
   /**
    * Build the display name for a worker tag
@@ -44,6 +47,39 @@ export default class ScopesScopeWorkersWorkerTagsController extends Controller {
     );
   }
 
+  updateApiTags() {
+    const currentApiTags = structuredClone(this.model.api_tags);
+    const { key, value } = this.modalTag;
+    const editedValues = this.editValue.split(',').map((value) => value.trim());
+
+    // remove the old value
+    currentApiTags[key] = currentApiTags[key].filter((val) => val !== value);
+
+    // remove key if value is empty array
+    if (currentApiTags[key].length === 0) {
+      delete currentApiTags[key];
+    }
+
+    // add the new value under the new key
+    if (this.editKey in currentApiTags) {
+      // if the new key exist, add the new values to
+      // its array if not already present
+      if (!currentApiTags[this.editKey].includes(this.editValue)) {
+        // currentApiTags[this.editKey].push(this.editValue);
+        editedValues.forEach((value) => {
+          if (!currentApiTags[this.editKey].includes(value)) {
+            currentApiTags[this.editKey].push(value);
+          }
+        });
+      }
+    } else {
+      // add to api tags if does not exist
+      currentApiTags[this.editKey] = editedValues;
+    }
+
+    return currentApiTags;
+  }
+
   /**
    * Removes selected tag from the worker.
    */
@@ -56,8 +92,20 @@ export default class ScopesScopeWorkersWorkerTagsController extends Controller {
       [this.modalTag.key]: [this.modalTag.value],
     });
     this.removeModal = false;
-    this.modalTag = null;
     this.removalConfirmation = null;
+  }
+
+  /**
+   * Edits selected tag on the worker
+   */
+  @action
+  @loading
+  @notifyError(({ message }) => message, { catch: true })
+  @notifySuccess('resources.worker.tags.edit.success')
+  async editApiTag() {
+    const apiTags = this.updateApiTags();
+    await this.model.setApiTags(apiTags);
+    this.editModal = false;
   }
 
   /**
@@ -69,5 +117,17 @@ export default class ScopesScopeWorkersWorkerTagsController extends Controller {
     this.removeModal = !this.removeModal;
     this.modalTag = apiTag;
     this.removalConfirmation = null;
+  }
+
+  /**
+   * Opens the edit tag modal.
+   * @param {object} apiTag
+   */
+  @action
+  toggleEditModal(apiTag) {
+    this.editModal = !this.editModal;
+    this.modalTag = apiTag;
+    this.editKey = apiTag?.key;
+    this.editValue = apiTag?.value;
   }
 }
--- a/ui/admin/app/styles/app.scss
+++ b/ui/admin/app/styles/app.scss
@@ -1025,18 +1025,28 @@
   text-overflow: ellipsis;
 }
 
-.confirmation-modal {
-  .hds-button-set {
-    margin-bottom: 0;
-  }
-
-  .hds-form-field--layout-vertical {
-    margin-top: 1.5rem;
-    margin-bottom: 0;
-  }
-}
-
 .worker-tags {
+  .confirmation-modal {
+    .hds-button-set {
+      margin-bottom: 0;
+    }
+
+    .hds-form-field--layout-vertical {
+      margin-top: 1.5rem;
+      margin-bottom: 0;
+    }
+  }
+
+  .edit-modal {
+    .hds-button-set {
+      margin-bottom: 0;
+    }
+
+    // div:nth-child(2) span {
+    //   margin-bottom: 1.5rem;
+    // }
+  }
+
   .hds-application-state {
     margin-top: 1.5rem;
   }
--- a/ui/admin/app/styles/app.scss
+++ b/ui/admin/app/styles/app.scss
@@ -1035,3 +1035,9 @@
     margin-bottom: 0;
   }
 }
+
+.worker-tags {
+  .hds-application-state {
+    margin-top: 1.5rem;
+  }
+}
--- a/ui/admin/app/templates/scopes/scope/workers/worker/tags.hbs
+++ b/ui/admin/app/templates/scopes/scope/workers/worker/tags.hbs
@@ -30,7 +30,7 @@
     />
   </page.actions>
 
-  <page.body>
+  <page.body class='worker-tags'>
     {{#if @model.allTags}}
       <Hds::Table
         @model={{@model.allTags}}
@@ -83,6 +83,18 @@
           </B.Tr>
         </:body>
       </Hds::Table>
+    {{else}}
+      <Hds::ApplicationState data-test-no-tags as |A|>
+        <A.Header @title={{t 'resources.worker.tags.messages.no-tags.title'}} />
+        <A.Body @text={{t 'resources.worker.tags.messages.no-tags.body'}} />
+        <A.Footer as |F|>
+          <F.LinkStandalone
+            @icon='plus'
+            @text={{t 'resources.worker.tags.actions.add'}}
+            @route='scopes.scope.workers.worker.create-tags'
+          />
+        </A.Footer>
+      </Hds::ApplicationState>
     {{/if}}
 
     {{#if this.removeModal}}
--- a/ui/admin/app/templates/scopes/scope/workers/worker/tags.hbs
+++ b/ui/admin/app/templates/scopes/scope/workers/worker/tags.hbs
@@ -71,6 +71,12 @@
                     @hasChevron={{false}}
                     @size='small'
                   />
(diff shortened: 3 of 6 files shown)


ui/admin/app/styles/app.scss
```diff
@@ -1025,13 +1025,29 @@
...
 
-  .hds-form-field--layout-vertical {
+  .edit-modal {
+    .hds-button-set {
+      margin-bottom: 0;
+    }
+
+    // div:nth-child(2) span {
+    //   margin-bottom: 1.5rem;
+    // }
```
DhariniJeeva, line 1047: is this needed?

ui/admin/app/styles/app.scss
```diff
@@ -1025,13 +1025,29 @@
...
+  }
+
+  .hds-application-state {
     margin-top: 1.5rem;
-    margin-bottom: 0;
```
DhariniJeeva, line 1035: does this change affect anywhere else we use app state component?

ui/admin/tests/acceptance/workers/tags-test.js
```diff
@@ -114,6 +135,84 @@ module('Acceptance | workers | worker | tags', function (hooks) {
 
     await click(API_TAG_ACTION_SELECTOR);
 
-    assert.dom(API_TAG_REMOVE_ACTION_SELECTOR).doesNotExist();
+    // Asserts first button is for editing and there is on second button
```
DhariniJeeva, line 138:
```suggestion
    // Asserts first button is for editing and there is no second button
```

ui/admin/tests/acceptance/workers/tags-test.js
```diff
@@ -114,6 +135,84 @@ module('Acceptance | workers | worker | tags', function (hooks) {
...
+  test('user cannot edit a tag without proper permission', async function (assert) {
+    instances.worker.authorized_actions =
+      instances.worker.authorized_actions.filter(
+        (item) => item !== 'set-worker-tags',
+      );
+    await visit(urls.tags);
+
+    await click(API_TAG_ACTION_SELECTOR);
+
+    // Asserts first button is for removing and there is on second button
```
DhariniJeeva, line 192:
```suggestion
    // Asserts first button is for removing and there is no second button
```

ui/admin/app/templates/scopes/scope/workers/worker/tags.hbs
```diff
@@ -64,25 +64,101 @@
...
+                      />
+                      {{#if perms.canEdit}}
+                        <dd.Interactive
+                          @text={{t 'resources.worker.tags.edit.title'}}
+                          {{on 'click' (fn this.toggleEditModal B.data)}}
+                        />
+                      {{/if}}
+                      {{#if perms.canRemove}}
+                        <dd.Interactive
+                          @text={{t 'resources.worker.tags.actions.remove'}}
```
DhariniJeeva, line 90: design question- should we use color critical for this to stand out?

ui/admin/app/controllers/scopes/scope/workers/worker/tags.js
```diff
@@ -70,4 +113,16 @@ export default class ScopesScopeWorkersWorkerTagsController extends Controller {
     this.modalTag = apiTag;
     this.removalConfirmation = null;
   }
+
+  /**
+   * Opens the edit tag modal.
```
DhariniJeeva, line 118: should prob say toggles the edit tag model since the method does both, thoughts?

ui/admin/app/controllers/scopes/scope/workers/worker/tags.js
```diff
@@ -70,4 +113,16 @@ export default class ScopesScopeWorkersWorkerTagsController extends Controller {
...
+   * @param {object} apiTag
+   */
+  @action
+  toggleEditModal(apiTag) {
+    this.editModal = !this.editModal;
+    this.modalTag = apiTag;
+    this.editKey = apiTag?.key;
+    this.editValue = apiTag?.value;
```
ZedLi, line 126: I'm not sure I understand why you need `editKey` and `editValue` if you're already saving the apiTag to the `modalTag`, couldn't you have just used only `modalTag` for the edited values?

Since you already have access to the data row from the table you always know what the original was

ui/admin/app/controllers/scopes/scope/workers/worker/tags.js
```diff
@@ -44,6 +47,34 @@ export default class ScopesScopeWorkersWorkerTagsController extends Controller {
     );
   }
 
+  updateApiTags() {
+    const apiTags = structuredClone(this.model.api_tags);
+    const { key, value } = this.modalTag;
+    const editedValues = this.editValue.split(',').map((value) => value.trim());
```
ZedLi, line 53: We can edit a tag to add more than one value at a time? Is that expected?

ui/admin/app/controllers/scopes/scope/workers/worker/tags.js
```diff
@@ -44,6 +47,34 @@ export default class ScopesScopeWorkersWorkerTagsController extends Controller {
...
+    apiTags[key] = apiTags[key].filter((val) => val !== value);
+
+    // remove key if value is empty array
+    if (apiTags[key].length === 0) {
+      delete apiTags[key];
+    }
+
+    // add new key/value to apiTags
+    if (this.editKey in apiTags) {
+      // if editKey exist, add the new values to its array if not already preset
```
ZedLi, line 65:
```suggestion
      // if editKey exists, add the new values to its array if not already present
```

ui/admin/app/controllers/scopes/scope/workers/worker/tags.js
```diff
@@ -44,6 +47,34 @@ export default class ScopesScopeWorkersWorkerTagsController extends Controller {
...
+      editedValues.forEach((value) => {
+        if (!apiTags[this.editKey].includes(value)) {
+          apiTags[this.editKey].push(value);
+        }
```
ZedLi, line 69: Rather than doing a check for each value, isn't it more efficient to just combine the arrays and make the array unique after?

ui/admin/app/styles/app.scss
```diff
@@ -1025,13 +1025,29 @@
...
+      margin-top: 1.5rem;
+      margin-bottom: 0;
+    }
   }
 
-  .hds-form-field--layout-vertical {
+  .edit-modal {
+    .hds-button-set {
+      margin-bottom: 0;
+    }
```
ZedLi, line 1043: How come we needed all this custom css for margins on the hds component? It looks like we're overriding our own override?

ui/admin/app/controllers/scopes/scope/workers/worker/tags.js
```diff
@@ -44,6 +47,34 @@ export default class ScopesScopeWorkersWorkerTagsController extends Controller {
     );
   }
 
+  updateApiTags() {
```
calcaide, line 50: Not blocking comment: do we plan to add test for this method?

ui/admin/app/styles/app.scss
calcaide: Indeed with @ZedLi comment, could we give a review to these styles? or explain why we have to add these overrides?

ui/admin/app/controllers/scopes/scope/workers/worker/tags.js
cameronperera: From discussions with product and design, yes. This follows the same pattern we have for creating tags. I feel if we didn't support it, users might try to type `dev, qa` like they can during creation and be confused as to why they cannot here.

ui/admin/app/controllers/scopes/scope/workers/worker/tags.js
cameronperera: Thanks. Fixed.

ui/admin/app/controllers/scopes/scope/workers/worker/tags.js
cameronperera: It absolutely should. Fixing now.

ui/admin/app/controllers/scopes/scope/workers/worker/tags.js
cameronperera: This method is used by `editApiTag` which does have tests so I didn't feel it needed. However, I can add some for this as well.

cameronperera: Tests added.

ui/admin/app/controllers/scopes/scope/workers/worker/tags.js
cameronperera: It sure is. Thanks for the suggestion.

ui/admin/app/styles/app.scss
cameronperera: Nope. Removed.

ui/admin/app/styles/app.scss
cameronperera: no this is limited to `.hds-application-state` within the worker/tags route.

ui/admin/app/templates/scopes/scope/workers/worker/tags.hbs
cameronperera: oh yes, good catch

ui/admin/tests/acceptance/workers/tags-test.js
cameronperera: Thanks. Fixed in both places.

ui/admin/app/controllers/scopes/scope/workers/worker/tags.js
ZedLi: Oh interesting. I had no idea you could do that in creation too, that wasn't obvious to me as a user

ui/admin/app/controllers/scopes/scope/workers/worker/tags.js
```diff
@@ -44,6 +51,40 @@ export default class ScopesScopeWorkersWorkerTagsController extends Controller {
     );
   }
 
+  /**
+   * Determines if a config tag is managed by HCP
+   * @param {object} tag
+   * @returns {boolean}
+   */
+  isHcpManaged(tag) {
```
ZedLi, line 59: I don't think we should assume those could be the only config values, shouldn't we be checking if we're running in HCP?

ui/admin/app/controllers/scopes/scope/workers/worker/tags.js
cameronperera: Discussed this with Zed and we do not have access to the row data so we are leaving how it is for now.

ui/admin/app/controllers/scopes/scope/workers/worker/tags.js
cameronperera: We cannot simply check if we are running in HCP as other non-hcp managed workers would show the HCP managed tooltip. We will work with cloud team to find a better solution in future releases.

ui/admin/app/controllers/scopes/scope/workers/worker/tags.js
```diff
@@ -44,6 +47,34 @@ export default class ScopesScopeWorkersWorkerTagsController extends Controller {
...
+  updateApiTags() {
+    const apiTags = structuredClone(this.model.api_tags);
+    const { key, value } = this.modalTag;
+    const editedValues = this.editValue.split(',').map((value) => value.trim());
+
+    // remove the old value
+    apiTags[key] = apiTags[key].filter((val) => val !== value);
+
+    // remove key if value is empty array
+    if (apiTags[key].length === 0) {
```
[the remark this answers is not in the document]
cameronperera: To clarify, did you essentially remove the value but keep the key, then click `Confirm`?

ZedLi (approved):

ui/admin/app/controllers/scopes/scope/workers/worker/tags.js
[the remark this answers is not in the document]
DhariniJeeva: yes that is correct

lisbet-alvarez (approved):
🙌

DhariniJeeva (approved):

ui/admin/app/controllers/scopes/scope/workers/worker/tags.js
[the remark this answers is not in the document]
cameronperera: Yeah that is a mirage issue. The api actually will not allow you to set that to an empty string but mirage will.

PR #2391, opened 2024-07-16
refactor: 💡 Remove stack traces and add logs for version info

ZedLi:
# Description
Added some logging to log the version info on startup. Also changed errors to be more uniform and return a message consistently. Stopped logging error objects so we don't log any stacktraces. 

## Screenshots 
Screenshot 2024-07-16 at 12 51 43 PM

## How to test
Start up desktop client and confirm version info is in logs. Try and generate errors either by throwing errors in code or creating situations in desktop client.

## Checklist

### DIFF
--- a/ui/desktop/electron-app/src/services/cache-daemon-manager.js
+++ b/ui/desktop/electron-app/src/services/cache-daemon-manager.js
@@ -11,6 +11,7 @@ const sanitizer = require('../utils/sanitizer.js');
 const { isWindows } = require('../helpers/platform.js');
 const treeKill = require('tree-kill');
 const log = require('electron-log/main');
+const generateErrorPromise = require('../utils/generateErrorPromise');
 
 class CacheDaemonManager {
   #socketPath;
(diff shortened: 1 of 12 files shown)


ui/desktop/electron-app/src/utils/jsonify.js
```diff
@@ -11,7 +11,6 @@ const jsonify = (data) => {
   try {
     return JSON.parse(data);
   } catch (e) {
-    log.warn('jsonify:', e);
```
calcaide, line 14: None blocking comment: we delete the entire log, is this what we want? or do we want to log `e.message` rathern than `e`?

calcaide (approved):
Looks good to me!!

DhariniJeeva (approved):
looks good!

ui/desktop/electron-app/src/utils/jsonify.js
ZedLi: It's not a useful log statement so I added log messages where from the callers of this method instead

ui/desktop/electron-app/src/services/cache-daemon-manager.js
```diff
@@ -23,15 +24,21 @@ class CacheDaemonManager {
...
+    this.#socketPath = parsedResponse?.item?.socket_address;
+
+    if (parsedResponse?.status_code === 200) {
+      return parsedResponse.item;
+    }
 
-    const status = jsonify(stdout);
-    this.#socketPath = status?.item?.socket_address;
-    return status;
+    log.warn('Cache Daemon Status:', stderr);
```
lisbet-alvarez, line 40: non-blocking: it looks like we will only know the version of the daemon if there is an error during this cmd. Im wondering if its possible to check the status at the very beginning along with the boundary cli version? Or perhaps knowing that info is not rly necessary ?

lisbet-alvarez (approved):
🙌

ui/desktop/electron-app/src/services/cache-daemon-manager.js
ZedLi: Maybe I should've specified this is meant for logging errors only, I actually do log the cache daemon status which includes the version [here](https://github.com/hashicorp/boundary-ui/pull/2391/files#diff-1261de05929d4dcb2a372b5d2587fa51007aebfff1d9fa8abd8f9df3e9ac297fR59).

It looks like this:
image

PR #2398, opened 2024-07-17
fix: 🐛 Fix breaking change for changing page size

ZedLi:
# Description
It seems `onPageSizeChange` is now required for pageSize changes. When I had read the [breaking changes](https://helios.hashicorp.design/whats-new/release-notes#400) in HDS originally, I took it as it's only an issue if you didn't manage the query parameters outside of the component but it turns out this was happening with internal logic regardless. The pages themselves work as they were just `a` links with query parameters. 

I added our own function to handle page size changes.

## Screenshots
https://github.com/user-attachments/assets/4fcb5da6-8451-4a84-9f97-5af431bcc20b

## How to Test
Use or start up an instance with many resources. Confirm that changing the page sizes works. 

## Checklist

### DIFF
--- a/addons/rose/addon/components/rose/pagination/index.hbs
+++ b/addons/rose/addon/components/rose/pagination/index.hbs
@@ -10,4 +10,5 @@
   @route={{this.router.currentRouteName}}
   @model={{@model}}
   @queryFunction={{this.paginationQueryParams}}
+  @onPageSizeChange={{this.handlePageSizeChange}}
 />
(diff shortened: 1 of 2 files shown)


calcaide (approved):
Thanks for fixing this

DhariniJeeva (approved):
thanks for the quick fix

lisbet-alvarez (approved):
Thank you for fixing this! I def. missed this when upgrading hds

PR #2406, opened 2024-07-19
feat: 🎸 Add debug logging for search times

ZedLi:
# Description
Added debug statements to allow a user to see how long search requests are taking. Also added a `BOUNDARY_DESKTOP_LOG_LEVEL` which they can set when starting DC to change log levels for now.

image

## How to Test
Create a build either with a CLI from main or 0.17. Run the DC from your terminal with the `BOUNDARY_DESKTOP_LOG_LEVEL` set to `debug`. 

My binary for mac is located at `boundary-ui/ui/desktop/electron-app/out/Boundary-darwin-arm64/Boundary.app/Contents/MacOS/` and I just run `BOUNDARY_DESKTOP_LOG_LEVEL=debug ./Boundary`

Also running the pipeline for a build [here](https://github.com/hashicorp/boundary-ui-releases/actions/runs/10012570932). Will need to test windows.

## Checklist

### DIFF
--- a/ui/desktop/electron-app/src/services/cache-daemon-manager.js
+++ b/ui/desktop/electron-app/src/services/cache-daemon-manager.js
@@ -117,9 +117,18 @@ class CacheDaemonManager {
     return generateErrorPromise(stderr);
   }
 
-  search(requestData) {
+  async search(requestData) {
+    const start = Date.now();
+    // Log request data, but clean up token from log
+    const { auth_token_id, token, ...logRequestData } = requestData;
+
     if (isWindows()) {
-      return searchCliCommand(requestData);
+      const result = await searchCliCommand(requestData);
+
+      const end = Date.now();
+      log.debug(`Search request took ${end - start} ms`, logRequestData);
+
+      return result;
     }
 
     delete requestData.token;
@@ -130,7 +139,12 @@ class CacheDaemonManager {
       path: `http://internal.boundary.local/v1/search?${queryString.toString()}`,
       socketPath: this.#socketPath,
     };
-    return unixSocketRequest(request);
+
+    const result = await unixSocketRequest(request);
+    const end = Date.now();
+    log.debug(`Search request took ${end - start} ms`, logRequestData);
+
+    return result;
   }
 }
 
--- a/ui/desktop/electron-app/src/index.js
+++ b/ui/desktop/electron-app/src/index.js
@@ -61,6 +61,7 @@ fixPath();
 // Setup logger
 log.initialize();
 log.transports.console.level = false;
+log.transports.file.level = process.env.BOUNDARY_DESKTOP_LOG_LEVEL ?? 'info';
 log.transports.file.format =
   '[{y}-{m}-{d}T{h}:{i}:{s}.{ms}{z}] [{level}] {text}';
 log.transports.file.fileName = 'desktop-client.log';


ui/desktop/electron-app/src/services/cache-daemon-manager.js
```diff
@@ -117,9 +117,18 @@ class CacheDaemonManager {
...
+    const start = Date.now();
+    // Log request data, but clean up token from log
+    const { auth_token_id, token, ...logRequestData } = requestData;
+
     if (isWindows()) {
-      return searchCliCommand(requestData);
+      const result = await searchCliCommand(requestData);
+
+      const end = Date.now();
+      log.debug(`Search request took ${end - start} ms`, logRequestData);
```
DhariniJeeva, line 129: do we need to mention that this is from `windows` or will that info already be there?

DhariniJeeva (approved):
looks great!

ui/desktop/electron-app/src/services/cache-daemon-manager.js
ZedLi: What do you mean by this is from `windows`?

ui/desktop/electron-app/src/services/cache-daemon-manager.js
```diff
@@ -117,9 +117,18 @@ class CacheDaemonManager {
...
+  async search(requestData) {
+    const start = Date.now();
+    // Log request data, but clean up token from log
+    const { auth_token_id, token, ...logRequestData } = requestData;
+
     if (isWindows()) {
-      return searchCliCommand(requestData);
+      const result = await searchCliCommand(requestData);
+
+      const end = Date.now();
```
calcaide, line 128: Do the `end` should be outside of this conditional? Or is the log intended just for windows platform?

I understand we are taking the start time `start` regardless the platform, but just taking `end` and logging if is windows, then we are just loggin for windows, correct?

ui/desktop/electron-app/src/services/cache-daemon-manager.js
```diff
@@ -117,9 +117,18 @@ class CacheDaemonManager {
...
+      log.debug(`Search request took ${end - start} ms`, logRequestData);
+
+      return result;
     }
 
     delete requestData.token;
```
calcaide, line 134: line 123 is already deleting token, isn't?

ZedLi: Yes it is but it doesn't get deleted for the windows search so I just did it again, otherwise i'd have to do two separate statements

ui/desktop/electron-app/src/services/cache-daemon-manager.js
DhariniJeeva: for us to identify that this comes from windows env, since we have two sets of logs one for windows and other OS

ZedLi: I'm not sure I'm following, there'll only ever be one set of logs. On windows this debug statement will run and on other OS's the other debug statement will run. They'll never both run

DhariniJeeva: you are right, I was thinking of a situation where we get a dump of logs from different machines to debug but that will never be the case. sorry for the confusion.

calcaide (approved):
LGTM! Thanks for taking care of this

cameronperera (approved):

PR #2409, opened 2024-07-19
Icu 14400 add a section for version info cli and dc

DhariniJeeva:
# Description
https://hashicorp.atlassian.net/browse/ICU-14400
This PR adds application information section on the user settings page in DC

Screenshot 2024-07-30 at 8 54 17 PM

with no daemon:
 
Screenshot 2024-07-30 at 8 48 09 PM
 

## Checklist

### DIFF
--- a/ui/desktop/app/components/settings-card/index.hbs
+++ b/ui/desktop/app/components/settings-card/index.hbs
@@ -23,14 +23,14 @@
       </div>
     {{/if}}
 
+    {{#if @description}}
+      <Hds::Text::Body @tag='p'>
+        {{@description}}
+      </Hds::Text::Body>
+    {{/if}}
   </div>
 
-  {{#if @description}}
-    <Hds::Text::Body @tag='p'>
-      {{@description}}
-    </Hds::Text::Body>
-  {{/if}}
-
+  {{! this yields the body from consumer }}
   {{yield}}
 
 </Hds::Card::Container>
--- a/ui/desktop/app/components/settings-card/index.hbs
+++ b/ui/desktop/app/components/settings-card/index.hbs
@@ -3,7 +3,7 @@
   SPDX-License-Identifier: BUSL-1.1
 }}
 
-<Hds::Card::Container @level='mid' @hasBorder={{true}} class='settings-card'>
+<Hds::Card::Container @level='mid' @hasBorder={{true}}>
 
   <div class='header-and-icon'>
     <Hds::IconTile @icon={{@icon}} @size='small' />
@@ -22,15 +22,13 @@
         <Hds::Badge @text={{@state}} @icon='loading' />
       </div>
     {{/if}}
-
-    {{#if @description}}
-      <Hds::Text::Body @tag='p'>
-        {{@description}}
-      </Hds::Text::Body>
-    {{/if}}
   </div>
+  {{#if @description}}
+    <Hds::Text::Body @tag='p'>
+      {{@description}}
+    </Hds::Text::Body>
+  {{/if}}
 
   {{! this yields the body from consumer }}
   {{yield}}
-
 </Hds::Card::Container>
--- a/ui/desktop/app/components/settings-card/index.hbs
+++ b/ui/desktop/app/components/settings-card/index.hbs
@@ -3,36 +3,34 @@
   SPDX-License-Identifier: BUSL-1.1
 }}
 
-<Hds::Card::Container @level="mid" @hasBorder={{true}} class="settings-card">
+<Hds::Card::Container @level='mid' @hasBorder={{true}} class='settings-card'>
 
-  <div class="header-and-icon">
-    <Hds::IconTile @icon={{@icon}} @size="small" />
+  <div class='header-and-icon'>
+    <Hds::IconTile @icon={{@icon}} @size='small' />
     <Hds::Text::Display @tag='h4'>
       {{@header}}
     </Hds::Text::Display>
 
     {{#if @action}}
-      <div class="action">
+      <div class='action'>
         {{yield}}
       </div>
     {{/if}}
 
     {{#if @state}}
-      <div class="state">
-        <Hds::Badge
-          @text={{@state}}
-          @icon='loading'
-        />
+      <div class='state'>
+        <Hds::Badge @text={{@state}} @icon='loading' />
       </div>
     {{/if}}
 
-    </div>
+  </div>
 
-    {{#if @description}}
-      <Hds::Text::Body @tag='p'>
-        {{@description}}
-      </Hds::Text::Body>
-    {{/if}}
+  {{#if @description}}
+    <Hds::Text::Body @tag='p'>
+      {{@description}}
+    </Hds::Text::Body>
+  {{/if}}
 
-    {{yield}}
+  {{yield}}
+
 </Hds::Card::Container>
--- a/ui/desktop/app/components/settings-card/index.hbs
+++ b/ui/desktop/app/components/settings-card/index.hbs
@@ -13,7 +13,7 @@
 
     {{#if @action}}
       <div class='action'>
-        {{yield}}
+        {{@action}}
       </div>
     {{/if}}
 
--- a/ui/desktop/app/components/settings-card/index.hbs
+++ b/ui/desktop/app/components/settings-card/index.hbs
@@ -1,0 +1,38 @@
+{{!
+  Copyright (c) HashiCorp, Inc.
+  SPDX-License-Identifier: BUSL-1.1
+}}
+
+<Hds::Card::Container @level="mid" @hasBorder={{true}} class="settings-card">
+
+  <div class="header-and-icon">
+    <Hds::IconTile @icon={{@icon}} @size="small" />
+    <Hds::Text::Display @tag='h4'>
+      {{@header}}
+    </Hds::Text::Display>
+
+    {{#if @action}}
+      <div class="action">
+        {{yield}}
+      </div>
+    {{/if}}
+
+    {{#if @state}}
+      <div class="state">
+        <Hds::Badge
+          @text={{@state}}
+          @icon='loading'
+        />
+      </div>
+    {{/if}}
+
+    </div>
+
+    {{#if @description}}
+      <Hds::Text::Body @tag='p'>
+        {{@description}}
+      </Hds::Text::Body>
+    {{/if}}
+
+    {{yield}}
+</Hds::Card::Container>
--- a/ui/desktop/app/components/settings-card/index.hbs
+++ b/ui/desktop/app/components/settings-card/index.hbs
@@ -13,6 +13,7 @@
 
     {{#if @action}}
       <div class='action'>
+        {{! this will contain buttons }}
         {{@action}}
       </div>
     {{/if}}
--- a/ui/desktop/app/components/settings-card/index.hbs
+++ b/ui/desktop/app/components/settings-card/index.hbs
@@ -11,10 +11,10 @@
       {{@header}}
     </Hds::Text::Display>
 
-    {{#if @action}}
+    {{#if (has-block 'action')}}
       <div class='action'>
         {{! this will contain buttons }}
-        {{@action}}
+        {{yield to='action'}}
       </div>
     {{/if}}
 
@@ -24,6 +24,7 @@
       </div>
     {{/if}}
   </div>
+
   {{#if @description}}
     <Hds::Text::Body @tag='p'>
       {{@description}}
@@ -31,8 +32,10 @@
   {{/if}}
 
   {{! this yields the body from consumer }}
-  <div class='content'>
-    {{yield}}
-  </div>
+  {{#if (has-block 'body')}}
+    <div class='content'>
+      {{yield to='body'}}
+    </div>
+  {{/if}}
 
 </Hds::Card::Container>
--- a/ui/desktop/app/components/settings-card/index.hbs
+++ b/ui/desktop/app/components/settings-card/index.hbs
@@ -31,5 +31,8 @@
   {{/if}}
 
   {{! this yields the body from consumer }}
-  {{yield}}
+  <div class='content'>
+    {{yield}}
+  </div>
+
 </Hds::Card::Container>
--- a/ui/desktop/app/controllers/scopes/scope/projects/settings/index.js
+++ b/ui/desktop/app/controllers/scopes/scope/projects/settings/index.js
@@ -4,10 +4,5 @@
  */
 
 import Controller from '@ember/controller';
-import { inject as service } from '@ember/service';
 
-export default class ScopesScopeProjectsSettingsIndexController extends Controller {
-  @service session;
-  @service ipc;
-  // Methods
-}
+export default class ScopesScopeProjectsSettingsIndexController extends Controller {}
--- a/ui/desktop/app/controllers/scopes/scope/projects/settings/index.js
+++ b/ui/desktop/app/controllers/scopes/scope/projects/settings/index.js
@@ -1,3 +1,8 @@
+/**
+ * Copyright (c) HashiCorp, Inc.
+ * SPDX-License-Identifier: BUSL-1.1
+ */
+
 import Controller from '@ember/controller';
 import { inject as service } from '@ember/service';
 import { getOwner } from '@ember/application';
--- a/ui/desktop/app/controllers/scopes/scope/projects/settings/index.js
+++ b/ui/desktop/app/controllers/scopes/scope/projects/settings/index.js
@@ -41,7 +41,6 @@ export default class ScopesScopeProjectsSettingsIndexController extends Controll
       .lookup('service:-document')
       .querySelector(rootElementSelector);
     this.session.set('data.theme', theme);
-    console.log(this.session.data.theme, 'theeeee')
     switch (theme) {
       case 'light':
         rootEl.classList.add('rose-theme-light');
--- a/ui/desktop/app/controllers/scopes/scope/projects/settings/index.js
+++ b/ui/desktop/app/controllers/scopes/scope/projects/settings/index.js
@@ -1,8 +1,59 @@
-/**
- * Copyright (c) HashiCorp, Inc.
- * SPDX-License-Identifier: BUSL-1.1
- */
-
 import Controller from '@ember/controller';
+import { inject as service } from '@ember/service';
+import { getOwner } from '@ember/application';
+import { action } from '@ember/object';
 
-export default class ScopesScopeProjectsSettingsIndexController extends Controller {}
+const THEMES = [
+  {
+    label: 'system',
+    value: 'system-default-theme',
+  },
+  {
+    label: 'light',
+    value: 'light',
+  },
+  {
+    label: 'dark',
+    value: 'dark',
+  },
+];
+
+export default class ScopesScopeProjectsSettingsIndexController extends Controller {
+  @service session;
+  @service ipc;
+  // Methods
+
+  /**
+   * Returns available themes
+   */
+  get themes() {
+    return THEMES;
+  }
+
+  /**
+   * Applies the specified color theme to the root ember element.
+   * @param {string} theme - "light", "dark", or nullish (system default)
+   */
+  @action
+  async toggleTheme({ target: { value: theme } }) {
+    const rootElementSelector = getOwner(this).rootElement;
+    const rootEl = getOwner(this)
+      .lookup('service:-document')
+      .querySelector(rootElementSelector);
+    this.session.set('data.theme', theme);
+    console.log(this.session.data.theme, 'theeeee')
+    switch (theme) {
+      case 'light':
+        rootEl.classList.add('rose-theme-light');
+        rootEl.classList.remove('rose-theme-dark');
+        break;
+      case 'dark':
+        rootEl.classList.add('rose-theme-dark');
+        rootEl.classList.remove('rose-theme-light');
+        break;
+      default:
+        rootEl.classList.remove('rose-theme-dark');
+        rootEl.classList.remove('rose-theme-light');
+    }
+  }
+}
--- a/ui/desktop/app/controllers/scopes/scope/projects/settings/index.js
+++ b/ui/desktop/app/controllers/scopes/scope/projects/settings/index.js
@@ -5,59 +5,9 @@
 
 import Controller from '@ember/controller';
 import { inject as service } from '@ember/service';
-import { getOwner } from '@ember/application';
-import { action } from '@ember/object';
-
-const THEMES = [
-  {
-    label: 'system',
-    value: 'system-default-theme',
-  },
-  {
-    label: 'light',
-    value: 'light',
-  },
-  {
-    label: 'dark',
-    value: 'dark',
-  },
-];
 
 export default class ScopesScopeProjectsSettingsIndexController extends Controller {
   @service session;
   @service ipc;
   // Methods
-
-  /**
-   * Returns available themes
-   */
-  get themes() {
-    return THEMES;
-  }
-
-  /**
-   * Applies the specified color theme to the root ember element.
-   * @param {string} theme - "light", "dark", or nullish (system default)
-   */
-  @action
-  async toggleTheme({ target: { value: theme } }) {
-    const rootElementSelector = getOwner(this).rootElement;
-    const rootEl = getOwner(this)
-      .lookup('service:-document')
-      .querySelector(rootElementSelector);
-    this.session.set('data.theme', theme);
-    switch (theme) {
-      case 'light':
-        rootEl.classList.add('rose-theme-light');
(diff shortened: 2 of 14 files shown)


ui/desktop/app/components/settings-card/index.hbs
```diff
@@ -0,0 +1,36 @@
...
+
+<Hds::Card::Container @level='mid' @hasBorder={{true}} class='settings-card'>
+
+  <div class='header-and-icon'>
+    <Hds::IconTile @icon={{@icon}} @size='small' />
+    <Hds::Text::Display @tag='h4'>
+      {{@header}}
+    </Hds::Text::Display>
+
+    {{#if @action}}
```
DhariniJeeva, line 14: this is not used right now, it is for the action buttons that we will have in the next tickets

ui/desktop/app/templates/scopes/scope/projects/settings/index.hbs
```diff
@@ -3,4 +3,61 @@
...
+      <div>
+        <Hds::Text::Display @tag='h5'>
+          {{t 'titles.cli'}}
+        </Hds::Text::Display>
+        <Hds::Text::Body @tag='p' @size='300'>
+          {{#if @model.cliExists}}
+            {{t 'states.installed'}}
+          {{else}}
+            {{t 'states.not-available'}}
+          {{/if}}
```
ZedLi, line 26: I don't think this is very useful, since it should never be the case where the CLI isn't installed. We can show the `not-available` message if for some reason it isn't since that should be an exceptional case, but shouldn't we display the CLI version here? And the Version on line 14 be the desktop client version?

ui/desktop/app/templates/scopes/scope/projects/settings/index.hbs
```diff
@@ -3,4 +3,61 @@
...
+      <div>
+        <Hds::Text::Display @tag='h5'>
+          {{t 'titles.cache-daemon'}}
+        </Hds::Text::Display>
+        <Hds::Text::Body @tag='p' @size='300'>
+          {{#if @model.getCacheDaemonVersion}}
+            {{@model.getCacheDaemonVersion}}
+            <Hds::Badge
+              @text={{t 'states.running'}}
+              @icon='loading'
```
ZedLi, line 38: Shouldn't this be a `running` icon instead of loading?

ui/desktop/electron-app/src/ipc/handlers.js
```diff
@@ -194,6 +194,26 @@ handle('isClientAgentRunning', async () => {
...
 
+/**
+ * Gets boundary cli version
+ */
+handle('getCliVersion', async () => boundaryCli.version());
+
+/**
+ * Gets Cache Daemon version
+ */
+handle('getCacheDaemonVersion', async () => {
```
ZedLi, line 205: Shouldn't we be returning the full status instead of just the version here and have the app handle parsing out the version? We likely will need more later

ui/desktop/app/controllers/scopes/scope/projects/settings/index.js
```diff
@@ -4,5 +4,60 @@
...
+  get themes() {
+    return THEMES;
+  }
+
+  /**
+   * Applies the specified color theme to the root ember element.
+   * @param {string} theme - "light", "dark", or nullish (system default)
+   */
+  @action
+  async toggleTheme({ target: { value: theme } }) {
```
ZedLi, line 43: Should we be removing the theme selector from the global dropdown now?

DhariniJeeva: yup, there was a note on figma that we should remove the global theme selector, I added a separate task for that

calcaide: just to clarify, is this `toggleTheme` function JUST for the theme selector on the global dropdown?

ui/desktop/app/components/settings-card/index.hbs
```diff
@@ -0,0 +1,36 @@
+{{!
+  Copyright (c) HashiCorp, Inc.
+  SPDX-License-Identifier: BUSL-1.1
+}}
+
+<Hds::Card::Container @level='mid' @hasBorder={{true}} class='settings-card'>
```
calcaide, line 6: After reviewing figma designs, I notice:
- All the cards have at least header and body/content.
- The headers in all the cards, at least have icon + text. Some have status and/or actions.
- The bodies/content seems to be a grid with a variety of content: text, components, etc

I will suggest within <hds::card>
We have 2 divs: `.header`, `.body/content` and try to follow and reproduce this pattern within all the cards, this will facilitate re use and clarify for the future CSS.

ui/desktop/app/templates/scopes/scope/projects/settings/index.hbs
calcaide: Indeed, I just randomly tested (for something else) last week to run the DC with no CLI and it's useless and it breaks, so I do strongly believe no one will be able to open DC, connect to a controller and show settings page unless there is a CLI present on the system.

ui/desktop/app/templates/scopes/scope/projects/settings/index.hbs
```diff
@@ -0,0 +1,63 @@
+{{!
+  Copyright (c) HashiCorp, Inc.
+  SPDX-License-Identifier: BUSL-1.1
+}}
+
+<SettingsCard @header={{t 'titles.application'}} @icon='boundary'>
+  <div class='content'>
+    <div class='section'>
```
calcaide, line 8: Can we ellaborate on the names we are assigning to clases here? `content` is clear, but then `section` and after `theme`? seems to me these is quite ad-hoc, but this is a pattern that we will reproduce in other settings cards, can we look at something more generic?

ui/desktop/app/styles/app.scss
```diff
@@ -662,3 +662,29 @@ $chevron-down: url("data:image/svg+xml;utf-8,<svg viewBox='0 0 24 24' fill='%231
   gap: 0.5rem;
   width: inherit;
 }
+
+// Settings
+.settings-card {
```
calcaide, line 667: I am not sure we need an specific class. What do you think on giving an id to the whole settings page? then we can select #settings > hds card. That way the CSS will apply to every hds card element within settings, which I think is what we want. Thoughts?

ui/desktop/electron-app/src/ipc/handlers.js
calcaide: Agreed, this could be `getCacheStatus` and then the model of an specific domain can parse out what's not need.

calcaide (changes requested):
Looking good!! I left some comments

ui/desktop/app/styles/app.scss
DhariniJeeva: this needs to be a `classname`, because this component will be re-used in building out the rest of the cards.

ui/desktop/app/templates/scopes/scope/projects/settings/index.hbs
DhariniJeeva: agreed it could be better, I updated this

ui/desktop/app/components/settings-card/index.hbs
DhariniJeeva: I have a header div block and yield the body/content from the consumers. Added a comment for more clarity

ui/desktop/app/templates/scopes/scope/projects/settings/index.hbs
DhariniJeeva: removed this

ui/desktop/electron-app/src/ipc/handlers.js
DhariniJeeva: thanks, updated to receive full status from cache

ui/desktop/app/controllers/scopes/scope/projects/settings/index.js
DhariniJeeva: 🤔 I don't think I follow, can you explain? 

we won't have a theme selector on the global dropdown going forward

calcaide: I was asking if `toggleTheme` functionality does toggle the theme only within the dropdown that is rendered within settings or if is to toggle the theme within the global dropdown?

ui/desktop/app/styles/app.scss
calcaide: What I mean is if the main element or [the div that wraps the page body](https://github.com/hashicorp/boundary-ui/pull/2409/files#diff-23bd88f09a10c21b4b497fa14821c75c0276e1e2553634c6dee953e4a78804deR6) we add an id such as `#settings` or `#settings-section` then, you do not need to add a classname to every card we will add, as we do here. You can reference this on the css as `#settings > hds_card` (or the default class HDS generated for hds_card). And those styles will apply to all the HDS cards that are within settings section, without need for `classname`

All good with the current implementation, just anticipating this section will be several cards, I though we could reduce classname usage.

ui/desktop/app/controllers/scopes/scope/projects/settings/index.js
DhariniJeeva: only within the settings page, I have another ticket that gets rid off the global dropdown

ui/desktop/electron-app/src/ipc/handlers.js
```diff
@@ -194,6 +194,33 @@ handle('isClientAgentRunning', async () => {
...
+});
+
+/**
+ * Gets Cache Daemon version
+ */
+handle('getCacheDaemonStatus', async () => {
+  let cache = '';
+  try {
+    cache = await cacheDaemonManager.status();
+  } catch (e) {
```
ZedLi, line 217: How come we're catching the error here and then just returning it?

ui/desktop/electron-app/src/ipc/handlers.js
```diff
@@ -194,6 +194,33 @@ handle('isClientAgentRunning', async () => {
...
+/**
+ * Returns desktop version
+ */
+handle('getDesktopVersion', async () => {
+  const desktopVersion = app.getVersion();
+  return desktopVersion;
+});
+
+/**
+ * Gets Cache Daemon version
```
ZedLi, line 211:
```suggestion
 * Returns the Cache daemon status
```

ui/desktop/electron-app/src/ipc/handlers.js
```diff
@@ -194,6 +194,33 @@ handle('isClientAgentRunning', async () => {
...
+ */
+handle('getCliVersion', async () => boundaryCli.version());
+
+/**
+ * Returns desktop version
+ */
+handle('getDesktopVersion', async () => {
+  const desktopVersion = app.getVersion();
+  return desktopVersion;
+});
```
ZedLi, line 208:
```suggestion
handle('getDesktopVersion', async () =>  app.getVersion());
```

Does `app.GetVersion` pull the correct version? We normally pull it from our generated config and the docs says it pulls it from the `package.json` which I don't believe we update.

ui/desktop/app/templates/scopes/scope/projects/settings/index.hbs
```diff
@@ -3,4 +3,54 @@
...
+        </Hds::Text::Body>
+      </div>
+      <div>
+        <Hds::Text::Display>
+          {{t 'titles.cache-daemon'}}
+        </Hds::Text::Display>
+        <Hds::Text::Body @tag='p' @weight='medium'>
+          {{#if @model.formattedCacheVersion}}
+            {{@model.formattedCacheVersion}}
+            <Hds::Badge
```
ZedLi, line 32: Should we display a not running with an icon as well if we don't get the version?

ui/desktop/app/templates/scopes/scope/projects/settings/index.hbs
```diff
@@ -3,4 +3,54 @@
...
+        <Hds::Text::Display>
+          {{t 'titles.version'}}
+        </Hds::Text::Display>
+        <Hds::Text::Body @tag='p' @weight='medium'>
+          {{@model.desktopVersion}}
+        </Hds::Text::Body>
+      </div>
+      <div>
+        <Hds::Text::Display>
+          {{t 'titles.cli'}}
```
ZedLi, line 19: Should this say CLI version?

ui/desktop/electron-app/src/ipc/handlers.js
DhariniJeeva: yeah, I wasn't sure which one to use, thought either would work, but good to know we don't keep track of the app one. I will switch this to pull from config

ui/desktop/app/templates/scopes/scope/projects/settings/index.hbs
DhariniJeeva: [figma](https://www.figma.com/design/28zD8tKqa9yUQVO13THm77/Transparent-Sessions-%26-Aliases?node-id=4623-123988&t=Eub6bjk8XC75GRHr-0) says CLI and cache daemon.  

@joshuaogle thoughts?  [this would be the headers in the application info section]

ui/desktop/app/routes/scopes/scope/projects/settings/index.js
```diff
@@ -4,5 +4,21 @@
...
+import { inject as service } from '@ember/service';
 
-export default class ScopesScopeProjectsSettingsIndexRoute extends Route {}
+export default class ScopesScopeProjectsSettingsIndexRoute extends Route {
+  @service ipc;
+
+  async model() {
+    const { versionNumber: cliVersion } =
+      await this.ipc.invoke('getCliVersion');
+    const getCacheDaemon = await this.ipc.invoke('getCacheDaemonStatus');
```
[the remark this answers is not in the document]
DhariniJeeva: updated!

ui/desktop/app/templates/scopes/scope/projects/settings/index.hbs
DhariniJeeva: updated to show not running and an icon for that

ui/desktop/app/templates/scopes/scope/projects/settings/index.hbs
DhariniJeeva: discussed offline and updated based on design feedback to use prefixes

ui/desktop/app/routes/scopes/scope/projects/settings/index.js
```diff
@@ -4,5 +4,21 @@
...
+    const { version } = getCacheDaemon;
+    // Format the version
+
+    const formattedCacheVersion = version && /(\d+\.\d+\.\d+)/.exec(version)[0];
```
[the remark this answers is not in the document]
DhariniJeeva: you are right, I did not need that. I was using that in case I did not have a version, but now modified the model to handle the case so it is not needed. thanks for the catch

ui/desktop/app/styles/app.scss
DhariniJeeva: agreed, I removed the classname from here and moved it to the wrapper. since we don't use `id` in our codebase for style, I kept it as `classname`. But I'm not opposed to using ID and this could be a team discussion on what the team prefers. adding to tech time

ui/desktop/app/routes/scopes/scope/projects/settings/index.js
```diff
@@ -4,5 +4,39 @@
...
+    formattedDesktopVersion = `v${desktopVersion}`;
+    try {
+      cacheDaemonStatus = await this.ipc.invoke('cacheDaemonStatus');
+    } catch (e) {
+      __electronLog?.warn('cache daemon is not running', e.message);
+      return {
+        formattedCliVersion,
+        formattedCacheVersion,
+        formattedDesktopVersion,
+        warning: true,
```
DhariniJeeva, line 31: I'm using this to show the warning alert in the UI, open to feedback on better ways to achieve this

ui/desktop/app/styles/app.scss
calcaide: Awesome @DhariniJeeva thanks for the clarification and bringing this topic to tech time 😃

calcaide (approved):

ui/desktop/app/routes/scopes/scope/projects/settings/index.js
ZedLi: It might be better to return the actual error so we can show any error messages?

DhariniJeeva: good call, I'm returning the err obj, we will display the message in the UI if there's any, let me know what you think

ui/desktop/app/routes/scopes/scope/projects/settings/index.js
```diff
@@ -4,5 +4,39 @@
...
+    const { versionNumber: cliVersion } =
+      await this.ipc.invoke('getCliVersion');
+    formattedCliVersion = `v${cliVersion}`;
+
+    const { desktopVersion } = await this.ipc.invoke('getDesktopVersion');
+    formattedDesktopVersion = `v${desktopVersion}`;
+    try {
+      cacheDaemonStatus = await this.ipc.invoke('cacheDaemonStatus');
+    } catch (e) {
+      __electronLog?.warn('cache daemon is not running', e.message);
```
ZedLi, line 26: I don't think this is strictly needed since we already have a log statement if the status fails in electron [here](https://github.com/hashicorp/boundary-ui/blob/10c128ee3baf695e350679f82122ef0cfb8cdd8a/ui/desktop/electron-app/src/services/cache-daemon-manager.js#L40)

ui/desktop/app/routes/scopes/scope/projects/settings/index.js
```diff
@@ -4,5 +4,39 @@
...
+      formattedCliVersion,
+      formattedDesktopVersion;
+    const { versionNumber: cliVersion } =
+      await this.ipc.invoke('getCliVersion');
+    formattedCliVersion = `v${cliVersion}`;
+
+    const { desktopVersion } = await this.ipc.invoke('getDesktopVersion');
+    formattedDesktopVersion = `v${desktopVersion}`;
+    try {
+      cacheDaemonStatus = await this.ipc.invoke('cacheDaemonStatus');
```
ZedLi, line 24: We can still get errors as part of a successful response like this but we can handle this in another ticket if you'd like.

```{
  "status_code": 200,
  "item": {
    "uptime": 2100252029000,
    "socket_address": "/Users/zhiheli/.boundary/socket/daemon.sock",
    "log_location": "/Users/zhiheli/.boundary/cache.log",
    "version": "Boundary v0.17.0 (126c4f55a006525ab4574742c517e7b3b76e0336)",
    "users": [
      {
        "id": "u_1234567890",
        "boundary_instance": {
          "address": "http://localhost:9200",
          "cache_support": "Supported",
          "last_support_check": 20486662000
        },
        "auth_tokens": [
          {
            "id": "at_zFm4apTRvb"
          },
          {
            "id": "at_9DZX9QLWrq",
            "keyringless_references": 1
          }
        ],
        "resources": [
          {
            "name": "resolvable-alias",
            "count": 3,
            "last_error": {
              "error": "cache.(Repository).refreshResolvableAliases: for token \"at_9DZX9QLWrq\": cache.defaultResolvableAliasFunc: unknown, unknown: error #0: ",
              "last_returned": 21007917000
            },
            "refresh_token": {
              "age": 328362859000,
              "last_used": 295585859000
            }
          },
          {
            "name": "target",
            "count": 2,
            "refresh_token": {
              "age": 22450018000,
              "last_used": 20487018000
            }
          },
          {
            "name": "session",
            "count": 0
          }
        ]
      }
    ]
  }
}
```

ui/desktop/tests/acceptance/projects/settings/index-test.js
```diff
@@ -9,10 +9,12 @@ import { setupApplicationTest } from 'ember-qunit';
 import setupMirage from 'ember-cli-mirage/test-support/setup-mirage';
 import { authenticateSession } from 'ember-simple-auth/test-support';
 import WindowMockIPC from '../../../helpers/window-mock-ipc';
+import setupStubs from 'api/test-support/handlers/cache-daemon-search';
 
 module('Acceptance | projects | settings | index', function (hooks) {
```
ZedLi, line 14: Should we have tests for the color theme picker here now instead? Maybe we should have the moving of the theme picker in a different PR completely

ui/desktop/app/templates/scopes/scope/projects/settings/index.hbs
```diff
@@ -3,4 +3,73 @@
   SPDX-License-Identifier: BUSL-1.1
 }}
 
-{{outlet}}
\ No newline at end of file
+<SettingsCard @header={{t 'titles.application'}} @icon='boundary'>
+  <div class='content'>
```
ZedLi, line 7: Should this div be part of the generic component?

ui/desktop/app/components/settings-card/index.hbs
```diff
@@ -0,0 +1,34 @@
...
+
+  <div class='header-and-icon'>
+    <Hds::IconTile @icon={{@icon}} @size='small' />
+    <Hds::Text::Display @tag='h4'>
+      {{@header}}
+    </Hds::Text::Display>
+
+    {{#if @action}}
+      <div class='action'>
+        {{yield}}
```
ZedLi, line 16: How's this work with the yield on line 33 as well?

ui/desktop/app/styles/app.scss
```diff
@@ -662,3 +662,31 @@ $chevron-down: url("data:image/svg+xml;utf-8,<svg viewBox='0 0 24 24' fill='%231
...
+  .content {
+    padding-left: 2.2rem;
+
+    .hds-form-field--layout-vertical {
+      margin-top: 2rem;
+    }
+  }
+
+  .app-information {
+    display: grid;
```
ZedLi, line 687: Should this be a grid specifically for 3 columns? I'm wondering if it makes sense to use flex instead? How were you planning on doing the other cards that only have 1 or 2 columns?

ui/desktop/app/styles/app.scss
```diff
@@ -662,3 +662,31 @@ $chevron-down: url("data:image/svg+xml;utf-8,<svg viewBox='0 0 24 24' fill='%231
...
+  .header-and-icon {
+    display: flex;
+    align-items: center;
+    gap: 0.5rem;
+  }
+
+  .content {
+    padding-left: 2.2rem;
+
+    .hds-form-field--layout-vertical {
```
ZedLi, line 681: Is this the right selector? It seems like we'll need margin for multiple things in different cards and this is very specific to the theme selector

ui/desktop/tests/acceptance/projects/settings/index-test.js
DhariniJeeva: moved it outside this PR as the tests might need some refactors to follow our most recent patterns

ui/desktop/app/components/settings-card/index.hbs
DhariniJeeva: changed it to `@action` as in the next tickets, we will add action button in this block

ui/desktop/app/styles/app.scss
DhariniJeeva: you are right, grid won't work in that case, changed to flex

ui/desktop/app/routes/scopes/scope/projects/settings/index.js
DhariniJeeva: yeah I'll make it a seprate ticket

ui/desktop/app/components/settings-card/index.hbs
ZedLi: Just want to understand, wouldn't we still need to yield content? What would we be passing to `@action`?

ui/desktop/app/routes/scopes/scope/projects/settings/index.js
ZedLi: Yeah it is, it's still a successful response and isn't blocking, errors can happen on any individual resource

ui/desktop/app/components/settings-card/index.hbs
DhariniJeeva: hmm, I didn't spend much time thinking this through, my initial thought was to pass in actions(signout/pause/restart) but that wouldn't work if there is more than one. now I'm thinking of passing in a named block to `@actions`?

ZedLi: Yup makes sense, I think we'd have to do named blocks in this component

DhariniJeeva: updated!

ZedLi (approved):
Looks great, thanks for getting things setup!

PR #2410, opened 2024-07-19
refactor: 💡 small text updates & loading state

lisbet-alvarez:
# Description

Minor text updates and generic loading state.
While testing with a lot of scopes it became apparent that a loading state is needed in the `scopes.scope.roles.role` route since that is where the scopes are pre loaded.

image

Uploading Screen Recording 2024-07-22 at 4.02.20 PM.mov…

## Checklist

### DIFF
--- a/addons/core/translations/resources/en-us.yaml
+++ b/addons/core/translations/resources/en-us.yaml
@@ -586,7 +586,7 @@ role:
         description: Select which scopes you want to add to this role.
         success: Scope selection saved
       keywords-selected:
-        description: Just a heads up, you can't select both add all descendants and add all children. You can only choose one or the other.
+        description: Just a heads up, you can't select both <strong>add all descendants</strong> and <strong>add all children</strong>. You can only choose one or the other.
         link: You can read more about it here
       manage-custom-scopes:
         title: Manage Custom Scopes
@@ -618,7 +618,7 @@ role:
     form:
       this:
         label: Add this scope
-        help: Add this scope.
+        help: Add {scopeDisplayName} scope.
       children:
         label: Add all children
         help:
--- a/addons/core/translations/resources/en-us.yaml
+++ b/addons/core/translations/resources/en-us.yaml
@@ -596,9 +596,9 @@ role:
         title: Manage Projects in {orgDisplayName}
         description: Select which projects in this org you want to apply to this role's grant sets.
         success: Projects in org have been updated
-      no-org-scopes:
-        title: No Org Scopes Available
-        description: Org scopes are not available for assignment.
+      no-scopes:
+        title: No {type} Scopes Available
+        description: '{type} scopes are not available for assignment.'
     keywords:
       this:
         sub-title: This scope
--- a/ui/admin/app/routes/scopes/scope/roles/role.js
+++ b/ui/admin/app/routes/scopes/scope/roles/role.js
@@ -5,6 +5,7 @@
 
 import Route from '@ember/routing/route';
 import { inject as service } from '@ember/service';
+import { action } from '@ember/object';
 
 export default class ScopesScopeRolesRoleRoute extends Route {
   // =services
@@ -39,5 +40,19 @@ export default class ScopesScopeRolesRoleRoute extends Route {
     if (role.scopeID !== scope.id) {
       this.router.replaceWith(transition.to.name, role.scopeID, role.id);
     }
+  }
+
+  // =actions
+
+  /**
+   * Event to determine whether the loading template should be shown.
+   * Only show the loading template when transitioning into the specified route.
+   * @param transition
+   * @returns {boolean}
+   */
+  @action
+  loading(transition) {
+    const to = transition.to?.name;
+    return to === 'scopes.scope.roles.role.index';
   }
 }
--- a/ui/admin/app/routes/scopes/scope/roles/role.js
+++ b/ui/admin/app/routes/scopes/scope/roles/role.js
@@ -46,13 +46,22 @@ export default class ScopesScopeRolesRoleRoute extends Route {
 
   /**
    * Event to determine whether the loading template should be shown.
-   * Only show the loading template when transitioning into the specified route.
+   * Only show the loading template during initial loads for specified
+   * routes. Don't show it when a user is just searching or
+   * filtering on the same page as it can be jarring.
    * @param transition
    * @returns {boolean}
    */
   @action
   loading(transition) {
     const to = transition.to?.name;
-    return to === 'scopes.scope.roles.role.index';
+    const from = transition.from?.name;
+    return (
+      (to === 'scopes.scope.roles.role.index' ||
+        to === 'scopes.scope.roles.role.manage-scopes.manage-custom-scopes' ||
+        to === 'scopes.scope.roles.role.manage-scopes.manage-org-projects' ||
+        to === 'scopes.scope.roles.role.scopes') &&
+      from !== to
+    );
   }
 }
--- a/ui/admin/app/routes/scopes/scope/roles/role/manage-scopes/manage-org-projects.js
+++ b/ui/admin/app/routes/scopes/scope/roles/role/manage-scopes/manage-org-projects.js
@@ -39,12 +39,19 @@ export default class ScopesScopeRolesRoleManageScopesManageOrgProjectsRoute exte
       scope_id: [{ equals: org_id }],
     };
 
-    const projectScopes = await this.store.query('scope', {
-      scope_id: org_id,
-      query: { search, filters },
-      page,
-      pageSize,
-    });
+    // All scopes have already been pre-loaded in the `scopes.scopes.roles.role` route.
+    // Therefore, we can simply peek into indexedDB to retrieve results.
+    const options = { peekIndexedDB: true };
+    const projectScopes = await this.store.query(
+      'scope',
+      {
+        scope_id: org_id,
+        query: { search, filters },
+        page,
+        pageSize,
+      },
+      options,
+    );
     const totalItems = projectScopes.meta?.totalItems;
     const totalItemsCount = await this.getTotalItemsCount(
       org_id,
--- a/ui/admin/app/routes/scopes/scope/roles/role/manage-scopes/manage-org-projects.js
+++ b/ui/admin/app/routes/scopes/scope/roles/role/manage-scopes/manage-org-projects.js
@@ -39,19 +39,13 @@ export default class ScopesScopeRolesRoleManageScopesManageOrgProjectsRoute exte
       scope_id: [{ equals: org_id }],
     };
 
-    // All scopes have already been pre-loaded in the `scopes.scopes.roles.role` route.
-    // Therefore, we can simply peek into indexedDB to retrieve results.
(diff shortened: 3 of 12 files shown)


ui/admin/tests/acceptance/roles/global-scope-test.js
```diff
@@ -138,7 +138,9 @@ module('Acceptance | roles | global-scope', function (hooks) {
     await waitUntil(
       () =>
         findAll(GRANT_SCOPE_ROW_SELECTOR(instances.role.grant_scope_ids[1]))
-          .length === 0,
```
lisbet-alvarez, line 141: two test cases started failing. im thinking its because a reload causes the loading state to momentarily appear which was causes the waitUntil to pass without actually waiting for the search request to complete.

ui/admin/app/templates/scopes/scope/roles/role/loading.hbs
```diff
@@ -0,0 +1,9 @@
+{{!
+  Copyright (c) HashiCorp, Inc.
+  SPDX-License-Identifier: BUSL-1.1
+}}
+
+<Hds::ApplicationState class='loading-indicator' as |A|>
+  <A.Header @title={{t 'states.loading.title'}} @icon='loading' />
+  <A.Body @text={{t 'states.loading.generic-description'}} />
+</Hds::ApplicationState>
```
lisbet-alvarez, line 9: we preload the scopes in the `scopes.scope.roles.role` route and if there are a lot of scopes then there is an apparent lag. So I added this generic loading state.

ui/admin/tests/acceptance/roles/global-scope-test.js
cameronperera: I played around with this a bit this morning and I think you are right. Seemed to be jumpy a bit with the loading screen.

cameronperera (approved):

ui/admin/app/templates/scopes/scope/roles/role/loading.hbs
ZedLi: Should this be in the `roles/loading` or `roles/role/loading`? I forget how the loading hierarchy works, is the roles loading indicator covered by `scopes/scope/loading.hbs` but role isn't?

ZedLi:
I'm also noticing I can't paginate on the projects table

https://github.com/user-attachments/assets/6a87c7d2-db16-4e3c-81ef-3a4e46036fd0

ui/admin/app/templates/scopes/scope/roles/role/loading.hbs
ZedLi, line 1: I'm also noticing possibly inefficient loading of scopes at the global level. It seems like we load just the scopes non recursively when entering `roles` which seems fine, but when going into a role inside of roles, we load scopes at the scope with recursive, i.e. ```{
  "scope_id": global,
  "recursive": true
} ```.

This makes sense to load the table but then when we goto manage the custom scopes, it once again uses the scopes non-recursively which is fine since that got covered by the scopes loaded from `roles`. But it seems when you go into projects, it will load scopes for that org, which if you have a lot of projects, is another long initial load if it's your first time entering. 

This seems like it could've been covered by the recursive scopes that were already loaded and you didn't need to make another unique scopes call, thoughts?

I was testing with 10000 projects in an org.

addons/core/translations/resources/en-us.yaml
```diff
@@ -586,7 +586,7 @@ role:
         description: Select which scopes you want to add to this role.
         success: Scope selection saved
       keywords-selected:
-        description: Just a heads up, you can't select both add all descendants and add all children. You can only choose one or the other.
+        description: Just a heads up, you can't select both <strong>add all descendants</strong> and <strong>add all children</strong>. You can only choose one or the other.
```
calcaide, line 589: None blocking comment: I do understand the <strong> element, I am curious why we are introducing its usage now and never before, do you mind providing a screenshot of this text rendered within PR description?
I do assume this was a suggestion from design?

ui/admin/app/templates/scopes/scope/roles/role/loading.hbs
calcaide: I agree with part of Zed comment, so I will wait to see what's the outcome to add my 2cents in case are need, but we should review this part.

lisbet-alvarez: @ZedLi  thanks for walking thru it over zoom. With a lot of projects I do notice that it was a very noticeable lag. Since everything is already loaded in the `role` route then we can just peek and avoid a network request all together. This also reduces lag. So far don't see any issues with going about it this way.

addons/core/translations/resources/en-us.yaml
lisbet-alvarez: yup this is the specific part that has the bolded text:
image

It was introduced as feedback and approved by design.

ui/admin/app/templates/scopes/scope/roles/role/loading.hbs
lisbet-alvarez: Yeah `roles` is covered by the loading template in the `scope` path tho im unsure why it doesn't work for the `role` route since it should bubble up to find a loading template. 🤷

Great catch, thank you for pointing out the `loading` action to prevent from showing the loading indicator in search/filtering.

ui/admin/app/routes/scopes/scope/roles/role/manage-scopes/manage-org-projects.js
```diff
@@ -39,12 +39,19 @@ export default class ScopesScopeRolesRoleManageScopesManageOrgProjectsRoute exte
...
-    const projectScopes = await this.store.query('scope', {
-      scope_id: org_id,
-      query: { search, filters },
-      page,
-      pageSize,
-    });
+    // All scopes have already been pre-loaded in the `scopes.scopes.roles.role` route.
+    // Therefore, we can simply peek into indexedDB to retrieve results.
+    const options = { peekIndexedDB: true };
+    const projectScopes = await this.store.query(
```
ZedLi, line 45: This wasn't working for me if a user has never loaded anything into indexed db and goes to the manage scopes for projects directly. I think it's because the scopes are being loaded in `role.index` and not `role` which means it's not considered a child of the route so it won't pre-load the scopes and doesn't return any results.

You could either load the scopes in `role` or honestly we could just include the `recursive` flag which would share the same query anyways since the token that's being stored would just be re-used

ui/admin/app/routes/scopes/scope/roles/role.js
```diff
@@ -40,4 +41,18 @@ export default class ScopesScopeRolesRoleRoute extends Route {
...
+  /**
+   * Event to determine whether the loading template should be shown.
+   * Only show the loading template when transitioning into the specified route.
+   * @param transition
+   * @returns {boolean}
+   */
+  @action
+  loading(transition) {
+    const to = transition.to?.name;
+    return to === 'scopes.scope.roles.role.index';
```
ZedLi, line 56: I think you need to check for both `from` and `to`, this seems to prevent a loading indicator going from the roles to the individual role route

ui/admin/app/routes/scopes/scope/roles/role/manage-scopes/manage-org-projects.js
lisbet-alvarez: well thats good to know

good catch! im leaning towards reusing the same query instead. there could be more scenarios i missed and that seems to be the safer option.

ui/admin/app/routes/scopes/scope/roles/role.js
lisbet-alvarez: im probably misunderstanding but when going from `roles` to `role` i am seeing the loading screen.

https://github.com/user-attachments/assets/c5f720f9-cda5-4bae-9eee-fa2206f7836a

ZedLi: Woops yeah I'm seeing it now, I think it was bc I forgot I moved the scopes loading to `roles` from `roles.index` for testing which changed the logic, looks fine now! Sorry false alarm!

But the other thing I'm noticing is no loading screens for any of the manage-scope screens, do you see that too?

lisbet-alvarez: ah yup ill fix that

ui/admin/app/templates/scopes/scope/roles/role/manage-scopes/loading.hbs
ZedLi, line 1: Hmm we need another template bc it doesn't use the one above in the hierarchy? 🤔

ui/admin/app/routes/scopes/scope/roles/role.js
```diff
@@ -40,4 +41,27 @@ export default class ScopesScopeRolesRoleRoute extends Route {
...
+  @action
+  loading(transition) {
+    const to = transition.to?.name;
+    const from = transition.from?.name;
+    return (
+      (to === 'scopes.scope.roles.role.index' ||
+        to === 'scopes.scope.roles.role.manage-scopes.manage-custom-scopes' ||
+        to === 'scopes.scope.roles.role.manage-scopes.manage-org-projects' ||
+        to === 'scopes.scope.roles.role.scopes') &&
+      from !== to
```
ZedLi, line 64: Not blocking but should some of these have been in another `loading` action in the manage scopes route instead?

ZedLi (approved):

ui/admin/app/templates/scopes/scope/roles/role/manage-scopes/loading.hbs
lisbet-alvarez: correct, it needs to be handled separately at every level

ui/admin/app/routes/scopes/scope/roles/role.js
lisbet-alvarez: i attempted to handle the loading in the separate routes however, the issue is that when transitioning in between routes it still uses the loading action in the `role` route which was complicating things so decided to handle it all together.

ui/admin/app/templates/scopes/scope/roles/role/manage-scopes/loading.hbs
ZedLi: I've read the docs quite a few times and still don't really understand it then haha, I'm not exactly sure why it worked for our other resources by just having it at `scopes/scope/loading` 🤷

lisbet-alvarez: i re-read those docs so many times 😭 what im getting from it is that it looks for the loading template in the folder for the common parent but not above that so im guessing thats why it shows the loading template when transitioning into the list routes but doesn't trigger for the searching in the list routes since thats one level deeper (index routes). i could be completely wrong tho 😌

ZedLi: Yeah I was looking at the common parent and thought it made sense for the manage-scopes and role one since they share one (role.index and role.manage-scopes), but maybe not

PR #2413, opened 2024-07-25
fix: 🐛 tag count in list view

lisbet-alvarez:
# Description

Before:

Screenshot 2024-07-24 at 6 24 32 PM

After:

image

## How to Test

1. start boundary dev instance
2. create an api tag that has the same key and value as an existing config tag
3. make sure that the count in list view for workers is accurate. 

## Checklist

### DIFF
--- a/addons/api/addon/models/worker.js
+++ b/addons/api/addon/models/worker.js
@@ -37,9 +37,6 @@ export default class WorkerModel extends GeneratedWorkerModel {
    * @type {number}
    */
   get tagCount() {
-    if (!this.config_tags && !this.api_tags) {
-      return 0;
-    }
     const allTags = [
       ...Object.values(this.config_tags ?? {}),
       ...Object.values(this.api_tags ?? {}),
--- a/addons/api/addon/models/worker.js
+++ b/addons/api/addon/models/worker.js
@@ -4,7 +4,6 @@
  */
 
 import GeneratedWorkerModel from '../generated/models/worker';
-import { attr } from '@ember-data/model';
 
 export const TYPE_WORKER_PKI = 'pki';
 
@@ -38,15 +37,12 @@ export default class WorkerModel extends GeneratedWorkerModel {
    * @type {number}
    */
   get tagCount() {
-    if (
-      !Object.keys(this.config_tags).length &&
-      !Object.keys(this.api_tags).length
-    ) {
+    if (!this.config_tags && !this.api_tags) {
       return 0;
     }
     const allTags = [
-      ...Object.values(this.config_tags),
-      ...Object.values(this.api_tags),
+      ...Object.values(this.config_tags ?? {}),
+      ...Object.values(this.api_tags ?? {}),
     ];
     return allTags.reduce(
       (previousCount, currentTags) => previousCount + currentTags.length,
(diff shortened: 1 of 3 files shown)


addons/api/addon/models/worker.js
```diff
@@ -34,14 +34,18 @@ export default class WorkerModel extends GeneratedWorkerModel {
...
 
   /**
-   * Returns the number of canonical tags present on the worker.
+   * Returns the number of tags present on the worker.
    * @type {number}
    */
   get tagCount() {
     if (!this.canonical_tags) {
       return 0;
     }
```
cameronperera, line 43: Thanks for spotting this bug!

Looking at this now, we need to change this check here. We updated the serializer to always set `canonical`, `config`, and `api` tags to an empty object if they are not present. Maybe something like this is what we need?
```suggestion
    if (
      !Object.keys(this.config_tags).length &&
      !Object.keys(this.api_tags).length
    ) {
      return 0;
    }
```

lisbet-alvarez: ah yes great catch, thank you!

addons/api/addon/models/worker.js
ZedLi, line 1: Should we have allowed duplicate API tag that's the same as a config tag? I'm wondering if we should even have allowed duplicates since `canonical` tags are what get used in the end and they de-dupe

addons/api/addon/models/worker.js
```diff
@@ -34,14 +34,21 @@ export default class WorkerModel extends GeneratedWorkerModel {
...
   /**
-   * Returns the number of canonical tags present on the worker.
+   * Returns the number of tags present on the worker.
    * @type {number}
    */
   get tagCount() {
-    if (!this.canonical_tags) {
+    if (
+      !Object.keys(this.config_tags).length &&
+      !Object.keys(this.api_tags).length
```
ZedLi, line 43: It seems like it's erroring out trying to goto the workers list page:

image

ZedLi (changes requested):

addons/api/addon/models/worker.js
lisbet-alvarez: If canonical tags are what get used then we should not allow user to create a duplicate api_tag. Ill reach out to foundations team since we rely on api for error checking.

addons/api/addon/models/worker.js
ZedLi: It doesn't look like it correctly normalizes the response for list responses right now but also, if these were getting normalized correctly, you shouldn't even need this if check.

addons/api/addon/models/worker.js
lisbet-alvarez: For now we'll just fix the count and once foundations gets back to us on this [thread](https://hashicorp.slack.com/archives/C014AHXUL3G/p1721926066351699) we can possibly make an improvement on the workflow.

addons/api/addon/models/worker.js
```diff
@@ -34,14 +33,18 @@ export default class WorkerModel extends GeneratedWorkerModel {
...
   /**
-   * Returns the number of canonical tags present on the worker.
+   * Returns the number of tags present on the worker.
    * @type {number}
    */
   get tagCount() {
-    if (!this.canonical_tags) {
+    if (!this.config_tags && !this.api_tags) {
       return 0;
     }
```
ZedLi, line 42: I don't believe this if check is actually necessary since you do a default on lines 44 and 45 which should work fine and just add up two empty arrays to 0 anyways. 

```suggestion
```

ZedLi (approved):

PR #2416, opened 2024-07-26
feat: 🎸 Setup initial desktop E2E tests

ZedLi:
✅ Closes: https://hashicorp.atlassian.net/browse/ICU-14061

# Description
Sets up the initial end to end tests for desktop client. I didn't use a helper class but instead integrated them as fixtures in combination with [page object models](https://playwright.dev/docs/pom).

These fixtures get loaded as part of each test and automatically setup and tear down the things needed for the electron app to be tested as well as provide helpers for common actions.

Forced Failure:
image

image

image

## How to Test
Run `npx playwright test --reporter=html` to get an html report. Force an error by adding something like `expect(1).toBe(2)` to a test to see an example of a trace and screenshot that gets captured.

Note that the trace is a bit buggy though the html viewer (still works, just images/css not loaded correctly for the snapshots though screenshots look fine). You can see it rendered properly by taking the trace in `tests/e2e/artifacts` and passing it to `npx playwright show-trace`

## Checklist


ui/desktop/tests/e2e/authentication.spec.js
```diff
@@ -0,0 +1,86 @@
...
+    clusterUrl,
+    username,
+    password,
+  }) => {
+    const loginPage = new LoginPage(electronPage);
+    await loginPage.setClusterUrl(clusterUrl);
+    await loginPage.loginWithPassword(username, password);
+
+    await expect(electronPage).toHaveURL(
+      /.*\/scopes\/global\/projects\/targets$/,
```
cameronperera, line 24: Are we using regex instead of a string because we can ignore the base?

ui/desktop/tests/e2e/authentication.spec.js
```diff
@@ -0,0 +1,86 @@
...
+    );
+
+    await electronPage
+      .locator('details')
+      .filter({ hasText: 'Sign Out' })
+      .click();
+    // await electronPage.locator('details').filter({ hasText: 'Sign Out' }).locator('summary').click();
```
cameronperera, line 31: Do we still need this?

ZedLi: Nope we don't, was leftover by accident

ui/desktop/tests/e2e/authentication.spec.js
ZedLi: Yup basically, and this assertion doesn't take glob so we have to use simple regex

ui/desktop/tests/e2e/fixtures/authenticateTest.js
```diff
@@ -0,0 +1,28 @@
...
+ * Copyright (c) HashiCorp, Inc.
+ * SPDX-License-Identifier: BUSL-1.1
+ */
+
+/* eslint-disable no-undef, no-empty-pattern */
+
+const { electronTest } = require('../fixtures/electronTest');
+const LoginPage = require('../pages/loginPage');
+
+exports.authenticateTest = electronTest.extend({
```
calcaide, line 11: C-L-E-A-N I really like it the usage of fixtures here!

ui/desktop/tests/e2e/fixtures/electronTest.js
```diff
@@ -0,0 +1,133 @@
...
+ */
+const getExecutablePath = (buildDirectory = 'out') => {
+  const rootDir = path.resolve('./electron-app');
+  // directory where the builds are stored
+  const outDir = path.resolve(rootDir, buildDirectory);
+  // list of files in the out directory
+  const builds = fs.readdirSync(outDir);
+
+  // list of platforms we support
+  const platforms = ['win32', 'darwin', 'linux'];
```
calcaide, line 24: Not blocking comment: Just heads up, once we merge `llb-improvements-and-transparent-session-changes` change within the pipeline, `win32` changes to `windows`.

Context on the change: `win32` is a not accepted by our blobl signature tool and has to be `windows`.

ui/desktop/tests/e2e/fixtures/electronTest.js
```diff
@@ -0,0 +1,133 @@
...
+const { test } = require('@playwright/test');
+const path = require('path');
+const fs = require('fs');
+
+/**
+ * Get the path to the Boundary Desktop executable
+ * @param buildDirectory
+ * @return {string}
+ */
+const getExecutablePath = (buildDirectory = 'out') => {
```
calcaide, line 16: Very clean!! way better than the previous hardcoded stuff 😅

ui/desktop/tests/e2e/authentication.spec.js
```diff
@@ -0,0 +1,86 @@
...
+
+/* eslint-disable no-undef */
+const { test } = require('./fixtures/baseTest');
+const LoginPage = require('./pages/loginPage');
+const { expect } = require('@playwright/test');
+
+/* eslint-disable no-undef */
+test.describe('user/password authentication tests', async () => {
+  test('Authenticates and signs out', async ({
+    electronPage,
```
cameronperera, line 14: Question for learning:
Looking through playwright docs for test and the [page](https://playwright.dev/docs/api/class-fixtures#fixtures-page). Are we calling this electronPage instead of page for any particular reason?

calcaide (approved):
Amazing work!!! very happy to see this running 😃

calcaide:
We can tweak or create new  yarn scripts to run specific reporter as the one you suggest: `npx playwright test --reporter=html` totally happy to swap the `e2e:dev` script to use the html reporter

ui/desktop/tests/e2e/authentication.spec.js
```diff
@@ -0,0 +1,86 @@
...
+    clusterUrl,
+    username,
+    password,
+  }) => {
+    const loginPage = new LoginPage(electronPage);
```
cameronperera, line 19: Question for learning:
We setup this `LoginPage` so we can use the predefined methods? However, it still points to the `electronPage` that we make our assertions against?

ZedLi:
> We can tweak or create new yarn scripts to run specific reporter as the one you suggest: `npx playwright test --reporter=html` totally happy to swap the `e2e:dev` script to use the html reporter

Yup good call, wasn't sure if we even wanted to make that default or a separate command since you could also just run it without the reporter as we were doing before

ui/desktop/tests/e2e/authentication.spec.js
```diff
@@ -0,0 +1,86 @@
...
+    await electronPage
+      .locator('details')
+      .filter({ hasText: 'Sign Out' })
+      .click();
+
+    await electronPage.getByRole('button', { name: 'Sign Out' }).click();
+    await expect(electronPage).toHaveURL(/.*\/scopes\/global\/authenticate\//);
+  });
+
+  test('Fails to authenticate with wrong password', async ({
```
cameronperera, line 74: Is this test actually needed since `loginWithLDAP` is calling `loginWithPassword`?

ui/desktop/tests/e2e/fixtures/authenticateTest.js
cameronperera: So this fixture is doing some basic setup before each test and authenticating for us correct?

ui/desktop/tests/e2e/settings.spec.js
```diff
@@ -0,0 +1,19 @@
+/**
+ * Copyright (c) HashiCorp, Inc.
+ * SPDX-License-Identifier: BUSL-1.1
+ */
+
+/* eslint-disable no-undef */
+const { test } = require('./fixtures/baseTest');
+const { expect } = require('@playwright/test');
+
+test.describe.fixme('Settings Page tests', async () => {
```
cameronperera, line 10: Are we using `fixme` here because we are still building out this page and it is in an llb?

ui/desktop/tests/e2e/targets.spec.js
```diff
@@ -0,0 +1,38 @@
...
+ * Copyright (c) HashiCorp, Inc.
+ * SPDX-License-Identifier: BUSL-1.1
+ */
+
+/* eslint-disable no-undef */
+const { test } = require('./fixtures/baseTest');
+const { expect } = require('@playwright/test');
+
+test.describe('Targets test', async () => {
+  test('Connects to a target and ends session', async ({ authedPage }) => {
```
cameronperera, line 11: Should we add a test for connecting from the targets route as well?

ui/desktop/tests/e2e/authentication.spec.js
ZedLi: Yup good question, it might be useful to provide some context as to what these fixtures even are. The `page` fixture you're referencing to is a helper to establish and setup a browser window (aka page). That's great for something like admin UI where we do use a browser. 

Unfortunately, the desktop client is an electron app so opening a browser window isn't quite what we want, we need something to actually open electron itself and then get a handle on the browser window _inside_ electron. I could've actually overriden the default `page` and have `electronPage` be called just `page` since I'm not aware of us needing to ever use it but I also didn't want to confuse people nor have people not realize it's actually using electron.  

`electronPage` comes from our own fixture to handle this setup, take a look at the code [here](https://github.com/hashicorp/boundary-ui/blob/5e9a706d7acbc7ca74221b983c842be9a1c294fe/ui/desktop/tests/e2e/fixtures/electronTest.js#L127). For context, look at how we [previously did it](https://github.com/hashicorp/boundary-ui/blob/219eba67bc36df4efe062809ff41b5e5bb96d114/ui/desktop/tests/end2end/authentication.spec.js#L43) and how we needed `beforeEach` hooks everytime to set up electron.

ui/desktop/tests/e2e/authentication.spec.js
ZedLi: Correct, I could have also called it `electronWindow` but it returns an actual `page` object in the end so that's what we're actually working in and asserting on. The `LoginPage` is meant to encapsulate the login screen itself, and we might have something like `TargetsPage` or `SettingsPage` that all still use the `electronPage` (more realistically `authedPage`) to handle common actions on that specific page.

ui/desktop/tests/e2e/authentication.spec.js
ZedLi: I'm not sure I quite understand the issue on why it wouldn't be needed. This test is to test authenticating with LDAP, which also happens to use a user and password to login (but is a different auth method). I could've not called `loginWithPassword` and keep that "only" for user/password login by duplicating the code but I didn't think it would ever change and would most likely affect both if it did.

Whether we login with OIDC, LDAP, or user/pass, we would still need to login with a username/password at the end of the day. Ideally we would also test OIDC but we have some limitations in being able to test that.

ui/desktop/tests/e2e/fixtures/authenticateTest.js
ZedLi: Yup, it's to simplify setup and not need to have `beforeAll/beforeEach` hooks for every test file

ui/desktop/tests/e2e/fixtures/electronTest.js
ZedLi: Yup good call, I can also add `windows` to the list for now and add a todo to remove `win32` later

ui/desktop/tests/e2e/settings.spec.js
ZedLi: Yup exactly, this is targeting main so settings won't exist yet. When the LLB gets rebased on top of main we can add tests then.

ui/desktop/tests/e2e/targets.spec.js
ZedLi: Yup probably. This wasn't to be an exhaustive list of tests, just to setup the test setups and fix what the previous existing tests were trying to do.

We'll need to add a lot more tests in the future but that's not the scope of this PR.

ui/desktop/tests/e2e/fixtures/electronTest.js
ZedLi: Actually I'm not sure it matters? I'm looking for the initial build made by electron forge which should be unmodified and not the one that we change in `postMake` hooks, does that sound right? It also looks like we already have the rename in `main`

ui/desktop/tests/e2e/pages/loginPage.js
```diff
@@ -0,0 +1,37 @@
...
+
+/* eslint-disable no-undef */
+
+class LoginPage {
+  constructor(page) {
+    this.page = page;
+  }
+
+  async setClusterUrl(clusterUrl) {
+    await this.page.evaluate(() => localStorage.clear());
```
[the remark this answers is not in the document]
moduli: It seems like I'm occasionally getting some failures here.

In some cases, it seems like the page loaded before the cluster url was cleared because it was showing the username/password screen. 

It one case, I got this error
```
1) tests/e2e/authentication.spec.js:13:3 › user/password authentication tests › Authenticates and signs out 

    Error: page.evaluate: Execution context was destroyed, most likely because of a navigation

       at tests/e2e/pages/loginPage.js:14

      12 |
      13 |   async setClusterUrl(clusterUrl) {
    > 14 |     await this.page.evaluate(() => localStorage.clear());
         |                     ^
      15 |     await this.page.getByLabel('Cluster URL').fill(clusterUrl);
      16 |     await this.page.getByRole('button', { name: 'Submit' }).click();
      17 |   }
```

ui/desktop/tests/e2e/fixtures/authenticateTest.js
```diff
@@ -0,0 +1,28 @@
...
+  clusterUrl: ({}, use) => {
+    use(process.env.BOUNDARY_ADDR ?? 'http://localhost:9200');
+  },
+  username: ({}, use) => {
+    use(process.env.E2E_PASSWORD_ADMIN_LOGIN_NAME ?? 'admin');
+  },
+  password: ({}, use) => {
+    use(process.env.E2E_PASSWORD_ADMIN_PASSWORD ?? 'password');
+  },
+  authedPage: async ({ electronPage, clusterUrl, username, password }, use) => {
```
moduli, line 22: Is there a way to use this [authentication state](https://playwright.dev/docs/auth) to log in once at the beginning of all tests and reuse that state? (not sure if this would work in the desktop client). We do use this in the Admin UI test suite.

ui/desktop/tests/e2e/authentication.spec.js
```diff
@@ -0,0 +1,86 @@
...
+    const loginPage = new LoginPage(electronPage);
+    await loginPage.setClusterUrl(clusterUrl);
+    await loginPage.loginWithPassword(username, 'wrong password!!');
+
+    await expect(electronPage).toHaveURL(/.*\/scopes\/global\/authenticate\//);
+    await expect(electronPage.getByText('Authentication Failed')).toBeVisible();
+  });
+});
+
+test.describe('LDAP authentication tests', async () => {
```
moduli, line 50: My initial thought was to avoid using `boundary dev` and have a full dockerized setup to truly be end-to-end. However, this would mean we would need to do some additional setup on boundary's end to create an LDAP auth method and the associated resources. This is where I thought some shared methods might make sense (perhaps some methods that use HTTP endpoints). 

Thoughts?

ui/desktop/tests/e2e/pages/loginPage.js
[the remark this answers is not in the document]
moduli: Also, it seems like something is not behaving correctly when using `npx playwright test --debug` (it opens to an authenticated state for me)

ui/desktop/tests/e2e/pages/loginPage.js
moduli, line 1: Page Object Model could be something we consider porting to on the Admin UI test suite side.

moduli (commented):
Happy to see some movement on this!

ui/desktop/tests/e2e/settings.spec.js
```diff
@@ -0,0 +1,19 @@
...
+  test('Authenticates using user and password method and signs out', async ({
```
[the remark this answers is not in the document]
ZedLi: Woops yeah, this was just a placeholder test once settings page is available, I can fix that.

ui/desktop/tests/e2e/pages/loginPage.js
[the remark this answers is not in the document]
ZedLi: Yeah that's the goal, it's to clear out the cluster url from localstorage (and any other state). I'll take a look to see if i can reproduce the occasional failure

ui/desktop/tests/e2e/pages/loginPage.js
```diff
@@ -0,0 +1,37 @@
...
+  async loginWithLDAP(username, password) {
+    await this.page
+      .getByRole('link', {
+        name: 'Generated global scope initial ldap auth method',
+      })
+      .click();
+    await this.loginWithPassword(username, password);
+  }
+
+  // Can't login with OIDC as we can't control the external browser that is opened
```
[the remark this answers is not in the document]
ZedLi: Yes unfortunately it's true. The issue is we don't open the new window in electron, we actually use the user's default browser they have set which means we can't have playwright hook into it to have control. It's similar to this issue [here](https://github.com/microsoft/playwright/issues/13014#issuecomment-1078363373)

ui/desktop/tests/e2e/fixtures/authenticateTest.js
ZedLi: Unfortunately this doesn't have support in electron, it works great if you're just using a browser but it seems electron support is lacking. 

It was the first thing I tried getting to work but unless I missed something, it doesn't look like it works properly in electron

ui/desktop/tests/e2e/authentication.spec.js
ZedLi: Yeah agreed. To be clear, this is just the start to get things setup and having the tests passing again. I think ideally in the long run we'd want to use a real boundary setup and probably enos to get an accurate end to end tests.

But in the meantime I was just using boundary dev and am likely to use some of the admin UI e2e cli methods to set some things up so we can get some tests working quicker

ui/desktop/tests/e2e/pages/loginPage.js
```diff
@@ -0,0 +1,37 @@
...
+    this.page = page;
+  }
+
+  async setClusterUrl(clusterUrl) {
+    await this.page.evaluate(() => localStorage.clear());
+    await this.page.getByLabel('Cluster URL').fill(clusterUrl);
+    await this.page.getByRole('button', { name: 'Submit' }).click();
+  }
+
+  async loginWithPassword(username, password) {
```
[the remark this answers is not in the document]
ZedLi: Good catch, updated!

ui/desktop/playwright.config.js
```diff
@@ -0,0 +1,14 @@
+/**
+ * Copyright (c) HashiCorp, Inc.
+ * SPDX-License-Identifier: BUSL-1.1
+ */
+
+/* eslint-disable no-undef */
+
+/** @type {import('@playwright/test').PlaywrightTestConfig} */
+const config = {
+  outputDir: './tests/e2e/artifacts',
```
DhariniJeeva, line 10: what is stored here?

ui/desktop/tests/e2e/fixtures/authenticateTest.js
```diff
@@ -0,0 +1,28 @@
...
+ * SPDX-License-Identifier: BUSL-1.1
+ */
+
+/* eslint-disable no-undef, no-empty-pattern */
+
+const { electronTest } = require('../fixtures/electronTest');
+const LoginPage = require('../pages/loginPage');
+
+exports.authenticateTest = electronTest.extend({
+  // TODO: Should we put these in a config file?
```
DhariniJeeva, line 12: I think a config file would make it easy for any future reference

ui/desktop/playwright.config.js
```diff
@@ -0,0 +1,14 @@
+/**
+ * Copyright (c) HashiCorp, Inc.
+ * SPDX-License-Identifier: BUSL-1.1
+ */
+
+/* eslint-disable no-undef */
+
+/** @type {import('@playwright/test').PlaywrightTestConfig} */
```
moduli, line 8: Is this line needed?

ui/desktop/playwright.config.js
moduli, line 1: Do we want this config file to live here? In Admin UI, this file is located in the `tests/e2e/ directory

ui/desktop/playwright.config.js
ZedLi: Any thing we save from tests, usually traces

ui/desktop/tests/e2e/fixtures/authenticateTest.js
ZedLi: Yup I think so too. I think once we start building out our config it might make sense to move them over

ui/desktop/tests/e2e/authentication.spec.js
```diff
@@ -0,0 +1,86 @@
...
+ * Copyright (c) HashiCorp, Inc.
+ * SPDX-License-Identifier: BUSL-1.1
+ */
+
+/* eslint-disable no-undef */
+const { test } = require('./fixtures/baseTest');
+const LoginPage = require('./pages/loginPage');
+const { expect } = require('@playwright/test');
+
+/* eslint-disable no-undef */
```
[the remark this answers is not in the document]
ZedLi: I think I'm going to try and clean these up in our eslint configs

ui/desktop/playwright.config.js
ZedLi: That's the type definition which is useful when writing the config out

ui/desktop/playwright.config.js
ZedLi: Yeah I saw that, I think this config should live in the root

ui/desktop/tests/e2e/authentication.spec.js
cameronperera: oooh ok that makes sense. I see the connection now and why we are calling it electronPage. I would agree that the naming fits. Thanks for pointing me in the right direction, I didn't notice that the first pass.

ui/desktop/tests/e2e/authentication.spec.js
cameronperera: ok never mind my original question. With the added context careful reading, I see the issue with my original thought and question. Please ignore 😆

cameronperera (approved):
Looks great. Thanks for the hard work getting this pattern set up.

ui/desktop/tests/e2e/authentication.spec.js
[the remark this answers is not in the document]
ZedLi: Updated our eslint config to clean this up a bit

ui/desktop/tests/e2e/pages/loginPage.js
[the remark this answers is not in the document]
ZedLi: I moved the localstorage clearing to earlier on as part of the electron test fixture setup, and it seems to have fixed it while we were debugging.

moduli (approved):

PR #2419, opened 2024-07-30
fix: 🐛 Fix intel macs not getting proper auto-update prompt

ZedLi:
# Description
Fixes an issue we were seeing where intel macs weren't properly getting auto-updated. We were returning `x64` instead of `amd64` and being unable to find the URL because of that. 

## How to Test
Creating a build [here](https://github.com/hashicorp/boundary-ui-releases/actions/runs/10170588076) and will test on our mac instance.

## Checklist

### DIFF
--- a/ui/desktop/electron-app/src/helpers/app-updater.js
+++ b/ui/desktop/electron-app/src/helpers/app-updater.js
@@ -35,7 +35,7 @@ const returnArchitectureToUpdate = () => {
   } catch (err) {
     log.error(`returnArchitectureToUpdate:`, err.message);
   }
-  return nodeArchitecture;
+  return 'amd64';
 };
 
 // Query releases url to find latest version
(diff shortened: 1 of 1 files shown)


ui/desktop/electron-app/src/helpers/app-updater.js
```diff
@@ -61,9 +61,11 @@ const findUpdateArchive = (version) => {
   const url = `${releasesUrl}${version}/boundary-desktop_${version}_darwin_${architecture}.zip`;
   return new Promise((resolve, reject) => {
     https.get(url, (response) => {
-      if (response.statusCode === 403)
```
ZedLi, line 64: We should be returning an error on any non `200` response. This likely made it harder for us to debug as we couldn't see this error out as it was a `404` response.

cameronperera (approved):

lisbet-alvarez (approved):
thx for fixing this

calcaide (approved):

DhariniJeeva (approved):

PR #2421, opened 2024-08-01
ICU-14496 add discard unsaved changes modal to manage scopes

lisbet-alvarez:
✅ Closes: https://hashicorp.atlassian.net/browse/ICU-14496

# Description

https://github.com/user-attachments/assets/be4e184b-f6ff-4029-9813-577b60f53113

## How to Test

1. Test "discard" and "cancel" workflows for manage-scopes page modal
2. Test "discard" and "cancel" workflows for manage-custom-scopes page modal
3. Test "discard" and "cancel" workflows for manage-org-projects page modal
4. Test "save" and "cancel" workflows on all pages.

## Checklist


ui/admin/app/components/form/role/manage-org-projects/index.js
```diff
@@ -31,14 +28,14 @@ export default class FormRoleManageOrgProjectsIndexComponent extends Component {
...
-        if (!includesId) this.selectedItems = [...this.selectedItems, key];
-      } else {
-        if (includesId)
-          this.selectedItems = this.selectedItems.filter(
-            (item) => item !== key,
-          );
+        if (index === -1) {
+          this.args.selectedItems.push(key);
+        }
+      } else if (index !== -1) {
```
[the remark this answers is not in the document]
lisbet-alvarez: yup it is needed cuz the check on line 34 is nested under another check. It is confusing to look at tho, ill use `&&` instead.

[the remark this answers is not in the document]
cameronperera: oh dang I need to pay closer attention to the formatting. My bad, you are absolutely right.

cameronperera (approved):
Tested, works great!

ui/admin/app/components/form/role/manage-custom-scopes/index.js
```diff
@@ -30,13 +27,14 @@ export default class FormRoleManageCustomScopesIndexComponent extends Component
...
-      const includesId = this.selectedItems.includes(key);
+      // If index is equal to -1 then key does not exist in the array.
+      const index = this.args.selectedItems.indexOf(key);
       if (isSelected) {
-        if (!includesId) this.selectedItems = [...this.selectedItems, key];
-      } else if (includesId) {
-          this.selectedItems = this.selectedItems.filter(
-            (item) => item !== key,
-          );
+        if (index === -1) {
```
ZedLi, line 33:
```
if (isSelected && index === -1) {
  this.args.selectedItems.push(key);
} else if (!isSelected && index !== -1) {
  this.args.selectedItems.splice(index, 1);
}
```

I think either way works but I think this might be a bit easier to read? It duplicates the check for `isSelected` but sometimes that makes it easier to understand, I also at first glance thought the same thing @cameronperera did

ui/admin/app/controllers/scopes/scope/roles/role/manage-scopes/index.js
```diff
@@ -31,14 +33,17 @@ export default class ScopesScopeRolesRoleManageScopesIndexController extends Con
...
     this.router.replaceWith('scopes.scope.roles.role.scopes');
   }
 
   /**
    * Redirect to role scopes as if nothing ever happened.
+   * @param {[string]} grantScopeIDs
    */
   @action
-  cancel() {
+  cancel(grantScopeIDs) {
```
ZedLi, line 45: Do these cancel actions need to take in any arguments? It looks like you pass in fields from the model which haven't been modified so couldn't you setup the controllers to have access to the models and just use them directly?

ui/admin/app/routes/scopes/scope/roles/role/manage-scopes/index.js
```diff
@@ -18,4 +40,34 @@ export default class ScopesScopeRolesRoleManageScopesIndexRoute extends Route {
...
   }
+
+  // =actions
+
+  /**
+   * Triggers confirm pop-up only when user has made changes and is trying to navigate away from current route.
+   * @param {object} transition
+   */
+  @action
+  async willTransition(transition) {
```
ZedLi, line 51: I see that you needed these in 3 routes, just out of curiosity, was it not possible to directly modify the models for these role grants (instead of using a `selectedItems`) so that way we can have it just be rolled back to the original array? Was it too complicated?

lisbet-alvarez: I briefly considered this approach but then decided not to go that route because the grant_scopes_ids field is readOnly and thought it would go against existing patterns to save intermediate state in that field. However, i think it could def. work and produce a cleaner result. Im going to try this approach out.

ZedLi: Is `grant_scopes_id` readonly? I'm not sure why that would be the case

lisbet-alvarez: woopsies i think when i was working on this i looked at the wrong field, i looked again and it is not readOnly.

ui/admin/app/components/form/role/manage-custom-scopes/index.js
lisbet-alvarez: great point! updated!

ui/admin/app/routes/scopes/scope/roles/role/manage-scopes/index.js
lisbet-alvarez: refactored to access the field in the model directly.

ui/admin/app/controllers/scopes/scope/roles/role/manage-scopes/index.js
lisbet-alvarez: i was thinking about this. The reason some controller action take in the model as a parameter is because that action is reused in other places. Im leaning towards not using `this.model` in the controller to keep the pattern consistent . Thoughts ?

ZedLi (approved):
Looks good!

ui/admin/app/routes/scopes/scope/roles/role/manage-scopes/manage-org-projects.js
```diff
@@ -53,16 +56,14 @@ export default class ScopesScopeRolesRoleManageScopesManageOrgProjectsRoute exte
       totalItems,
     );
 
-    const { selectedProjectIDs, remainingProjectIDs } =
-      await this.getSelectedProjects(role, org_id);
+    const remainingProjectIDs = await this.getSelectedProjects(role, org_id);
```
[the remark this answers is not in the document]
lisbet-alvarez: ohhh great point, ill make that change.

ui/admin/app/routes/scopes/scope/roles/role/manage-scopes/manage-custom-scopes.js
```diff
@@ -128,4 +131,29 @@ export default class ScopesScopeRolesRoleManageScopesManageCustomScopesRoute ext
...
   }
+
+  // =actions
+
+  /**
+   * Triggers confirm pop-up only when user has made changes and is trying to navigate away from current route.
+   * @param {object} transition
+   */
+  @action
+  async willTransition(transition) {
```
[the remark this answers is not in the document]
lisbet-alvarez: yes that is exactly correct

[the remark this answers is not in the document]
ZedLi: Hmm I have a wild idea, it looks like the transition abort logic is nearly the same, what if we passed in something to [data](https://api.emberjs.com/ember/4.12/classes/Transition/properties/data?anchor=data) in the `willTransition`, something like:

`transition.data = { model: role };`

and then also check that in the application event handler so we have some sort of escape hatch?

`const maybeModel = transition?.data?.model ?? transition?.from?.attributes;`

That way these `willTransition` don't have to redo the same logic that's _basically_ the same.

Not sure if this is a good idea but just something I thought of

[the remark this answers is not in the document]
lisbet-alvarez: wow yeah this does seem like the perfect scenario for the data attribute, i didn't realize it was an option. Thanks for linking that. I don't see any cons with this approach and it think its very helpful for future scenarios (which i'm sure there will be) where the model hook doesn't return only the model.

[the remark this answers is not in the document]
lisbet-alvarez: thanks for the suggestion, this has been updated.

ui/admin/app/routes/application.js
```diff
@@ -44,14 +44,21 @@ export default class ApplicationRoute extends Route {
       const fromName = transition?.from?.name;
       const toName = transition?.to?.name;
       const maybeModel = transition?.from?.attributes;
-      if (fromName !== toName && maybeModel?.hasDirtyAttributes) {
+      const dataModel = transition?.data?.model;
```
ZedLi, line 47: Could you have done `const maybeModel = transition?.data?.model ?? transition?.from?.attributes;`?

That way you don't have to check two models in the if check or rolling back and you just decide which model to use in the very beginning

lisbet-alvarez: great point, updated!

PR #2424, opened 2024-08-01
chore: move theme selection to user settings page

DhariniJeeva:
# Description
this PR removes theme selection from menu bar and moves to settings page

 :tickets: [Jira ticket](https://hashicorp.atlassian.net/browse/ICU-14452)  

Screenshot 2024-08-01 at 9 38 11 AM
Screenshot 2024-08-01 at 9 38 05 AM

## Checklist

### DIFF
--- a/ui/desktop/app/controllers/scopes/scope/projects/settings/index.js
+++ b/ui/desktop/app/controllers/scopes/scope/projects/settings/index.js
@@ -3,6 +3,68 @@
  * SPDX-License-Identifier: BUSL-1.1
  */
 
-import Controller from '@ember/controller';
+import Controller, { inject as controller } from '@ember/controller';
+import { inject as service } from '@ember/service';
+import { getOwner } from '@ember/application';
+import { action } from '@ember/object';
 
-export default class ScopesScopeProjectsSettingsIndexController extends Controller {}
+const THEMES = [
+  {
+    label: 'system',
+    value: 'system-default-theme',
+  },
+  {
+    label: 'light',
+    value: 'light',
+  },
+  {
+    label: 'dark',
+    value: 'dark',
+  },
+];
+
+export default class ScopesScopeProjectsSettingsIndexController extends Controller {
+  @controller('scopes/scope/projects/settings/index') settings;
+  // Services
+
+  @service session;
+  @service ipc;
+
+  // Methods
+
+  /**
+   * Returns available themes
+   */
+  get themes() {
+    return THEMES;
+  }
+
+  /**
+   * Applies the specified color theme to the root ember element.
+   * @param {string} theme - "light", "dark", or nullish (system default) or an event object
+   */
+  @action
+  toggleTheme(theme) {
+    if (typeof theme === 'object') {
+      theme = theme.target.value;
+    }
+    const rootElementSelector = getOwner(this).rootElement;
+    const rootEl = getOwner(this)
+      .lookup('service:-document')
+      .querySelector(rootElementSelector);
+    this.session.set('data.theme', theme);
+    switch (theme) {
+      case 'light':
+        rootEl.classList.add('rose-theme-light');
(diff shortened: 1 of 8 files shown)


ui/desktop/app/routes/application.js
```diff
@@ -34,9 +33,13 @@ export default class ApplicationRoute extends Route {
...
     await this.session.setup();
-    const theme = this.session.get('data.theme');
-    this.toggleTheme(theme);
     await this.clusterUrl.updateClusterUrl();
+    const theme = this.session.get('data.theme');
+    /* eslint-disable-next-line ember/no-controller-access-in-routes */
+    const controller = this.controllerFor(
+      'scopes.scope.projects.settings.index',
+    );
+    controller.toggleTheme(theme);
```
DhariniJeeva, line 42: I had to do this so the color selection is applied when the app refreshes, on refresh we take the user to the landing page. If I were to set this in `settings` route `beforemodel` hook, then after a refresh, the user would only see their selection when they navigate to the settings route which didn't seem like a good UX.  Let me know your thoughts on this approach

DhariniJeeva: on second thought, maybe we shouldn't take them to the landing page(targets) on refresh? 🤔 By refresh I mean, a forced manual refresh (cmd+ r)

ui/desktop/app/controllers/scopes/scope/projects/settings/index.js
```diff
@@ -3,6 +3,68 @@
...
+    value: 'dark',
+  },
+];
+
+export default class ScopesScopeProjectsSettingsIndexController extends Controller {
+  @controller('scopes/scope/projects/settings/index') settings;
+  // Services
+
+  @service session;
+  @service ipc;
```
ZedLi, line 31: Is this ipc service needed?

ui/desktop/app/controllers/scopes/scope/projects/settings/index.js
```diff
@@ -3,6 +3,68 @@
...
+    value: 'light',
+  },
+  {
+    label: 'dark',
+    value: 'dark',
+  },
+];
+
+export default class ScopesScopeProjectsSettingsIndexController extends Controller {
+  @controller('scopes/scope/projects/settings/index') settings;
```
ZedLi, line 27: What was this for? Doesn't this refer to the same controller?

ui/desktop/app/routes/application.js
```diff
@@ -34,9 +33,13 @@ export default class ApplicationRoute extends Route {
   async beforeModel() {
     this.intl.setLocale(['en-us']);
     await this.session.setup();
-    const theme = this.session.get('data.theme');
-    this.toggleTheme(theme);
     await this.clusterUrl.updateClusterUrl();
+    const theme = this.session.get('data.theme');
+    /* eslint-disable-next-line ember/no-controller-access-in-routes */
+    const controller = this.controllerFor(
```
ZedLi, line 39: Hmm I don't think we should reach down into a child controller. I'm thinking the `toggleTheme` action you have in the controller should either be in a service or in this `application` controller and then you call it from the settings controller

DhariniJeeva: thanks for the feedback, I moved it to app controller

ui/desktop/app/controllers/scopes/scope/projects/settings/index.js
DhariniJeeva: yup this wasn't needed, I forgot to remove this

ui/desktop/app/controllers/scopes/scope/projects/settings/index.js
DhariniJeeva: thanks, not needed, removed!

ZedLi (approved):
Looks good!

PR #2425, opened 2024-08-02
chore: 🤖 add user info

DhariniJeeva:
# Description

Adds a section for username info 

:tickets: [Jira ticket](https://hashicorp.atlassian.net/browse/ICU-14401) 

Screenshot 2024-08-05 at 11 30 52 AM
Screenshot 2024-08-05 at 11 31 48 AM

Screenshot 2024-08-01 at 5 51 15 PM
## Checklist


ui/desktop/app/styles/app.scss
```diff
@@ -665,23 +665,29 @@ $chevron-down: url("data:image/svg+xml;utf-8,<svg viewBox='0 0 24 24' fill='%231
...
   .content {
     padding-left: 2.2rem;
   }
 
   .app-information {
-    display: flex;
-    justify-content: space-between;
+    display: grid;
     margin: 1rem 2rem 1rem 0;
+    grid-template-columns: repeat(3, 1fr);
```
DhariniJeeva, line 688: I changed this back to grid because with flex, I had to set width to control the space between columns and it didn't match with other rows. seemed simpler with grid, also works when we have two or three columns.

ZedLi: I think this works right now but flexbox offers you the flexibility to not _have_ to be in a multiple of the exact `fr` you defined (i.e. it takes up `1fr` or `2fr`, etc). For example, I see that we'll have a server card with a URL that could actually take up two thirds of the card most likely, we'd now need extra css to have that element span two `fr` to appear correct while flex requires no extra css for that.

Instead of width, I'd recommend using `flex-basis` so that these elements can take up more space than a column width if needed

ui/desktop/app/styles/app.scss
```diff
@@ -665,23 +665,30 @@ $chevron-down: url("data:image/svg+xml;utf-8,<svg viewBox='0 0 24 24' fill='%231
...
   .hds-card__container {
     padding: 1rem;
+    margin-bottom: 1rem;
   }
 
   .header-and-icon {
     display: flex;
-    align-items: center;
-    gap: 0.5rem;
+    align-items: flex-start;
```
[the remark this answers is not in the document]
DhariniJeeva: I thought it wasn't centered in figma yesterday and changed it, but now I can't really tell again. I think I will undo this change

ui/desktop/app/controllers/scopes/scope/projects/settings/index.js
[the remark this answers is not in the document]
DhariniJeeva: sounds good, I like this approach, I will refactor the template

ui/desktop/app/components/settings-card/index.hbs
```diff
@@ -11,18 +11,19 @@
       {{@header}}
     </Hds::Text::Display>
 
+    {{#if (has-block 'state')}}
+      <div class='state'>
```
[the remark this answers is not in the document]
DhariniJeeva: it wasn't, removed it !

ui/desktop/app/styles/app.scss
DhariniJeeva: thanks for the flex-basis reminder, I changed it to flex

ui/desktop/app/controllers/scopes/scope/projects/settings/index.js
```diff
@@ -4,5 +4,52 @@
...
+   */
+  get isAuthenticated() {
+    const { isAuthenticated } = this.session;
+    return isAuthenticated;
+  }
+  /**
+   * Returns the type of authmethod used by the user
+   * @type {string}
+   */
+  get authMethod() {
```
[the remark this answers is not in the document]
DhariniJeeva: I'm still working on this - Design confirmed that we display name + type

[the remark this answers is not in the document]
DhariniJeeva: updated!

ui/desktop/app/components/settings-card/user/index.js
```diff
@@ -0,0 +1,66 @@
...
+   * @type {string}
+   */
+  get authMethod() {
+    const {
+      data: {
+        authenticated: { authenticator },
+      },
+    } = this.session;
+    const formattedAuthenticator = authenticator.split(':')[1];
+    if (
```
ZedLi, line 48: Could we also capitalize the password type (`Password` instead of `password`)? I'm guessing you could just have the else statement capitalize everything else

ui/desktop/app/components/settings-card/user/index.hbs
```diff
@@ -0,0 +1,49 @@
...
+    </div>
+  </:body>
+
+  <:action>
+    <Hds::Button
+      data-test-signout-button
+      @text={{t 'actions.deauthenticate'}}
+      @color='critical'
+      @icon='sign-out'
+      @size='small'
```
ZedLi, line 45: Question, should this be small or medium? It seems a _little_ small but not sure if medium is too large?

ui/desktop/tests/acceptance/projects/settings/index-test.js
```diff
@@ -94,4 +101,30 @@ module('Acceptance | projects | settings | index', function (hooks) {
...
+
+  test('displays userinfo when authenticated', async function (assert) {
+    authenticateSession({ username: 'testuser' });
+    assert.expect(2);
+    await visit(urls.settings);
+    assert.ok(currentSession().isAuthenticated);
+    assert.dom(USERNAME_SELECTOR).hasText('testuser');
+  });
+
+  test('displays type of auth method when authenticated', async function (assert) {
```
ZedLi, line 113: I wonder if these should be integration tests instead, since it doesn't seem like it needs the rest of the page to be tested?

And that way we can also add tests that OIDC, LDAP, and password auth methods are displayed correctly

ZedLi (approved):
Looks good, just left a few minor comments

ui/desktop/app/components/settings-card/user/index.js
```diff
@@ -0,0 +1,66 @@
...
+      formattedAuthenticator === 'oidc' ||
+      formattedAuthenticator === 'ldap'
+    ) {
+      return formattedAuthenticator.toUpperCase();
+    }
+    return formattedAuthenticator;
```
ZedLi, line 54: Could we also capitalize the password type (`Password` instead of `password`)? I'm guessing you could just have this statement capitalize everything else

ui/desktop/app/components/settings-card/user/index.hbs
DhariniJeeva: good catch, medium looks better

ui/desktop/app/components/settings-card/user/index.js
DhariniJeeva: I ended up using translation for the types here to avoid the extra processing and updated the var to `authenticatorType`

ui/desktop/tests/acceptance/projects/settings/index-test.js
DhariniJeeva: good point, moved to integration and added an assert for types display

PR #2426, opened 2024-08-02
refact(e2e): Update Admin UI tests to use Page Objects

moduli:
This PR refactors the Admin UI e2e test suite to use [Page Objects](https://playwright.dev/docs/pom). This was done to reorganize the growing `boundary-ui.js` file, while also aligning with a pattern being proposed for the Desktop e2e tests. 
- `boundary-ui.js` was split into multiple page objects in `pages/`
- tests were updated to reference the new methods

All tests were run locally to confirm that the refactor has not broken anything.
```
yarn run e2e:ce:docker
yarn run e2e:ce:aws
yarn run e2e:ent:docker
yarn run e2e:ent:aws
```

https://hashicorp.atlassian.net/browse/ICU-14468


ui/admin/tests/e2e/helpers/boundary-ui.js
moduli, line 1: There's one method here that I'm still debating how to handle

We have a generic method for deleting resources. Do we want to create multiple methods that do the same thing?
- `HostCatalogsPage:deleteHostCatalogs`
- `HostCatalogsPage:deleteHostSets`
- `HostCatalogsPage:deleteHosts`
- ...

Or do we create some kind of `global.js` file for methods like this?

```
/**
 * Uses the UI to delete a Boundary resource. Assume you have selected the desired resource.
 * Note: For a resource to be deleted using this method,
 * the resource page should allow to delete the resource using the Manage button.
 */
exports.deleteResource = async (page) => {
  await page.getByText('Manage').click();
  await page.getByRole('button', { name: /^(Delete|Remove Worker)/ }).click();
  await page.getByText('OK', { exact: true }).click();
  await expect(
    page.getByRole('alert').getByText('Success', { exact: true }),
  ).toBeVisible();
  await page.getByRole('button', { name: 'Dismiss' }).click();
};
```

ui/admin/tests/e2e/pages/aliases.js
```diff
@@ -0,0 +1,65 @@
...
+    await this.page.getByLabel('Name').fill(aliasName);
+    await this.page.getByLabel('Description').fill('This is an automated test');
+    await this.page.getByLabel('Alias Value').fill(alias);
+    await this.page.getByLabel('Target ID').fill(targetId);
+
+    await this.page.getByRole('button', { name: 'Save' }).click();
+    await expect(
+      this.page.getByRole('alert').getByText('Success', { exact: true }),
+    ).toBeVisible();
+    await this.page.getByRole('button', { name: 'Dismiss' }).click();
```
moduli, line 54: Per the other comment about a `global.js` file, is it worth creating a helper method to dismiss these success alerts? (this is used all over)

ui/admin/tests/e2e/helpers/boundary-ui.js
ZedLi: I'm thinking if there are generic methods that are common among pages, we can have a base page that we extend so you can reuse these methods, thoughts on that? And then, for example, the one for workers could override it and just have the `Remove Worker` instead of delete so they can be specific too when it needs to be.

moduli: Yup. Makes sense. I do think that means I need to break down the pages even further. For example
```
Host Catalogs Page -> Host Catalogs Page + Host Sets Page + Hosts Page
```

moduli: @ZedLi Completed your suggestion. Please have a look when you have a moment.

ZedLi (approved):
Looks great! Much more organized and easier to navigate. Thanks for this work!

PR #2429, opened 2024-08-07
fix: 🐛 targets not displaying even after they are cached

lisbet-alvarez:
# Description

The main issue is that the daemon returns empty array for targets since it times out. But even after the targets are finished getting cached, they are not displayed in the UI. That is because in first attempt to load the `this.allTargets` variable got set to empty array which was causing the `this.makeAllTargetsQuery` to not even run. As a result the `this.showFilters` check in the template was coming back as false.

Before:

https://github.com/user-attachments/assets/77e9eb06-5f21-4e42-8780-f85bac467e52

After:

https://github.com/user-attachments/assets/0bee5a97-74ae-4b68-bd06-a0719cc0e78c

## Checklist

### DIFF
--- a/ui/desktop/app/routes/scopes/scope/projects/sessions/index.js
+++ b/ui/desktop/app/routes/scopes/scope/projects/sessions/index.js
@@ -105,10 +105,10 @@ export default class ScopesScopeProjectsSessionsIndexRoute extends Route {
     const totalItems = sessions.meta?.totalItems;
 
     // Query all sessions and all targets for defining filtering values if entering route for the first time
-    if (!this.allSessions) {
+    if (!this.allSessions?.length) {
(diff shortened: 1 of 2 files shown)


ZedLi (approved):

cameronperera (approved):
Great catch!

ZedLi (commented):
Should we also do the same in the sessions page? I think it also _could_ have the same issue but a lot less likely?

lisbet-alvarez:
> Should we also do the same in the sessions page? I think it also _could_ have the same issue but a lot less likely?

ah i didn't check the sessions page, but thats a great point. Rn i looked and that page also retrieves `allTargets` and `allSessions` so we should modify as well

PR #2430, opened 2024-08-07
refact(e2e): Split up boundary-cli helper methods

moduli:
This PR refactors the Admin UI e2e test suite to split up  the `boundary-cli.js` helper methods file. This was done to better organize the growing file as well as to set it up for being shared with the Desktop e2e tests.

All tests were run locally to confirm that the refactor has not broken anything.
```
yarn run e2e:ce:docker
yarn run e2e:ce:aws
yarn run e2e:ent:docker
yarn run e2e:ent:aws
```

https://hashicorp.atlassian.net/browse/ICU-14743


ui/admin/tests/e2e/helpers/boundary-cli.js
moduli, line 1: Curious to hear thoughts on this pattern.

ZedLi: Could we just have the exports be like `export * from './boundary-cli/accounts';` instead of doing named ones for each file to make it easier?

ui/admin/tests/e2e/helpers/boundary-cli/accounts.js
```diff
@@ -0,0 +1,31 @@
...
+ */
+
+const { execSync } = require('child_process');
+
+/**
+ * Creates a new password account
+ * @param {string} authMethodId ID of the auth-method that the new account will be created for.
+ * @returns {Promise<string>} new account's ID
+ */
+export async function createPasswordAccountCli(authMethodId) {
```
ZedLi, line 13: I want to mention these are actually ESM exports and not commonJS, but we're still importing in commonJS style throughout (i.e. using `require`). I just realized, was there any reason we needed to do that? Can all these e2e tests be swapped over to just use ESM imports/exports?

I'm not sure if the node modules will have any issue (e.g. `import { execSync } from 'node:child_process'`) but if they do, maybe we can even just swap over whole sale to ESM in admin UI now and declare in the `package.json` the type as `module`.

moduli: @ZedLi I was just following an example set out here: https://github.com/hashicorp/boundary-ui/pull/2407/files. Happy to change to whatever is appropriate.

ZedLi: Can we try switching over to ESM and see if we can get any issues? That would be much nicer and I'm not sure if there was any particular reason we needed to use commonJS imports

moduli: Yup. Will try it out.

moduli: @ZedLi Attempted to make some updates per your suggestion. Let me know if it's aligned with what you're thinking. Still need to test and make sure I didn't break anything.

moduli: Ah right. I forgot about this, but last time I tried changing some of these imports, it would break the ember tests. It looks like it's doing it again. Do you have any insight to that?

ui/admin/package.json
```diff
@@ -33,7 +33,7 @@
     "pretest": "yarn clean:coverage",
     "test": "concurrently \"npm:test:*\" --names \"test:\"",
     "test:ember": "COVERAGE=true ember exam --test-port 0 --split=4 --parallel=1 --random",
-    "e2e": "yarn playwright test --config ./tests/e2e/playwright.config.js",
+    "e2e": "yarn playwright test --config ./tests/e2e/playwright.config.ts",
```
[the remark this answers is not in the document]
moduli: When I tried to set that to `.mjs`, it complained that there was a `require.resolve('./global-setup')`. Do you know of the equivalent? 

I'm seeing this article: https://stackoverflow.com/questions/54977743/do-require-resolve-for-es-modules.

ui/admin/tests/e2e/helpers/boundary-cli/accounts.js
ZedLi: Yeah the main issue is we don't declare what type of module system to use so it will default to commonJS. The issue then is that the tests now use ESM which don't know how to properly import/export so using `.mjs` forces it. 

I was hoping we could force it to use ESM by changing the type in package.json but it seems like some packages have some interoperability issues. We would have to force them to `.cjs` through the whole chain which is a bigger headache than it's worth so I think just using `.mjs` in the e2e tests is easier.

ui/admin/package.json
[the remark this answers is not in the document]
ZedLi: I see, do you need to actually `require.resolve` for this? It seems like the location should just work according to the [docs](https://playwright.dev/docs/api/class-testconfig#test-config-global-setup).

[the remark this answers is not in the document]
moduli: Ah. I was using this example: https://playwright.dev/docs/test-global-setup-teardown#option-2-configure-globalsetup-and-globalteardown.

I'll try a few things out and see what works.

PR #2431, opened 2024-08-08
Icu 14403 add dropdown to select and save log verbosity

ZedLi:
✅ Closes: https://hashicorp.atlassian.net/browse/ICU-14403

# Description
Added card to be able to change log levels of app as well as show where the logs are saved. 

image

image

## How to Test
Start desktop client and confirm that changing log levels works. Debug should show search request times. Confirm the log location is correct as well. 

Confirm that log level config is saved correctly, it should be located at: `~/Library/Application Support/Boundary/config.json`

## Checklist

### DIFF
--- a/ui/desktop/app/components/settings-card/logs/index.js
+++ b/ui/desktop/app/components/settings-card/logs/index.js
@@ -1,3 +1,8 @@
+/**
+ * Copyright (c) HashiCorp, Inc.
+ * SPDX-License-Identifier: BUSL-1.1
+ */
+
 import Component from '@glimmer/component';
 import { action } from '@ember/object';
 import { inject as service } from '@ember/service';
--- a/ui/desktop/app/components/settings-card/logs/index.js
+++ b/ui/desktop/app/components/settings-card/logs/index.js
@@ -14,24 +14,7 @@ export default class SettingsCardLogsComponent extends Component {
   // =attributes
 
   get logLevels() {
-    return [
-      {
-        label: 'error',
-        value: 'error',
-      },
-      {
-        label: 'warn',
-        value: 'warn',
-      },
-      {
-        label: 'info',
-        value: 'info',
-      },
-      {
-        label: 'debug',
-        value: 'debug',
-      },
-    ];
+    return ['error', 'warn', 'info', 'debug'];
   }
 
   // =methods
--- a/ui/desktop/app/components/settings-card/logs/index.js
+++ b/ui/desktop/app/components/settings-card/logs/index.js
@@ -1,0 +1,37 @@
+import Component from '@glimmer/component';
+import { action } from '@ember/object';
+import { inject as service } from '@ember/service';
+
+export default class SettingsCardLogsComponent extends Component {
+  // =services
+  @service ipc;
+
+  // =attributes
+
+  get logLevels() {
+    return [
+      {
+        label: 'error',
+        value: 'error',
+      },
+      {
+        label: 'warn',
+        value: 'warn',
+      },
+      {
+        label: 'info',
+        value: 'info',
+      },
+      {
+        label: 'debug',
+        value: 'debug',
+      },
+    ];
+  }
+
+  // =methods
+  @action
+  async selectLogLevel({ target: { value } }) {
+    await this.ipc.invoke('setLogLevel', value);
+  }
+}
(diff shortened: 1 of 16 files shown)


ui/desktop/electron-app/package.json
```diff
@@ -26,6 +26,7 @@
     "electron-is-dev": "^2.0.0",
     "electron-log": "^5.1.2",
     "electron-squirrel-startup": "^1.0.0",
+    "electron-store": "^8.2.0",
```
ZedLi, line 29: I went with `electron-store` as it seemed small with minimal dependencies and a decent amount of usage compared to other packages. It also under the hood just uses the `app.getPath` [userData](https://www.electronjs.org/docs/latest/api/app#appgetpathname) to store the config files.

ui/desktop/app/routes/scopes/scope/projects/settings/index.js
```diff
@@ -13,28 +13,34 @@ export default class ScopesScopeProjectsSettingsIndexRoute extends Route {
...
     try {
       cacheDaemonStatus = await this.ipc.invoke('cacheDaemonStatus');
     } catch (e) {
-      return {
-        formattedCliVersion,
-        formattedCacheVersion,
-        formattedDesktopVersion,
-        error: e,
-      };
+      error = e;
```
cameronperera, line 29: non-blocking: Could we change line 28 from `e` to `error`? I know we typically use `e` but in this case I don't see an issue with defining the exception as `error`.

ui/desktop/tests/integration/components/settings-card/user-test.js
```diff
@@ -0,0 +1,75 @@
...
+        data = data;
+        isAuthenticated = true;
+      },
+    );
+
+    await render(hbs`<SettingsCard::User />`);
+
+    assert.dom('.hds-badge__text').hasText('Authenticated');
+    assert.dom('[data-test-auth-method]').hasText('Password');
+    assert.dom('[data-test-username]').hasText('admin');
```
cameronperera, line 37: non-blocking: Could we move these to const variables in this file? This will help during test selector refactor work in the future.

ui/desktop/electron-app/src/index.js
```diff
@@ -61,7 +62,7 @@ fixPath();
 // Setup logger
 log.initialize();
 log.transports.console.level = false;
-log.transports.file.level = process.env.BOUNDARY_DESKTOP_LOG_LEVEL ?? 'info';
```
cameronperera, line 64: clarifying question: We would no longer be able to set the log level with this env variable now since we can do it via the settings page?

ui/desktop/app/components/settings-card/logs/index.js
```diff
@@ -0,0 +1,42 @@
...
+        value: 'warn',
+      },
+      {
+        label: 'info',
+        value: 'info',
+      },
+      {
+        label: 'debug',
+        value: 'debug',
+      },
```
DhariniJeeva, line 33: since both key/value are the same here, should we just use an array?

ui/desktop/electron-app/src/ipc/handlers.js
```diff
@@ -210,17 +211,42 @@ handle('getDesktopVersion', async () => ({ desktopVersion: releaseVersion }));
...
+
+/**
+ * Returns the path to the log file
+ */
+handle('getLogPath', () => {
+  switch (os.platform()) {
+    case 'win32':
+      return `${
+        process.env.USERPROFILE ?? '%USERPROFILE%'
+      }\\AppData\\Roaming\\Boundary\\logs\\desktop-client.log`;
```
DhariniJeeva, line 232: why are we using double slashes here?

DhariniJeeva (approved):
looks great, left a couple of non blocking comments

ui/desktop/tests/integration/components/settings-card/user-test.js
ZedLi: Sure, I tried not to extract to variables since I expected them only to be used once but I can still move them to consts

ui/desktop/electron-app/src/index.js
ZedLi: Correct, this will remove the need for using the env variable

ui/desktop/electron-app/src/ipc/handlers.js
ZedLi: It's just escaping the backslashes, so the user only sees the one backslash, I have an example of what it looks like on windows in the PR description

ui/desktop/app/components/settings-card/logs/index.js
ZedLi: Yup good call, updated!

lisbet-alvarez (approved):
Tested locally w/ macos and everything working like it should!

PR #2432, opened 2024-08-09
refactor: 💡 Updated Manage Dropdown in auth-methods accounts

CaptFalc:
✅ Closes: https://hashicorp.atlassian.net/browse/ICU-14599

# Description
Updated Manage Dropdown in auth-methods account form to HDS

### **Before**
Screenshot 2024-08-09 at 12 50 20 PM
Screenshot 2024-08-09 at 12 50 31 PM
### **After**
Screenshot 2024-08-09 at 12 52 44 PM
Screenshot 2024-08-09 at 12 52 50 PM

## How to Test
1. Navigate to Auth methods
2. Select auth method resource
3. Select account tab
4. Select account from tab
5. HDS Design should reflect in dropdown

## Checklist

~- [ ] I have added JSON response output for API changes
~- [ ] I have added steps to reproduce and test for bug fixes in the description
~- [ ] I have commented on my code, particularly in hard-to-understand areas
~- [X ] My changes generate no new warnings

### DIFF
--- a/ui/admin/tests/acceptance/accounts/delete-test.js
+++ b/ui/admin/tests/acceptance/accounts/delete-test.js
@@ -81,7 +81,7 @@ module('Acceptance | accounts | delete', function (hooks) {
     instances.account.authorized_actions =
       instances.account.authorized_actions.filter((item) => item !== 'delete');
     await visit(urls.account);
-    assert.notOk(find(MANAGE_DROPDOWN_SELECTOR), find(DELETE_ACTION_SELECTOR));
+    assert.dom(MANAGE_DROPDOWN_SELECTOR).doesNotExist();
   });
 
   test('errors are displayed when delete on account fails', async function (assert) {
--- a/ui/admin/tests/acceptance/accounts/delete-test.js
+++ b/ui/admin/tests/acceptance/accounts/delete-test.js
@@ -20,6 +20,11 @@ module('Acceptance | accounts | delete', function (hooks) {
   setupApplicationTest(hooks);
   setupMirage(hooks);
 
+  const MANAGE_DROPDOWN_SELECTOR =
+    '[data-test-manage-account-auth-methods] div:first-child button';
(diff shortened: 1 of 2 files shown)


ui/admin/tests/acceptance/accounts/delete-test.js
```diff
@@ -67,17 +72,16 @@ module('Acceptance | accounts | delete', function (hooks) {
...
   });
 
   test('cannot delete an account without proper authorization', async function (assert) {
     instances.account.authorized_actions =
       instances.account.authorized_actions.filter((item) => item !== 'delete');
     await visit(urls.account);
-    assert.notOk(
-      find('.rose-layout-page-actions .rose-dropdown-button-danger'),
-    );
+    assert.notOk(find(MANAGE_DROPDOWN_SELECTOR), find(DELETE_ACTION_SELECTOR));
```
ZedLi, line 84: `notOk` actually takes in only [one argument](https://qunitjs.com/api/assert/notOk/). 

If you want to check for both, I would recommend using two assertions for this and using the [dom assertions](https://github.com/mainmatter/qunit-dom/blob/master/API.md#assertdom). The dom assertions provide a more natural way of asserting the element doesn't exist versus asserting the truthiness of a found element. 

As an example:

`assert.dom(MANAGE_DROPDOWN_SELECTOR).doesNotExist();`

CaptFalc: Sounds good, thanks for the heads up Zed!

So in the case I implemented it; the line never ran?

cameronperera (approved):

ui/admin/tests/acceptance/accounts/delete-test.js
CaptFalc: Exactly that, think that the way you wrote it works best for future scenarios handling those asserts.

PR #2435, opened 2024-08-12
ICU-13883 add remove org modal

lisbet-alvarez:
✅ Closes: https://hashicorp.atlassian.net/browse/ICU-13883

# Description

Add a modal for when a user deselects a singular org with projects. 
Add a separate modal for when a user deselects all and orgs have projects attached to them.

https://github.com/user-attachments/assets/8ca51091-58c6-4320-af79-fc88239622d6

## How to Test

1. Make sure you have an org created that has various projects
2. Go to roles page -> click on a role
3. Click "manage" dropdown -> "manage scopes" -> "manage custom scopes"
4. For the Org with many projects click the link under "Projects selected"
5. Select and save a couple projects.
6. After saving, you should now be on the "manage custom scopes" page again.
7. If you deselect the org with projects then the modal should appear.
8. click "Remove org and projects" and now project count should be zero
9. Click "save" and scopes tab table should not have those projects

## Checklist

### DIFF
--- a/addons/core/translations/resources/en-us.yaml
+++ b/addons/core/translations/resources/en-us.yaml
@@ -617,6 +617,8 @@ role:
       apply: Apply changes to org
       remove-org-and-projects: Remove org and projects
       remove-org-only: Remove just the org
+      remove-orgs-and-projects: Remove orgs and projects
+      remove-orgs-only: Remove just the orgs
     form:
       this:
         label: Add this scope
@@ -630,8 +632,11 @@ role:
         label: Add all descendants
         help: Add all of the orgs and the projects in the orgs that are underneath this scope.
     remove-org:
-      title: Remove projects under this org?
-      description: Do you want to remove the projects under this org from this role? Or do you just want to remove the org from this role?
+      title: Remove this org?
+      description: You're about to remove {orgDisplayName}. Do you want to remove the projects under {orgDisplayName} as well? Or do you just want to remove {orgDisplayName} from this project?
+    remove-all-orgs:
+      title: Remove all selected orgs?
+      description: You're about to remove all the selected orgs. Do you want to remove the projects under these orgs as well? Or do you just want to remove just the orgs?
 target:
   title: Target
   title_plural: Targets
--- a/addons/core/translations/resources/en-us.yaml
+++ b/addons/core/translations/resources/en-us.yaml
@@ -615,6 +615,8 @@ role:
         help: Customize which scopes are associated with this role.
       view-projects: '{selected}/{total} total projects'
       apply: Apply changes to org
+      remove-org-and-projects: Remove org and projects
+      remove-org-only: Remove just the org
     form:
       this:
         label: Add this scope
@@ -627,6 +629,9 @@ role:
       descendants:
         label: Add all descendants
         help: Add all of the orgs and the projects in the orgs that are underneath this scope.
+    remove-org:
+      title: Remove projects under this org?
+      description: Do you want to remove the projects under this org from this role? Or do you just want to remove the org from this role?
 target:
   title: Target
   title_plural: Targets
--- a/ui/admin/app/components/form/role/manage-custom-scopes/index.js
+++ b/ui/admin/app/components/form/role/manage-custom-scopes/index.js
@@ -99,7 +99,7 @@ export default class FormRoleManageCustomScopesIndexComponent extends Component 
    */
   @action
   toggleRemoveOrgModal() {
-    this.selectedOrg = [];
+    this.selectedOrg = '';
   }
 
   /**
--- a/ui/admin/app/components/form/role/manage-custom-scopes/index.js
+++ b/ui/admin/app/components/form/role/manage-custom-scopes/index.js
@@ -10,52 +10,117 @@ import { tracked } from '@glimmer/tracking';
 export default class FormRoleManageCustomScopesIndexComponent extends Component {
   // =attributes
 
-  @tracked showRemoveOrgModal = false;
-  @tracked selectedOrg;
+  @tracked selectedOrgs = [];
+  @tracked selectedOrg = '';
+
+  /**
+   * Returns the display name of the selectedOrg.
+   * @type {string}
+   */
+  get orgDisplayName() {
+    const org = this.args.model.orgScopes.find(
+      ({ id }) => id === this.selectedOrg,
+    );
+    return org.displayName;
+  }
 
   // =actions
 
   /**
    * Handles the selection changes for the paginated table.
    * @param {object} selectableRowsStates
+   * @param {object} selectionKey
    */
   @action
   selectionChange({ selectableRowsStates, selectionKey }) {
     const { role, projectTotals } = this.args.model;
-    selectableRowsStates.forEach((row) => {
-      const { isSelected, selectionKey: key } = row;
-      const includesId = role.grant_scope_ids.includes(key);
-      if (isSelected && !includesId) {
-        // Add the org id if it was selected and it doesn't exist as a grant scope.
-        role.grant_scope_ids = [...role.grant_scope_ids, key];
-      } else if (!isSelected && includesId) {
-        // Remove org id if it was deselected and it does exist as a grant scope.
-        role.grant_scope_ids = role.grant_scope_ids.filter(
-          (item) => item !== key,
-        );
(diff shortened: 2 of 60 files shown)


addons/core/translations/resources/en-us.yaml
```diff
@@ -627,6 +629,9 @@ role:
       descendants:
         label: Add all descendants
         help: Add all of the orgs and the projects in the orgs that are underneath this scope.
+    remove-org:
+      title: Remove projects under this org?
+      description: Do you want to remove the projects under this org from this role? Or do you just want to remove the org from this role?
```
cameronperera, line 634: I could be reading it wrong but when I read the first question, it feels to me like we can remove just the projects but keep the org selected, not remove the projects AND the org.

ui/admin/app/routes/scopes/scope/roles/role/manage-scopes/manage-custom-scopes.js
```diff
@@ -7,7 +7,7 @@ import Route from '@ember/routing/route';
 import { action } from '@ember/object';
 import { inject as service } from '@ember/service';
 import { TYPE_SCOPE_PROJECT } from 'api/models/scope';
-
+import { TrackedObject } from 'tracked-built-ins';
```
cameronperera, line 10: small nitpick: Could we add a space between the imports and class declaration?

ui/admin/tests/acceptance/roles/global-scope-test.js
```diff
@@ -509,6 +514,72 @@ module('Acceptance | roles | global-scope', function (hooks) {
...
+
+    await click(SAVE_BTN_SELECTOR);
+
+    assert.strictEqual(currentURL(), urls.manageScopes);
+    assert.dom(BUTTON_ICON_SELECTOR).isVisible();
+
+    await click(SAVE_BTN_SELECTOR);
+
+    assert.strictEqual(currentURL(), urls.roleScopes);
+    assert.strictEqual(findAll(TABLE_ROW_SELECTOR).length, 0);
```
cameronperera, line 580: I think we can do the same thing here with `assert.dom(TABLE_ROW_SELECTOR).exist({ count: 0 })`. I know we have used it in past work, however I can see that it isn't as clear. It makes me wonder if `doesNotExist()` would also work and be more clear.
```suggestion
    assert.dom(TABLE_ROW_SELECTOR).exist({ count: 0 });
```

ui/admin/app/routes/scopes/scope/roles/role/manage-scopes/manage-custom-scopes.js
lisbet-alvarez: of course!

ui/admin/app/components/form/role/manage-custom-scopes/index.js
```diff
@@ -5,27 +5,57 @@
 
 import Component from '@glimmer/component';
 import { action } from '@ember/object';
+import { tracked } from '@glimmer/tracking';
 
 export default class FormRoleManageCustomScopesIndexComponent extends Component {
+  // =attributes
+
+  @tracked showRemoveOrgModal = false;
+  @tracked selectedOrg;
```
ZedLi, line 14: Is the boolean needed if you are also keeping track of the `selectedOrg`?

ui/admin/app/components/form/role/manage-custom-scopes/index.js
```diff
@@ -5,27 +5,57 @@
...
+        // Add the org id if it was selected and it doesn't exist as a grant scope.
         role.grant_scope_ids = [...role.grant_scope_ids, key];
       } else if (!isSelected && includesId) {
+        // Remove org id if it was deselected and it does exist as a grant scope.
         role.grant_scope_ids = role.grant_scope_ids.filter(
           (item) => item !== key,
         );
+        // If the org id was deselected and has selected projects then trigger the modal.
+        const selected = projectTotals[key]?.selected;
+        if (selectionKey === key && selected?.length) {
```
ZedLi, line 38: I have two questions:
1. I didn't look too closely before but seeing this key logic and looking at the docs I'm now curious, why did we have to iterate over all the `selectableRowsStates` when we get the `selectionKey` for each row? Why was it necessary to iterate over all the rows each time a row is being checked or unchecked? Is it to handle the `all` rows check option?
2. If a user clicks the all checkbox, do we need to handle removing all projects if a user unselects all orgs with it?

lisbet-alvarez: 1. Now that you point it out, there is a more efficient way of doing this by separating the cases that use `all`. I'll experiment with it more.
2. I think that's another good point and i'll check in with design.

cameronperera (approved):

ui/admin/app/components/form/role/manage-custom-scopes/index.js
lisbet-alvarez: good point, the bool has been removed.

ui/admin/app/components/form/role/manage-custom-scopes/index.js
lisbet-alvarez: Update: a new modal has been added for the case in which the user clicks the all checkbox. Thank u for the feedback.

ui/admin/app/components/form/role/manage-custom-scopes/index.js
```diff
@@ -10,52 +10,117 @@ import { tracked } from '@glimmer/tracking';
...
-    selectableRowsStates.forEach((row) => {
-      const { isSelected, selectionKey: key } = row;
-      const includesId = role.grant_scope_ids.includes(key);
-      if (isSelected && !includesId) {
-        // Add the org id if it was selected and it doesn't exist as a grant scope.
-        role.grant_scope_ids = [...role.grant_scope_ids, key];
-      } else if (!isSelected && includesId) {
-        // Remove org id if it was deselected and it does exist as a grant scope.
+
+    const addOrRemoveValues = (add, remove, orgId) => {
```
ZedLi, line 38: Was it intentional that selecting orgs automatically selects all projects underneath it now? Before this change it didn't auto select any projects and you had to actually go in now.

It seems like now there's no easy way to just select only orgs?

lisbet-alvarez: yeah it was intentional, originally i had asked about that change being a modal in this [slack thread](https://hashicorp.slack.com/archives/C06LR1KKN4E/p1723568423341309).

ui/admin/app/components/form/role/manage-custom-scopes/index.js
lisbet-alvarez: @ZedLi tagging for visibility to the comment above

ui/admin/app/components/form/role/manage-custom-scopes/index.js
ZedLi: So there's no way to just select all the orgs only?

lisbet-alvarez: Nope, not on the manage custom scopes page. The only way would be to select the org save it (it auto selects projects) then go into the manage projects page and deselect all projects. Would you prefer if i move this specific change to a different PR?

lisbet-alvarez: changes around this have been removed! thank u for the feedback!

ui/admin/app/components/form/role/manage-custom-scopes/index.js
```diff
@@ -5,27 +5,108 @@
...
+      this.toggleRemoveOrgModal();
+    }
+  }
+
+  /**
+   * Toggles the modal to remove an org and projects off.
+   */
+  @action
+  toggleRemoveOrgModal() {
+    this.selectedOrg = [];
```
ZedLi, line 102: Should this be an empty string instead of array?

ZedLi (approved):
Works great!

ui/admin/app/components/form/role/manage-custom-scopes/index.js
lisbet-alvarez: oh yup good catch, i was experimenting with something and forgot to change it back.

PR #2438, opened 2024-08-15
Icu 14781 add a section for server information

DhariniJeeva:
# Description
Adds server information https://hashicorp.atlassian.net/browse/ICU-14781

HCP:
 
Screenshot 2024-08-15 at 3 39 29 PM

 
Screenshot 2024-08-15 at 3 39 53 PM

## Checklist

### DIFF
--- a/ui/desktop/app/components/settings-card/index.hbs
+++ b/ui/desktop/app/components/settings-card/index.hbs
@@ -23,12 +23,6 @@
 
   </div>
 
-  {{#if @description}}
-    <Hds::Text::Body @tag='p' class='description' @color='faint'>
-      {{@description}}
-    </Hds::Text::Body>
-  {{/if}}
-
   {{! this yields the body from consumer }}
   {{#if (has-block 'body')}}
     <div class='content'>
--- a/ui/desktop/app/components/settings-card/index.hbs
+++ b/ui/desktop/app/components/settings-card/index.hbs
@@ -23,6 +23,12 @@
 
   </div>
 
+  {{#if @description}}
+    <Hds::Text::Body @tag='p' @color='faint'>
+      {{@description}}
+    </Hds::Text::Body>
+  {{/if}}
+
   {{! this yields the body from consumer }}
   {{#if (has-block 'body')}}
     <div class='content'>
--- a/ui/desktop/app/components/settings-card/index.hbs
+++ b/ui/desktop/app/components/settings-card/index.hbs
@@ -17,7 +17,6 @@
 
     {{#if (has-block 'action')}}
       <div class='action'>
-        {{! this will contain buttons }}
         {{yield to='action'}}
       </div>
     {{/if}}
@@ -25,7 +24,7 @@
   </div>
 
   {{#if @description}}
-    <Hds::Text::Body @tag='p'>
+    <Hds::Text::Body @tag='p' class='description' @color='faint'>
       {{@description}}
     </Hds::Text::Body>
   {{/if}}
--- a/ui/desktop/app/components/settings-card/server/index.hbs
+++ b/ui/desktop/app/components/settings-card/server/index.hbs
@@ -3,9 +3,19 @@
   SPDX-License-Identifier: BUSL-1.1
 }}
 
-<SettingsCard @header={{t 'settings.server'}} @icon='cloud'>
+<SettingsCard
+  @header={{t 'settings.server.title'}}
+  @icon='cloud'
+  @description={{t 'settings.server.description'}}
+>
   <:state>
-    {{! TODO: connected }}
+    {{#if @model.serverInformation}}
+      <Hds::Badge
+        @text={{t 'settings.connected'}}
+        @icon='check'
+        @color='success'
+      />
+    {{/if}}
   </:state>
   <:body>
     <div>
--- a/ui/desktop/app/components/settings-card/server/index.hbs
+++ b/ui/desktop/app/components/settings-card/server/index.hbs
@@ -1,0 +1,31 @@
+{{!
+  Copyright (c) HashiCorp, Inc.
+  SPDX-License-Identifier: BUSL-1.1
+}}
+
+<SettingsCard @header={{t 'settings.server'}} @icon='cloud'>
+  <:state>
+    {{! TODO: connected }}
+  </:state>
+  <:body>
+    <div>
+      <Hds::Text::Display>
+        {{t 'settings.provider'}}
+      </Hds::Text::Display>
+      <Hds::Text::Body @tag='p' @weight='medium' data-test-server-name>
+        {{this.providerName}}
+      </Hds::Text::Body>
+    </div>
+    <div class='large-column'>
+      <Hds::Text::Display>
+        {{t 'settings.url'}}
+      </Hds::Text::Display>
+      <Hds::Copy::Snippet
+        data-test-server-url
+        @textToCopy={{@model.serverInformation}}
+        @isFullWidth={{true}}
+      />
+    </div>
+  </:body>
+
+</SettingsCard>
--- a/ui/desktop/app/components/settings-card/server/index.hbs
+++ b/ui/desktop/app/components/settings-card/server/index.hbs
@@ -3,7 +3,11 @@
   SPDX-License-Identifier: BUSL-1.1
 }}
 
-<SettingsCard @header={{t 'settings.server.title'}} @icon='cloud'>
+<SettingsCard
+  @header={{t 'settings.server.title'}}
+  @icon='cloud'
+  @description={{t 'settings.server.description'}}
+>
   <:state>
     {{#if @model.serverInformation}}
       <Hds::Badge
@@ -14,9 +18,6 @@
     {{/if}}
   </:state>
   <:body>
-    <Hds::Text::Body @tag='p' @color='faint'>
-      {{t 'settings.server.description'}}
-    </Hds::Text::Body>
     <div>
       <Hds::Text::Display>
         {{t 'settings.provider'}}
--- a/ui/desktop/app/components/settings-card/server/index.hbs
+++ b/ui/desktop/app/components/settings-card/server/index.hbs
@@ -3,11 +3,7 @@
   SPDX-License-Identifier: BUSL-1.1
 }}
 
-<SettingsCard
-  @header={{t 'settings.server.title'}}
-  @icon='cloud'
-  @description={{t 'settings.server.description'}}
->
+<SettingsCard @header={{t 'settings.server.title'}} @icon='cloud'>
   <:state>
     {{#if @model.serverInformation}}
       <Hds::Badge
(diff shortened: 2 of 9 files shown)


ui/desktop/app/components/settings-card/server/index.hbs
```diff
@@ -0,0 +1,41 @@
...
+  @description={{t 'settings.server.description'}}
+>
+  <:state>
+    {{#if @model.serverInformation}}
+      <Hds::Badge
+        @text={{t 'settings.connected'}}
+        @icon='check'
+        @color='success'
+      />
+    {{/if}}
```
lisbet-alvarez, line 18: Im not sure what the intended functionality is, but when i shut down my self-managed instance the "Connected" badge does not disappear. 

https://github.com/user-attachments/assets/a2871fb2-4fcb-4960-84ea-30c73a88002f

ui/desktop/app/components/settings-card/server/index.hbs
```diff
@@ -0,0 +1,41 @@
...
+      <Hds::Text::Display>
+        {{t 'settings.provider'}}
+      </Hds::Text::Display>
+      <Hds::Text::Body @tag='p' @weight='medium' data-test-server-name>
+        {{this.providerName}}
+      </Hds::Text::Body>
+    </div>
+    <div class='large-column'>
+      <Hds::Text::Display>
+        {{t 'settings.url'}}
```
lisbet-alvarez, line 31: non-blocking: to be consistent with the "Welcome to Boundary" card should we use "Cluster URL" here as well?

DhariniJeeva: good point, checking with design

ui/desktop/app/components/settings-card/server/index.hbs
DhariniJeeva: hmmm good catch, so right now I'm just using the cluster URL to display the `connected`, I'll check what we do in other pages to throw the error and display disconnected. 

Also, now I'm wondering why there is no error in the application block

ui/desktop/app/components/settings-card/server/index.js
```diff
@@ -0,0 +1,26 @@
...
+  @service intl;
+
+  // =attributes
+  /**
+   * Returns the provider name for HCP and generic name for non HCP servers
+   * @type {string}
+   */
+  get providerName() {
+    const regex = /hashicorp\.cloud/;
+    if (this.args.model.serverInformation?.match(regex)) {
```
calcaide, line 20: Do we anticipate having to know if is HCP or not somewhere else?

Asking just in case we want to apply the pattern we have in some models I.e: add a function isHcp where it receives a value, perhaps serverUrl, and returns true or false if belongs to HCP.

This way we can re-use the logic that determinate if is HCP.

calcaide (changes requested):
Left some comments.

The settings page is looking sharp!! Amazing job!!

calcaide:
Just heads up, if you open the DC in the browser and navigate to the settings page, there is an error related to ipcMock that brakes the settings section.

ui/desktop/app/components/settings-card/server/index.hbs
```diff
@@ -0,0 +1,41 @@
...
+  <:body>
+    <div>
+      <Hds::Text::Display>
+        {{t 'settings.provider'}}
+      </Hds::Text::Display>
+      <Hds::Text::Body @tag='p' @weight='medium' data-test-server-name>
+        {{this.providerName}}
+      </Hds::Text::Body>
+    </div>
+    <div class='large-column'>
```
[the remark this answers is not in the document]
DhariniJeeva: this was added to keep the URLs in full width and to occupy both second and third column, we have a similar layout for logs level section

ui/desktop/app/components/settings-card/server/index.js
DhariniJeeva: hmm, I don't know if this belongs in the model since this seems like a parsing that the UI needs for this specific use case.

ui/desktop/app/components/settings-card/server/index.hbs
DhariniJeeva: done, thank you!

ui/desktop/app/components/settings-card/index.hbs
```diff
@@ -17,15 +17,14 @@
...
-        {{! this will contain buttons }}
         {{yield to='action'}}
       </div>
     {{/if}}
 
   </div>
 
   {{#if @description}}
-    <Hds::Text::Body @tag='p'>
+    <Hds::Text::Body @tag='p' class='description' @color='faint'>
```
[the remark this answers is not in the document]
DhariniJeeva: good catch, moved it body, thanks for the call out

ui/desktop/app/components/settings-card/index.hbs
```diff
@@ -17,19 +17,12 @@
...
     {{#if (has-block 'action')}}
       <div class='action'>
-        {{! this will contain buttons }}
         {{yield to='action'}}
       </div>
     {{/if}}
 
   </div>
 
-  {{#if @description}}
```
ZedLi, line 27: Just a heads up I actually did use this block, I think it makes sense to have it. What I had done to avoid another class was I modified the CSS to just bake it in as part of the headers class. 

I can handle merging these though since your PR can make it in first

ui/desktop/app/components/settings-card/server/index.js
```diff
@@ -0,0 +1,26 @@
...
+  // =services
+  @service intl;
+
+  // =attributes
+  /**
+   * Returns the provider name for HCP and generic name for non HCP servers
+   * @type {string}
+   */
+  get providerName() {
+    const regex = /hashicorp\.cloud/;
```
ZedLi, line 19: Should this also take into account int and dev environments as well?

ui/desktop/app/components/settings-card/index.hbs
DhariniJeeva: ahh ok thanks for the headsup-  I will undo this description change and not do the styles and after you merge your PR, it will have the correct margin, does that make it easy?

ZedLi: I can just merge it normally, you don't have to change it back or you could just add the CSS i used:

```css
  .header-and-icon {
    display: flex;
    align-items: center;
    gap: 1rem;

    ~ * {
      padding-left: 2.8rem;
    }
  }
```

And I changed the `.content` margin to be `margin: 1rem 0;`, that way all the padding in the body is handled by the `~ *`

DhariniJeeva: thanks, added the styles

ui/desktop/app/components/settings-card/server/index.js
DhariniJeeva: good call, included regex to check for `boundary and hcp/hashicorp` in the URL string
```
.boundary.hcp.dev/
.boundary.hcp.to/
.boundary.hashicorp.cloud/
```

DhariniJeeva:
> Just heads up, if you open the DC in the browser and navigate to the settings page, there is an error related to ipcMock that brakes the settings section.

thanks for the catch Carlos, added mock methods and don't see the console errors

ui/desktop/app/components/settings-card/server/index.hbs
ZedLi: What would error out in the application block? Nothing there would be hitting the controller I believe. I'm not sure if this is one that matters too much if I were to be honest since it likely would occur only locally since you have to bring down the controller

DhariniJeeva: yeah, that is right unless we refresh the view, I don't see a way to check if the controller is down.   

should we remove the `connected` word to avoid any confusion?

ui/desktop/app/routes/scopes/scope/projects/settings/index.js
```diff
@@ -48,6 +48,7 @@ export default class ScopesScopeProjectsSettingsIndexRoute extends Route {
 
     const logLevel = await this.ipc.invoke('getLogLevel');
     const logPath = await this.ipc.invoke('getLogPath');
+    const serverInformation = await this.ipc.invoke('getClusterUrl');
```
ZedLi, line 51: What's the difference between this and using the [renderer cluster url](https://github.com/hashicorp/boundary-ui/blob/9b1f052a569fc1579befffda87de3a91ba5e417b/ui/desktop/app/services/cluster-url.js#L30) from our ember service? I don't recall how the URL works

calcaide:
> > Just heads up, if you open the DC in the browser and navigate to the settings page, there is an error related to ipcMock that brakes the settings section.
> 
> thanks for the catch Carlos, added mock methods and don't see the console errors

@DhariniJeeva 

Amazing fix!! tested manually and now everything works like a charm 😄

ui/desktop/app/components/settings-card/server/index.js
calcaide: yep. I was wondering in case we have more cases within the DC where we need to determinate if the controller is hcp or not hcp we could do something more generic.
But seems at the moment, this is the only case we have, so all good KISS 😄

ui/desktop/app/styles/app.scss
```diff
@@ -674,6 +674,10 @@ $chevron-down: url("data:image/svg+xml;utf-8,<svg viewBox='0 0 24 24' fill='%231
     display: flex;
     align-items: center;
     gap: 1rem;
+
+    ~ * {
```
calcaide, line 678: None blocking comment: did we check this does not affect other elements? asking because it is a bold selector lol, just checking 😉

calcaide (approved):
Awesome!! this is looking great 😄

ui/desktop/app/components/settings-card/server/index.hbs
DhariniJeeva: good point, I will leave this as is for now. But will check with design/PM and include this for our polishing work if the decision is to add a button

ui/desktop/app/styles/app.scss
ZedLi: Yup, it should only be the ones after the header. It's not a common selector but it basically does the same thing as if we did `.hds-container > :not(.header-and-icon)` but just a bit more succinct assuming the header/icon is the first element

lisbet-alvarez (approved):
Works great!! 🙌

ui/desktop/app/routes/scopes/scope/projects/settings/index.js
DhariniJeeva: good point, I was not aware we have a service for cluster-url, it seems like we process the cluster url within the service here in updateClusterUrl in the application [route](https://github.com/hashicorp/boundary-ui/blob/a9d110a1b6bb9e266b21594c5f780ed1f0dbc079/ui/desktop/app/routes/application.js#L38), I'll update the route to use the url from the service

ui/desktop/app/components/settings-card/server/index.hbs
DhariniJeeva: checked with design, we will be removing the label `connected` to avoid any confusion.  added a separate ticket for this

ZedLi (approved):
Looks great!

PR #2440, opened 2024-08-16
Automate copyright headers addition

calcaide:
# Description

This PR automates the addition of copyright headers as discussed at Tech time.

## How to Test

I run the test procedure described below within [this dummy PR](https://github.com/hashicorp/boundary-ui/pull/2441), you can see [the automated commit added here](https://github.com/hashicorp/boundary-ui/pull/2441/commits/fe0ee7c1287e5b5f60b9acd3f87d141a7269f806).

- Create a new branch checking out from this PR.
- Add a JS file within the codebase. i.e: `boundary-ui/ui/admin/app/random.js`.
- Push the branch and start a PR.
- Review your PR and make sure: There is an automated commit that add headers to the `random.js` file.


.github/workflows/copyright-validate.yaml
```diff
@@ -1,10 +1,15 @@
-name: Check Copyright Headers
+name: Add Copyright Headers
 on: pull_request
 jobs:
   copywrite:
     runs-on: ubuntu-latest
+    permissions:
+      contents: write
```
ZedLi, line 7: Is this needed to be able to commit?

ZedLi (approved):
Thanks for simplifying our workflow!

cameronperera (approved):
Thanks for this work!

DhariniJeeva (approved):
looks great!! thanks for this

.github/workflows/copyright-validate.yaml
calcaide: [More infor about GH action permissions](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/controlling-permissions-for-github_token#overview).

This allows the action, use the `GITHUB_TOKEN` with write permissions within the content of the repo, so it can change files and commit them.

But I based the decission of adding it as per documentation, I did not perform a run to test if we can write the files, generate a commit and pushed without this rule.

PR #2450, opened 2024-08-21
Add CLI_SETUP and delete BYPASS_CLI_BUILD

calcaide:
# Description

Add CLI_SETUP and delete BYPASS_CLI_BUILD.

Note: This PR needs to be sync [with a counterpart within boundary-ui-releases](https://github.com/hashicorp/boundary-ui-releases/pull/107), since the BYPASS_CLI_BUILD is use there.

Note 2: This is merging to a LLB which will contain all the changes to achieve Desktop Client with no CLI

:tickets: [Jira ticket](https://hashicorp.atlassian.net/browse/ICU-14904)

## How to Test
- Navigate to `electron-app/cli` and clean up the directory.
- Run `SETUP_CLI=true`
- Desktop client should download the CLI and start.
- To validate CLI has been downloaded, navigate to `electron-app/cli` and confirm the CLI is present.
- Run `yarn start:desktop`, in the terminal you should see a warning: `WARNING: Bypassing cli setup`.
- Desktop client should start no problem.

### DIFF
--- a/ui/desktop/README.md
+++ b/ui/desktop/README.md
@@ -83,12 +83,12 @@ To run as a desktop app:
 
 - `yarn start:desktop`
 
-The Boundary CLI is NOT downloaded by default, to download and extract the CLI to `electron-app/cli` folder as part of the build, you need to set environment variable `CLI_SETUP` to true. Example: `CLI_SETUP=true yarn start:desktop`.
-CLI version is defined in `electron-app/config.cli.js`.
+The Boundary CLI is NOT downloaded by default, to download and extract the CLI to `electron-app/cli` folder as part of the build, you need to set the environment variable `CLI_SETUP` to true. Example: `CLI_SETUP=true yarn start:desktop`.
+The CLI version is defined in `electron-app/config/cli/VERSION`.
 
 ### Developing Using Non-Release Versions of Boundary
 
-To develop using a non-release version of Boundary, download the Boundary CLI version you want to use and extract it at `electron-app/cli` folder. You may need to create the directory or to cleaning it up beforehand.
+To develop using a non-release version of Boundary, download the Boundary CLI version you want to use and extract it to the `electron-app/cli` folder. You may need to create the directory or clean it up beforehand.
 
 ### Environment Variables (DEV)
 
--- a/ui/desktop/README.md
+++ b/ui/desktop/README.md
@@ -83,7 +83,7 @@ To run as a desktop app:
 
 - `yarn start:desktop`
 
-The Boundary CLI is NOT downloaded by default, to download and extract the CLI to `electron-app/cli` folder as part of the build, you need to set the environment variable `CLI_SETUP` to true. Example: `CLI_SETUP=true yarn start:desktop`.
+The Boundary CLI is NOT downloaded by default, to download and extract the CLI to `electron-app/cli` folder as part of the build, you need to set the environment variable `SETUP_CLI` to true. Example: `SETUP_CLI=true yarn start:desktop`.
 The CLI version is defined in `electron-app/config/cli/VERSION`.
 
 ### Developing Using Non-Release Versions of Boundary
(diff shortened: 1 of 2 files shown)


ui/desktop/README.md
```diff
@@ -106,7 +100,7 @@ These environment variables may be used to customized the build.
 | `APP_UPDATER_CURRENT_VERSION` | Version of client. |
 | `APP_UPDATER_LATEST_VERSION_TAG` | Next version for comparison with current version. |
 | `APP_UPDATER_LATEST_VERSION_LOCATION` | Location of app release to use for updating client. Can be a filepath or url. |
-| `BYPASS_CLI_SETUP` | Disable download and extraction of cli. For development use only. |
+| `CLI_SETUP` | Enable download and extraction of CLI. For development use only. |
```
ZedLi, line 103: Do you think this should be called `SETUP_CLI` instead since it's nicer to read with the verb in the beginning?

Is it still considered `"For development use only."` or should we remove that now?

ui/desktop/README.md
```diff
@@ -83,18 +83,12 @@ To run as a desktop app:
 
 - `yarn start:desktop`
 
-The Boundary CLI is downloaded and extracted to `electron-app/cli/` folder as part of
-build. CLI version is defined in `electron-app/config/cli.js`.
+The Boundary CLI is NOT downloaded by default, to download and extract the CLI to `electron-app/cli` folder as part of the build, you need to set environment variable `CLI_SETUP` to true. Example: `CLI_SETUP=true yarn start:desktop`.
```
ZedLi, line 86:
```suggestion
The Boundary CLI is NOT downloaded by default, to download and extract the CLI to `electron-app/cli` folder as part of the build, you need to set the environment variable `CLI_SETUP` to true. Example: `CLI_SETUP=true yarn start:desktop`.
```

ui/desktop/README.md
```diff
@@ -83,18 +83,12 @@ To run as a desktop app:
...
+CLI version is defined in `electron-app/config.cli.js`.
```
ZedLi, line 87:
```suggestion
The CLI version is defined in `electron-app/config/cli/VERSION`.
```
Don't we define it in the `VERSION` file?

ui/desktop/README.md
```diff
@@ -83,18 +83,12 @@ To run as a desktop app:
...
 ### Developing Using Non-Release Versions of Boundary
 
-You can also develop using a non-release version of Boundary - bypassing the
-above CLI download behavior - by manually placing the version of Boundary that
-you want to develop with in the `electron-app/cli/` directly (you may need to
-create the directory).
-
-After doing this, run yarn with `BYPASS_CLI_SETUP=true`; example:
-`BYPASS_CLI_SETUP=true yarn start:desktop`.
+To develop using a non-release version of Boundary, download the Boundary CLI version you want to use and extract it at `electron-app/cli` folder. You may need to create the directory or to cleaning it up beforehand.
```
ZedLi, line 91:
```suggestion
To develop using a non-release version of Boundary, download the Boundary CLI version you want to use and extract it to the `electron-app/cli` folder. You may need to create the directory or clean it up beforehand.
```

ui/desktop/electron-app/config/cli.js
```diff
@@ -95,20 +95,20 @@ const extract = async (artifactPath, destination) => {
 
 module.exports = {
   setup: async () => {
-    if (process.env.BYPASS_CLI_SETUP) {
+    if (process.env.CLI_SETUP) {
```
ZedLi, line 98: Shouldn't this check for `if (process.env.CLI_SETUP === 'true')` since it would be a string? Otherwise any flag would pass this check

ui/desktop/README.md
calcaide: You are correct, I made the confusion here on wording. When I mean the version is defined within config.cli.js I mean the script will generate what version to download via the url, but you are totally correct, the CLI version is defined within the VERSION file. Accepting your changes 😉 

Thanks for the catch!

ui/desktop/README.md
calcaide: Absolutely agree. `SETUP_CLI` sounds and read way better. I will provide the change.

Yes!!! God catch, I will delete the "for dev only"

ui/desktop/electron-app/config/cli.js
calcaide: You ar right, this wask a mistake on my end, thanks for the catch!

cameronperera (approved):
Thanks for updating this along with the README.

ZedLi (approved):
Looks good!

DhariniJeeva (approved):
looks great!

lisbet-alvarez (approved):
Thank you for these updates!

PR #2452, opened 2024-08-21
refact(e2e): Use fixtures for environment variables

moduli:
This PR does some refactoring in the Admin UI test suite to use fixtures for environment variables. 

https://hashicorp.atlassian.net/browse/ICU-14759

### DIFF
--- a/ui/admin/tests/e2e/playwright.config.mjs
+++ b/ui/admin/tests/e2e/playwright.config.mjs
@@ -3,7 +3,7 @@
  * SPDX-License-Identifier: BUSL-1.1
  */
 
-import { defineConfig, devices } from '@playwright/test';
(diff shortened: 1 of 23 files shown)


moduli:
@ZedLi Is this what you were thinking? If so, I can apply to this to the rest of the tests.

ui/admin/tests/e2e/playwright.config.mjs
ZedLi, line 1: @moduli We could totally do this, playwright also lets you use the [config](https://playwright.dev/docs/release-notes#breaking-change-custom-config-options) to provide defaults as well but it doesn't sound like we'd need them for these

ZedLi (approved):
Looks good!

PR #2458, opened 2024-08-22
chore: 🤖 update ember-cli-mirage

cameronperera:
# Description
This PR updates `ember-cli-mirage` to v3.0.3. This is a minor change that will help with updating packages for this [dependabot alert](https://github.com/hashicorp/boundary-ui/security/dependabot/140). 

I saw this warning while running test. I removed the entry for fireworm from yarn.lock [here](https://github.com/hashicorp/boundary-ui/pull/2458/files#diff-51e4f558fae534656963876761c95b83b6ef5da5103c4adef6768219ed76c2deR13025-R13027) and ran `yarn install` again. This cleared up the warning.
Screenshot 2024-08-22 at 4 55 40 PM

## How to Test

Please navigate around the Admin UI and desktop client Vercel builds. I have done local testing and have not see any issues and this [build](https://github.com/hashicorp/boundary-ui-releases/actions/runs/10529040740) was successful.

## Checklist


addons/api/package.json
```diff
@@ -49,6 +50,7 @@
     "ember-cli": "^4.12.1",
     "ember-cli-dependency-checker": "^3.3.1",
     "ember-cli-inject-live-reload": "^2.1.0",
+    "ember-cli-mirage": "3.0.3",
```
ZedLi, line 53: We have this now both in `devDependencies` and `dependencies`

ZedLi (changes requested):

addons/api/package.json
cameronperera: Apologies. I will fix that now.

addons/api/package.json
```diff
@@ -41,6 +41,7 @@
     "@ember/string": "^3.0.1",
     "@ember/test-helpers": "^2.9.3",
     "@embroider/test-setup": "^2.1.1",
+    "@faker-js/faker": "^8.0.2",
```
calcaide, line 44: What is the reason for adding faker here, it was not here before, right?

calcaide: Oh I see before it was on line 67 and now we move it to 44?

calcaide:
@cameronperera 
I do see the build you link within the description has an error when generating release metadata. I am pretty sure it is not related, but I will feel more confident to approve this PR if we can do a run with all green (or just re-run the failed step there).

For context:
The release metadata geneartion step is prettys straight forward, B-U-T the piece of software we use to do so is very picky (as it should) with namings, arch's and platforms. Sometimes, this job will fail because we changed something that affected the naming generation of a zip or a binary, which usually means we broke something else, that's why my comment above. I am pretty sure in this case is a hiccup, but I would like if you can re-run.

calcaide (dismissed):
Left a couple of commetns

addons/api/package.json
cameronperera: yeah I am not exactly sure why it moved. It just did.

cameronperera:
> @cameronperera I do see the build you link within the description has an error when generating release metadata. I am pretty sure it is not related, but I will feel more confident to approve this PR if we can do a run with all green (or just re-run the failed step there).
> 
> For context: The release metadata geneartion step is prettys straight forward, B-U-T the piece of software we use to do so is very picky (as it should) with namings, arch's and platforms. Sometimes, this job will fail because we changed something that affected the naming generation of a zip or a binary, which usually means we broke something else, that's why my comment above. I am pretty sure in this case is a hiccup, but I would like if you can re-run.

I fixed the naming of my test release branch and that fixed the issue. I updated the description with the new build and it was successful all the way through!

addons/api/package.json
ZedLi: I'm confused, why do we now have one pinned to `3.0.3` in `devDependencies` and `^3.0.3` in `dependencies`?

cameronperera: I did some git stashing and I missed the devDep one. Let me try this again. Apologies, I thought I had it all fixed.

cameronperera: Fixed and test build ran successfully and local tests working.

lisbet-alvarez (approved):
Thanks for updating this!

ZedLi (approved):

addons/api/package.json
ZedLi: It's because it's supposed to be alphabetical ordering so it re-ordered it

cameronperera: Oh yeah. I would have expected to see that more often though so I didn't think to check for the order.

PR #2459, opened 2024-08-22
fix: 🐛 Fix windows hanging when client agent isn't running

ZedLi:
# Description
There was a change where we manually timeout requests to client agent status as currently it can take too long. However, it turns out out on windows, when a `SIGTERM` was sent to the client agent status it didn't actually return anything. On mac an error response was returned on `stderr` but because nothing was returned on windows, our handler for `spawn` was waiting indefinitely for a response. 

This change seeks to at least always resolve the promise if the child process ends so a caller isn't waiting forever. 

## How to Test
Run DC in windows without client agent running and it should not hang up when logging in and going to settings page.

## Checklist

### DIFF
--- a/ui/desktop/electron-app/src/helpers/spawn-promise.js
+++ b/ui/desktop/electron-app/src/helpers/spawn-promise.js
@@ -90,9 +90,9 @@ module.exports = {
       },
     });
 
-    const { stdout, stderr } = childProcess;
+    const { stdout, stderr, error } = childProcess;
 
-    return { stdout: stdout?.toString(), stderr: stderr?.toString() };
+    return { stdout: stdout?.toString(), stderr: stderr?.toString(), error };
   },
 
   /**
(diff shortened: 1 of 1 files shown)


ui/desktop/electron-app/src/helpers/spawn-promise.js
```diff
@@ -90,9 +90,9 @@ module.exports = {
       },
     });
 
-    const { stdout, stderr } = childProcess;
+    const { stdout, stderr, error } = childProcess;
```
ZedLi, line 93: This wasn't relevant to the PR but I realized we weren't returning the error

ui/desktop/electron-app/src/helpers/spawn-promise.js
```diff
@@ -111,6 +111,19 @@ module.exports = {
...
+      // a timeout occurs so this guarantees we return a response to the caller.
+      // Otherwise this should not get hit as we should be returning a response
+      // from one of the handlers above.
+      childProcess.on('close', () =>
+        resolve({
+          childProcess,
+          stderr: JSON.stringify({ error: 'Process was closed.' }),
+        }),
+      );
     });
```
lisbet-alvarez, line 127: non-blocking: should a similar handler for `close` be added to the `spawnAsyncJSONPromise` helper since it also returns a promise ?

lisbet-alvarez (approved):
Tested on windows, works great!

DhariniJeeva (approved):

PR #2463, opened 2024-08-26
refactor: 💡 update test selectors for storage-buckets list

cameronperera:
✅ Closes: https://hashicorp.atlassian.net/browse/ICU-14034

# Description

update test selectors for storage-buckets list
this introduces a few new storage bucket specific selectors and common selectors.

## Checklist


ui/admin/tests/acceptance/storage-buckets/selectors.js
```diff
@@ -30,3 +31,8 @@ export const FIELD_WORKER_FILTER = '[name=worker_filter]';
 export const FIELD_WORKER_FILTER_ERROR = '[data-test-worker-filter-error]';
 export const TOAST = '[role=alert] div';
 export const TOAST_SUCCESSFULL_VALUE = 'Saved successfully.';
+
+export const TABLE_FIRST_ROW_ACTION_DROPDOWN =
+  'tbody tr:first-child td:last-child button';
+export const TABLE_FIRST_ROW_ACTION_FIRST_ITEM =
+  'tbody tr:first-child td:last-child ul li:first-child a';
```
cameronperera, line 38: These selector ultimately could fall under "common selectors" but when it comes to tables and using selectors to interact or assert on certain elements, we should keep it close to the resource we are working with. Happy to discuss this more and/or change to a common selector.

ui/admin/tests/helpers/selectors.js
```diff
@@ -15,4 +15,10 @@ export const TABLE_FIRST_ROW_RESOURCE_LINK = 'tbody tr:first-child a';
 export const RESOURCE_NOT_FOUND_SUBTITLE = '.rose-message-subtitle';
 export const RESOURCE_NOT_FOUND_VALUE = 'Error 404';
 
+export const PAGE_MESSAGE_DESCRIPTION = '.rose-message-description';
+export const PAGE_MESSAGE_LINK = '.rose-message-body .hds-link-standalone';
+
 export const HREF = (url) => `[href="${url}"]`;
+
+export const SIDEBAR_NAV_NTH_LINK = (i) =>
+  `[title="General"] a:nth-of-type(${i})`;
```
cameronperera, line 24: For this selector, we could create a selector for FIRST, SECOND, THIRD, etc. However, I felt this is a simpler way to use it. Thoughts?

ui/admin/tests/acceptance/storage-buckets/selectors.js
calcaide: Agreed, this granularity of selector with pseudo-class is best to keep it close to its domain.
Something like: `tbody button.deleteElement` can fit better as common selector. Since it's using very general selector + a class.

ui/admin/tests/helpers/selectors.js
calcaide: For something as important as the sidenav, main navigation element of the app, I will aim for more specific selector. I.E: `SIDEBAR_ORGS`, `SIDEBAR_STORAGE_BUCKETS`.

My thinking:
- Easier to read/understand. I.E: to know what section we where clicking in each test, I had to open the UI, check I had the same flags and version as the test and then check what section belongs to the index 2 or 3. 
- The sidebar, despite not often, can suffer changes, therefore we will need to refactor all these indexes and that would not be easy for the reason I mentioned above.
- I am not saying it can't be a dynamic option such as a function, but if you want to go that route I will suggest use descriptive names and not indexes. I.E: `.dom(commonSelectors.SIDEBAR_NAV_LINK('storage_buckets'))`.


Happy to discuss this at tech time 😃

calcaide: As discussed within our 1:1 I will wait for changes within this function 😉

cameronperera: This has been updated. Please take another look when you have a chance.

lisbet-alvarez (approved):
looks great!! 🙌

ui/admin/tests/acceptance/storage-buckets/selectors.js
```diff
@@ -3,6 +3,7 @@
  * SPDX-License-Identifier: BUSL-1.1
  */
 
+export const STORAGE_BUCKET_TITLE = 'Storage Buckets';
```
calcaide, line 6: Since this selector is within the storage bucket specific selectors file, adding STORAGE_BUCKET_* to the selector name, seems redundant.
What are your thoughts on renaming this selector to TITLE, TITLE_SECTION or something like that?

calcaide (approved):
Awesome work!! left a comment but nothing major 😉

ui/admin/tests/acceptance/storage-buckets/selectors.js
cameronperera: Great thought. I opted to include `STORAGE_BUCKET` in the title name because we are calling the import `selectors`. If we maybe changed the import name to `SBselectors` we could leave `STORAGE_BUCKET` out of the title name. This is something we can discuss during tech time maybe?

PR #2468, opened 2024-08-30
Implement DC with no CLI and using users path CLI

calcaide:
# Description
Implement BYPASS_CLI_BUILD and using users path CLI.

:tickets: [Jira ticket](https://hashicorp.atlassian.net/browse/ICU-14909)

## How to test:

- I perform a Mac local built with no cli directory present and `SETUP_CLI=false`. So we DO NOT download and include CLI. We use users path CLI.
- I perform a Mac local built with `SETUP_CLI=true`. Se we download and include CLI. We use built in CLI.

For testing, I performed local build, get the output binary and use debugtron to debug the output and not the dev environment.

## Screenshots:
Mac build with built-in CLI:
Screenshot of the app working + console logs reporting what CLI is been use and in which path.

mac-built-in-cli

Mac build with NO built-in CLI:
Screenshot of the app working + console logs reporting what CLI is been use and in which path.

mac-system-cli


ui/desktop/electron-app/config/forge.config.js
```diff
@@ -39,7 +42,7 @@ module.exports = {
         done();
       },
     ],
-    extraResource: ['./cli'],
+    extraResource,
```
calcaide, line 45: I though on leaving the `beforeAsar` regardless, so we make sure we always clean the CLI directory and we do not pack it by error. Let me know if you will like to discuss this.

ZedLi: Yeah it shouldn't matter much, we prevent errors if the path doesn't exist with `force: true`

ui/desktop/electron-app/config/forge.config.js
```diff
@@ -4,6 +4,9 @@ const { isMac } = require('../src/helpers/platform.js');
 const fs = require('fs');
 const path = require('path');
 
+// When BYPASS_CLI_BUILD we do not copy the CLI to the output binary.app/
+const extraResource = process.env.BYPASS_CLI_BUILD === 'true' ? [] : ['./cli'];
```
ZedLi, line 8: Do we need another flag here? Could we have just always made the `cli` folder and leave it empty if we don't package the CLI? Or even better can't we actually just check if the directory exists here and conditionally decide whether to include the `extraResource`instead?

ZedLi:
Is changing the auto update check happening in a separate PR?

DhariniJeeva (approved):
looks good!

ui/desktop/electron-app/config/forge.config.js
calcaide: My initial idea with this approach was to provide an specific flag to guarantee a build with no CLI regardless where the build process was run, but mostly thinking in the pipeline.

But after reading your comment, you are right. This could be simple just checking if the cli directory is empty or not.

Since we already have `SETUP_CLI`, we can control not downloading a CLI, therefore cli directory will be empty.

Thanks for the suggestion, I will implement it without `BYPASS_CLI_BUILD`.

ui/desktop/electron-app/config/forge.config.js
```diff
@@ -4,6 +4,13 @@ const { isMac } = require('../src/helpers/platform.js');
 const fs = require('fs');
 const path = require('path');
 
+// When SETUP_CLI=false we do not download the CLI. If there is no cli directory present
```
calcaide, line 7: @ZedLi 
Change the approach as per suggestion.

Now, if the cli directory exists, we will copy the cli in. If the directory does not exists we don't.

This translates the responsibility of having CLI or not to whoever runs the build process.

Changes on boundary-ui-releases to use this on tis way.

calcaide: I just realize, assuming we are performing 2 builds, once with CLI other without (reasons explained below) I think we want to differentiate the build with no CLI.
The binary current name is: Boundary.app
I do propose for no cli: Boundary-nocli.app

thoughts?

Why we need to perform 2 builds?

It was suggested to just perform one build, and then to provide no CLI just use the normal build and delete the CLI within Boundary.app/contents/... but after quick testing that is not feasible, or at least not as easy as we though. I will put the details of why is not within RFC within abandoned ideas.

ui/desktop/electron-app/src/cli/path.js
```diff
@@ -6,15 +6,20 @@
...
+    const builtInCliPath = isDev
+      ? path.resolve(__dirname, '..', '..', 'cli', binName)
+      : path.resolve(process.resourcesPath, 'cli', binName);
+    const isBuiltInCli = existsSync(builtInCliPath);
 
-    // The CLI will be located in a different location once packaged up in an ASAR
-    return isDev
-      ? path.resolve(__dirname, '..', '..', 'cli', name)
-      : path.resolve(process.resourcesPath, 'cli', name);
+    return isBuiltInCli ? builtInCliPath : binName;
```
lisbet-alvarez, line 23: Perhaps a small change of `binName` to `binaryName` could help with readability. Also just to clarify: we return just the binary name when the cli folder does NOT exist because of the case where the user used the installer to download the binary ?

ZedLi (approved):

ui/desktop/electron-app/src/cli/path.js
calcaide: @lisbet-alvarez 
Absolutley!! Change done! thanks for the suggestion.

In case the DC has a built in CLI we return the absolute path to it. Notice the path is different when you are in dev mode.

In case the DC does not have a built in CLI we return the binary name, assuming the boundary binary will be available at the users $Path, so boundary is present within users systems.

For more context: it is a "safe" assumption to made, since the DC with no CLI will be ONLY available for the installer. Will not be public as DC binary.

It is true, we will need to add a check, if the DC without CLI can't find it, it should gracefully let the user know. ATM it returns an error within the error modal, not ideal, but I will provide soon something better.

lisbet-alvarez (approved):
🙌 Nice work!

cameronperera (approved):

calcaide:
> Is changing the auto update check happening in a separate PR?

@ZedLi 

My apologize I forgot to answer this and to create the ticket for it. I will be taking care of it in [a separate ticket](https://hashicorp.atlassian.net/browse/ICU-15109)

PR #2471, opened 2024-09-03
test: 💍 add TODO for a11y violations

cameronperera:
# Description

This will help address the false negative a11y violations we are seeing when tests are run in dark mode.

```
// TODO: Failing due to a11y violation while in dark mode.
// Investigating issue with styles not properly
// being applied during test.
const session = this.owner.lookup('service:session');
session.set('data.theme', 'light');
```

You will see that I have repeated the above TODO and session lookup, sometimes in the same test file more than once. This is to ensure I am only changing it for that exact test and we can easily remove it from each test as we address these issues. 
I was going to use [QUnit's test.todo](https://qunitjs.com/api/QUnit/test.todo/) but that would be for a WIP test that we expect to fail. With the change of the theme for the specific tests, they will pass every time now so `test.todo` doesn't exactly work here.

## How to Test

To verify that this is working, pull down these changes, change system theme to dark, and run `yarn test`. All test should be passing as expected.

## Checklist


ZedLi (approved):
> I was going to use [QUnit's test.todo](https://qunitjs.com/api/QUnit/test.todo/) but that would be for a WIP test that we expect to fail. With the change of the theme for the specific tests, they will pass every time now so test.todo doesn't exactly work here.

You would just let it fail by not manually setting the theme in this case. But this also works fine

cameronperera:
> > I was going to use [QUnit's test.todo](https://qunitjs.com/api/QUnit/test.todo/) but that would be for a WIP test that we expect to fail. With the change of the theme for the specific tests, they will pass every time now so test.todo doesn't exactly work here.
> 
> You would just let it fail by not manually setting the theme in this case. But this also works fine

That was my original plan but technically the test are working as expected. Just the false negative from the a11yAudit so the `test.todo` felt incorrect here.

DhariniJeeva (approved):

lisbet-alvarez (approved):
thx for adding this!

ui/admin/tests/acceptance/accounts/change-password-test.js
```diff
@@ -77,6 +77,11 @@ module('Acceptance | accounts | change password', function (hooks) {
 
   test('can change password for account', async function (assert) {
     assert.expect(2);
+    // TODO: Failing due to a11y violation while in dark mode.
```
calcaide, line 80: Do we have a jira ticket to cover such investigation? if so It will good to add the jira ticket link, so no need to repeat all the 
```
// TODO: Failing due to a11y violation while in dark mode.
    // Investigating issue with styles not properly
    // being applied during test.
```

and we can say something like `// TODO: Issue fix ICU-12345`

PR #2475, opened 2024-09-04
Use updated cache daemon

ZedLi:
# Description
There have been quite a few changes to how we handle responses. The cache daemon will now truncate any response that returns more than 250 items. They will also now return a `incomplete` response to indicate that the response was truncated as well as a `refresh_status` to indicate if they're still building the cache in the background. The latter will also be used eventually when partial results are returned.

These two new fields are independent which means for example we could have an incomplete response and still be refreshing in the background or a complete response and still be refreshing in the background. 

I've also made a change to always show the filters so the user has access to the refresh button. This is mainly needed as we will eventually return partial responses and for users to refresh while waiting for the cache to build. Previously we had no way for letting users to refresh without using the electron refresh hot keys. 

The dynamic scope filters for targets now just returns all the scopes the user have access to instead of ones just associated with targets. Foundations added `implicit-scopes` which are all scopes associated with targets/sessions but we would have to duplicate the scope models to be able to use it (otherwise all scope calls would go through the daemon) so I ended up just loading all scopes. It might be better to just wait until scopes are part of the daemon as it's just an extra call right now that likely won't change much for the filters.

Example response from cache daemon with over 250 results:
```json
{
  "status_code": 200,
  "item": {
    "targets": [
      {
        "id": "ttcp_01un6InYFN",
        "scope_id": "p_t9Wd9rDQgW",
        "scope": {
          "id": "p_t9Wd9rDQgW",
          "type": "project",
          "name": "project0",
          "description": "Test project scope #0",
          "parent_scope_id": "o_a0XERM5apw"
        },
        "name": "Test TCP Target #141984",
        "created_time": "2024-08-28T20:42:25.194206Z",
        "updated_time": "2024-08-28T20:42:28.553402Z",
        "version": 3,
        "type": "tcp",
        "session_max_seconds": 2147483647,
        "session_connection_limit": -1,
        "attributes": {
          "default_port": 22
        },
        "authorized_actions": [
          "remove-credential-sources",
          "authorize-session",
          "update",
          "remove-host-sources",
          "delete",
          "add-credential-sources",
          "no-op",
          "read",
          "set-credential-sources",
          "add-host-sources",
          "set-host-sources"
        ]
      }
    ],
    "incomplete": true,
    "refresh_status": "not-refreshing"
  }
}
```

image
image

## How to Test
Build from the `llb-cache-speedup` branch in `boundary` and copy it to the desktop CLI folder. Use a cluster that has a decent amount of targets. My [HCP cluster](https://27425825-515c-4d8c-beb6-aa91cbc20c5a.boundary.hcp.to) has ~50k targets if you'd like to try with that. If you want 500k+ targets you'll need to use the boundary scale [repo](https://github.com/hashicorp/boundary-scale-tests). I can also whitelist your IP if you want to use my cluster. 

## Checklist

### DIFF
--- a/addons/api/addon-test-support/handlers/cache-daemon-search.js
+++ b/addons/api/addon-test-support/handlers/cache-daemon-search.js
@@ -7,6 +7,7 @@ import sinon from 'sinon';
 import { typeOf } from '@ember/utils';
 import { resourceNames } from 'api/handlers/cache-daemon-handler';
 import { singularize } from 'ember-inflector';
+import { underscore } from '@ember/string';
 
 /**
  * This test helper can be used to help setup your sinon stubs in your tests.
@@ -69,16 +70,18 @@ export default function setupStubs(hooks) {
           .withArgs('searchCacheDaemon')
           .onCall(i)
           .returns({
-            [resourceName]: models.map((model) => {
-              // Use internal serializer to serialize the model correctly
-              // according to our mirage serializers
-              const modelData =
-                this.server.serializerOrRegistry.serialize(model);
+            [underscore(resourceNames[singularize(resourceName)])]: models.map(
+              (model) => {
+                // Use internal serializer to serialize the model correctly
+                // according to our mirage serializers
+                const modelData =
+                  this.server.serializerOrRegistry.serialize(model);
 
-              // Serialize the data properly to standard JSON as that is what
-              // we're expecting from the cache daemon response
-              return JSON.parse(JSON.stringify(modelData));
-            }),
+                // Serialize the data properly to standard JSON as that is what
+                // we're expecting from the cache daemon response
+                return JSON.parse(JSON.stringify(modelData));
+              },
+            ),
           });
       });
     };
--- a/addons/api/addon/handlers/cache-daemon-handler.js
+++ b/addons/api/addon/handlers/cache-daemon-handler.js
@@ -8,6 +8,7 @@ import { inject as service } from '@ember/service';
 import { getOwner, setOwner } from '@ember/application';
 import { generateMQLExpression } from '../utils/mql-query';
 import { paginateResults } from '../utils/paginate-results';
+import { underscore } from '@ember/string';
 
 /**
  * Not all types are yet supported by the cache daemon so we'll
@@ -94,12 +95,14 @@ export default class CacheDaemonHandler {
         const sessionData = this.session.data?.authenticated;
         const auth_token_id = sessionData?.id;
         const token = sessionData?.token;
+        const resourceName = resourceNames[type];
+
         remainingQuery = {
           ...remainingQuery,
           query: searchQuery,
           auth_token_id,
           token,
-          resource: resourceNames[type],
+          resource: resourceName,
         };
 
         let cacheDaemonResults = {};
@@ -130,7 +133,7 @@ export default class CacheDaemonHandler {
             }
           } else {
             __electronLog?.error(
-              'Failed to search cache daemon; falling back to search controller',
+              `Failed to search cache daemon for ${resourceName}. Error:`,
               e.message,
             );
 
@@ -138,10 +141,7 @@ export default class CacheDaemonHandler {
           }
         }
 
-        // Currently returns with a singular top level field with resource name
-        // e.g. { targets: [...] } or { sessions: [..] }
-        // So this just unwraps to the array, or undefined
-        const [results] = Object.values(cacheDaemonResults);
+        const results = cacheDaemonResults[underscore(resourceName)];
         const payload = { items: paginateResults(results, page, pageSize) };
 
         const schema = store.modelFor(type);
@@ -167,7 +167,15 @@ export default class CacheDaemonHandler {
         // This isn't conventional but is better than returning an ArrayProxy
         // or EmberArray since the ember store query method asserts it has to be an array
         // so we can't just return an object.
-        records.meta = { totalItems: results?.length ?? 0 };
+        // Also include whether the results are incomplete or is still refreshing
+        // as part of the response
+        records.meta = {
+          totalItems: results?.length ?? 0,
+          isLoadIncomplete: cacheDaemonResults.incomplete ?? false,
+          // Sometimes the refresh status returns an error status if it takes too long but it's still refreshing
+          isRefreshing: cacheDaemonResults.refresh_status !== 'not-refreshing',
+        };
+
         return records;
       }
       default:
--- a/addons/core/translations/states/en-us.yaml
+++ b/addons/core/translations/states/en-us.yaml
@@ -18,3 +18,6 @@ loading:
   title: Loading all items...
   description: All of your {resource} are still loading. Please wait while we finish loading everything on screen.
   generic-description: All of your resources are still loading. Please wait while we finish loading everything on screen.
+incomplete-loading:
+  limit: We've reached the result limit. Please narrow down your search or filters if the result you're looking for is not shown
+  refreshing: We're still currently loading the results. Please wait a moment before refreshing again.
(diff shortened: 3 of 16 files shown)


addons/api/addon/handlers/cache-daemon-handler.js
```diff
@@ -167,7 +167,14 @@ export default class CacheDaemonHandler {
         // This isn't conventional but is better than returning an ArrayProxy
         // or EmberArray since the ember store query method asserts it has to be an array
         // so we can't just return an object.
-        records.meta = { totalItems: results?.length ?? 0 };
+        // Also include whether the results are incomplete or is still refreshing
+        // as part of the response
+        records.meta = {
```
calcaide, line 172: Like a P-R-O! nice!

addons/api/addon-test-support/handlers/cache-daemon-search.js
```diff
@@ -69,16 +70,18 @@ export default function setupStubs(hooks) {
           .withArgs('searchCacheDaemon')
           .onCall(i)
           .returns({
-            [resourceName]: models.map((model) => {
-              // Use internal serializer to serialize the model correctly
-              // according to our mirage serializers
-              const modelData =
-                this.server.serializerOrRegistry.serialize(model);
+            [underscore(resourceNames[singularize(resourceName)])]: models.map(
```
calcaide, line 73: none blocking question: What is the reason behind adding underscore? Did we get case sensitive issues? or is for visual/design purposes?

BTW I didn't know this was available through `@ember/string`

addons/core/translations/states/en-us.yaml
```diff
@@ -18,3 +18,6 @@ loading:
   title: Loading all items...
   description: All of your {resource} are still loading. Please wait while we finish loading everything on screen.
   generic-description: All of your resources are still loading. Please wait while we finish loading everything on screen.
+incomplete-loading:
+  limit: We've reached the result limit. Please narrow down your search or filters if the result you're looking for is not shown.
```
calcaide, line 22: None blocking: it is "reached the results limit" or "reached the result limit"? Asking because in the `refreshing` string we say "results"

ui/desktop/app/routes/scopes/scope/projects/sessions/index.js
```diff
@@ -97,41 +96,60 @@ export default class ScopesScopeProjectsSessionsIndexRoute extends Route {
       page,
       pageSize,
       force_refresh: true,
+      max_result_set_size: -1,
```
DhariniJeeva, line 99: what does -1 indicate?

calcaide (approved):
I tested manually, as per my understanding of the PR successfully. But I will explain a bit just in case you see a none expected behaviour.

I tested using your HCP cluster.

- First load: open the DC client and in targets list I got the message: still loading...
- I waited a couple of minutes and click refresh again, now 250 targets are load and the message now says: "We've reached the result limit..."
- I start playing around with the search bar.
- Searching for a term like "blah" returns the "We've reached the result limit..." So over 250 results.
- Searching for a term like "understanding", does not show any message, does show the list of targets that contain that term. 30 targets are return.

Great work!!! 🙌 thanks for taking care of this.

addons/api/addon-test-support/handlers/cache-daemon-search.js
ZedLi: So the main issue is the change [here](https://github.com/hashicorp/boundary-ui/pull/2475/files#diff-b16b26b10ee0d49fd5b21a15ae0557b83b56f724edf52ab924b41737972ad124L144). Previously I was just array destructuring the value out of the return. Unfortunately now that they added more fields to the object, I can't guarantee getting the correct field so I had to grab it by key. This complicated things as this didn't accurately map the field in mirage to the actual response type now for some fields. 

The issue is for multiple word types. For example, we make the call for a resource type called `alias` as `store.query('alias')` which we map to `resolvable-aliases` for the daemon but the actual response field type is `resolvable_aliases`. This is further complicated by the usage of mirage where the resource types are plural and it's _`aliases`_ (instead of `alias`) which we need to return as `resolvable_aliases`.

Let me know if that's unclear, it's a bit convoluted but it's how our types are being mapped across our models both from API and mirage

addons/core/translations/states/en-us.yaml
ZedLi: Hmm I was referring to "results" in the other statement in the plural sense but in this case I did mean the "result limit" in the singular sense (i.e. the result limit is 250, a singular number)

ui/desktop/app/routes/scopes/scope/projects/sessions/index.js
ZedLi: Ahh this isn't obvious but bc the cache daemon will truncate responses by default, the `-1` just indicates to still return back all results. I still wanted all results for the `allSessions` call as I need to collect up all the possible targets to filter by and bc this is filtered by user, there shouldn't be that many total sessions.

ui/desktop/app/routes/scopes/scope/projects/targets/index.js
```diff
@@ -79,8 +77,12 @@ export default class ScopesScopeProjectsTargetsIndexRoute extends Route {
    * projects for filtering options.
    *
    * @returns {Promise<{totalItems: number, targets: [TargetModel], projects: [ScopeModel], allTargets: [TargetModel] }> }
+   * @return {Promise<{totalItems: number, targets: [TargetModel], projects: [ScopeModel],
+   * isCacheDaemonRunning: boolean, isLoadIncomplete: boolean, isRefreshing: boolean}>}
    */
   async model({ search, scopes, availableSessions, types, page, pageSize }) {
+    const isCacheRunningPromise = this.ipc.invoke('isCacheDaemonRunning');
```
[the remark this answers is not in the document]
ZedLi: Yeah, throughout these routes, I tried not to await any promise unless it needed to be done immediately so we can kick off multiple calls at once, that way we're not awaiting all requests sequentially.

i.e.
```
Call A --> Call B --> Call C   Total
100ms      100ms     100ms    300ms
```

vs
```
Call A --> Call B       Total
Call C
100ms      100ms      200ms
```

addons/api/addon/handlers/cache-daemon-handler.js
ZedLi: This isn't ideal IMO and I quite dislike it 😬, it's very hacky but I still don't have a better way of doing it unfortunately

DhariniJeeva (approved):
looks good to me!

addons/api/addon/handlers/cache-daemon-handler.js
calcaide: Not judjing at all, what I like it is in a few lines it is pretty clear what you ar defining and the origin of it, so I find it very descriptive an easy to understand

calcaide (approved):
I re-tested after the latest changes and working as expected. Once again, nice work solving this issue!

ui/desktop/app/routes/scopes/scope/projects/sessions/index.js
```diff
@@ -97,43 +97,70 @@ export default class ScopesScopeProjectsSessionsIndexRoute extends Route {
...
     }
-    this.allTargets = await this.store.query('target', allTargetsQuery, {
-      pushToStore: false,
-    });
+
+    const uniqueTargetIds = [
+      ...new Set(this.allSessions.map((session) => session.target_id)),
+    ];
+    // Chunk target IDs in case there's a lot of sessions as cache daemon syntax can't take too many
+    const chunkedTargetIds = chunk(uniqueTargetIds, 50);
```
lisbet-alvarez, line 136: If im understanding this correctly, we are separating the target ids into chunks of 50, why not 250 since that is the limit ? Is it because its faster to do more smaller chunks than less bigger chunks ?

PR #2480, opened 2024-09-09
fix: 🐛 Fix daemon getting killed if started before DC

ZedLi:
# Description
I missed a check when the client daemon was renamed to cache daemon. This accidentally caused us to kill a user started cache daemon. 

## How to Test
Start the daemon before the DC and close the DC after logging in. The daemon should still be running. 

## Checklist


DhariniJeeva (approved):
thanks for the fix!

lisbet-alvarez (approved):
thx for fixing this!

PR #2483, opened 2024-09-11
Add duration unit format

cameronperera:
NOTE: I did upgrade webpack in this branch to verify that this fixed the build issues and it... DOES! Once this PR is merged, I will get the webpack PR updated and verify it is working.

# Description

This PR is replacing `intl-unofficial-duration-unit-format` with `@formatjs/intl-durationformat`. This is to help us get the [webpack upgrade](https://github.com/hashicorp/boundary-ui/security/dependabot/158) ([PR](https://github.com/hashicorp/boundary-ui/pull/2466)) into a working state.

The issue with `intl-unofficial-duration-unit-format` is due to an [export path](https://github.com/piuccio/intl-unofficial-duration-unit-format/issues/33) format in the released versions. Instead of patching this fix, we took this time to replace this package with the newer `@formatjs/intl-durationformat` polyfill.

One small difference between these packages is the polyfill ([spec for Int.DurationFormat](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/DurationFormat/DurationFormat)) does not seem to have the same ability to define the format we want. Before if the session had a `max_duration` greater than 24 hours, it would keep adding to the hours and not show days. The new polyfill will display days now.
Before: `24:01:26 remaining`
After: `1 day, 0:01:26 remaining`

## How to Test

Use Vercel or an instance with session recordings.
In Admin UI, visit the session recordings pages and look for the different places we display the duration of the session recording.
In the Desktop Client, start a session and you should see "xx:xx:xx remaining" when viewing the session details.

## Checklist

### DIFF
--- a/addons/core/addon/helpers/time-remaining.js
+++ b/addons/core/addon/helpers/time-remaining.js
@@ -5,7 +5,8 @@
 
 import Helper from '@ember/component/helper';
 import { inject as service } from '@ember/service';
-import DurationUnitFormat from 'intl-unofficial-duration-unit-format';
+import { DurationFormat } from '@formatjs/intl-durationformat';
+import { secondsToDuration } from 'core/utils/seconds-to-duration';
 
 export default class extends Helper {
   // =services
@@ -29,14 +30,13 @@ export default class extends Helper {
       difference > 0 ? Math.floor(difference / 1000) : 0;
 
     // Format the duration
-    const duration = new DurationUnitFormat(this.intl.primaryLocale, {
-      style: 'timer',
-      format: `{hours}:{minutes}:{seconds} ${this.intl.t(
-        'resources.session.remaining',
-      )}`,
+    const duration = new DurationFormat(this.intl.primaryLocale, {
+      style: 'digital',
     });
+    const formattedTime = duration.format(
+      secondsToDuration(differenceInSeconds),
+    );
 
-    // Return the formatted duration
-    return duration.format(differenceInSeconds);
+    return `${formattedTime} ${this.intl.t('resources.session.remaining')}`;
   }
 }
--- a/addons/core/tests/integration/helpers/time-remaining-test.js
+++ b/addons/core/tests/integration/helpers/time-remaining-test.js
@@ -32,6 +32,17 @@ module('Integration | Helper | time-remaining', function (hooks) {
     );
   });
 
+  test('it calculates the correct time remaining with more than a week', async function (assert) {
+    this.set('expirationTime', new Date(Date.now() + 1001 * 60 * 60 * 24 * 8));
+
+    await render(hbs`{{time-remaining this.expirationTime}}`);
+
+    assert.strictEqual(
+      this.element.textContent.trim(),
+      '1 wk, 1 day, 0:11:31 remaining',
+    );
+  });
+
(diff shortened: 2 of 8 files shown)


lisbet-alvarez (approved):
Looks great! 🙌

addons/core/addon/helpers/time-remaining.js
```diff
@@ -29,14 +30,13 @@ export default class extends Helper {
       difference > 0 ? Math.floor(difference / 1000) : 0;
 
     // Format the duration
-    const duration = new DurationUnitFormat(this.intl.primaryLocale, {
-      style: 'timer',
-      format: `{hours}:{minutes}:{seconds} ${this.intl.t(
```
ZedLi, line 34: Hmm I forget, was there a reason we only limited it to hours before? I kind of like having more than hours

addons/core/tests/integration/helpers/time-remaining-test.js
```diff
@@ -13,19 +13,26 @@ module('Integration | Helper | time-remaining', function (hooks) {
...
     this.set('expirationTime', new Date(Date.now() + 1000 * 60 * 60 * 24));
 
     await render(hbs`{{time-remaining this.expirationTime}}`);
 
     assert.strictEqual(this.element.textContent.trim(), '23:59:59 remaining');
   });
 
-  test('it handles negative time remaining', async function (assert) {
-    assert.expect(1);
+  test('it calculates the correct time remaining with more than 24 hours', async function (assert) {
```
ZedLi, line 24: Could we also have a test for more than a week to see how it's formatted?

addons/core/tests/unit/utils/seconds-to-duration-test.js
```diff
@@ -0,0 +1,69 @@
+/**
+ * Copyright (c) HashiCorp, Inc.
+ * SPDX-License-Identifier: BUSL-1.1
+ */
+
+import { secondsToDuration } from 'core/utils/seconds-to-duration';
+import { module, test } from 'qunit';
+
+module('Unit | Utility | seconds-to-duration', function () {
+  test('it handles 0 correctly', function (assert) {
```
ZedLi, line 10: Can we add a test of it handling a negative number correctly?

addons/core/addon/helpers/time-remaining.js
cameronperera: I don't remember the reasoning but I think the new format allows for easier readability.

addons/core/tests/integration/helpers/time-remaining-test.js
cameronperera: Done.

addons/core/tests/unit/utils/seconds-to-duration-test.js
cameronperera: Done.

addons/core/addon/helpers/time-remaining.js
calcaide: I remember a discussion, I do not remember the reason for it.

calcaide (approved):
Thanks for the work!

ZedLi (approved):
Looks good!

PR #2486, opened 2024-09-12
Upgrade node engines to work with node 20 and 22

calcaide:
UPDATE Friday January 31: Tested this changes on the `boundary-deskto-releases` successfuly, [here the run](https://github.com/hashicorp/boundary-desktop-releases/actions/runs/13078071926).

# Description
Upgrade node engines to work with node 20 and 22.


cameronperera:
Since node 21 is already past it's [maintenance window](https://github.com/nodejs/release?tab=readme-ov-file#release-schedule), should we try going straight to v22 instead?

DhariniJeeva:
should we also consider using the same version in boundary-ui-releases?

DhariniJeeva (approved):

calcaide:
@cameronperera 

Good point. 

Some thinking loud:
- Right now 20 is LTS, so we should keep node 20 support IMHO.
- As you mention 22 is current.
- So we should keep support for: 20 and 22.

is that correct?

cameronperera:
> @cameronperera
> 
> Good point.
> 
> Some thinking loud:
> 
> * Right now 20 is LTS, so we should keep node 20 support IMHO.
> * As you mention 22 is current.
> * So we should keep support for: 20 and 22.
> 
> is that correct?

Correct. If we are going to upgrade our supported node version, we should probably stick to the LTS versions. Plus v22 doesn't start it's maintenance schedule until a year from now. So this should hold us for a while.

calcaide:
@cameronperera 

Awesome!! pushing changes 😃

calcaide:
@DhariniJeeva 

We keep supporting node 20. For now, I will suggest to keep the pipeline with node 20. Upgrading node in the pipeline requires more work, specially testing. We just upgrade the pipeline from 18 to 20 so IMHo its fine for a while.
Something I will like to explore once TS is released.

Thoughts?

DhariniJeeva:
> @DhariniJeeva
> 
> We keep supporting node 20. For now, I will suggest to keep the pipeline with node 20. Upgrading node in the pipeline requires more work, specially testing. We just upgrade the pipeline from 18 to 20 so IMHo its fine for a while. Something I will like to explore once TS is released.
> 
> Thoughts?

that sounds good!

lisbet-alvarez (dismissed):

cameronperera (approved):

calcaide:
@DhariniJeeva 
Due to reverting the node20 upgrade changes within `boundary-ui-releases` I will hold this PR until we can run node20 within the pipeline.

cameronperera (approved):

calcaide:
> @calcaide did you run a script to update node v in different files, or was this manual? Just wondering how we handle upgrades.

@laurenolivia 

I did perform this manually.

This PR updates [the engines key within package.json](https://docs.npmjs.com/cli/v11/configuring-npm/package-json#engines), it will be developer responsability to upgrade acordingly the node version use within the dev enviornment.

PR #2494, opened 2024-09-18
fix: show username first and then password when credentials are listed

DhariniJeeva:
✅ Closes: https://hashicorp.atlassian.net/browse/ICU-15105

# Description

This PR moves up `username` field  

before:

Screenshot 2024-09-17 at 6 37 46 PM

after:
Screenshot 2024-09-17 at 6 31 53 PM

with a JSON blob: 
Screenshot 2024-09-18 at 6 22 21 PM

## Checklist

### DIFF
--- a/ui/desktop/app/components/credentials-panel/index.hbs
+++ b/ui/desktop/app/components/credentials-panel/index.hbs
@@ -33,6 +33,7 @@
         as |CB|
       >
         <CB.Title>
+
           {{#if credential.source.name}}
             {{credential.source.name}}
           {{else}}
@@ -61,7 +62,7 @@
 
       <ul class='credential-secret'>
         {{#if credential.secrets}}
-          {{#each credential.secrets as |secret|}}
+          {{#each (order-by credential.secrets 'username') as |secret|}}
             <li class='secret-container'>
               <span class='secret-key'>{{secret.key}}</span>
               <HiddenSecret @secret={{secret.value}} />
--- a/ui/desktop/app/components/credentials-panel/index.hbs
+++ b/ui/desktop/app/components/credentials-panel/index.hbs
@@ -62,7 +62,7 @@
 
       <ul class='credential-secret'>
         {{#if credential.secrets}}
-          {{#each (order-by credential.secrets 'username') as |secret|}}
+          {{#each (order-by credential.secrets this.customOrder) as |secret|}}
             <li class='secret-container'>
               <span class='secret-key'>{{secret.key}}</span>
               <HiddenSecret @secret={{secret.value}} />
--- a/ui/desktop/app/components/credentials-panel/index.js
+++ b/ui/desktop/app/components/credentials-panel/index.js
@@ -8,6 +8,8 @@ import { tracked } from '@glimmer/tracking';
 import { action } from '@ember/object';
 
 export default class CredentialsPanelIndexComponent extends Component {
+  customOrder = ['username', 'password'];
+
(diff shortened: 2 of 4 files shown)


DhariniJeeva:
I created this against `main` since this does not impact any TS changes, but let me know if you think otherwise, happy to move this to llb

cameronperera (approved):
LGTM. Thanks for getting this updated!

ui/desktop/tests/integration/helpers/order-by-test.js
```diff
@@ -0,0 +1,35 @@
...
+import { hbs } from 'ember-cli-htmlbars';
+
+module('Integration | Helper | order-by', function (hooks) {
+  setupRenderingTest(hooks);
+
+  test('it renders', async function (assert) {
+    this.set('inputValue', [
+      { key: 'username', value: 1 },
+      { key: 'password', value: 2 },
+      { key: 'none', value: 3 },
```
ZedLi, line 18: What happens if you have other keys that are in between username and password? Does that order make sense?

ui/desktop/app/components/credentials-panel/index.hbs
```diff
@@ -61,7 +62,7 @@
 
       <ul class='credential-secret'>
         {{#if credential.secrets}}
-          {{#each credential.secrets as |secret|}}
+          {{#each (order-by credential.secrets 'username') as |secret|}}
```
ZedLi, line 65: Shouldn't username and password be next to each other? How do we handle if a user has many different keys and username and password might be far apart from each other?

Should we make this more generic to handle any arbitrary sorting order?

DhariniJeeva: that's a valid point, looking into this

DhariniJeeva: changed the function logic to follow the custom order array we pass. So now, the objects will be ordered based on whatever custom order is passed to the function. Let me know what you think

ui/desktop/tests/integration/helpers/order-by-test.js
```diff
@@ -0,0 +1,37 @@
...
+
+import { module, test } from 'qunit';
+import { setupRenderingTest } from 'ember-qunit';
+import { render } from '@ember/test-helpers';
+import { hbs } from 'ember-cli-htmlbars';
+
+module('Integration | Helper | order-by', function (hooks) {
+  setupRenderingTest(hooks);
+
+  test('it renders', async function (assert) {
```
ZedLi, line 14: Could we add a few more test cases to show the different scenarios, e.g. preserves sort order if no keys are in the list, some keys are in the list, etc

ZedLi (approved):
Looks good!

ui/desktop/tests/integration/helpers/order-by-test.js
DhariniJeeva: good call, added!

ui/desktop/app/components/credentials-panel/index.js
```diff
@@ -8,6 +8,8 @@ import { tracked } from '@glimmer/tracking';
 import { action } from '@ember/object';
 
 export default class CredentialsPanelIndexComponent extends Component {
+  customOrder = ['username', 'password'];
```
calcaide, line 11: I really like this! Amazing

calcaide (approved):
Terrific job!! Thank you for this helper and the "issue" it fixes.

PR #2496, opened 2024-09-19
feat: 🎸 Update alert designs

ZedLi:
# Description
Changed the two alerts at the top of the page to a tooltip and nearer pagination.
I also updated the toolbar refreshing tool to use HDS components which will also affect admin UI.

﻿Before:
image

After:
Screenshot 2024-09-19 at 4 06 46 PM

Admin UI:
image

## How to Test
Start up desktop client with a CLI built from the [boundary cache speedup llb](https://github.com/hashicorp/boundary/tree/llb-cache-speedup). Test against a cluster with at least 50k targets, you can use my HCP cluster [here](https://27425825-515c-4d8c-beb6-aa91cbc20c5a.boundary.hcp.to), or I can whitelist you on my 1M targets cluster. 

Also confirm the refresh button works in admin UI on the sessions and workers page.

## Checklist

### DIFF
--- a/ui/desktop/app/templates/scopes/scope/projects/targets/index.hbs
+++ b/ui/desktop/app/templates/scopes/scope/projects/targets/index.hbs
@@ -62,23 +62,21 @@
         </Dropdown>
       </S.Generic>
     </Hds::SegmentedGroup>
-    <span>
+    {{#if @model.isCacheRefreshing}}
+      <ToolbarRefresher @onClick={{route-action 'refreshAll'}}>
+        <Hds::TooltipButton
+          @text={{t 'states.incomplete-loading.refreshing.tooltip'}}
+        >
+          <Hds::Text::Body @tag='p' @color='faint'>
+            {{t 'states.incomplete-loading.refreshing.description'}}
+          </Hds::Text::Body>
+        </Hds::TooltipButton>
+      </ToolbarRefresher>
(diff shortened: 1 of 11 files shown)


calcaide (approved):
Tested manually using the provided HCP cluster.

Works as expected, showing alerts with new design.

Play around applying filters, cleaning up them, forcing refreshes. Everything works as expected, with some delays between actions, but that's expected due to the magnitude of this cluster and all the delays are loading times which are reflected within the load status bar, blue bar below the header.

Thank you for the hard work!

ui/desktop/app/templates/scopes/scope/projects/targets/index.hbs
```diff
@@ -174,6 +172,14 @@
       </B.Tr>
     </:body>
   </Hds::Table>
+  {{#if @model.isLoadIncomplete}}
+    <Hds::Alert @type='compact' as |A|>
```
DhariniJeeva, line 176: do we not want to show it as a warning type in the updated design?

ZedLi: Yeah I think we're changing it from a warning now to just be informational

DhariniJeeva (approved):
looks great, thanks for making the changes!

PR #2498, opened 2024-09-24
icu-14735 add search & pagination to aliases

lisbet-alvarez:
✅ Closes: https://hashicorp.atlassian.net/browse/ICU-14735

# Description

Add search & pagination to aliases.
NOTE: We will need to update db version so it doesn't break for users. (There is a draft pr to do this so we might want to revisit that PR and update it if neccessary).
There will still be some upcoming changes to ensure search & pagination does not break when there are 500k+ aliases.

https://github.com/user-attachments/assets/6b4d556b-8594-4c68-a49c-548500a09938

## How to Test

1. delete current indexedDB since alias index was added.
2. go to "Aliases"
3. experiment with searching & switching between pages.

## Checklist

### DIFF
--- a/addons/api/addon/services/indexed-db.js
+++ b/addons/api/addon/services/indexed-db.js
@@ -29,6 +29,8 @@ export const modelIndexes = {
     '&id, attributes.created_time, attributes.type, attributes.name, attributes.description, attributes.scope.scope_id, attributes.plugin.name',
   'auth-method':
     '&id, attributes.created_time, attributes.type, attributes.name, attributes.description, attributes.is_primary, attributes.scope.scope_id',
+  alias:
+    '&id, attributes.created_time, attributes.type, attributes.value, attributes.name, attributes.description, attributes.destination_id, attributes.scope.scope_id',
 };
 
 export const formatDbName = (userId, clusterUrl) =>
--- a/ui/admin/app/routes/scopes/scope/aliases/index.js
+++ b/ui/admin/app/routes/scopes/scope/aliases/index.js
@@ -41,7 +41,7 @@ export default class ScopesScopeAliasesIndexRoute extends Route {
     const { id: scope_id } = scope;
 
     let aliases;
-    let aliasesExist;
+    let doAliasesExist = false;
     let totalItems = 0;
 
     if (
@@ -63,24 +63,26 @@ export default class ScopesScopeAliasesIndexRoute extends Route {
           hash({
             alias,
             target: alias.destination_id
-              ? this.store.findRecord('target', alias.destination_id)
+              ? this.store.findRecord('target', alias.destination_id, {
+                  backgroundReload: false,
+                })
               : null,
           }),
         ),
       );
-      aliasesExist = await this.getAliasesExist(scope_id, totalItems);
+      doAliasesExist = await this.getDoAliasesExist(scope_id, totalItems);
     }
 
-    return { aliases, aliasesExist, totalItems };
+    return { aliases, doAliasesExist, totalItems };
   }
 
   /**
-   * Sets aliasesExist to true if there exists any aliases.
+   * Sets doAliasesExist to true if there exists any aliases.
    * @param {string} scope_id
    * @param {number} totalItems
    * @returns {Promise<boolean>}
    */
-  async getAliasesExist(scope_id, totalItems) {
+  async getDoAliasesExist(scope_id, totalItems) {
     if (totalItems > 0) {
       return true;
     }
--- a/ui/admin/app/routes/scopes/scope/aliases/index.js
+++ b/ui/admin/app/routes/scopes/scope/aliases/index.js
@@ -77,7 +77,7 @@ export default class ScopesScopeAliasesIndexRoute extends Route {
   }
 
   /**
-   * Sets doAliasesExist to true if there exists any aliases.
+   * Sets doAliasesExist to true if there are any aliases.
    * @param {string} scope_id
    * @param {number} totalItems
    * @returns {Promise<boolean>}
--- a/ui/admin/app/routes/scopes/scope/aliases/index.js
+++ b/ui/admin/app/routes/scopes/scope/aliases/index.js
@@ -11,9 +11,7 @@ export default class ScopesScopeAliasesIndexRoute extends Route {
   // =services
 
   @service store;
-  @service session;
   @service can;
-  @service router;
 
   // =attributes
 
--- a/ui/admin/app/routes/scopes/scope/aliases/index.js
+++ b/ui/admin/app/routes/scopes/scope/aliases/index.js
@@ -4,9 +4,99 @@
  */
 
 import Route from '@ember/routing/route';
+import { inject as service } from '@ember/service';
+import { hash } from 'rsvp';
 
 export default class ScopesScopeAliasesIndexRoute extends Route {
+  // =services
+
+  @service store;
+  @service session;
+  @service can;
+  @service router;
+
+  // =attributes
+
+  queryParams = {
+    search: {
+      refreshModel: true,
+      replace: true,
+    },
+    page: {
+      refreshModel: true,
+    },
+    pageSize: {
+      refreshModel: true,
+    },
+  };
+
   // =methods
+
+  /**
+   * Loads queried aliases and the number of aliases under current scope.
+   * @returns {Promise<{totalItems: number, aliases: [object], aliasesExist: boolean}>}
+   */
+  async model({ search, page, pageSize }) {
+    const scope = this.modelFor('scopes.scope');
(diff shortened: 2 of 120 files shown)


addons/api/addon/services/indexed-db.js
```diff
@@ -29,6 +29,8 @@ export const modelIndexes = {
     '&id, attributes.created_time, attributes.type, attributes.name, attributes.description, attributes.scope.scope_id, attributes.plugin.name',
   'auth-method':
     '&id, attributes.created_time, attributes.type, attributes.name, attributes.description, attributes.is_primary, attributes.scope.scope_id',
+  alias:
+    '&id, attributes.created_time, attributes.type, attributes.value, attributes.name, attributes.description, attributes.destination_id, attributes.scope.scope_id',
 };
```
lisbet-alvarez, line 34: Since a change was made to the index we need to update the indexedDb version. I saw that there is already a draft PR for that. Perhaps reliable-workers pod can look into turning that into an open PR?

ui/admin/app/routes/scopes/scope/aliases/index.js
```diff
@@ -4,10 +4,100 @@
...
+        ),
+      );
+      aliasesExist = await this.getAliasesExist(scope_id, totalItems);
+    }
+
+    return { aliases, aliasesExist, totalItems };
+  }
+
+  /**
+   * Sets aliasesExist to true if there exists any aliases.
```
DhariniJeeva, line 78:
```suggestion
   * Sets aliasesExist to true if there any aliases.
```

ui/admin/app/routes/scopes/scope/aliases/index.js
```diff
@@ -4,10 +4,100 @@
...
+        aliases.map((alias) =>
+          hash({
+            alias,
+            target: alias.destination_id
+              ? this.store.findRecord('target', alias.destination_id)
+              : null,
+          }),
+        ),
+      );
+      aliasesExist = await this.getAliasesExist(scope_id, totalItems);
```
DhariniJeeva, line 71: is this to check if there are any aliases in that particular scope?

ui/admin/app/templates/scopes/scope/aliases/index.hbs
```diff
@@ -18,155 +18,181 @@
...
-                (hash
-                  canUpdate=(can 'update model' B.data.alias)
-                  canDelete=(can 'delete model' B.data.alias)
-                )
-                as |perms|
-              }}
-                {{#if (or perms.canUpdate perms.canDelete)}}
-                  <Hds::Dropdown as |dd|>
+  <page.body class='search-filtering'>
+    {{#if @model.aliasesExist}}
```
DhariniJeeva, line 29: Hmm, can we name this differently? right now, `@model.aliases` and `@model.aliasesExist` read the same to me. Naming it with more context would tell us why we can't use `@model.aliases` to toggle the search bar visibility

DhariniJeeva:
Thanks for introducing pagination to Aliases 👏  
I added a couple of comments, looks great otherwise!

ui/admin/app/routes/scopes/scope/aliases/index.js
```diff
@@ -4,10 +4,100 @@
...
+      });
+      totalItems = aliases.meta?.totalItems;
+      // since we don't receive target info from aliases list API,
+      // we query the store to fetch target information based on the destination id
+      aliases = await Promise.all(
+        aliases.map((alias) =>
+          hash({
+            alias,
+            target: alias.destination_id
+              ? this.store.findRecord('target', alias.destination_id)
```
ZedLi, line 66: Do you think we should set `backgroundReload` to false to help prevent extra queries if it already exists in the store? I imagine we shouldn't need to try and refresh data that was already in the store, especially if people are going between pages

ui/admin/app/templates/scopes/scope/aliases/index.hbs
```diff
@@ -18,155 +18,181 @@
...
+                  <Hds::Text::Body
+                    @tag='p'
+                    @size='100'
+                    @weight='medium'
+                    @color='faint'
+                  >
+                    {{t 'titles.not-set'}}
+                  </Hds::Text::Body>
+                {{/if}}
+              </B.Td>
```
ZedLi, line 89: I know this is a limitation of us not wanting to load all targets for the filter right now but without the filter, people are definitely going to think it's a bug that when they search the target name it doesn't filter down the table 😕

ui/admin/app/routes/scopes/scope/aliases/index.js
lisbet-alvarez: yup correct

ui/admin/app/templates/scopes/scope/aliases/index.hbs
cameronperera: Would something like `@model.doAliasesExist` read better? Try to make it more clear that this is a boolean compared to an array of alias models.

ui/admin/app/templates/scopes/scope/aliases/index.hbs
cameronperera: For other resources that don't require a name, we display the resource ID. I know the ID isn't as helpful as a name but it could still be better than `Not Set`.

ZedLi: I think `Not Set` is different as there could be an alias that's not pointing to any target yet so it's in the absence of the target. `displayName` would still display the ID if the target name doesn't exist

ui/admin/app/templates/scopes/scope/aliases/index.hbs
ZedLi: I will say we have been using this naming pattern across other pages so if we want to change it, we should apply it everywhere.

lisbet-alvarez: good points, i think `doAliasesExist` is clear. Ill also update in other places since its a simple change.

ui/admin/app/routes/scopes/scope/aliases/index.js
lisbet-alvarez: good point, i didnt realize that was an option, originally thought that not includeing `reload: true` was sufficient to prevent extra queries.

ui/admin/app/templates/scopes/scope/aliases/index.hbs
lisbet-alvarez: hmm good thinking, perhaps an `Hds::TooltipButton` in the `Points to` table header letting user know they can only search on id and not name? I'll reach out to design on this.

lisbet-alvarez: update: no changes are going to be made in regards to this, [slack thread ](https://hashicorp.slack.com/archives/C012GTH1C4E/p1727301278916459?thread_ts=1726778606.934579&cid=C012GTH1C4E) for context

ui/admin/tests/acceptance/aliases/list-test.js
```diff
@@ -43,15 +50,32 @@ module('Acceptance | aliases | list', function (hooks) {
...
-
+    urls.alias = `${urls.aliases}/${instances.alias.id}`;
+    urls.aliasWithTarget = `${urls.aliases}/${instances.aliasWithTarget.id}`;
     intl = this.owner.lookup('service:intl');
 
     authenticateSession({});
   });
 
   test('users can navigate to aliases with proper authorization', async function (assert) {
+    this.server.schema.aliases.all().destroy();
```
cameronperera, line 78: non-blocking: Why are we destroying all aliases?

DhariniJeeva (approved):
looks great!

ui/admin/tests/acceptance/aliases/list-test.js
lisbet-alvarez: Ah good question, those test cases also test the empty app state, before there were no aliases but for search & created some in `beforeEach` hook

ui/admin/app/routes/scopes/scope/aliases/index.js
```diff
@@ -4,10 +4,102 @@
...
+import { inject as service } from '@ember/service';
+import { hash } from 'rsvp';
 
 export default class ScopesScopeAliasesIndexRoute extends Route {
+  // =services
+
+  @service store;
+  @service session;
+  @service can;
+  @service router;
```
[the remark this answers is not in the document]
lisbet-alvarez: ahh thank you, great catch!

cameronperera (approved):
LGTM

ZedLi (approved):
Looks good! 1M aliases is incredibly slow but at least it loads

PR #2501, opened 2024-09-25
test(e2e): Add Admin UI test for vault credential libraries (enterprise)

moduli:
This PR adds a test to the Admin UI e2e test suite to check the creation of Vault credential libraries using `boundary-enterprise`. We already had a test for `boundary`, and this adds a similar check for the enterprise version, while accounting for the slight differences that this version has
- need to use the enterprise variant of creating targets (there's a selector for target type (i.e. tcp, ssh))
- need to use the enterprise variant of creating credential libraries (there's a selector for library type (i.e. generic secrets, ssh certificates))

## Testing
One test was added to enterprise
```
yarn run e2e:ent:docker
```

https://hashicorp.atlassian.net/browse/ICU-14761


calcaide (approved):
Tested manually and works as expecetd 😎 

Thanks for the work Michael!!

PR #2502, opened 2024-09-25
Icu 15181 add pagination to session recordings

cameronperera:
# Description

This adds pagination support for Session Recordings along with search for `session-recording.id` and filters for user and target.

Another change that is made in this PR is how we handle redirects for the `scopes.scope.session-recordings.session-recording.channels-by-connection.channel` route. If the user has entered a URL where the session recording ID is wrong, we redirect the user back to the list view and display a toast message.

:tickets: [Jira ticket](https://hashicorp.atlassian.net/browse/ICU-15181)

Search and filter:

https://github.com/user-attachments/assets/8c748e27-898f-4443-a978-e01f571824c7

Redirect change:

https://github.com/user-attachments/assets/41eb6cde-7e64-4c94-96dc-619da173683a

## How to Test

Create sessions with targets that have recording enabled. After creating a few, you should see the table of session recordings with new filters above.

## Checklist

### DIFF
--- a/addons/api/addon/services/indexed-db.js
+++ b/addons/api/addon/services/indexed-db.js
@@ -30,7 +30,7 @@ export const modelIndexes = {
   'auth-method':
     '&id, attributes.created_time, attributes.type, attributes.name, attributes.description, attributes.is_primary, attributes.scope.scope_id',
   'session-recording':
-    '&id, attributes.created_time, attributes.type, attributes.state, attributes.start_time, attributes.end_time, attributes.duration, attributes.scope.scope_id, attributes.create_time_values.user.id, attributes.create_time_values.user.name, attributes.create_time_values.host.id, attributes.create_time_values.target.id, attributes.create_time_values.target.name',
+    '&id, attributes.created_time, attributes.type, attributes.state, attributes.start_time, attributes.end_time, attributes.duration, attributes.scope.scope_id, attributes.create_time_values.user.id, attributes.create_time_values.user.name, attributes.create_time_values.target.id, attributes.create_time_values.target.name, attributes.create_time_values.target.scope.id, attributes.create_time_values.target.scope.name',
 };
 
 export const formatDbName = (userId, clusterUrl) =>
--- a/addons/api/addon/services/indexed-db.js
+++ b/addons/api/addon/services/indexed-db.js
@@ -29,6 +29,8 @@ export const modelIndexes = {
     '&id, attributes.created_time, attributes.type, attributes.name, attributes.description, attributes.scope.scope_id, attributes.plugin.name',
   'auth-method':
     '&id, attributes.created_time, attributes.type, attributes.name, attributes.description, attributes.is_primary, attributes.scope.scope_id',
+  'session-recording':
+    '&id, attributes.created_time, attributes.type, attributes.state, attributes.start_time, attributes.end_time, attributes.duration, attributes.scope.scope_id, attributes.create_time_values.user, attributes.create_time_values.host, attributes.create_time_values.target',
 };
 
 export const formatDbName = (userId, clusterUrl) =>
--- a/addons/api/addon/services/indexed-db.js
+++ b/addons/api/addon/services/indexed-db.js
@@ -30,7 +30,7 @@ export const modelIndexes = {
   'auth-method':
     '&id, attributes.created_time, attributes.type, attributes.name, attributes.description, attributes.is_primary, attributes.scope.scope_id',
   'session-recording':
-    '&id, attributes.created_time, attributes.type, attributes.state, attributes.start_time, attributes.end_time, attributes.duration, attributes.scope.scope_id, attributes.create_time_values.user, attributes.create_time_values.host, attributes.create_time_values.target',
+    '&id, attributes.created_time, attributes.type, attributes.state, attributes.start_time, attributes.end_time, attributes.duration, attributes.scope.scope_id, attributes.create_time_values.user.id, attributes.create_time_values.user.name, attributes.create_time_values.host.id, attributes.create_time_values.target.id, attributes.create_time_values.target.name',
 };
 
 export const formatDbName = (userId, clusterUrl) =>
--- a/addons/api/addon/services/indexed-db.js
+++ b/addons/api/addon/services/indexed-db.js
@@ -113,7 +113,7 @@ export default class IndexedDbService extends Service {
     }
 
     this.#db = new Dexie(dbName);
-    this.#db.version(1).stores(modelIndexes);
+    this.#db.version(2).stores(modelIndexes);
   }
 
   /**
--- a/ui/admin/app/controllers/scopes/scope/session-recordings/index.js
+++ b/ui/admin/app/controllers/scopes/scope/session-recordings/index.js
@@ -32,8 +32,8 @@ export default class ScopesScopeSessionRecordingsIndexController extends Control
   get filters() {
     return {
       allFilters: {
-        users: this.userOptions,
-        targets: this.targetOptions,
+        users: this.filterOptions('user'),
+        targets: this.filterOptions('target'),
       },
       selectedFilters: {
         users: this.users,
@@ -43,35 +43,17 @@ export default class ScopesScopeSessionRecordingsIndexController extends Control
   }
 
   /**
-   * Returns all users for session recordings
+   * Returns all filter options for key for session recordings
+   * @param {string} key
    * @returns {[object]}
    */
-  get userOptions() {
+  @action
+  filterOptions(key) {
     const uniqueMap = new Map();
     this.model.allSessionRecordings.forEach(
       ({
         create_time_values: {
-          user: { id, name },
-        },
-      }) => {
-        if (!uniqueMap.has(id)) {
-          uniqueMap.set(id, { id, name });
-        }
-      },
-    );
-    return Array.from(uniqueMap.values());
-  }
-
-  /**
-   * Returns all targets for session recordings
-   * @returns {[object]}
-   */
-  get targetOptions() {
-    const uniqueMap = new Map();
-    this.model.allSessionRecordings.forEach(
-      ({
-        create_time_values: {
-          target: { id, name },
+          [key]: { id, name },
         },
       }) => {
         if (!uniqueMap.has(id)) {
--- a/ui/admin/app/controllers/scopes/scope/session-recordings/index.js
+++ b/ui/admin/app/controllers/scopes/scope/session-recordings/index.js
@@ -19,7 +19,7 @@ export default class ScopesScopeSessionRecordingsIndexController extends Control
    * @type {boolean}
    */
   get hasSessionRecordings() {
-    return this.model?.sessionRecordings?.length;
+    return !!this.model?.sessionRecordings?.length;
   }
 
   /**
@@ -27,6 +27,6 @@ export default class ScopesScopeSessionRecordingsIndexController extends Control
    * @type {boolean}
    */
   get hasSessionRecordingsConfigured() {
-    return this.model?.storageBuckets?.length;
+    return !!this.model?.storageBuckets?.length;
   }
 }
--- a/ui/admin/app/controllers/scopes/scope/session-recordings/index.js
+++ b/ui/admin/app/controllers/scopes/scope/session-recordings/index.js
@@ -19,7 +19,7 @@ export default class ScopesScopeSessionRecordingsIndexController extends Control
 
   queryParams = [
     'search',
-    { times: { type: 'array' } },
+    'time',
     { users: { type: 'array' } },
     { scopes: { type: 'array' } },
     { targets: { type: 'array' } },
@@ -28,7 +28,7 @@ export default class ScopesScopeSessionRecordingsIndexController extends Control
   ];
 
   @tracked search = '';
-  @tracked times = [];
+  @tracked time = null;
   @tracked users = [];
   @tracked scopes = [];
   @tracked targets = [];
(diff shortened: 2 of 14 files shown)


ui/admin/app/templates/scopes/scope/session-recordings/index.hbs
```diff
@@ -15,118 +15,240 @@
...
-              <div
-                class='session-recording-list-cell-item target-column hds-font-weight-semibold'
+  <page.body class='search-filtering'>
+    {{#if @model.sessionRecordingsExist}}
+      <div class='search-filtering-toolbar'>
+        <Hds::SegmentedGroup data-test-session-recordings-bar as |S|>
+          <S.TextInput
+            @value={{this.search}}
+            @type='search'
+            placeholder={{t 'actions.search-session-recording'}}
```
lisbet-alvarez, line 25: non-blocking: just to clarify, is the reason why "Search recording ID" is the placeholder to let users know they can also search by id (since its not in the table) ?

cameronperera: That is correct!

addons/api/addon/services/indexed-db.js
```diff
@@ -29,6 +29,8 @@ export const modelIndexes = {
     '&id, attributes.created_time, attributes.type, attributes.name, attributes.description, attributes.scope.scope_id, attributes.plugin.name',
   'auth-method':
     '&id, attributes.created_time, attributes.type, attributes.name, attributes.description, attributes.is_primary, attributes.scope.scope_id',
+  'session-recording':
+    '&id, attributes.created_time, attributes.type, attributes.state, attributes.start_time, attributes.end_time, attributes.duration, attributes.scope.scope_id, attributes.create_time_values.user.id, attributes.create_time_values.user.name, attributes.create_time_values.target.id, attributes.create_time_values.target.name, attributes.create_time_values.target.scope.id, attributes.create_time_values.target.scope.name',
```
lisbet-alvarez, line 33: non-blocking: so basically for the scopes filter, we mainly using target scope because  we expect user scope id to be the same?

ui/admin/app/routes/scopes/scope/session-recordings/index.js
```diff
@@ -4,5 +4,160 @@
...
+   * Load all session recordings.
+   * @return {Promise<{ totalItems: number, sessionRecordings: [SessionRecordingModel], sessionRecordingsExist: boolean, storageBucketsExist: boolean }>}
+   */
+  async model({ search, times, users, scopes, targets, page, pageSize }) {
+    const scope = this.modelFor('scopes.scope');
+    const { id: scope_id } = scope;
+    let sessionRecordings;
+    let totalItems = 0;
+    let sessionRecordingsExist = false;
+    let storageBucketsExist = false;
```
lisbet-alvarez, line 59: in the other PR we changed the pattern to something like `dostorageBucketsExist`

ui/admin/app/routes/scopes/scope/session-recordings/index.js
```diff
@@ -4,5 +4,160 @@
...
+    const filters = {
+      scope_id: [{ equals: scope_id }],
+      created_time: [],
+      'create_time_values.user.id': [],
+      'create_time_values.target.scope.id': [],
+      'create_time_values.target.id': [],
+    };
+    times.forEach((time) => {
+      filters.created_time.push({ gte: new Date(time) });
```
lisbet-alvarez, line 68: non-blocking: out of curisoity, is conversion to Date object necessary?

ui/admin/app/routes/scopes/scope/session-recordings/index.js
```diff
@@ -4,5 +4,160 @@
...
+    targets.forEach((target) => {
+      filters['create_time_values.target.id'].push({ equals: target });
+    });
+
+    if (
+      this.can.can('list scope', scope, {
+        collection: 'session-recordings',
+      })
+    ) {
+      const queryOptions = {
```
ZedLi, line 85: Don't these requests need to be `recursive`? Session recordings inherit their scope from the storage bucket which can be more than just the global scope but we only ever access these from global right now.

ui/admin/app/routes/scopes/scope/session-recordings/index.js
```diff
@@ -4,5 +4,160 @@
...
+      sessionRecordings = await this.store.query(
+        'session-recording',
+        queryOptions,
+      );
+      totalItems = sessionRecordings.meta?.totalItems;
+      // Query all session reocordings for filtering values if entering route for the first time
+      if (!this.allSessionRecordings) {
+        await this.getAllSessionRecordings(scope_id);
+      }
+      sessionRecordingsExist = !!this.allSessionRecordings.length;
```
ZedLi: I don't think we've traditionally used `!!`, haven't we mainly used `Boolean`?

addons/api/addon/services/indexed-db.js
cameronperera: I believe that is partially true. If we used the scope of the user that may not be helpful. Especially since an admin user's scope could be global. So we use the scope tied to the target that was used to generate the session recording. There was even a comment in the designs about only showing the projects in the filter but I used our standard scope filter of showing the org and what projects are part of that org since that is how we already display the scope filter in the desktop client.

ui/admin/app/routes/scopes/scope/session-recordings/index.js
cameronperera: Right. I just saw that in your PR. I will update now.

ui/admin/app/routes/scopes/scope/session-recordings/index.js
cameronperera: I could have been doing something wrong with my dates and the format but I had to do this to get the filtering to work properly.

ui/admin/app/routes/scopes/scope/session-recordings/index.js
cameronperera: They do. Thanks for catching that.

ui/admin/app/routes/scopes/scope/session-recordings/index.js
cameronperera: Yep we have. Updated this as well.

ZedLi:
Can you try with 200k session recordings? The page won't load for me and I'm curious if you see the same behavior

ui/admin/app/templates/scopes/scope/session-recordings/index.hbs
ZedLi: Hmm is that needed? It makes me think it will only search by recording ID but it will search across all the fields you have indexed

cameronperera:
> Can you try with 200k session recordings? The page won't load for me and I'm curious if you see the same behavior

Fixed this issue and got it working. This was not a scale issue but filtering related. For reference, here is a quick video of filtering with 200k session recordings.

https://github.com/user-attachments/assets/6bb7c1c7-339a-4ec4-8164-baf2a388dcb3

ui/admin/app/templates/scopes/scope/session-recordings/index.hbs
cameronperera: That is true. Thinking on this, you might be right that this will make the user think they can only search by id. I can change it to the normal search placeholder.

ui/admin/app/routes/scopes/scope/session-recordings/index.js
ZedLi: Do you know if it's just a string or an actual date object in indexed db?

ZedLi: Can we actually also add some tests using dates to `indexed-db-query-test.js` to confirm this behavior? This is the first time we've used dates

ui/admin/app/controllers/scopes/scope/session-recordings/index.js
```diff
@@ -4,21 +4,189 @@
...
+    { times: { type: 'array' } },
+    { users: { type: 'array' } },
+    { scopes: { type: 'array' } },
+    { targets: { type: 'array' } },
+    'page',
+    'pageSize',
+  ];
+
+  @tracked search = '';
+  @tracked times = [];
```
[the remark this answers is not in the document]
cameronperera: Apologies, I didn't think to add context for this to the PR. If you look at the [design](https://www.figma.com/design/lvhT00hEVGZA59PBLlX991/Search-and-Filtering-V1-%2B-V2?node-id=3933-14227&node-type=frame&t=KXFTa75SUwRJ3V1e-0), the plan to is include a "from" and "to" option for filtering by time. For the sake of the release wanted to keep it to the today, last 3 days, and last 7 days options. With that in mind, I decided to keep it as an array for now.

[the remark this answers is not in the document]
ZedLi: Does it make sense to even be an array then in light of the design? If it has a `from` and `to` we wouldn't be using an array?

Also I noticed the design says it should have last 7 days selected by default, are we going to be doing that?

ui/admin/app/routes/scopes/scope/session-recordings/index.js
cameronperera: It is stored as a date object in indexed db. I will work on some tests as well.

ui/admin/app/controllers/scopes/scope/session-recordings/index.js
```diff
@@ -4,21 +4,189 @@
...
+
+  /**
+   * Looks up org by ID and returns the org name
+   * @param {string} orgID
+   * @returns {string}
+   */
+  @action
+  orgName(orgID) {
+    const org = this.store.peekRecord('scope', orgID);
+    return org.name;
```
lisbet-alvarez, line 125: Should this return `org.displayName` instead ?

ui/admin/app/routes/scopes/scope/session-recordings/index.js
```diff
@@ -4,5 +4,156 @@
...
+        page,
+        pageSize,
+      };
+
+      sessionRecordings = await this.store.query(
+        'session-recording',
+        queryOptions,
+      );
+      totalItems = sessionRecordings.meta?.totalItems;
+      // Query all session reocordings for filtering values if entering route for the first time
```
lisbet-alvarez, line 97:
```suggestion
      // Query all session recordings for filtering values if entering route for the first time
```

ui/admin/app/routes/scopes/scope/session-recordings/index.js
```diff
@@ -4,5 +4,156 @@
...
+        sessionRecordings,
+        doSessionRecordingsExist: doSessionRecordingsExist,
+        allSessionRecordings: this.allSessionRecordings,
+        totalItems,
+        doStorageBucketsExist: doStorageBucketsExist,
+      };
+    }
+  }
+
+  async getAllSessionRecordings(scope_id) {
```
lisbet-alvarez, line 114: small nitpick: this method is missing a comment

ui/admin/app/controllers/scopes/scope/session-recordings/index.js
```diff
@@ -4,21 +4,189 @@
...
+   * Returns array of time options for time filter
+   * @returns {[object]}
+   */
+  get timeOptions() {
+    const now = new Date();
+    const midnight = new Date(now.getFullYear(), now.getMonth(), now.getDate());
+    const threeDaysPast = new Date(midnight);
+    threeDaysPast.setDate(midnight.getDate() - 3);
+    const sevenDaysPast = new Date(midnight);
+    sevenDaysPast.setDate(midnight.getDate() - 7);
```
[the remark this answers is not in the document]
cameronperera: Correct. I took today to mean just that calendar day. If we changed it to "Last 24 hours" I would change the text as well. I can chat with product and/or design in the morning to see if they have a preference.

ui/admin/app/controllers/scopes/scope/session-recordings/index.js
cameronperera: It should return `org.displayName`. Thanks

ui/admin/app/routes/scopes/scope/session-recordings/index.js
cameronperera: Fixed.

ui/admin/app/routes/scopes/scope/session-recordings/index.js
cameronperera: Fixed.

ui/admin/app/controllers/scopes/scope/session-recordings/index.js
cameronperera: Fixed.

ui/admin/app/controllers/scopes/scope/session-recordings/index.js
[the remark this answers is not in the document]
cameronperera: Good points. I will play around with changing this up and setting a default of 7 days.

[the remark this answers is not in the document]
cameronperera: I updated the filter to no longer be an array. I will talk with design about the default time shortcut as well. In the design doc, it says if the user were to clear out the from and to options, we should default back to the 7 day shortcut. However, we may not want to do that if we don't include the from and to options right now.

lisbet-alvarez (approved):
Thx for all this work!!

ZedLi (approved):
Looks good!

PR #2504, opened 2024-09-27
appupdater cli check

calcaide:
# Description

[ICU-15109](https://hashicorp.atlassian.net/browse/ICU-15109)

Add check on the appUpdater to skip it if the Desktop client is NOT using the built in CLI.

**IMPORTANT**
The first approach I tried was to use node [fs.lstatSync](https://nodejs.org/api/fs.html#fslstatsyncpath-options). The issue I encounter is that the lstat object returned, in POSIX systems, the key `file` always `false` value.
This happens because on POSIX system, the binary name we return is `boundary` without file extension, therefore the lstat object mark it as directory, whatever we were passing `boundary` or `electron-app/cli/boundary`.

Decided to rely on [parsing the path](https://nodejs.org/api/path.html#pathparsepath). If the parsed object returns values for the keys `dir` and `root`, means it is indeed the CLI built in the desktop client.

Happy to discuss other approaches.

## How to Test

For Mac:

Use the built in CLI:
- Set `SETUP_CLI=true` and run `yarn start:desktop`.
- The desktop client should prompt the app updater.

Use system CLI:
- Set `SETUP_CLI=false` and run `yarn start:desktop`.
- The desktop client should NOT prompt the app updater.

For Windows:

Notice in windows, we have the appUpdater disabled, so it will not be possible to test without code modifications.

Code modifications:
- Above [this line](https://github.com/hashicorp/boundary-ui/blob/feat-ICU-15109-appupdater-cli-check/ui/desktop/electron-app/src/helpers/app-updater.js#L177) add a `console.log(`isBuiltInCli: `, boundaryCli.isBuiltInCli())`.

Test procedure:
- Set `SETUP_CLI=true` and run `yarn start:desktop`. The `console.log` should show TRUE.
- Set `SETUP_CLI=false` and run `yarn start:desktop`. The `console.log` should show FALSE.

[ICU-15109]: https://hashicorp.atlassian.net/browse/ICU-15109?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/ui/desktop/electron-app/src/cli/index.js
+++ b/ui/desktop/electron-app/src/cli/index.js
@@ -4,7 +4,7 @@
  */
 
 const { spawnSync } = require('../helpers/spawn-promise.js');
-const { path } = require('./path.js');
+const { path, isBuiltInCli } = require('./path.js');
 const pathPosix = require('path');
 const pathWin = require('path/win32');
 const { isWindows } = require('../helpers/platform.js');
@@ -13,18 +13,7 @@ module.exports = {
   // Check boundary cli existence
   exists: () => Boolean(path()),
 
-  // Return true if the CLI in usage is the built in. False if relies on system CLI.
-  isBuiltInCli: () => {
-    const pathToParse = path();
-    const pathParsed = isWindows()
-      ? pathWin.parse(pathToParse)
-      : pathPosix.parse(pathToParse);
-
-    if (pathParsed.dir === '' && pathParsed.root === '') {
-      return false;
-    }
-    return true;
-  },
+  isBuiltInCli,
 
   // Returns JSON-formatted version information from the CLI
   version: () => {
--- a/ui/desktop/electron-app/src/cli/index.js
+++ b/ui/desktop/electron-app/src/cli/index.js
@@ -5,9 +5,6 @@
 
 const { spawnSync } = require('../helpers/spawn-promise.js');
 const { path, isBuiltInCli } = require('./path.js');
-const pathPosix = require('path');
-const pathWin = require('path/win32');
-const { isWindows } = require('../helpers/platform.js');
 
 module.exports = {
   // Check boundary cli existence
--- a/ui/desktop/electron-app/src/cli/index.js
+++ b/ui/desktop/electron-app/src/cli/index.js
@@ -5,10 +5,26 @@
 
 const { spawnSync } = require('../helpers/spawn-promise.js');
 const { path } = require('./path.js');
+const pathPosix = require('path');
+const pathWin = require('path/win32');
+const { isWindows } = require('../helpers/platform.js');
 
 module.exports = {
   // Check boundary cli existence
   exists: () => Boolean(path()),
+
+  // Return true if the CLI in usage is the built in. False if relies on system CLI.
+  isBuiltInCli: () => {
+    const pathToParse = path();
+    const pathParsed = isWindows()
+      ? pathWin.parse(pathToParse)
+      : pathPosix.parse(pathToParse);
+
+    if (pathParsed.dir === '' && pathParsed.root === '') {
+      return false;
+    }
+    return true;
+  },
 
   // Returns JSON-formatted version information from the CLI
   version: () => {
--- a/ui/desktop/electron-app/src/cli/index.js
+++ b/ui/desktop/electron-app/src/cli/index.js
@@ -8,7 +8,7 @@ const { path, isBuiltInCli } = require('./path.js');
 
 module.exports = {
   // Check boundary cli existence
-  exists: () => Boolean(path()),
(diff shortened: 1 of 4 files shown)


ui/desktop/electron-app/src/cli/index.js
```diff
@@ -5,11 +5,27 @@
...
+const pathWin = require('path/win32');
+const { isWindows } = require('../helpers/platform.js');
 
 module.exports = {
   // Check boundary cli existence
   exists: () => Boolean(path()),
 
+  // Return true if the CLI in usage is the built in. False if relies on system CLI.
+  isBuiltInCli: () => {
+    const pathToParse = path();
```
ZedLi, line 18: Do we need to use the path from `path()`?

Couldn't we have pulled out the logic for `isBuiltInCli` to another method and used that?

```javascript
    const builtInCliPath = isDev
      ? path.resolve(__dirname, '..', '..', 'cli', binaryName)
      : path.resolve(process.resourcesPath, 'cli', binaryName);
    const isBuiltInCli = existsSync(builtInCliPath);
```

calcaide: @ZedLi 

If understand correctly, you suggest to move the `isBuiltInCli` to the `path.js` file? 
Initially I though of extracting to a function the current logic under `path()`, and re-use it for method path and isBuiltInCli, since is the same logic.
But discarted since that logic lives in `path.js` and not in `index.js`, where the current `exists()` method lives.

Let me know if you mean to move the logic to `path.js` and I will proceed, or feel free to ping me and clarify through zoom.

Thank you

ZedLi: Yeah basically, I didn't realize the names are the same but I was specifically talking about the logic of `isBuiltInCli` from `path()`.

I'm not sure we need the actual path itself as much as the logic that checks if we have our own CLI, which means we could then skip doing the parsing of the path after

calcaide: 100%. Yeah Indeed, I will proceed to move `isBuiltInCli` to `path.js` so we can share the logic.

This, despite is an easy check, I had doubts where will be best approach and best place to make it happend. 

Iterations will provide the better answer 😃

calcaide:
@ZedLi The PR has been updated with the suggested [changes within your feedback](https://github.com/hashicorp/boundary-ui/pull/2504#discussion_r1781861505) , let me know your thoughts.

Thank you

ui/desktop/electron-app/src/cli/path.js
```diff
@@ -8,18 +8,21 @@ const { isWindows } = require('../helpers/platform.js');
...
   path: () => {
-    const binaryName = isWindows() ? 'boundary.exe' : 'boundary';
-    const builtInCliPath = isDev
-      ? path.resolve(__dirname, '..', '..', 'cli', binaryName)
-      : path.resolve(process.resourcesPath, 'cli', binaryName);
-    const isBuiltInCli = existsSync(builtInCliPath);
-
     return isBuiltInCli ? builtInCliPath : binaryName;
   },
+  isBuiltInCli,
```
ZedLi, line 27: This will only be called once I believe when this file is `require`d which I kind of like since it doesn't need to keep doing the check each time this is called but is there any concern we would ever need to re-calculate this during runtime? I'm thinking no but just wanted to make sure we're doing this intentionally. 

That also brings up should `path()` still be a function as well

calcaide: @ZedLi 

That is correct, this is intentionally. When working on this, I realize path() was call (therefore re-calculate) several ties just to execute DC. I found that unnecessary and wanted isBuiltInCli to not follow that approach.

I don't find any concern following this approach when it comes to CLI checks, let me explain my thinking here, perhaps I missed something.

- In case DC includes CLI: re-check this is unncessary, since the CLI will always live in the same place. If any day we want to change CLI localtion, we will release a new DC, so no problem here.
- In case DC does not include CLI, we rely on user's path: all users will use the instaler, so it is safe to assume most users will not be changing CLI location manually.
If an advanced user decides to change CLI location, I assume, it will manually update boundary location within user path and "restart" boundary, boundary client agent and DC client.

That is correct, as per the thinking explained above, I do strongly believe path() should be just a const and not a function.
I wanted to wait on the windows bug resolution before introducing such change, but after seeing Dharini's work, we will need to resolve conflicts anyway, so adding it within this PR.
The path change, has been successfully tested manually and using the DC e2e tests.

Let me know if you think otherwise, happy to discuss other approaches.

ZedLi (approved):
Looks good!

PR #2508, opened 2024-10-02
Fix windows no cli crash

DhariniJeeva:
# Description
[ticket](https://hashicorp.atlassian.net/browse/ICU-15235) 

This PR fixes the app crashing on launch in windows

## How to Test

- pull this branch to a Windows machine/parallels/ec2 instance
- install CLI locally using [this](https://github.com/hashicorp/boundary-installer/actions/runs/10966402384) click on the Windows build and follow the instructions to install
- delete `CLI` folder from `ui/desktop/electron-app` to make sure we don't package cli in our DC build 
- build it locally `yarn build` within ui/desktop
- go to the directory where the out/boundary binary is stored and click on boundary.exe app 
- the app should run as expected and not crash 

note: Plz reach out to me if you need a cluster for testing

## Checklist

### DIFF
--- a/ui/desktop/electron-app/src/cli/path.js
+++ b/ui/desktop/electron-app/src/cli/path.js
@@ -7,6 +7,8 @@ const path = require('path');
 const { isWindows } = require('../helpers/platform.js');
 const isDev = require('electron-is-dev');
 const { existsSync } = require('node:fs');
+const which = require('which');
+const log = require('electron-log/main');
 
 const binaryName = isWindows() ? 'boundary.exe' : 'boundary';
 const builtInCliPath = isDev
@@ -19,8 +21,17 @@ const isBuiltInCli = existsSync(builtInCliPath);
 /**
  * Returns Boundary CLI path if the CLI is built in or the Boundary binary name
  * if not, so we assume boundary is available within user $PATH.
+ * For Windows, we filter out the path that is within the current working directory.
  */
-const pathBoundary = isBuiltInCli ? builtInCliPath : binaryName;
+const pathBoundary = isBuiltInCli
+  ? builtInCliPath
+  : isWindows()
+    ? which.sync(binaryName, { nothrow: true, all: true })
+      .filter((binary) => !binary.startsWith(process.cwd()))[0]
+    : binaryName;
+
+log.info('pathBoundary', pathBoundary)
+console.log(pathBoundary)
 
 module.exports = {
   path: pathBoundary,
--- a/ui/desktop/electron-app/src/cli/path.js
+++ b/ui/desktop/electron-app/src/cli/path.js
@@ -20,11 +20,14 @@ const isBuiltInCli = existsSync(builtInCliPath);
 /**
  * Returns Boundary CLI path if the CLI is built in or the Boundary binary name
  * if not, so we assume boundary is available within user $PATH.
- * For Windows, we filter out the path that is within the current working directory.
  */
 const pathBoundary = isBuiltInCli
   ? builtInCliPath
-  : isWindows()
+  : // On Windows, we need to filter out the path that is within the current working directory.
+  // This is necessary because the spawn methods may not always find the correct binary path.
+  // The `which` module returns an array of all available paths for the 'boundary' binary.
+  // We then filter out the path that is within the current working directory.
+  isWindows()
   ? which
       .sync(binaryName, { nothrow: true, all: true })
       .filter((binary) => !binary.startsWith(process.cwd()))[0]
--- a/ui/desktop/electron-app/src/cli/path.js
+++ b/ui/desktop/electron-app/src/cli/path.js
@@ -8,7 +8,6 @@ const { isWindows } = require('../helpers/platform.js');
 const isDev = require('electron-is-dev');
 const { existsSync } = require('node:fs');
 const which = require('which');
-const log = require('electron-log/main');
 
 const binaryName = isWindows() ? 'boundary.exe' : 'boundary';
 const builtInCliPath = isDev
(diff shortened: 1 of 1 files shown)


ui/desktop/electron-app/src/cli/path.js
```diff
@@ -20,6 +21,15 @@ module.exports = {
       : path.resolve(process.resourcesPath, 'cli', binaryName);
     const isBuiltInCli = existsSync(builtInCliPath);
 
-    return isBuiltInCli ? builtInCliPath : binaryName;
+    // Filter out the current working directory from the list of binaries as spawn method picks up incorrect binary path
+    const pathToRemove = isWindows()
+      ? path.join(process.cwd(), binaryName).replace(/\\/g, '\\\\')
+      : process.cwd();
+    const binaries = which.sync(binaryName, { nothrow: true, all: true });
```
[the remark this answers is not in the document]
ZedLi: Actually I think it's fine, it's probably better to not have to change the existing callers to use await

ui/desktop/electron-app/src/cli/path.js
```diff
@@ -20,6 +21,15 @@ module.exports = {
       : path.resolve(process.resourcesPath, 'cli', binaryName);
     const isBuiltInCli = existsSync(builtInCliPath);
 
-    return isBuiltInCli ? builtInCliPath : binaryName;
+    // Filter out the current working directory from the list of binaries as spawn method picks up incorrect binary path
+    const pathToRemove = isWindows()
+      ? path.join(process.cwd(), binaryName).replace(/\\/g, '\\\\')
```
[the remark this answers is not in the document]
DhariniJeeva: ah yes, that works here, updated ty!

calcaide (approved):
Tested and works like a charm!!

I could connect to 2 different clusters no problem :)

P.S: I have issues with the installer not adding to the user's path boundary, at this point I am thinking in my environment, so rather than using the installer, I download 0.17.0 from our release channel, install it and add it manually to the users Path. 

P.S.2: I am working on [this PR](https://github.com/hashicorp/boundary-ui/pull/2504) which will change a bit the `path.js` file. But we can work together if there are conflicts, let me know if you encounter any issue/conflicts

ui/desktop/electron-app/src/cli/path.js
```diff
@@ -20,6 +21,16 @@ module.exports = {
       : path.resolve(process.resourcesPath, 'cli', binaryName);
     const isBuiltInCli = existsSync(builtInCliPath);
```
ZedLi, line 22: I think we can check for `isBuiltInCli` to return `builtInCliPath` immediately here so that way we don't have to do two ternary statements after

ui/desktop/electron-app/src/cli/path.js
```diff
@@ -20,6 +21,16 @@ module.exports = {
...
 
+    // Filter out the current working directory from the list of binaries as spawn method picks up incorrect binary path
+    if (isWindows()) {
+      const binaries = which.sync(binaryName, { nothrow: true, all: true });
+
+      const filteredBinaryPath = binaries.filter(
+        (binary) => !binary.startsWith(process.cwd()),
```
ZedLi, line 29: I just noticed there's actually 3 results returned by `which`, do you also see a `boundary.exe` returned? Is that something we need to filter out as well?

ui/desktop/electron-app/src/cli/path.js
```diff
@@ -20,6 +21,16 @@ module.exports = {
...
+      );
+      const resolvedBinaryPath = path.resolve(filteredBinaryPath[0]);
```
ZedLi, line 31: What does `path.resolve` do here? Isn't the `filteredBinaryPath[0]` already an absolute path?

ui/desktop/electron-app/src/cli/path.js
DhariniJeeva: Hmm, I tested this a few times and only saw two results from `which`, any idea where we get this `boundary.exe` from?

ZedLi: Hmm maybe it's just my env? I'm not exactly sure since it seemed like a relative path which I assumed was the DC itself. Either way it was the last one in the list so it wasn't getting picked so I'm not sure if it's an issue?

Would be good if someone else could confirm

ui/desktop/electron-app/src/cli/path.js
DhariniJeeva: you are right, we don't need that.

ui/desktop/electron-app/src/cli/path.js
```diff
@@ -19,8 +20,15 @@ const isBuiltInCli = existsSync(builtInCliPath);
...
  * if not, so we assume boundary is available within user $PATH.
+ * For Windows, we filter out the path that is within the current working directory.
  */
-const pathBoundary = isBuiltInCli ? builtInCliPath : binaryName;
+const pathBoundary = isBuiltInCli
+  ? builtInCliPath
+  : isWindows()
+  ? which
+      .sync(binaryName, { nothrow: true, all: true })
+      .filter((binary) => !binary.startsWith(process.cwd()))[0]
```
lisbet-alvarez, line 30: question for understanding: what is line 30 doing, not sure if i understand the condition for filtering out?

lisbet-alvarez (approved):
Tested in parallels & the app opens without crashing!! Thank you!

ui/desktop/electron-app/src/cli/path.js
DhariniJeeva: so we want to filter out the binary path that is in our current directory as that is not correct and we only need the path from the CLI that is installed if there's one packaged within the build. 

`which` command returns all available paths of `boundary` in array format.  

I will explain in the comment a bit more since it is not clear

lisbet-alvarez: That helps explain it thank you!

ZedLi (approved):
Looks good!

PR #2510, opened 2024-10-03
fix: 🐛 deleting session recording causes error

lisbet-alvarez:
✅ Closes: https://hashicorp.atlassian.net/browse/ICU-15236

# Description

Fixes transition aborted errors when re-applying and deleteing.

https://github.com/user-attachments/assets/0344df34-0ba2-488c-b2b3-1bec6e4fb5d9

## How to Test

1. create a session recording
2. create storage policy to allow deletion and apply in "Global settings"
3. Reapply policy to session recording
4. Delete session recording.
## Checklist

### DIFF
--- a/addons/core/translations/resources/en-us.yaml
+++ b/addons/core/translations/resources/en-us.yaml
@@ -454,7 +454,6 @@ storage-bucket:
       help_org: This storage bucket will be available for use with targets within this org.
     provider:
       label: Provider
-      help: Boundary currently only supports Amazon S3 storage.
     endpoint_url:
       label: Endpoint URL
       help: Cannot be changed later.
--- a/addons/core/translations/resources/en-us.yaml
+++ b/addons/core/translations/resources/en-us.yaml
@@ -706,7 +706,7 @@ target:
       help: The local proxy port on which to listen by default when a session is started on a client.
     sidebar:
       label: Session Recording
-      help: Record SSH sessions on this target and save them to an Amazon S3 bucket.
+      help: Record SSH sessions on this target and save them to a storage bucket.
       link: Session Recording settings
   questions:
     delete-host-sources:
--- a/ui/admin/app/routes/scopes/scope/session-recordings/index.js
+++ b/ui/admin/app/routes/scopes/scope/session-recordings/index.js
@@ -96,7 +96,10 @@ export default class ScopesScopeSessionRecordingsIndexRoute extends Route {
       if (!this.allSessionRecordings) {
         await this.getAllSessionRecordings(scope_id);
       }
-      doSessionRecordingsExist = Boolean(this.allSessionRecordings.length);
+      doSessionRecordingsExist = await this.getDoSessionRecordingsExist(
+        scope_id,
+        totalItems,
+      );
       doStorageBucketsExist = await this.getDoStorageBucketsExist(scope_id);
 
       return {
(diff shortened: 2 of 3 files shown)


ui/admin/app/routes/scopes/scope/session-recordings/index.js
```diff
@@ -125,6 +128,31 @@ export default class ScopesScopeSessionRecordingsIndexRoute extends Route {
     );
   }
 
+  /**
+   * Sets doSessionRecordingsExist to true if there are any session recordings.
+   * @param {string} scope_id
+   * @param {number} totalItems
+   * @returns {Promise<boolean>}
+   */
+  async getDoSessionRecordingsExist(scope_id, totalItems) {
```
ZedLi, line 137: What was the purpose again of using this method? It looks like we still have to make the request for all sessions regardless

addons/core/translations/resources/en-us.yaml
```diff
@@ -706,7 +706,7 @@ target:
       help: The local proxy port on which to listen by default when a session is started on a client.
     sidebar:
       label: Session Recording
-      help: Record SSH sessions on this target and save them to an Amazon S3 bucket.
+      help: Record SSH sessions on this target and save them to a storage bucket.
```
ZedLi, line 709: Good catch! I wonder if there's any other ones in the codebase?

I found this [one](https://github.com/hashicorp/boundary-ui/blob/c4e58fb780405eba2ee5f2e8f71268bcc0e24c40/addons/core/translations/resources/en-us.yaml#L450) but I don't think it's used?

ZedLi (approved):
Looks great, thanks for the bug fix!

ui/admin/app/routes/scopes/scope/session-recordings/index.js
lisbet-alvarez: Good question, the query for all session recordings only happens on initial load or refresh. There was a bug where if on initial load if there were no session-recordings then on a subsequent load when I do have session-recordings they would not show up.

addons/core/translations/resources/en-us.yaml
lisbet-alvarez: good catch! the help translation was def. not used anywhere.

ui/admin/app/routes/scopes/scope/session-recordings/index.js
```diff
@@ -125,6 +128,31 @@ export default class ScopesScopeSessionRecordingsIndexRoute extends Route {
...
+    if (totalItems > 0) {
+      return true;
+    }
+    const options = { pushToStore: false, peekIndexedDB: true };
+    const sessionRecordings = await this.store.query(
+      'session-recording',
+      {
+        scope_id,
+        query: { filters: { scope_id: [{ equals: scope_id }] } },
```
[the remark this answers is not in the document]
lisbet-alvarez: AH okay, I didn't realize that, thank you for explaining it.

ui/admin/app/routes/scopes/scope/session-recordings/index.js
cameronperera: Could we fix that bug by checking the length of the response in the `getAllSessionRecordings` method? If the length is 0, we set `this.allSessionRecordings = null`?  That would fix the bug and remove the need for this check as well.

lisbet-alvarez: Then there could be the case where length was >0 but the session recordings were deleted and user navigates back to the list page and it'll show the filters and search box but no table. I'll also test this out to make sure.

cameronperera (approved):
Thanks for fixing these few things

ZedLi (approved):

PR #2511, opened 2024-10-07
Add worker type dynamic host catalog

DhariniJeeva:
# Description
[jira](https://hashicorp.atlassian.net/browse/ICU-15257)

API changes:
Screenshot 2024-10-07 at 3 11 44 PM

UI changes: 
before: 
Screenshot 2024-10-07 at 3 13 00 PM

after:
 
Screenshot 2024-10-07 at 4 13 29 PM

error:

Screenshot 2024-10-07 at 3 41 32 PM

## How to Test

1.  use staged build 
2. pull this branch down and click on new host catalog
3. you should be able to add worker filter for AWS type host catalog [reach out to me if you need creds for AWS]
4. you should not see this option for static or Azure type host catalog 
5. you should be able to update/delete worker filter once created and save successfully 
## Checklist

### DIFF
--- a/addons/api/addon/generated/models/host-catalog.js
+++ b/addons/api/addon/generated/models/host-catalog.js
@@ -66,6 +66,13 @@ export default class GeneratedHostCatalogModel extends BaseModel {
   // AWS specific
 
   @attr('string', {
+    for: 'aws',
+    description:
+      'An expression used to filter the workers that have network access to a service that is hosting the external object store.',
+  })
+  worker_filter;
+
+  @attr('string', {
     isNestedAttribute: true,
     description: '',
   })
--- a/addons/api/addon/generated/models/host-catalog.js
+++ b/addons/api/addon/generated/models/host-catalog.js
@@ -66,7 +66,7 @@ export default class GeneratedHostCatalogModel extends BaseModel {
   // AWS specific
 
   @attr('string', {
-    for: 'aws',
+    for: 'plugin',
     description:
       'An expression used to filter the workers that have network access to a service that is hosting the external object store.',
   })
--- a/addons/api/addon/serializers/host-catalog.js
+++ b/addons/api/addon/serializers/host-catalog.js
@@ -11,6 +11,7 @@ const fieldsByType = {
     'region',
     'access_key_id',
     'secret_access_key',
+    'worker_filter',
   ],
   azure: [
     'disable_credential_rotation',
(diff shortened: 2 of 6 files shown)


madeleinetle:
Looks good! Small nitpick (and not sure if you've gotten to this yet), but the "Learn more" text should lead to this link: https://developer.hashicorp.com/boundary/docs/concepts/filtering/worker-tags

DhariniJeeva:
> Looks good! Small nitpick (and not sure if you've gotten to this yet), but the "Learn more" text should lead to this link: https://developer.hashicorp.com/boundary/docs/concepts/filtering/worker-tags

thanks for the link, I originally thought we didn't have it and added a TODO for that, but looks like we already use the link in our repo. Learn more should now point to the docs link.

addons/core/translations/resources/en-us.yaml
```diff
@@ -225,6 +225,9 @@ host-catalog:
     azure_client_id:
       label: Client/Application ID
       help: The client (application) ID of an Azure service principal that Boundary will use to authenticate and discover hosts from Azure.
+    worker_filter:
+      label: Worker Filter
+      help: Insert a worker filter in order to route requests to the correct worker. Learn more.
```
lisbet-alvarez, line 230: `Learn more` text appears twice. I think this one can be removed.
```suggestion
      help: Insert a worker filter in order to route requests to the correct worker.
```

lisbet-alvarez (approved):
left a small suggestion, otherwise works great!

addons/api/addon/serializers/host-catalog.js
```diff
@@ -38,19 +39,32 @@ export default class HostCatalogSerializer extends ApplicationSerializer {
...
-    // This deletes any fields that don't belong to the record type
+
+    // Delete any fields that don't belong to the compositeType
+    if (
+      isPlugin &&
+      json &&
+      options?.for &&
+      !options.for.includes(compositeType)
+    ) {
+      if (!fieldsByType[compositeType].includes(key)) {
```
cameronperera, line 50: non-blocking: Could this if statement be included with the if statement above it?

cameronperera (approved):
Left a small non-blocking comment

addons/core/translations/resources/en-us.yaml
DhariniJeeva: ah thanks for the catch, I totally missed that

addons/api/addon/serializers/host-catalog.js
DhariniJeeva: ah yes we can, updated!

addons/api/addon/generated/models/host-catalog.js
```diff
@@ -65,6 +65,13 @@ export default class GeneratedHostCatalogModel extends BaseModel {
 
   // AWS specific
 
+  @attr('string', {
+    for: 'plugin',
```
ZedLi, line 69: What's the point of this `for` since we have to do a custom serializer to handle the `worker_filter` for azure plugins?

addons/api/addon/serializers/host-catalog.js
```diff
@@ -38,19 +39,31 @@ export default class HostCatalogSerializer extends ApplicationSerializer {
...
     const { options } = attribute;
-    // Nested conditionals for readability
-    // This deletes any fields that don't belong to the record type
+
+    // Delete any fields that don't belong to the compositeType
+    if (
+      isPlugin &&
+      json &&
+      options?.for &&
+      !options.for.includes(compositeType) &&
```
ZedLi, line 48: Does this do anything? Since the composite type will be `azure` or `aws`, not `plugin`.
It seems to me that it should have been fine to just add it to the `fieldsByType` and that would be it?

We could then add the usage of `compositeType` to be handled by `for` in the application serializer as part of a refactor in the future to remove the need for `fieldsByType`

PR #2518, opened 2024-10-10
fix: 🐛 Push associated targets in sessions to ember store

ZedLi:
# Description
Fixes the bug here described in this slack [thread](https://hashicorp.slack.com/archives/C05EHAH84HE/p1728507679515329).

As users can now start sessions with transparent sessions, we aren't guaranteed that they have navigated to the target first in the DC so it may not yet be in the store.

Before:
image

After:
image

## How to Test
Use the installer to install the client agent. Make sure you're using a cluster that has enough targets where it won't be able to load all targets on one page. (Use my [int cluster](https://27425825-515c-4d8c-beb6-aa91cbc20c5a.boundary.hcp.to) if you'd like). 

Connect to a generic TCP target (make sure it wasn't loaded on the first page) transparently and go to sessions (use www.hashicorp.com in your browser if using my int cluster). It should have the target name displayed. 

## Checklist

### DIFF
--- a/ui/desktop/app/routes/scopes/scope/projects/sessions/index.js
+++ b/ui/desktop/app/routes/scopes/scope/projects/sessions/index.js
@@ -136,27 +136,21 @@ export default class ScopesScopeProjectsSessionsIndexRoute extends Route {
     const chunkedTargetIds = chunk(uniqueTargetIds, 50);
 
     const associatedTargetsPromises = chunkedTargetIds.map((targetIds) =>
-      this.store.query(
-        'target',
-        {
-          scope_id: orgScope.id,
-          recursive: true,
-          force_refresh: true,
-          query: {
-            filters: {
-              id: {
-                logicalOperator: 'or',
-                values: targetIds.map((targetId) => ({
-                  equals: targetId,
-                })),
-              },
+      this.store.query('target', {
+        scope_id: orgScope.id,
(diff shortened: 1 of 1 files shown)


ui/desktop/app/routes/scopes/scope/projects/sessions/index.js
```diff
@@ -136,27 +136,21 @@ export default class ScopesScopeProjectsSessionsIndexRoute extends Route {
...
+            id: {
+              logicalOperator: 'or',
+              values: targetIds.map((targetId) => ({
+                equals: targetId,
+              })),
             },
           },
         },
-        {
-          pushToStore: false,
```
ZedLi, line 157: I made the decision to just push all the targets from this request to the store. It _should_ be fine since it's limited to the user's sessions so there shouldn't be that many.

We could alternatively just not push to the store for this request still, and then make a separate request to indexed db for the current page's sessions and just push those targets to the store but I'm assuming for the majority of cases that's just an extra request.

cameronperera: Yeah I think this is a safe solution. Likely a single user isn't going to have hundreds of thousands of sessions that still exist so the performance shouldn't be hurt.

cameronperera (approved):

PR #2520, opened 2024-10-14
Refactor a11yAudit for auth-method list test

cameronperera:
# Description

This PR introduces the new changes to our a11y auditing in our acceptance tests. I have also added some documentation for future reference. Please make any recommendations if the README entry isn't clear or complete.

I kept this PR small for easy review of this new change. In the future, refactoring the whole CRUDL of a specific resource should be fine.

## Checklist


ui/admin/tests/acceptance/auth-methods/list-test.js
```diff
@@ -69,13 +71,75 @@ module('Acceptance | auth-methods | list', function (hooks) {
...
+
+      await visit(urls.globalAuthMethods);
+      await a11yAudit();
+
+      currentSession().set('data.theme', 'dark');
+      await a11yAudit();
+    });
   });
 
   test('users can navigate to auth methods with proper authorization', async function (assert) {
```
calcaide, line 145: None blocking comment: the group of tests which do not belong to the a11Audit aren't group by a `module` statement. Is this in purpose?
Following the pattern we have in the acceptance test, i.e: [users list-test](https://github.com/hashicorp/boundary-ui/blob/main/ui/admin/tests/acceptance/users/list-test.js), [auth-methods delete-test](https://github.com/hashicorp/boundary-ui/blob/main/ui/admin/tests/acceptance/auth-methods/delete-test.js#L21), we wrap them in a module, such as `'Acceptance | auth-methods | list'`.

calcaide (approved):
Left a none blocking comment.

Thanks for the work!! 🙌

ui/admin/tests/acceptance/auth-methods/list-test.js
cameronperera: Thanks for pointing this out as looking at this is confusing. All the tests in this file are wrapped in a module named `'Acceptance | auth-methods | list'`. What I did was add another module within that just for the `a11yAudit` tests. The main reason for this is we get access to all of the work done in the `beforeEach`. I wonder if maybe I should move the whole `a11yAudit` module to the bottom of the parent test module to make this less confusing?

ZedLi (approved):
Looks to me like a very reasonable refactor for getting focused tests on a11y audits. Thanks for looking into this!

cameronperera:
Closing due to new and improved a11y changes coming soon!

PR #2523, opened 2024-10-17
Update credentials

DhariniJeeva:
# Description

allow credential update after creation 
https://hashicorp.atlassian.net/browse/ICU-15299

## 

https://github.com/user-attachments/assets/39dff852-ded3-495a-959f-12ee5f3d75db

https://github.com/user-attachments/assets/e0c26b3b-52e5-4696-b037-98da405c9abe

How to Test

## Checklist

### DIFF
--- a/addons/api/addon/generated/models/host-catalog.js
+++ b/addons/api/addon/generated/models/host-catalog.js
@@ -97,12 +97,14 @@ export default class GeneratedHostCatalogModel extends BaseModel {
 
   // AWS dynamic credentials
   @attr('string', {
+    for: 'plugin',
     isNestedAttribute: true,
     description: 'The role ARN to use.',
   })
   role_arn;
 
   @attr('string', {
+    for: 'plugin',
     isNestedAttribute: true,
     description: 'The role external ID to use.',
   })
--- a/addons/api/addon/generated/models/host-catalog.js
+++ b/addons/api/addon/generated/models/host-catalog.js
@@ -58,6 +58,7 @@ export default class GeneratedHostCatalogModel extends BaseModel {
   // AWS & Azure
 
   @attr('boolean', {
+    for: 'plugin',
     isNestedAttribute: true,
     description: '',
   })
@@ -66,12 +67,14 @@ export default class GeneratedHostCatalogModel extends BaseModel {
   // AWS specific
 
   @attr('string', {
+    for: 'plugin',
     description:
       'An expression used to filter the workers that have network access to a service that is hosting the external object store.',
   })
   worker_filter;
 
   @attr('string', {
+    for: 'plugin',
     isNestedAttribute: true,
     description: '',
   })
@@ -79,12 +82,14 @@ export default class GeneratedHostCatalogModel extends BaseModel {
 
   // AWS static credentials
   @attr('string', {
+    for: 'plugin',
     isNestedSecret: true,
     description: '',
   })
   access_key_id;
 
   @attr('string', {
+    for: 'plugin',
     isNestedSecret: true,
     description: '',
   })
@@ -104,12 +109,14 @@ export default class GeneratedHostCatalogModel extends BaseModel {
   role_external_id;
 
   @attr('string', {
+    for: 'plugin',
     isNestedAttribute: true,
     description: 'The role session to use.',
   })
   role_session_name;
 
   @attr('object-as-array', {
+    for: 'plugin',
     isNestedAttribute: true,
     description: 'The role tags to use.',
   })
@@ -118,30 +125,35 @@ export default class GeneratedHostCatalogModel extends BaseModel {
   // Azure specific
 
   @attr('string', {
+    for: 'plugin',
     isNestedAttribute: true,
     description: '',
   })
   tenant_id;
 
   @attr('string', {
+    for: 'plugin',
     isNestedAttribute: true,
     description: '',
   })
   client_id;
 
   @attr('string', {
+    for: 'plugin',
     isNestedAttribute: true,
     description: '',
   })
   subscription_id;
 
   @attr('string', {
+    for: 'plugin',
     isNestedSecret: true,
     description: '',
   })
   secret_id;
 
   @attr('string', {
+    for: 'plugin',
     isNestedSecret: true,
     description: '',
   })
--- a/addons/api/addon/serializers/host-catalog.js
+++ b/addons/api/addon/serializers/host-catalog.js
@@ -115,16 +115,4 @@ export default class HostCatalogSerializer extends ApplicationSerializer {
 
     return serialized;
   }
-
-  // normalize(typeClass, hash, ...rest) {
-  //   const normalizedHash = structuredClone(hash);
-  //   const normalized = super.normalize(typeClass, normalizedHash, ...rest);
-
-  //   if (!normalized.data.attributes.role_arn) {
-  //     hostCatalog.credentialType = TYPE_STATIC_CREDENTIAL;
-  //   }
-  //   normalized.data.attributes.access_key_id = null;
-  //   normalized.data.attributes.secret_access_key = null;
-  //   return normalized;
-  // }
 }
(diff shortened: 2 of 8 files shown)


addons/api/addon/serializers/host-catalog.js
```diff
@@ -68,15 +68,36 @@ export default class HostCatalogSerializer extends ApplicationSerializer {
...
+        ) {
+          json.attributes[key] = null;
+        }
+      }
+
+      // If compositeType is Azure, delete AWS fields from the json
+      if (
+        compositeType === TYPE_HOST_CATALOG_PLUGIN_AZURE &&
+        !azureFields.includes(key)
+      ) {
```
DhariniJeeva, line 100: cleaner way would be to introduce a  new `compositeType` field in model `@atrr` 

if you are ok with that idea, I can refactor that in a separate PR. Let me know your thoughts on this approach

addons/api/addon/generated/models/host-catalog.js
```diff
@@ -104,12 +109,14 @@ export default class GeneratedHostCatalogModel extends BaseModel {
   role_external_id;
```
lisbet-alvarez, line 109: do `role_arn` and `role_external_id` also need `for: 'plugin'` option or is there a reason to not add it?

addons/api/addon/serializers/host-catalog.js
lisbet-alvarez: Not sure i'm following along what introducing new `compositeType` entails. Does this mean that instead of `compositeType` being just `aws`  it would change to be something more specific depending on credential type ?

addons/api/addon/serializers/host-catalog.js
```diff
@@ -98,7 +119,7 @@ export default class HostCatalogSerializer extends ApplicationSerializer {
     // Delete unnecessary fields for static host-catalog
     delete serialized.attributes;
     delete serialized.secrets;
-    delete serialized.worker_filter;
+
```
lisbet-alvarez, line 122: question for understanding: why was `delete serialized.worker_filter` removed ?

addons/api/addon/serializers/host-catalog.js
```diff
@@ -68,15 +68,36 @@ export default class HostCatalogSerializer extends ApplicationSerializer {
...
         ? azureFields
         : AWSfieldsWithCredentialType[credentialType];
 
-    // Delete any nested attribute fields that don't belong to the record type
+    // Delete any nested attribute fields that don't belong to the record's compositeType
     if (isPlugin && options.isNestedAttribute && json.attributes) {
-      if (!fields.includes(key)) {
+      // If compositeType is AWS, delete azure fields from the json
+      if (compositeType === TYPE_HOST_CATALOG_PLUGIN_AWS) {
+        if (
```
calcaide, line 80: I see we are checking twice the condition `!AWSfieldsWithCredentialType[credentialType].includes(key)` alongside with a second condition.

Could this be simplified to check such condition only once?
i.e:

```javascript
if (compositeType === TYPE_HOST_CATALOG_PLUGIN_AWS) {
  if (!AWSfieldsWithCredentialType[credentialType].includes(key)) {
    if (azureFields.includes(key)) {
      delete json.attributes[key];
    }
    json.attributes[key] = null;
  }
}
```
What do you think?

ui/admin/app/controllers/scopes/scope/host-catalogs/index.js
```diff
@@ -63,6 +64,10 @@ export default class ScopesScopeHostCatalogsIndexController extends Controller {
     isNew ? 'notifications.create-success' : 'notifications.save-success',
   )
   async save(hostCatalog) {
+    // If the role_arn is empty, then the credential type should be static
```
calcaide, line 67: Asking for learning purposes: If role_arn empty, means credential type should be static, shouldn't be doing this at the model or normlaizer? Or what is the reason for doing this within the controller?

addons/api/addon/serializers/host-catalog.js
DhariniJeeva: yeah, right now we pass `for` which is the recordType. So if we could pass `compositeType` or `credentialType` from the model it would be a lot cleaner compared to how we handle this in a resource serializer

addons/api/addon/generated/models/host-catalog.js
DhariniJeeva: ah good catch, I forgot to add `plugin` for those two fields. wiIl update it

addons/api/addon/serializers/host-catalog.js
DhariniJeeva: Since I added type `static` or `plugin` in the model attr via `for`, We don't need to explicitly delete this field from static serializer anymore. This case is handled in the [application serializer](https://github.com/hashicorp/boundary-ui/blob/main/addons/api/addon/serializers/application.js#L76) where we automatically discard fields that don't belong to the record type

ui/admin/app/controllers/scopes/scope/host-catalogs/index.js
DhariniJeeva: this is only for the UI, on page refresh this works fine as this gets correctly set in the [model](https://github.com/hashicorp/boundary-ui/blob/main/addons/api/addon/models/host-catalog.js#L54)

addons/api/addon/serializers/host-catalog.js
DhariniJeeva: 💯 this works here, I just added `else` condition to this block. updated!

addons/api/addon/serializers/host-catalog.js
```diff
@@ -98,7 +112,19 @@ export default class HostCatalogSerializer extends ApplicationSerializer {
...
+  //   const normalizedHash = structuredClone(hash);
+  //   const normalized = super.normalize(typeClass, normalizedHash, ...rest);
+
+  //   if (!normalized.data.attributes.role_arn) {
+  //     hostCatalog.credentialType = TYPE_STATIC_CREDENTIAL;
+  //   }
+  //   normalized.data.attributes.access_key_id = null;
+  //   normalized.data.attributes.secret_access_key = null;
+  //   return normalized;
+  // }
```
lisbet-alvarez, line 129: is this supposed to be commented out?

DhariniJeeva: oops we don't need that, removed, ty!

addons/api/addon/serializers/host-catalog.js
```diff
@@ -68,15 +68,29 @@ export default class HostCatalogSerializer extends ApplicationSerializer {
...
+    // Delete any nested attribute fields that don't belong to the record's compositeType
     if (isPlugin && options.isNestedAttribute && json.attributes) {
-      if (!fields.includes(key)) {
+      // If compositeType is AWS, delete azure fields from the json, and set the rest that don't belong to respective compositeType to null
+      if (compositeType === TYPE_HOST_CATALOG_PLUGIN_AWS) {
+        if (!AWSfieldsWithCredentialType[credentialType].includes(key)) {
+          if (azureFields.includes(key)) {
+            delete json.attributes[key];
+          } else {
+            json.attributes[key] = null;
```
lisbet-alvarez, line 84: question for understanding: why do the attrs that don't belong to the seleted aws cred. type need to be set to null, why not delete the attr ? Is it because that is what the api expects ?

DhariniJeeva: yeah in patch calls, we need to send null otherwise API won't know if we updated that field. 
For instance, if we have a dynamic credential type and I change it to static in the UI, I need to send null for the dynamic credential fields to inform the API that we want to change it to static credentials

lisbet-alvarez (approved):
update workflow works great!

addons/api/addon/serializers/host-catalog.js
lisbet-alvarez: thank you for the explanation!

addons/api/addon/serializers/host-catalog.js
lisbet-alvarez: ah okay that makes sense, sounds like a good idea and would help with readability

PR #2526, opened 2024-10-18
Upgrade electron-forge to 7.5.0

calcaide:
# Description

UPDATE February 12 2025: this PR has been previously approved + a [test run on the pipeline](https://github.com/hashicorp/boundary-desktop-releases/actions/runs/13529843926).

- Upgrade electron-forge to`7.5.0`.
- Delete lodash patch within `electron-app`.

The reason behind this changes is electron-forge/packer changed the way it validates semver for windows, therefore, the pattern we used before turns out invalid and returns an error when building the desktop client. [More info about the issue within this GH ticket](https://github.com/electron/packager/issues/1714#issuecomment-2091266284). [Code that validates the semVer](https://github.com/electron/packager/blob/d9655d4bab5a9328391e24c98235ef0d241f1e71/src/resedit.ts#L21-L33), within packager.

Due to this change, don't support non-numeric version, we add the function `returnSemVerFromReleaseVersion` to clean up the provided releaseVersion and bein compliance.

Delete lodash patch for electron-app its not need anymore. The new version of electron-forge resolves lodash `4.17.21`, as [per snyk scan, does not have security vulnerabilities](https://security.snyk.io/package/npm/lodash).

## How to Test

- Perform a pipeline run succesfully.
- Test pipeline run, using the new branch naming pattern.
- Unit testing (manually) `returnSemVerFromReleaseVersion` with next string:
    - `1.2.3-beta.test.node.pty` = `1.2.3`;
    - `0.2.3-beta.electron.forge` = `0.2.3`;
    - `v2.0.3` = `2.0.3`;
    - `v0.10.3` = `0.10.3`;


ui/desktop/electron-app/config/desktop.js
```diff
@@ -2,6 +2,17 @@ const fs = require('fs');
...
 
+const returnSemVerFromReleaseVersion = (releaseVersion) => {
+  let [major, minor, patch] = releaseVersion.split('.').slice(0, 3);
+
+  // Delete, if present, v from Major
+  if (!parseInt(major)) {
+    major = major.replace('v', '');
+  }
+
+  return major.concat('.', minor, '.', parseInt(patch));
```
ZedLi, line 13: Do we `parseInt(patch)` to basically remove any text that may have been attached to it? e.g. `1-beta` in `3.2.1-beta`? If so, perhaps we could add a comment to explain that?

Could this also have just used a template string?
```suggestion
  return `${major}.${minor}.${parseInt(patch)}`
```

calcaide: Yes.

- The `parseInt` for `patch` was add for that exact reason. Good catch, added a comment.
- I also move the comment to explain `returnSemVerFromReleaseVersion`.

cameronperera (dismissed):
LGTM. Thanks for digging into this issue and coming up with a solution for the version.

ZedLi (approved):

DhariniJeeva (approved):

ZedLi (dismissed):

e2e-tests/desktop/tests/credentials.spec.js
```diff
@@ -303,11 +303,11 @@ test.describe('Credential Panel tests', () => {
     // Empty string and null should not be visible
     await expect(
       authedPage.getByRole('listitem').filter({ hasText: 'nested.key3' }),
-    ).not.toBeVisible();
+    ).toBeHidden();
```
DhariniJeeva, line 306: I believe this is from a recent PR that was merged into main

DhariniJeeva (approved):
looks great!!

e2e-tests/desktop/tests/credentials.spec.js
calcaide: You are right. I merged main into this branch to update, and now is showing this as change I will look into it, it shouldn't show it.

ui/desktop/electron-app/config/desktop.js
```diff
@@ -11,11 +25,19 @@ const createConfig = () => {
     releaseVersion: process.env.RELEASE_VERSION,
     releaseCommit: process.env.RELEASE_COMMIT,
     executableName: 'Boundary',
+    releaseVersionRaw: process.env.RELEASE_VERSION,
```
calcaide, line 28: @ZedLi As [per your comment](https://github.com/hashicorp/boundary-desktop-releases/pull/128#discussion_r1953494199) I think this can be a good approach.

I created the `releaseVersionRaw` which as its name shows, its the same as `releaseVersion` but raw, without extracting the semVer.

`releaseVersionRaw` its ONLY use by electron-forge, within the `postMake` hook to copy artifacts.

Let me know your thoughts.

ZedLi (approved):
Looks good!

DhariniJeeva (approved):

PR #2530, opened 2024-10-21
fix: 🐛 upgrade cookie package

cameronperera:
# Description

This resolves this [dependabot alert](https://github.com/hashicorp/boundary-ui/security/dependabot/166)
We had 2 versions of cookie, `0.6.0` & `0.4.1`.
By upgrading `socket.io` that upgraded `engine.io` which upgraded `cookie` to `0.7.2`.
Upgrading `express` to `cookie` to `0.7.1`.

Anything > `0.7.0` fixes our the security alert.

[Test release](https://github.com/hashicorp/boundary-ui-releases/actions/runs/11448516342)

## Checklist

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -7566,15 +7566,15 @@ cookie-signature@1.0.6:
   resolved "https://registry.yarnpkg.com/cookie-signature/-/cookie-signature-1.0.6.tgz#e303a882b342cc3ee8ca513a79999734dab3ae2c"
   integrity sha512-QADzlaHc8icV8I7vbaJXJwod9HWYp8uCqf1xa4OfNu1T7JVxQIrUgOWtHdNDtPiywmFbiS12VjotIXLrKM3orQ==
 
-cookie@0.6.0:
-  version "0.6.0"
-  resolved "https://registry.yarnpkg.com/cookie/-/cookie-0.6.0.tgz#2798b04b071b0ecbff0dbb62a505a8efa4e19051"
-  integrity sha512-U71cyTamuh1CRNCfpGY6to28lxvNwPG4Guz/EVjgf3Jmzv0vlDp1atT9eS5dDjMYHucpHbWns6Lwf3BKz6svdw==
+cookie@0.7.1:
+  version "0.7.1"
+  resolved "https://registry.yarnpkg.com/cookie/-/cookie-0.7.1.tgz#2f73c42142d5d5cf71310a74fc4ae61670e5dbc9"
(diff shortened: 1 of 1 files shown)


lisbet-alvarez (approved):

ZedLi (approved):

calcaide (changes requested):
After running `yarn why cookie` looks like we are bringing this old versions because third party dependencies of `ember-cli`.

Screenshot 2024-10-22 at 12 57 10 PM

My question is: Do we need to provide a fix urgently? otherwise, I will advocate for attempting to upgrade `ember-cli` and validate such upgrade brings versions of `cookie` without security vulnerabilities.

Let me know your thoughts.

cameronperera:
> After running `yarn why cookie` looks like we are bringing this old versions because third party dependencies of `ember-cli`.
> 
> Screenshot 2024-10-22 at 12 57 10 PM
> My question is: Do we need to provide a fix urgently? otherwise, I will advocate for attempting to upgrade `ember-cli` and validate such upgrade brings versions of `cookie` without security vulnerabilities.
> 
> Let me know your thoughts.

That is a good point. Upgrading `ember-cli` might just resolve a good deal of our security alerts. I opted to go this route since we do have a hard push to resolve our security alerts but I can see how these are more bandaid fixes while an `ember-cli` upgrade is likely the real fix. Would you be ok with me doing that work in a separate branch? We could allow the smaller security alert related PRs to be merged, then focus on just upgrading `ember-cli`. I did see you have a ticket for it so I can use that.

calcaide:
> > After running `yarn why cookie` looks like we are bringing this old versions because third party dependencies of `ember-cli`.
> > Screenshot 2024-10-22 at 12 57 10 PM
> > My question is: Do we need to provide a fix urgently? otherwise, I will advocate for attempting to upgrade `ember-cli` and validate such upgrade brings versions of `cookie` without security vulnerabilities.
> > Let me know your thoughts.
> 
> That is a good point. Upgrading `ember-cli` might just resolve a good deal of our security alerts. I opted to go this route since we do have a hard push to resolve our security alerts but I can see how these are more bandaid fixes while an `ember-cli` upgrade is likely the real fix. Would you be ok with me doing that work in a separate branch? We could allow the smaller security alert related PRs to be merged, then focus on just upgrading `ember-cli`. I did see you have a ticket for it so I can use that.

Hey @cameronperera 

Thank you for the work, totally agree with doing it in a separate PR.

Totally understandable about the push to fix. I am totally open to "force" versions or other kind of "fixes" for the sake of delivering within a due date, but if we know it's a fix and the real problem could be a dependenc upgrade, I will ask to create a ticket in jira to upgrade such dependency when we have time, and move on with the fix. All good on that front 😉

cameronperera:
This was fixed with the `ember-cli` upgrade so I am closing this out.

PR #2531, opened 2024-10-21
chore: 🤖 audit most recent resolutions

lisbet-alvarez:
# Description

Resolutions Audited:
* `loader-utils`
    * Upgraded `babel-loader` to 9.2.1 which removes `loader-utils` dep.
    * Forced `babel-loader` version `^8.0.6` to resolve to 8.4.1 which uses `loader-utils` patched version
        * `babel-loader` is a dependency of `ember-auto-import`
* `minimatch`
    * Resolution not needed anymore since all occurrences resolved to patched version
* `prismjs`
    * Was not needed since all versions resolved to 1.27.0 which is patched
* `underscore -> nomnom -> jsonlint`
    * `jsonlint` seems unmaintained we should replace but resolution still needed for now
* `trim`
    * Resolution not needed & dependency actually not needed anywhere (no matches)
* `micromatch`
    * Resolution still needed
    * micromatch@^3.1.4 -> anymatch@^2.0.0 -> sane@^4.0.0 -> broccoli@^3.5.2 -> ember-cli@^5.12.0
    * micromatch@^3.1.4 -> sane@^4.0.0 -> broccoli@^3.5.2 -> ember-cli@^5.12.0
    * micromatch@^3.1.4 -> find-yarn-workspace-root@^1.2.1 -> ember-cli-dependency-checker@^3.3.2
* `send`
    * Resolution not needed since forcing `express` to recently released version fixes it.

## How to Test

boundary-ui-releases: [test run](https://github.com/hashicorp/boundary-ui-releases/actions/runs/11488957922)

## Checklist


yarn.lock
```diff
@@ -5740,12 +5740,12 @@ babel-jest@^29.7.0:
     slash "^3.0.0"
 
 babel-loader@^8.0.6:
-  version "8.2.2"
-  resolved "https://registry.yarnpkg.com/babel-loader/-/babel-loader-8.2.2.tgz#9363ce84c10c9a40e6c753748e1441b60c8a0b81"
-  integrity sha512-JvTd0/D889PQBtUXJ2PXaKU/pjZDMtHA9V2ecm+eNRmmBCMR09a+fmpGTNwnJtFmFl5Ei7Vy47LjBb+L0wQ99g==
+  version "8.4.1"
```
calcaide, line 5743: Rather than force `babel-loader` within the yarn.lock, can we catually upgrade it properly within our packages.json? It is present, [here](https://github.com/hashicorp/boundary-ui/blob/main/ui/desktop/package.json#L67), [here](https://github.com/hashicorp/boundary-ui/blob/main/ui/admin/package.json#L64) and [here](https://github.com/hashicorp/boundary-ui/blob/main/addons/rose/package.json#L57)

calcaide:
I am not understanding from the PR description `underscore -> nomnom -> jsonlint`. 

If I run `yarn why jsonlint` the only result I get is from Rose, and we are actually resolving `jsonlint` as dependency.
Screenshot 2024-10-22 at 1 09 18 PM

So I am confuse by your comment. Are you saying that underscore contains nommon which contains jsonlint? the resolution we have is the other way around `**/nomnom/underscore`, can you please elaborate a bit? thank you

calcaide:
I follow up the breadcrumbs with micromatch, as despite we still need it today, if we perform an `ember-cli` upgrade, most likely we will not need it anymore.

After deleting the resolution of micromatch, reinstalling deps and running `yarn why micromatch`, you can see most deps that use `micromatch` are using a version with no vulnerabilities `4.0.8`.

The only 2 deps that bring micromatch versions with security vulnerabilities, are dependencies from `ember-cli`. I am not 100% sure, since needs to be checked, but I bet upgrading `ember-cli` can solve this issue and the resolutions will not be need anymore.

Screenshot 2024-10-22 at 1 12 24 PM

I am not mentioning this to do perform that work now, but to audit with breadcrumbs, so we can asses how to get rid of the resolution in the future.

yarn.lock
```diff
@@ -10347,23 +10347,23 @@ exponential-backoff@^3.1.1:
   integrity sha512-dX7e/LHVJ6W3DE1MHWi9S1EYzDESENfLrYohG2G++ovZrYOkm4Knwa0mc1cn84xJOR4KEU0WSchhLbd0UklbHw==
 
 express@^4.10.7, express@^4.18.1:
-  version "4.20.0"
-  resolved "https://registry.yarnpkg.com/express/-/express-4.20.0.tgz#f1d08e591fcec770c07be4767af8eb9bcfd67c48"
-  integrity sha512-pLdae7I6QqShF5PnNTCVn4hI91Dx0Grkn2+IAsMTgMIKuQVte2dN9PeGSSAME2FR8anOhVA62QDIUaWVfEXVLw==
+  version "4.21.1"
```
calcaide, line 10350: Express is used just because `ember-cli`, rather than force express within `yarn.lock` can we look into upgrade `ember-cli` to bring a newer version of express?

calcaide (changes requested):
Left some comments. Thanks for taking care of this work

yarn.lock
lisbet-alvarez: great catch, didn't realize it was a dev. dep. ill update this.

lisbet-alvarez:
> I am not understanding from the PR description `underscore -> nomnom -> jsonlint`.
> 
> If I run `yarn why jsonlint` the only result I get is from Rose, and we are actually resolving `jsonlint` as dependency. Screenshot 2024-10-22 at 1 09 18 PM
> 
> So I am confuse by your comment. Are you saying that underscore contains nommon which contains jsonlint? the resolution we have is the other way around `**/nomnom/underscore`, can you please elaborate a bit? thank you

i write the dependencies backwards because its easier for me to understand that way, but it means that `jsonlint` has `nomnom` as a dep and `nomnom` has `underscore` as a dep. Apologies for the confusion.

yarn.lock
lisbet-alvarez: thats a fair point, @cameronperera opened a PR to upgrade ember-cli. I'll convert this PR to draft and update this PR once the ember-cli upgrade is merged.

calcaide:
> > I am not understanding from the PR description `underscore -> nomnom -> jsonlint`.
> > If I run `yarn why jsonlint` the only result I get is from Rose, and we are actually resolving `jsonlint` as dependency. Screenshot 2024-10-22 at 1 09 18 PM
> > So I am confuse by your comment. Are you saying that underscore contains nommon which contains jsonlint? the resolution we have is the other way around `**/nomnom/underscore`, can you please elaborate a bit? thank you
> 
> i write the dependencies backwards because its easier for me to understand that way, but it means that `jsonlint` has `nomnom` as a dep and `nomnom` has `underscore` as a dep. Apologies for the confusion.

Ahh gotcha! All good 😉

yarn.lock
calcaide: Sure! Yeah, not saying it will, I did not follow the breadcrumbs what version of express is bringing the newest version of `ember-cli` but in case it does problem solve 😃 otherwise, forcing the version of express is totally fine 😉

addons/api/package.json
```diff
@@ -24,7 +24,7 @@
   },
   "dependencies": {
     "dexie": "^3.2.4",
-    "ember-auto-import": "^2.6.3",
+    "ember-auto-import": "^2.8.1",
```
lisbet-alvarez, line 27: this change is mostly unrelated, saw that ember-auto-import was already resolving to newer version so decided to add the change in the package.json files to newest version.

ZedLi (approved):
Awesome work! Thanks for auditing these resolutions

cameronperera (approved):
Nice work!

addons/api/package.json
laurenolivia: @lisbet-alvarez Since our upgrades are manual, I'm wondering which command you're running to determine which dependencies need updating?

laurenolivia (commented):
See comment [here](https://github.com/hashicorp/boundary-ui/pull/2531/files#r1819549402)

addons/api/package.json
lisbet-alvarez: `yarn why {dep. name}` is a good resource since it tells you what version is being used and where its being used. Tho in the case of `ember-auto-import` I noticed it was resolving to a newer minor version in our `yarn.lock` file.

laurenolivia: I imagine you are not using `yarn why <package-name>` unless you've already determined the package needs updating, right?

calcaide (approved):
Thanks for the aweswome work!

addons/api/package.json
laurenolivia: We cleared up why packages end up in [resolutions](https://github.com/hashicorp/boundary-ui/blob/main/package.json#L51) 👍 thanks!

DhariniJeeva (approved):
looks great!

PR #2532, opened 2024-10-22
test: 💍 update test selectors for accounts delete

cameronperera:
✅ Closes: https://hashicorp.atlassian.net/browse/ICU-14038

# Description

update test selectors for accounts delete

## Checklist

### DIFF
--- a/ui/admin/tests/acceptance/accounts/selectors.js
+++ b/ui/admin/tests/acceptance/accounts/selectors.js
@@ -6,10 +6,12 @@
 export const APP_HEADER_DROPDOWN =
   '.rose-header-utilities details:first-child summary';
 
-export const MANAGE_DROPDOWN =
+export const MANAGE_DROPDOWN_AUTH_METHOD =
   '[data-test-manage-auth-methods-dropdown] button';
-export const MANAGE_DROPDOWN_NEW_ACCOUNT = (url) =>
-  `[data-test-manage-auth-methods-dropdown] a[href="${url}"]`;
+export const MANAGE_DROPDOWN_ACCOUNT =
+  '[data-test-manage-account-auth-methods] button';
+export const MANAGE_DROPDOWN_DELETE_ACCOUNT =
+  '[data-test-manage-account-auth-methods] ul li button';
 
 export const FIELD_CURRENT_PASSWORD = '[name=currentPassword]';
 export const FIELD_CURRENT_PASSWORD_VALUE = 'current password';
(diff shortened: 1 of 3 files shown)


ui/admin/tests/acceptance/accounts/selectors.js
```diff
@@ -6,10 +6,12 @@
 export const APP_HEADER_DROPDOWN =
   '.rose-header-utilities details:first-child summary';
 
-export const MANAGE_DROPDOWN =
+export const MANAGE_DROPDOWN_AUTH_METHOD =
   '[data-test-manage-auth-methods-dropdown] button';
-export const MANAGE_DROPDOWN_NEW_ACCOUNT = (url) =>
-  `[data-test-manage-auth-methods-dropdown] a[href="${url}"]`;
```
cameronperera, line 12: Realized I add this but ended up using a different selector so I removed this.

ui/admin/tests/acceptance/accounts/selectors.js
```diff
@@ -6,10 +6,12 @@
 export const APP_HEADER_DROPDOWN =
   '.rose-header-utilities details:first-child summary';
 
-export const MANAGE_DROPDOWN =
+export const MANAGE_DROPDOWN_AUTH_METHOD =
```
laurenolivia, line 9: Nice. This is a much clearer name for the env var 👍

ui/admin/tests/acceptance/accounts/selectors.js
```diff
@@ -6,10 +6,12 @@
...
   '[data-test-manage-auth-methods-dropdown] button';
-export const MANAGE_DROPDOWN_NEW_ACCOUNT = (url) =>
-  `[data-test-manage-auth-methods-dropdown] a[href="${url}"]`;
+export const MANAGE_DROPDOWN_ACCOUNT =
```
laurenolivia, line 11: nit: Wondering what you think about including `AUTH` in these as well, if the test selector includes `*-auth-*`? Similar to your var above.

laurenolivia (commented):
Small nits. Nice work ✨

calcaide (approved):
Great job!!! 🙌

ui/admin/tests/acceptance/accounts/selectors.js
cameronperera: I opted not to include `AUTH` just to keep it shorter. Since this selector file is specific to accounts, I figured that would be a safe bet. Otherwise, we would have `MANAGE_DROPDOWN_AUTH_METHOD_ACCOUNT` which is very long. While the specificity in the `data-test` attribute is needed in my opinion, the name of the test selector doesn't need to be. We could potentially even rename that `data-test` attribute.

laurenolivia: Fair & non-blocking. I interpreted `auth` as being unique to these selectors, specifically. Either way ✅ !

laurenolivia (approved):
✨ 🦄

PR #2533, opened 2024-10-22
chore: 🤖 upgrade ember-cli version

cameronperera:
✅ Closes: https://hashicorp.atlassian.net/browse/ICU-15393

# Description

chore: 🤖 upgrade ember-cli version
This upgrade should resolve a few different security alerts that we have. I will address those in a separate PR as I want to keep changes in packages low as we run tests, build, test locally, and test the release pipeline.

[Release test run](https://github.com/hashicorp/boundary-ui-releases/actions/runs/11468678848)

## Checklist

### DIFF
--- a/addons/api/package.json
+++ b/addons/api/package.json
@@ -47,7 +47,7 @@
     "broccoli-asset-rev": "^3.0.0",
     "concurrently": "^8.0.1",
     "doctoc": "^2.2.0",
-    "ember-cli": "^4.12.2",
+    "ember-cli": "^5.12.0",
     "ember-cli-dependency-checker": "^3.3.2",
     "ember-cli-inject-live-reload": "^2.1.0",
     "ember-cli-sri": "^2.1.1",
--- a/addons/auth/package.json
+++ b/addons/auth/package.json
@@ -39,7 +39,7 @@
     "concurrently": "^8.0.1",
     "doctoc": "^2.2.0",
     "ember-auto-import": "^2.6.3",
-    "ember-cli": "^4.12.2",
+    "ember-cli": "^5.12.0",
     "ember-cli-dependency-checker": "^3.3.2",
     "ember-cli-inject-live-reload": "^2.1.0",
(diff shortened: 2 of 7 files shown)


ZedLi (approved):
Nice!

lisbet-alvarez (approved):
🥳

calcaide:
@cameronperera 

We have some `resolutions` related to `ember-cli`, [here the code](https://github.com/hashicorp/boundary-ui/blob/main/package.json#L70-L74). Do we plan on audit/check if these are necessary after the `ember-cli` upgrade? I am fine if that happens in a separate PR, just checking 😉

calcaide (approved):
Left a couple of comments.

Great work!!

cameronperera:
> @cameronperera After the `ember-cli` upgrade, seems to me we do not need anymore `lodash` patch.
> 
> * We can delete the `devDependencies` [here code](https://github.com/hashicorp/boundary-ui/blob/main/package.json#L51-L52).
> * We can delete the `post-install` script, [here code](https://github.com/hashicorp/boundary-ui/blob/main/package.json#L42).
> * We can delete the patch itself, [here code](https://github.com/hashicorp/boundary-ui/blob/main/patches/lodash.template%2B4.5.0.patch).
> 
> To validate such thing, I pull this branch, delete the mentioned above, perform a `rm -rf node_modules/ && yarn install` then run `yarn why lodash` and saw that we are resolving lodash `4.17.21` without using the patch.
> 
> Mentioning this because you have more context than me within the patch work, so if this validation process makes sense to you, we can get rid of it 😃 .
> 
> For the patch within `electron-app`, do not worry, because I will get rid of it [within this PR](https://github.com/hashicorp/boundary-ui/pull/2526/files) since after `electron-forge` upgrade seems not need anymore.
> 
> Let me know if I am missing something.

Thanks for the detailed response and yes, the plan was to remove the patch and resolutions once this is merged in. I will work on that right after I get this merged.

PR #2534, opened 2024-10-23
refactor: 💡 remove route-action usage in admin ui

lisbet-alvarez:
✅ Closes: https://hashicorp.atlassian.net/browse/ICU-12945

# Description

Remove remaining route-action usage in admin ui (filtering).
Remove route-action helper.

## How to Test

1. Test scope filtering in roles -> role -> add-principals page.
2. Test scope filtering in groups -> group -> add-members page.
3. Test tag filtering in workers page.

## Checklist


ui/admin/app/controllers/scopes/scope/groups/group/add-members.js
```diff
@@ -57,4 +57,22 @@ export default class ScopesScopeGroupsGroupAddMembersController extends Controll
...
     await this.router.replaceWith('scopes.scope.groups.group.members');
   }
+
+  /**
+   * Calls filterBy action located in the route.
+   * @param {string} field
+   * @param {[ScopeModel]} value
+   */
+  @action
+  callFilterBy(field, value) {
```
ZedLi, line 67: Do we need to call these `call<action>`? Did you think it might be confusing to just call it what it originally was and just proxy it through to the route? For example `filterBy` and then calling `this.send('filterBy')`?

ZedLi (approved):
Amazing work!! Great to see this done

ui/admin/app/controllers/scopes/scope/groups/group/add-members.js
lisbet-alvarez: we mainly just need to call it a different name or else it causes an infinite loop and calls itself instead of the action in the route. 
ex. error msg:
image

ZedLi: Ahh I didn't know that, I assumed send would only look at routes

cameronperera (approved):
Nice work!

PR #2540, opened 2024-10-27
Upgrade HDS dependency to 4.13 & migrate FlightIcon to Hds::Icon

zamoore:
# Description
Upgrade HDS dependency to 4.13 & migrate FlightIcon to Hds::Icon

### DIFF
--- a/addons/rose/index.js
+++ b/addons/rose/index.js
@@ -27,7 +27,7 @@ module.exports = {
 
   /**
    * Due to a limitation in how ember treats nested addons (see https://github.com/ember-cli/ember-cli/issues/4475)
-   * this is neeeded to reach down into @hashicorp/ember-flight-icons' contentFor hook to run the logic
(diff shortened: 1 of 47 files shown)


addons/rose/index.js
```diff
@@ -32,7 +32,6 @@ module.exports = {
    */
   contentFor(type, config) {
```
lisbet-alvarez, line 33: thank you all this work!!

small suggestion: can the comment for this hook (line 28-32) be updated to not reference `@hashicorp/ember-flight-icons`?

lisbet-alvarez (approved):
awesome work!! 🎊
[boundary-ui-release test](https://github.com/hashicorp/boundary-ui-releases/actions/runs/11634850156) run passes as well!

cameronperera (approved):

addons/rose/index.js
zamoore: No problem. Thanks!

PR #2544, opened 2024-11-01
Icu 15480 groups create test

DhariniJeeva:
# Description
update test selectors for groups create

https://hashicorp.atlassian.net/browse/ICU-15480

## Checklist

### DIFF
--- a/ui/admin/tests/acceptance/groups/create-test.js
+++ b/ui/admin/tests/acceptance/groups/create-test.js
@@ -4,17 +4,15 @@
  */
 
 import { module, test } from 'qunit';
-import { visit, currentURL, click, fillIn, find } from '@ember/test-helpers';
+import { visit, currentURL, click, fillIn } from '@ember/test-helpers';
 import { setupApplicationTest } from 'ember-qunit';
 import setupMirage from 'ember-cli-mirage/test-support/setup-mirage';
 import { setupIndexedDb } from 'api/test-support/helpers/indexed-db';
 import { Response } from 'miragejs';
-import {
-  authenticateSession,
-  // These are left here intentionally for future reference.
-  //currentSession,
-  //invalidateSession,
-} from 'ember-simple-auth/test-support';
+import { authenticateSession } from 'ember-simple-auth/test-support';
+
+import * as commonSelectors from 'admin/tests/helpers/selectors';
+import * as selectors from './selectors';
 
 module('Acceptance | groups | create', function (hooks) {
   setupApplicationTest(hooks);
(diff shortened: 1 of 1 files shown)


cameronperera (commented):
Looks great for the test selector changes. One thing I try to also update while in these test files is the readability by using Arrange, Act, Assert. This just adds space around the 3 different steps we typically have for each test. Example of what this looks like.
```
  test('can cancel new group creation', async function (assert) {
    // This is our Arrange
    const groupsCount = this.server.schema.groups.all().models.length;
    await visit(urls.newGroup);
    
    // This is our Act
    await fillIn(commonSelectors.FIELD_NAME, commonSelectors.FIELD_NAME_VALUE);
    await click(commonSelectors.CANCEL_BTN);
    
    // This is our Assert
    assert.strictEqual(currentURL(), urls.groups);
    assert.strictEqual(
      this.server.schema.groups.all().models.length,
      groupsCount,
    );
  });
```

ui/admin/tests/acceptance/groups/create-test.js
```diff
@@ -52,11 +49,14 @@ module('Acceptance | groups | create', function (hooks) {
   });
 
   test('can create new group', async function (assert) {
-    const groupsCount = this.server.db.groups.length;
+    const groupsCount = this.server.schema.groups.all().models.length;
     await visit(urls.newGroup);
-    await fillIn('[name="name"]', 'group name');
-    await click('[type="submit"]');
-    assert.strictEqual(this.server.db.groups.length, groupsCount + 1);
+    await fillIn(commonSelectors.FIELD_NAME, commonSelectors.FIELD_NAME_VALUE);
```
calcaide, line 54: None blocking comment: since a group has also the field Description as optional, like name, I will suggest to also `fillIn` the field description.

ui/admin/tests/acceptance/groups/create-test.js
```diff
@@ -52,11 +49,14 @@ module('Acceptance | groups | create', function (hooks) {
...
+    await click(commonSelectors.SAVE_BTN);
+    assert.strictEqual(
+      this.server.schema.groups.all().models.length,
```
calcaide, line 57: We are repeating this within the test file. Could we define a let outside the test but within the describe context and tests reuse it?
[As we did here](https://github.com/hashicorp/boundary-ui/blob/main/ui/admin/tests/acceptance/host-catalogs/hosts/create-test.js#L72) and use it [here](https://github.com/hashicorp/boundary-ui/blob/main/ui/admin/tests/acceptance/host-catalogs/hosts/create-test.js#L77) and [here](https://github.com/hashicorp/boundary-ui/blob/main/ui/admin/tests/acceptance/host-catalogs/hosts/create-test.js#L87)

DhariniJeeva: ah yes good point, updated!

calcaide (approved):
Great work!! 🙌

PR #2563, opened 2024-11-12
refactor: 💡 Fix linting errors in e2e

ZedLi:
# Description
Fixed some of the linting errors that were present in our e2e tests.

## Checklist

### DIFF
--- a/e2e-tests/admin/tests/credential-store-vault-ent.spec.js
+++ b/e2e-tests/admin/tests/credential-store-vault-ent.spec.js
@@ -4,10 +4,7 @@
  */
 
 import { test } from '../playwright.config.js';
-import { expect } from '@playwright/test';
 import { execSync } from 'child_process';
-import { nanoid } from 'nanoid';
-import { readFile } from 'fs/promises';
 
 import { authenticatedState } from '../global-setup.js';
 import {
--- a/e2e-tests/admin/tests/worker-tags.spec.js
+++ b/e2e-tests/admin/tests/worker-tags.spec.js
@@ -37,9 +37,9 @@ test('Worker Tags @ce @ent @aws @docker', async ({ page }) => {
       .getByRole('columnheader')
       .nth(i)
       .innerText();
-    if (header == 'Worker') {
+    if (header === 'Worker') {
       workerListWorkerIndex = i;
-    } else if (header == 'Tags') {
+    } else if (header === 'Tags') {
       workerListTagIndex = i;
     }
   }
@@ -58,7 +58,6 @@ test('Worker Tags @ce @ent @aws @docker', async ({ page }) => {
     .getByRole('table')
     .getByRole('columnheader')
     .count();
-  let dialogTagIndex;
   let dialogTypeIndex;
   for (let i = 0; i < dialogHeadersCount; i++) {
     const header = await page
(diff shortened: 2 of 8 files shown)


e2e-tests/admin/tests/credential-store-vault-ent.spec.js
```diff
@@ -4,10 +4,7 @@
  */
 
 import { test } from '../playwright.config.js';
-import { expect } from '@playwright/test';
```
ZedLi, line 7: I was surprised some of the tests weren't asserting anything, we likely should be asserting _something_ happened

e2e-tests/admin/tests/worker-tags.spec.js
```diff
@@ -67,9 +66,7 @@ test('Worker Tags @ce @ent @aws @docker', async ({ page }) => {
       .getByRole('columnheader')
       .nth(i)
       .innerText();
-    if (header == 'Tag') {
-      dialogTagIndex = i;
```
ZedLi, line 71: Not sure if this was previously used and is vestigial but some of these variables weren't used at all and tests ran fine without them

e2e-tests/package.json
```diff
@@ -22,13 +22,14 @@
...
     "desktop": "yarn workspace desktop build && yarn run desktop:dev",
-    "desktop:dev": "npx playwright test --config desktop/playwright.config.js --reporter=html"
+    "desktop:dev": "yarn playwright test --config desktop/playwright.config.js --reporter=html"
   },
   "devDependencies": {
     "@babel/core": "^7.25.2",
     "@babel/eslint-parser": "^7.21.3",
     "@playwright/test": "^1.43.0",
     "concurrently": "^9.1.0",
+    "dotenv": "^16.4.5",
```
ZedLi, line 32: I added this purely for my own convenience so I can use env files when running tests directly in my IDE

e2e-tests/admin/tests/credential-store-vault-ent.spec.js
moduli: My initial thought was that a lot of the methods invoked by these tests have `expects` in them and that the goal for the test is to see if they're able to successfully complete this workflow. 

So if we were to try that in this test for example, would you expect something like... ?
```
let session = await authorizeSessionByTargetIdCli(targetId);
expect(session).toBeTruthy();
```

moduli (approved):

e2e-tests/admin/tests/credential-store-vault-ent.spec.js
ZedLi: I think it would be fine if the helper methods asserted, I only checked the `credential-store-vault-ent` test but I didn't really see any assertions regarding the session establishment.

I originally added something like `expect(session.status_code).toBe(200)` but then after thinking about it more felt like we probably should be actually going to the sessions page and validating it that way, any thoughts?

calcaide: My 2 cents:
If a test runs a procedure, let's say create a credential library, an assertion can be checking the DOM for the Successfully created toast? (attached image for better understanding) or as mentioned, after the creation, go to the list page and validate the resource has been created?

cred-lib

calcaide (approved):

e2e-tests/admin/tests/credential-store-vault-ent.spec.js
moduli: @calcaide Generally, operations will already check for the success toast (1: to confirm that the operation was successful. 2: to clear out the notification so that they don't pile up and block elements on the page).

To @ZedLi's point, we could have some additional assertions here. We have one test that DOES navigate to the sessions page to confirm that a session was created. Other tests (like the one Zed pointed out) just confirm whether or not running a CLI command will throw an exception. I created a ticket on my backlog to do a pass of the tests and see if we want to add any additional assertions.

PR #2565, opened 2024-11-13
Icu 15743 tag workflow

cameronperera:
# Description

This PR adds the workflow for creating a tag specific worker filter

https://github.com/user-attachments/assets/68185c0d-b113-47e8-ad75-e43e6a3e0f74

## How to Test

1. Go to the workers tab on a Target, toggle the "show generator" toggle
2. Enter in the key and value
3. Should be able to copy the formatted result

## Checklist

### DIFF
--- a/addons/core/translations/en-us.yaml
+++ b/addons/core/translations/en-us.yaml
@@ -143,10 +143,10 @@ worker-filter-generator:
     link: Learn more about formatting a worker filter.
     input-values:
       title: Input values
-      description: Next, input what you want to format
+      description: Next, input what you want to format.
     formatted-result:
       title: Formatted result
-      description: Copy and paste this into the editor above
+      description: Copy and paste this into the editor above.
       label: Generated result
     tag:
       label: Tag
--- a/addons/core/translations/en-us.yaml
+++ b/addons/core/translations/en-us.yaml
@@ -141,6 +141,13 @@ worker-filter-generator:
     title: Filter generator
     description: Choose what you want to format into a filter.
     link: Learn more about formatting a worker filter.
+    input-values:
+      title: Input values
+      description: Next, input what you want to format
+    formatted-result:
+      title: Formatted result
+      description: Copy and paste this into the editor above
+      label: Generated result
     tag:
       label: Tag
       helper: Format a worker tag
--- a/ui/admin/app/components/worker-filter/index.hbs
+++ b/ui/admin/app/components/worker-filter/index.hbs
@@ -66,7 +66,7 @@
       </G.HelperText>
       <G.RadioField
         @value={{this.generatorTagType}}
-        checked={{eq this.filterGeneratorType this.generatorTagType}}
+        checked={{eq this.selectedGeneratorType this.generatorTagType}}
         as |F|
       >
         <F.Label>{{t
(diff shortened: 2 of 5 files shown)


addons/core/translations/en-us.yaml
```diff
@@ -141,6 +141,13 @@ worker-filter-generator:
     title: Filter generator
     description: Choose what you want to format into a filter.
     link: Learn more about formatting a worker filter.
+    input-values:
+      title: Input values
+      description: Next, input what you want to format
```
lisbet-alvarez, line 146:
```suggestion
      description: Next, input what you want to format.
```

addons/core/translations/en-us.yaml
```diff
@@ -141,6 +141,13 @@ worker-filter-generator:
...
+    formatted-result:
+      title: Formatted result
+      description: Copy and paste this into the editor above
```
lisbet-alvarez, line 149:
```suggestion
      description: Copy and paste this into the editor above.
```

ui/admin/app/components/worker-filter/index.hbs
```diff
@@ -73,6 +77,63 @@
...
+            'worker-filter-generator.filter-generator.input-values.title'
+          }}</F.Legend>
+        <F.HelperText>{{t
+            'worker-filter-generator.filter-generator.input-values.description'
+          }}</F.HelperText>
+        <F.Control>
+          <Hds::Form::TextInput::Field
+            @value={{this.tagKey}}
+            name='tag_key'
+            @width='320px'
```
lisbet-alvarez, line 93: im wondering if we should leave this as pixels or use rem instead. Both work, just wondering what the convention is.

ui/admin/app/components/worker-filter/index.hbs
```diff
@@ -73,6 +77,63 @@
...
+          'worker-filter-generator.filter-generator.formatted-result.description'
+        }}</F.HelperText>
+      <F.Control>
+        <Hds::Text::Body @tag='p' @weight='semibold'>{{t
+            'worker-filter-generator.filter-generator.formatted-result.label'
+          }}</Hds::Text::Body>
+        <div class='generated-results-container'>
+          <Hds::Form::TextInput::Base
+            readonly
+            id='generatedValue'
```
lisbet-alvarez, line 128: i think for id values we usually use dashes `id='generated-value'`. Does that work?

ui/admin/app/components/worker-filter/index.hbs
cameronperera: I looked around and we primarily use `%` for the `@width` property. The issue with that is the input would get much smaller if the user wants change the width of the screen for split screen for example. I don't think `rem` would give us much of a change. I would lean towards leaving it as is.

ui/admin/app/components/worker-filter/index.hbs
cameronperera: I will test that now.

cameronperera: It worked, updated.

lisbet-alvarez (approved):
🥳

PR #2583, opened 2024-11-21
fix vault brokered credentials display

DhariniJeeva:
# Description
This PR displays the vault brokered credentials in a consumable format. Previously this showed in a data and metadata JSON, but it is in a key/value copyable format. We decided against showing metadata info as it does not add any value to the users.  

https://hashicorp.atlassian.net/browse/ICU-15855

before: 
Screenshot 2024-11-25 at 12 51 01 PM

after:
Screenshot 2024-11-25 at 12 46 28 PM

using mocks: [you can see the email.address key here, which is a nested object within `data`, if we ever encounter a nested object inside data, it will be displayed using `parentPath`]

 
Screenshot 2024-11-26 at 10 03 41 AM

## How to Test

1. Use Enos to set up the vault cred store, which brokers SSH key creds. Reach out to me if you would like to test it, it is super easy(~5 min task) if you already have Enos running. 

## Checklist


addons/api/addon/models/session.js
```diff
@@ -75,16 +75,30 @@ class SessionCredential {
...
 
+  /**
+   * Extracts secrets from the payload secret JSON object.
+   * @param {object} secretJSON - The payload secret JSON object.
+   * @returns {SessionCredential.SecretItem[]} - The array of secret items.
+   */
+  extractSecrets(secretJSON) {
+    // If `decoded` contains a `data` object and if that object has a `username` and `password`,
+    // then we need to return those as key-value pairs
+    // TODO: will there be other labels besides `username` and `password`?
```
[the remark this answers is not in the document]
DhariniJeeva: we discussed this offline, but I'll add the gist for context-  we will be displaying everything that we get from `data` in a key/value copyable format(similar to how we show static creds types such as username/password), and we will not show metadata as that is not useful to the user. 

right now, we will show with the parent level key  

```
data.nested: 'more data'  --> copyable 
other: 'wow' --> copyable 
```

let me your thoughts on this approach.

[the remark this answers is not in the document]
ZedLi: Are we making this change only for vault cred store? I think we should keep the logic the same for static cred since it can be any arbitrary values but it seems like vault returns back it as a specific format with the `data` and `metadata`.

Are you able to return any arbitrary nested object from vault? I'm curious if we need to do the `data.nested` and recurse through the object (and whether we should)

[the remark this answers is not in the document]
DhariniJeeva: the logic is the same for static and vault creds. The JSON blob from static credentials will also now be displayed in key/value format. Pretty much everything sent to decoded will be displayed in key/value format, except `metadata`  as we decided against displaying that to the user. 

I removed the path prepend logic and now show it as a flat single object key and value, so it will be like the below example. I also added screenshots, where I use different cred types from both static and vault and how they would look with this new change

for ex: if the user passes this as JSON static credential,
```
{
  data: {
    nested: "more data"
  },
  other: "wow"
}
```
we will see
```
nested: 'more data'  --> copyable 
other: 'wow' --> copyable 
```

addons/api/addon/utils/flatten-nested-object.js
```diff
@@ -0,0 +1,29 @@
...
+    if (key === 'metadata') {
+      continue;
+    }
+    // Check if the value is an object and it is not null
+    if (typeof obj[key] === 'object' && obj[key] !== null) {
+      // Recursively flatten the object
+      flattenObject(obj[key], result);
+    } else if (obj[key]) {
+      // Only add to result if the value is not null
+      result[key] = obj[key];
```
ZedLi, line 25: What happens if you have nested keys that are the same name?

addons/api/addon/models/session.js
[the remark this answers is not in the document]
ZedLi: I think my point was that static cred stores also have arbitrary json types, should we be unfurling those recursively? Should we just limit this to vault types and only look at the `data` field? What happens with collisions?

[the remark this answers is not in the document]
DhariniJeeva: oops, yes you are right, I will limit it to just vault. I misunderstood when I was looking at static JSON creds, I thought we didn't support key/value and only displayed the JSON. 

 what sort of collisions are you thinking? can you give me an example?

addons/api/addon/utils/flatten-nested-object.js
DhariniJeeva: ohh yeah it will overwrite the previous entry. I will have to preserve the path of the child obj. I will add back the path logic I originally had

DhariniJeeva: but I wonder if it's a good UX to show the `parentKey.key` as the field name 🤔  But if we want to preserve uniqueness and avoid being overwritten then we might have to go with this approach

addons/api/addon/utils/flatten-nested-object.js
```diff
@@ -0,0 +1,25 @@
+/**
+ * Copyright (c) HashiCorp, Inc.
+ * SPDX-License-Identifier: BUSL-1.1
+ */
+
+/**
+ * Flattens a nested object into a single-level object.
+ * @param obj
+ * @param result
```
calcaide, line 9: Are we missing `parentPath` as @param?

calcaide:
The add unit test file `Unit | Utility | flatten-nested-object` its not running. I can't see it run locally, neither on the CI. Even using `ember test --filter="Unit | Utility | flatten-nested-object"` within `api` returns no test found.

addons/api/addon/utils/flatten-nested-object.js
DhariniJeeva: ah yes, added the param

addons/api/tests/unit/utils/flatten-nested-object-test.js
```diff
@@ -0,0 +1,33 @@
...
+      'b.c': 2,
+      'b.d': 3,
+    };
+    assert.deepEqual(flattenObject(input), expected);
+  });
+
+  test('removes keys with null/empty values', function (assert) {
+    const input = {
+      a: null,
+      b: { c: 2 },
```
calcaide, line 26: Can we add a an object that is null at first level?
```
const input = {
      b: { c: null },
    };
```
Just to cover the case when receiving a null object.

addons/api/addon/utils/flatten-nested-object.js
```diff
@@ -0,0 +1,26 @@
...
+ * @param parentPath
+ * @returns {Object}
+ */
+
+export const flattenObject = (obj, result = {}, parentPath = '') => {
```
calcaide, line 14: None blocking comment: I was trying to understand what `parentPath` is, now I do, asking your opinion, do you think `parentKey` can be more descriptive?

My thinking:
if `b: { c: 2, d: 3 },` transforms to `{'b.c': 2, 'b.d': 3}` seems more `parentKey` is b while `key` is c or d, dependending the case

DhariniJeeva: yeah I can change to key, I think that was the name I originally had lol

calcaide (approved):
LGTM!

addons/api/addon/utils/flatten-nested-object.js
```diff
@@ -0,0 +1,26 @@
...
+  for (const key in obj) {
+    // Check if the value is an object and it is not null
+    if (typeof obj[key] === 'object' && obj[key] !== null) {
```
ZedLi, line 17: Won't this also include arrays with the `typeof` check? Ember also has a typeOf [function](https://api.emberjs.com/ember/release/functions/@ember%2Futils/typeOf) to check for specific values

DhariniJeeva: thanks for the link, I did not know ember has its own typeof, I will use that here

addons/api/tests/unit/utils/flatten-nested-object-test.js
```diff
@@ -0,0 +1,39 @@
+/**
+ * Copyright (c) HashiCorp, Inc.
+ * SPDX-License-Identifier: BUSL-1.1
+ */
+
+import { flattenObject } from 'api/utils/flatten-nested-object';
+import { module, test } from 'qunit';
+
+module('Unit | Utility | flatten-nested-object', function () {
+  test('it flattens a nested object to single level object', function (assert) {
```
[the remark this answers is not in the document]
DhariniJeeva: added!

addons/api/addon/models/session.js
```diff
@@ -75,16 +76,31 @@ class SessionCredential {
...
+  /**
+   * Extracts secrets from the payload secret JSON object.
+   * We only need to flatten the object if the type is vault-generic
+   * or vault-ssh-certificate and we only need the data object.
+   * @param {object} secretJSON - The payload secret JSON object.
+   * @returns {SessionCredential.SecretItem[]} - The array of secret items.
+   */
+  extractSecrets(secretJSON, type) {
+    const source =
+      type.includes('vault') && secretJSON?.data
```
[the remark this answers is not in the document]
DhariniJeeva: good call, added this

PR #2587, opened 2024-11-21
fix: 🐛 default to light theme for acceptance tests

cameronperera:
# Description

This is a temporary fix our a11yAudits failing due to the dark color theme. I have work to address that but until that work is started, this will set the default theme to the light color theme. This shouldn't be too much of a change outside of the test that specifically test the theme toggle.

## Checklist


ui/admin/tests/helpers/index.js
```diff
@@ -21,9 +25,10 @@ function setupApplicationTest(hooks, options) {
   // For example, if you need an authenticated session for each
   // application test, you could do:
   //
-  // hooks.beforeEach(async function () {
-  //   await authenticateSession(); // ember-simple-auth
-  // });
+  hooks.beforeEach(function () {
+    authenticateSession();
```
calcaide, line 29: The precious commented code had `await`, do we need the await in this case?

calcaide: Also, if we are adding `authenticateSession()` within the `beforeEach` wrapper, shouldn't we delete the `authenticateSession()` present in each test `beforeEach`?

cameronperera: I removed the `await` because we don't await it anywhere else in our tests. I have run some test to check timing and there wasn't a big difference at all. A few seconds of difference between `await` or no `await`. 

To the other comment, I was going to remove them but I realized keeping them doesn't cause any issues. Some of the `authenticateSession` calls actually have specific data passed into them so I felt it best to keep that local to those test modules.

Another reason to keep them is because this is a temporary change. Eventually, soon I hope, I will start work on the a11y refactor and this code will be removed.

calcaide: @cameronperera 

Regarding the await for `authenticateSession`. You are right, it is not a pattern we have been following. But after doing some research, seems we should be awaiting everywhere we use `authenticateSession`. My thinking:

[As per documentation](https://github.com/mainmatter/ember-simple-auth?tab=readme-ov-file#testing), `authenticateSession` is an async function, therefore we should await to guarantee the session is authenticated before continuing with the test. I do believe is the only way to guarantee we will never run on a race condition.

I understand this specific topic, due to change on pattern, goes outside the topic of this PR. I will start an async discussion within slack to clarify what pattern we should be following, so this will not be a blocker comment.


About the second comment:
Gotcha, you are right, it acts as an "aggregator" to the initial one

calcaide (dismissed):
Awesome work! left a none blocking comment

ui/admin/tests/helpers/index.js
cameronperera: @calcaide Thanks for the detailed response. I also read the thread in slack about this and I am on board. I will update this one spot in my PR so you won't have to catch it in yours. Thanks for getting those together.

calcaide (approved):

DhariniJeeva (approved):
thanks for the fix!

PR #2589, opened 2024-11-22
refactor: 💡 remove rose table references

lisbet-alvarez:
✅ Closes: https://hashicorp.atlassian.net/browse/ICU-15644

# Description
Rose table component is not used anymore. This PR removes the component from rose and any reference to it (css, selectors, etc.)

## Checklist


cameronperera (approved):
Thanks for cleaning all of this up. Great work!

ui/admin/tests/acceptance/credential-store/credentials/read-test.js
```diff
@@ -126,7 +133,13 @@ module('Acceptance | credential-stores | credentials | read', function (hooks) {
         (item) => item != 'read',
       );
     await visit(urls.credentials);
-    assert.dom('.rose-table-body  tr:nth-child(2) a').doesNotExist();
```
calcaide, line 129: None blocking comment: I think, this was a false positive? its been quite a bit we changed those tables to HDS but we did not updated the tests. Because the assertion was a `doesNotExists` it has been passing when the component does not even exists.
Not as part of this PR, but it makes me think we might need to re-think these tests.

calcaide (approved):
Terrific clean up Lis!!!

DhariniJeeva (commented):
thanks for wrapping up table migration, awesome work!

DhariniJeeva (approved):

ui/admin/tests/acceptance/credential-store/credentials/read-test.js
lisbet-alvarez: yes you are correct. The test was using the rose table class names for the selectors to check whether the link was not present when `read` ability was removed . For now, I updated the selectors to look for the link in the table and used the element names instead of class names. I'll add this as a tech time topic.

ZedLi (dismissed):
Thanks for this big cleanup!!!

DhariniJeeva (approved):

PR #2594, opened 2024-11-26
Icu 14777 add more target tests

ZedLi:
# Description
Added some more target tests that utilize enos for e2e. I decided to setup much of the boundary bits needed for testing using HTTP requests. Open to discussion about this.

There's realistically 3 ways to interact with boundary for a lot of our test setups:
1. Using the boundary CLI.
2. Make HTTP requests to the boundary controller
3. Use admin UI

I tried the boundary CLI first but I didn't see much of a benefit using it over HTTP besides @moduli having already built out most of the commands. It also introduces more overhead having to spin up shells to execute the commands.

I tried using admin UI to setup tests which also has the huge benefit of having @moduli already built most of what we need and it worked well except for a few quirks:
1. I had to reach in the admin side of e2e tests to access the pages which I wasn't a fan of, it didn't feel quite right to pull those out to the root either.
2. Some of the pages did make a few assumptions of how they were used so I had to modify them which also requires me to verify the admin UI tests weren't affected. I'm not a fan of having these pages now affecting both sets of tests.
3. The biggest con is it's _slow_ and using an actual browser to set things up takes forever.

I decided mainly to keep things separate between admin and desktop as well as to keep test setups much faster and to just build out the boundary HTTP client requests. It setups up quite fast and and is still straight forward to interact with.

I also updated the README into a shared one at the root.
 
## How to Test
Run through the admin README to setup enos using enterprise boundary. At the e2e root folder, run `yarn desktop`.

## Checklist

### DIFF
--- a/e2e-tests/admin/tests/login.spec.js
+++ b/e2e-tests/admin/tests/login.spec.js
@@ -3,7 +3,7 @@
  * SPDX-License-Identifier: BUSL-1.1
  */
 
-import { test } from '../playwright.config.js';
+import { test } from '../../global-setup.js';
 import { expect } from '@playwright/test';
 
 import { LoginPage } from '../pages/login.js';
--- a/e2e-tests/admin/tests/worker-tags.spec.js
+++ b/e2e-tests/admin/tests/worker-tags.spec.js
@@ -7,7 +7,7 @@ import { test } from '../playwright.config.js';
 import { expect } from '@playwright/test';
 import { customAlphabet } from 'nanoid';
 
-import { authenticatedState } from '../global-setup.js';
+import { authenticatedState } from '../../global-setup.js';
 import { WorkersPage } from '../pages/workers.js';
 
 test.use({ storageState: authenticatedState });
--- a/e2e-tests/admin/tests/worker-tags.spec.js
+++ b/e2e-tests/admin/tests/worker-tags.spec.js
@@ -3,11 +3,10 @@
  * SPDX-License-Identifier: BUSL-1.1
  */
 
-import { test } from '../playwright.config.js';
+import { test, authenticatedState } from '../../global-setup.js';
 import { expect } from '@playwright/test';
 import { customAlphabet } from 'nanoid';
 
-import { authenticatedState } from '../../global-setup.js';
 import { WorkersPage } from '../pages/workers.js';
 
 test.use({ storageState: authenticatedState });
--- a/e2e-tests/desktop/playwright.config.js
+++ b/e2e-tests/desktop/playwright.config.js
@@ -6,6 +6,18 @@ import { defineConfig } from '@playwright/test';
 
 /** @type {import('@playwright/test').PlaywrightTestConfig} */
 export default defineConfig({
+  globalSetup: '../global-setup',
   outputDir: './artifacts',
   workers: 1, // Tests need to be run in serial, otherwise there may be conflicts when using the CLI
+  use: {
+    baseURL: process.env.BOUNDARY_ADDR,
+    storageState: './admin/artifacts/authenticated-state.json',
+    extraHTTPHeaders: {
+      // This token is set in global-setup.js
+      Authorization: `Bearer ${process.env.E2E_TOKEN}`,
+    },
+  },
+  expect: {
+    timeout: 10000,
+  },
 });
--- a/e2e-tests/desktop/tests/targets.spec.js
+++ b/e2e-tests/desktop/tests/targets.spec.js
@@ -149,7 +149,7 @@ test.describe('Targets test', async () => {
         .getByRole('listitem')
         .filter({ hasText: 'private_key' })
         .locator('pre'),
-    ).toContainText('BEGIN OPENSSH PRIVATE KEY');
+    ).toContainText(/BEGIN (OPENSSH|RSA) PRIVATE KEY/);
 
     await authedPage.getByRole('button', { name: 'End Session' }).click();
     await expect(authedPage.getByText('Canceled successfully.')).toBeVisible();
@@ -202,7 +202,7 @@ test.describe('Targets test', async () => {
           .getByRole('listitem')
           .filter({ hasText: 'private_key' })
           .locator('pre'),
-      ).toContainText('BEGIN OPENSSH PRIVATE KEY');
+      ).toContainText(/BEGIN (OPENSSH|RSA) PRIVATE KEY/);
 
       await authedPage.getByRole('button', { name: 'End Session' }).click();
       await expect(
--- a/e2e-tests/desktop/tests/targets.spec.js
+++ b/e2e-tests/desktop/tests/targets.spec.js
@@ -4,18 +4,227 @@
  */
 
 import { expect, test } from '../fixtures/baseTest.js';
+import * as boundaryHttp from '../../helpers/boundary-http';
+
+const hostName = 'Host name for test';
+let org;
+let targetWithHost;
+let targetWithTwoHosts;
+let sshTarget;
+let credential;
+
+test.beforeAll(
+  async ({ request, targetAddress, targetPort, sshUser, sshKeyPath }) => {
+    org = await boundaryHttp.createOrg(request);
+    const project = await boundaryHttp.createProject(request, org.id);
+
+    // Create host set
+    const hostCatalog = await boundaryHttp.createStaticHostCatalog(
+      request,
+      project.id,
+    );
+    const host = await boundaryHttp.createHost(request, {
+      hostCatalogId: hostCatalog.id,
+      address: targetAddress,
+    });
+    const host2 = await boundaryHttp.createHost(request, {
+      hostCatalogId: hostCatalog.id,
+      address: targetAddress,
+      name: hostName,
+    });
+
+    const hostSet = await boundaryHttp.createHostSet(request, hostCatalog.id);
+    await boundaryHttp.addHostToHostSet(request, {
+      hostSet,
+      hostIds: [host.id],
+    });
+
+    // Create a host set with 2 hosts
+    let hostSetWithTwoHosts = await boundaryHttp.createHostSet(
+      request,
+      hostCatalog.id,
+    );
+    hostSetWithTwoHosts = await boundaryHttp.addHostToHostSet(request, {
+      hostSet: hostSetWithTwoHosts,
+      hostIds: [host.id],
+    });
+    hostSetWithTwoHosts = await boundaryHttp.addHostToHostSet(request, {
+      hostSet: hostSetWithTwoHosts,
+      hostIds: [host2.id],
+    });
+
+    // Create a static credential store and key pair credential
+    const credentialStore = await boundaryHttp.createStaticCredentialStore(
+      request,
+      project.id,
+    );
+    credential = await boundaryHttp.createStaticCredentialKeyPair(request, {
+      credentialStoreId: credentialStore.id,
+      username: sshUser,
+      sshKeyPath,
+    });
+
+    // Create tcp target with host set and 1 host
+    targetWithHost = await boundaryHttp.createTarget(request, {
+      scopeId: project.id,
+      type: 'tcp',
+      port: targetPort,
+    });
+    targetWithHost = await boundaryHttp.addHostSource(request, {
+      target: targetWithHost,
+      hostSourceIds: [hostSet.id],
+    });
+    targetWithHost = await boundaryHttp.addBrokeredCredentials(request, {
+      target: targetWithHost,
+      credentialIds: [credential.id],
+    });
+
+    // Create tcp target with host set and 2 hosts
+    targetWithTwoHosts = await boundaryHttp.createTarget(request, {
+      scopeId: project.id,
+      type: 'tcp',
+      port: targetPort,
+    });
+    targetWithTwoHosts = await boundaryHttp.addHostSource(request, {
+      target: targetWithTwoHosts,
+      hostSourceIds: [hostSetWithTwoHosts.id],
+    });
+    targetWithTwoHosts = await boundaryHttp.addBrokeredCredentials(request, {
+      target: targetWithTwoHosts,
+      credentialIds: [credential.id],
+    });
+
+    // Create an SSH target with address
+    sshTarget = await boundaryHttp.createTarget(request, {
+      scopeId: project.id,
+      type: 'ssh',
+      port: targetPort,
+      address: targetAddress,
+    });
+    sshTarget = await boundaryHttp.addInjectedCredentials(request, {
+      target: sshTarget,
+      credentialIds: [credential.id],
+    });
+  },
+);
+
+test.afterAll(async ({ request }) => {
+  if (org) {
+    await boundaryHttp.deleteOrg(request, org.id);
+  }
+});
 
 test.describe('Targets test', async () => {
-  test('Connects to a target and ends session', async ({ authedPage }) => {
-    await authedPage
-      .getByRole('link', { name: 'Generated target with a direct address' })
-      .click();
-
+  test('Connects to a tcp target with one host', async ({ authedPage }) => {
+    await authedPage.getByRole('link', { name: targetWithHost.name }).click();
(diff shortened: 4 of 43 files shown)


e2e-tests/desktop/tests/targets.spec.js
```diff
@@ -4,19 +4,228 @@
...
+import * as boundaryHttp from '../../helpers/boundary-http';
+
+const hostName = 'Host name for test';
+let org;
+let targetWithHost;
+let targetWithTwoHosts;
+let sshTarget;
+let credential;
+
+test.beforeAll(
```
ZedLi, line 16: Only created one set of resources across all tests as DC doesn't really modify the actual resources themselves. But I could still have these in a `beforeEach` or for individual tests if we don't like the sharing of resources.

e2e-tests/desktop/tests/targets.spec.js
```diff
@@ -4,19 +4,228 @@
  */
 
 import { expect, test } from '../fixtures/baseTest.js';
+import * as boundaryHttp from '../../helpers/boundary-http';
```
ZedLi, line 7: I renamed all the `boundary-http` methods to remove the `http` suffix as I didn't feel the consumer needed to know everytime they call each function as you know it by just importing it from the specific folder. 

I used this as an example but I prefer doing `import * as boundaryHttp` since we have both http and CLI.
I could also do the same for the CLI methods, let me know if you have any thoughts on this or prefer the suffixes @moduli.

e2e-tests/desktop/tests/targets.spec.js
```diff
@@ -4,19 +4,228 @@
...
+    targetWithTwoHosts = await boundaryHttp.addHostSource(request, {
+      target: targetWithTwoHosts,
+      hostSourceIds: [hostSetWithTwoHosts.id],
+    });
+    targetWithTwoHosts = await boundaryHttp.addBrokeredCredentials(request, {
+      target: targetWithTwoHosts,
+      credentialIds: [credential.id],
+    });
+
+    // Create an SSH target with address
```
moduli, line 97: Are we going to assume these tests are always going to run against an enterprise boundary instance?

e2e-tests/desktop/tests/targets.spec.js
```diff
@@ -4,19 +4,228 @@
...
+    await authedPage.getByRole('link', { name: sshTarget.name }).click();
     await authedPage.getByRole('button', { name: 'Connect' }).click();
 
     await expect(authedPage).toHaveURL(
       /.*\/scopes\/global\/projects\/sessions/,
     );
 
+    await authedPage.getByRole('tab', { name: 'Shell' }).click();
+    // TODO: Research a better way to test canvas elements for the shell,
+    //  would it be too brittle to assert a snapshot of an expected image?
```
moduli, line 227: snapshots feel potentially brittle, especially with that session timer and the ephemeral port on that page? 
Screenshot 2024-11-26 at 11 26 27 AM

Are we able to get the text contents of the shell? Could we we wait and ensure that the shell gets a shell prompt?

moduli (commented):
I figured the http methods would be the way to go for desktop tests. 

Ran the tests locally. Pretty niceeeeeee.

e2e-tests/desktop/tests/targets.spec.js
ZedLi: Alternatively we can still just call the import as `* as boundary` and only refer to the http or cli when both are getting imported since they realistically are doing nearly the same thing and just changes how it communicates with the controller

moduli: Doesn't feel like that makes much of a difference? I'm cool with how it is now.

e2e-tests/desktop/tests/targets.spec.js
ZedLi: Yeah, that's one other thing I wanted to discuss, as I haven't added a README yet. It assumes right now it's only against an enterprise controller since SSH targets (and some of the vault setup) are pretty important that always needs to be tested which is why I didn't add any tags like in admin and anything we test in CE should work the same in ent.

I'm not sure how much value there is for testing against a CE controller right now but maybe it's something we can split out down the line with tagging if it's needed?

e2e-tests/desktop/tests/targets.spec.js
ZedLi: Unfortunately you can't just pull the text contents out otherwise I would totally do that. Canvas elements are basically images so there's not much to do with it unless you mock it out

e2e-tests/desktop/tests/targets.spec.js
moduli: Yea that makes sense. Just wanted to make sure there weren't any major differences.

moduli (dismissed):

e2e-tests/admin/tests/worker-tags.spec.js
```diff
@@ -3,11 +3,10 @@
  * SPDX-License-Identifier: BUSL-1.1
  */
 
-import { test } from '../playwright.config.js';
+import { test, authenticatedState } from '../../global-setup.js';
 import { expect } from '@playwright/test';
 import { customAlphabet } from 'nanoid';
```
DhariniJeeva, line 8: this is a non blocker- I was just curious why we use `nanoid` instead of the `generateId` method from our mirage? looks like both return alphanumeric chars

e2e-tests/admin/tests/login.spec.js
```diff
@@ -3,7 +3,7 @@
  * SPDX-License-Identifier: BUSL-1.1
  */
 
-import { test } from '../playwright.config.js';
+import { test } from '../../global-setup.js';
 import { expect } from '@playwright/test';
 
 import { LoginPage } from '../pages/login.js';
```
DhariniJeeva, line 9: this is a nitpick- should it say `resources` instead of pages?

e2e-tests/desktop/playwright.config.js
```diff
@@ -6,6 +6,18 @@ import { defineConfig } from '@playwright/test';
 
 /** @type {import('@playwright/test').PlaywrightTestConfig} */
 export default defineConfig({
+  globalSetup: '../global-setup',
   outputDir: './artifacts',
   workers: 1, // Tests need to be run in serial, otherwise there may be conflicts when using the CLI
+  use: {
+    baseURL: process.env.BOUNDARY_ADDR,
+    storageState: './admin/artifacts/authenticated-state.json',
```
DhariniJeeva, line 14: 🤔  do we use the same auth state for both desktop and admin? In that should we delete the artifacts folder within ec2/desktop?

e2e-tests/desktop/playwright.config.js
```diff
@@ -6,6 +6,18 @@ import { defineConfig } from '@playwright/test';
...
+    extraHTTPHeaders: {
+      // This token is set in global-setup.js
+      Authorization: `Bearer ${process.env.E2E_TOKEN}`,
+    },
+  },
+  expect: {
+    timeout: 10000,
```
DhariniJeeva, line 21: is there any difference in using timeout within expect vs using `timeout` option in admin config? 

`  timeout: 90000, // Each test is given 90s to complete`

DhariniJeeva:
curious to know if you considered sharing `playwright.config` between Admin and Desktop? It seems like most things are common.

e2e-tests/desktop/tests/targets.spec.js
DhariniJeeva: 🤔  do we have access to these canvas methods? I think mocking these methods (if we have access) is better than visually testing the shell

ZedLi: We do not, but either way that would defeat the purpose of E2E tests

e2e-tests/admin/tests/worker-tags.spec.js
ZedLi: Yeah this was a convention that was already being used, in practice it doesn't matter too much which one we use, we also could have used `uuid` and maybe should consolidate.

But note that this folder does not have access to the ember addons or helpers anymore so we'd just use the standard packages for it.

e2e-tests/desktop/playwright.config.js
ZedLi: Yes, this is for specifically the timeout of expect statements, while the timeout at the root is for an overall test timeout.

e2e-tests/admin/tests/login.spec.js
ZedLi: No, this is a playwright [convention](https://playwright.dev/docs/pom) for organizing actions

e2e-tests/desktop/playwright.config.js
ZedLi: I was thinking of doing a shared artifacts folder but we use them _slightly_ different right now so haven't pursued it yet, but we could do that.

ZedLi:
> curious to know if you considered sharing `playwright.config` between Admin and Desktop? It seems like most things are common.

I have, right now our configs diverge enough where I didn't want to share it

e2e-tests/desktop/tests/targets.spec.js
```diff
@@ -4,19 +4,228 @@
...
+        .getByRole('link', { name: targetWithTwoHosts.name })
+        .click();
+      await authedPage.getByRole('button', { name: 'Connect' }).click();
+
+      await expect(authedPage).toHaveURL(
+        /.*\/scopes\/global\/projects\/targets/,
+      );
+      await authedPage.getByRole('button', { name: host }).click();
+
+      await expect(authedPage).toHaveURL(
```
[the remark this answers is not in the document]
ZedLi: We could, I also was debating removing the `toHaveURL` assertions so it might be a moot point in the end

e2e-tests/global-setup.js
```diff
@@ -0,0 +1,75 @@
...
+  process.env.E2E_TOKEN = state.authenticated.token;
+
+  await browser.close();
+}
+
+export default globalSetup;
+
+export const authenticatedState = './admin/artifacts/authenticated-state.json';
+
+export const test = baseTest.extend({
```
calcaide, line 45: None blocking comment: can we add a brief comment of what is the purpose of extend these vars? i.e: provide local vars to test context.
I think this is nice, I understand it as a "hub" to provide the test context with values that live in environment variables but without having to access env variables constantly.

e2e-tests/admin/tests/worker-tags.spec.js
calcaide: @DhariniJeeva @ZedLi 

None blocking comment: nice catch about the usage of different libraries for similar purposes. After reading both comments, I think its a good topic to discuss outside the PR context, since it was not introduced within the PR but already present.
Mentioning because recently I am working on a similar consolidation to `uuid` ([ICU-15818](https://hashicorp.atlassian.net/browse/ICU-15818)) and this felt familiar.
If you don't mind I will bring this topic to our next tech time: discuss uuid, nanoid, ember-uuid, generateId and standard crypto.randomUUID. All these libraries/packages are being used on Boundary UI while we could consolidate.

e2e-tests/desktop/playwright.config.js
calcaide: Agreed on the shared artifacts folder being a separate effort.

e2e-tests/desktop/tests/targets.spec.js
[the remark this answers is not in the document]
calcaide: I do think, unless we are having a need for having test selectors within e2e-test, I will keep it simple for now. But if we see we are re-using or repeating these here and there, I will vote for them.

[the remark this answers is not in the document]
calcaide: @ZedLi 
About removing `toHaveURL` at least within this test seems more an assertion to "validate" we are in a expected URL than actually an expectation of the test perse, am I understanding it right?

e2e-tests/helpers/boundary-http/credentials.js
```diff
@@ -0,0 +1,39 @@
...
+ */
+export async function createStaticCredentialKeyPair(
+  request,
+  { credentialStoreId, username, sshKeyPath },
+) {
+  const privateKey = await readFile(sshKeyPath, {
+    encoding: 'utf-8',
+  });
+
+  const response = await request.post(`/v1/credentials`, {
```
calcaide, line 26: @ZedLi 
Perhaps outside the context of this PR, but request has been deprecated and its in maintenance mode, [here link to the info](https://github.com/request/request/issues/3142).
Did we consider using other package to perform http requests? I am not saying we shouldn't use request at all, isn't rocket science and it solves our need, but asking if we consider others due to the request situation.

calcaide (approved):
Awesome work!! very excited for this, the move makes e2e test being first class citizen!! Thank you

Left some none blocking comments 😉

e2e-tests/global-setup.js
ZedLi: Yup I'll do that

e2e-tests/desktop/tests/targets.spec.js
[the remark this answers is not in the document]
ZedLi: Yeah I originally used `toHaveURL` as a smoke test in general when I just wrote the tests to make sure I was in the right place but it's not a great validation since users can't even see the URL normally and this is now reaching into implementation details and we should be validating the page itself

e2e-tests/global-setup.js
```diff
@@ -0,0 +1,75 @@
...
+
+async function globalSetup() {
+  await checkEnv([
+    'BOUNDARY_ADDR',
+    'E2E_PASSWORD_ADMIN_LOGIN_NAME',
+    'E2E_PASSWORD_ADMIN_PASSWORD',
+    'E2E_PASSWORD_AUTH_METHOD_ID',
+  ]);
+
+  // Log in and save the authenticated state to reuse in tests
```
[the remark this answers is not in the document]
ZedLi: Sure, I can extract that out but that's basically the entire `globalSetup` haha

e2e-tests/admin/tests/worker-tags.spec.js
ZedLi: Yeah I agree we should definitely consolidate our usages, we have it scattered in quite a few places.

calcaide:
After running the tests I got 3 of them failing, is this expected?

Screenshot 2024-11-27 at 12 00 55 PM

calcaide:
More info about one of the test failures mentioned above:
Screenshot 2024-11-27 at 12 01 58 PM
Screenshot 2024-11-27 at 12 02 07 PM
Screenshot 2024-11-27 at 12 02 17 PM

ZedLi:
> After running the tests I got 3 of them failing, is this expected?

Ahh I made the assumption the key says `BEGIN OPENSSH KEY`,  I can make that more general

e2e-tests/global-setup.js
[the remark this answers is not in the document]
calcaide: I am fine if you just add a comment saying in case the `globalSetup` grows, we can extract the login part.

I like the suggested separation of concerns because it's a very specific action, log in and take the token. Which now seems most, but not all, part `globalSetup`, to me is cleaner to understand. As mentioned on the comment, I will not block the PR for it 😉

e2e-tests/global-setup.js
calcaide: Oh thanks for the context! one of the few e2e test PR's I missed!

Thanks for the addition

e2e-tests/desktop/tests/targets.spec.js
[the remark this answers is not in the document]
calcaide: 100% agreed

e2e-tests/helpers/boundary-http/credentials.js
calcaide: lol sorry for my fast wrong assumption!

calcaide (approved):

DhariniJeeva (approved):
looks good!

PR #2597, opened 2024-11-27
refactor: 💡 Migrate ember-uuid to uuid

calcaide:
# Description

Migrate ember-uuid to uuid to consolidate since we are already using uuid but ember-uuid only within rose

:tickets: [Jira ticket](https://hashicorp.atlassian.net/browse/ICU-15818)

## How to Test
- Run unit tests within rose.
- Run unit tests within Admin.

### DIFF
--- a/addons/api/mirage/route-handlers/auth.js
+++ b/addons/api/mirage/route-handlers/auth.js
@@ -4,7 +4,6 @@
  */
 
 // import config from '../../config/environment';
-// import { v1 } from 'ember-uuid';
 import { Response } from 'miragejs';
 
 // /**
--- a/addons/rose/addon/utilities/component-auto-id.js
+++ b/addons/rose/addon/utilities/component-auto-id.js
@@ -3,12 +3,12 @@
  * SPDX-License-Identifier: BUSL-1.1
  */
 
-import { v1 } from 'ember-uuid';
+import { v4 as uuidv4 } from 'uuid';
 
 /**
  * A function that generates a component ID, based on UUID.
  * @return {string}
  */
 export function generateComponentID() {
-  return `component-${v1()}`;
+  return `component-${uuidv4()}`;
 }
--- a/ui/admin/package.json
+++ b/ui/admin/package.json
@@ -37,7 +37,8 @@
   },
   "dependencies": {
     "ember-named-blocks-polyfill": "^0.2.5",
-    "lodash": "^4.17.21"
+    "lodash": "^4.17.21",
+    "uuid": "^11.0.3"
   },
   "devDependencies": {
     "@babel/core": "^7.25.2",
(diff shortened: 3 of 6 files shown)


ui/admin/package.json
```diff
@@ -37,7 +37,8 @@
   },
   "dependencies": {
     "ember-named-blocks-polyfill": "^0.2.5",
-    "lodash": "^4.17.21"
+    "lodash": "^4.17.21",
+    "uuid": "^11.0.3"
```
calcaide, line 41: Interesting enough, I add `uuid` as dependency as a safe guard, but without adding it as dep, the test above still working. I think, it works because uuid is already within `node_modules` due to being a dependency on desktop.

Let me knoe your thoughts on this

addons/rose/addon/utilities/component-auto-id.js
```diff
@@ -3,12 +3,12 @@
  * SPDX-License-Identifier: BUSL-1.1
  */
 
-import { v1 } from 'ember-uuid';
+import { v4 as uuidv4 } from 'uuid';
```
calcaide, line 6: In this specific case, I decided to go with `v4`, but `uuid` has `v1` available in case we feel we want to respect the uuid v generated before.

ui/admin/package.json
ZedLi: Yeah we share the same dependencies/node_modules since they're part of the overall monorepo (except for the electron app).

addons/rose/addon/utilities/component-auto-id.js
ZedLi: I would also just use `v4`

yarn.lock
ZedLi, line 1: I think you might have done a `yarn add` without an `install`. It looks like this yarn lock didn't get updated properly

ZedLi (dismissed):

yarn.lock
calcaide: You are right, thanks for the catch!

addons/api/mirage/route-handlers/auth.js
```diff
@@ -4,7 +4,6 @@
  */
 
 // import config from '../../config/environment';
```
DhariniJeeva, line 6: I think we can remove this comment?

calcaide: I didn't want to delete this comment or the rest of commented stuff within this file since is out of the scope of migrating `ember-uuid` to `uuid`.
`config` is used within `authenticateHandler` (which is commented) and I do not know why is commented or even the scope of it.

I will create a ticket to clean up `authenticateHandler`.

calcaide: [Here link to ticket](https://hashicorp.atlassian.net/browse/ICU-15896)

PR #2605, opened 2024-11-29
fix: 🐛 Delete peerDeps within our addons

calcaide:
# Description

Delete `peerDependencies` within our addoncs.

IMHO they are not necessary, they where add when we upgrade Ember to 4.x, when using the blueprints.

Since then, we have seen warnings when performing `yarn install` about code, api and rose have `ember-source` as unmet peerDep, but `ember-source` is installed within all the apps and addons and still.

:tickets: [Jira ticket](https://hashicorp.atlassian.net/browse/ICU-15886)

## Screenshots:

Before:
peerDeps_before

After:
peerDeps_after

## How to Test
- Install deps successfully.


ZedLi (approved):
Yeah it makes no sense especially if they're already a `dependency` it looks like. Though should they have been `peerDeps` and `devDependencies` instead?

DhariniJeeva (approved):

calcaide:
> Yeah it makes no sense especially if they're already a `dependency` it looks like. Though should they have been `peerDeps` and `devDependencies` instead?

@ZedLi 
Not sure I fully understand, but I will try my best to answer:
- lodash: is moved to dependency within `addons/core` to be packed with, since core uses `lodash/debounce`.
- ember-source: is declared as `devDep` in all our projects.

ZedLi:
> ember-source: is declared as `devDep` in all our projects.

Wait I misread it, I thought `ember-source` was a dep. I meant we can either make it a dependency or a combination of `devDependency` and `peerDependency`, so that way the addon has access to it but still only pull in the consuming application's version, is it right that it's only a `devDependency? Isn't it right that it's both a `devDependency` and `peerDependency`?

calcaide:
> > ember-source: is declared as `devDep` in all our projects.
> 
> Wait I misread it, I thought `ember-source` was a dep. I meant we can either make it a dependency or a combination of `devDependency` and `peerDependency`, so that way the addon has access to it but still only pull in the consuming application's version, is it right that it's only a `devDependency`? Isn't it right that it's both a `devDependency` and `peerDependency`?

@ZedLi 

The behaviour by default when creating an addon is next:
- `ember-source` is add as `devDependency`.
- `ember-source` is add as `peerDependency`.

So answering your question, yes, it is right that is both, a `devDependency` and a `peerDependency`.
But for this case, where the addon is built at development time and `ember-source` is declared as `devDep` within the addon, I don't think is incorrect or not right to not include `ember-source` as peerDependency. Neither I have nothing against leaving it as `peerDependency`, but seems like yarn has it.

Before deleting the `peerDependency`, I tried to add as dependency `ember-source: ^4.0.0` to the consumers of the add-ons, but the warnings still showing.
As per my understanding on how `peerDependencies` work is the consumer of a package that has a `peerDependency` is responsible of bring it, we do, but the package manager still warning us we are not doing it? So I though this could be a solution, I believe is the right one, but open to discuss it.

ZedLi:
Hmm yeah I think I forgot that ember handles this differently. I also forgot that `ember-source` was a `devDependency` in the apps. Should lodash (and other deps) have been under `devDependency` in the UI apps as well?

Either way this is fine!

calcaide:
> Hmm yeah I think I forgot that ember handles this differently. I also forgot that `ember-source` was a `devDependency` in the apps. Should lodash (and other deps) have been under `devDependency` in the UI apps as well?
> 
> Either way this is fine!

@ZedLi 

Admin and Desktop have lodash as dependency, In both UI's is use, not much tho, in [Admin its just use in one file](https://github.com/hashicorp/boundary-ui/blob/main/ui/admin/app/routes/scopes/scope/workers/index.js#L46-L49).

My goal with this PR is to tackle peerDeps warnings. But I can create a ticket to review deps vs devDeps in our UI's. Let me know yout thoughts.

Or even to reconsider the usage of it. As mentioned, [in Admin we are importing lodash to use it in one line](https://github.com/hashicorp/boundary-ui/blob/main/ui/admin/app/routes/scopes/scope/workers/index.js#L46-L49).

PR #2606, opened 2024-11-29
fix: 🐛 Fix peerDeps within Rose

calcaide:
# Description

Fix peerDeps within Rose.

Fixed 2 of 3 warnings. The last one, requires us to downgrade `ember-test-helpers` within rose. The peerDep requires `3.x` and we are using `4.x`.

:tickets: [Jira ticket](https://hashicorp.atlassian.net/browse/ICU-15887)

## Screenshots:

Before:
rose_before

After:
rose_after

## How to Test
- Install deps successfully.


cameronperera (approved):

addons/rose/package.json
```diff
@@ -59,11 +59,13 @@
     "broccoli-asset-rev": "^3.0.0",
     "concurrently": "^9.1.0",
     "doctoc": "^2.2.0",
+    "ember-basic-dropdown": "^8.1.0",
```
ZedLi, line 62: Should we be adding peer dependencies for HDS's dependencies? This looks like it's 2 layers down and doesn't feel like something we should be adding

cameronperera: That is a good point. Those 2 packages are in HDS components devDependencies. Maybe they need to be moved to regular dependencies?

calcaide: As @cameronperera  mentioned, both deps are defined as `devDependencies` within HDS, therefore not being packed.
I will reach out to #team-design-systems with the query and check with them on this issue.

I do think, if a library we use decides to carry down the `peerDeps` to the consumer, we haven't much option but to add them. 
As mentioned above, these deps are not packed within HDS but need for HDS to properly work. In this case, kept working bc other deps we use, packed `ember-basic-dropdwon` and `ember-concurrency` so they are present.

ZedLi: Yup, I think it's fine if HDS passes on their peer dependencies to us and it's explicit, but it looked like from the description it was just HDS' dependencies warning us, it sounds like if that's the case HDS should be clearer about what is required

calcaide: @ZedLi @cameronperera 
After checking with #team-design-systems seems like they are working on it, in next versions of HDS this should be solved. [Here slack thread](https://hashicorp.slack.com/archives/C7KTUHNUS/p1733361363288529)

PR #2611, opened 2024-12-05
chore: handle nested data from static json credentials

DhariniJeeva:
# Description
This PR handles nested data from JSON type credentials same as Vault 
https://hashicorp.atlassian.net/browse/ICU-15854

before:
Screenshot 2024-12-04 at 12 45 41 PM

after:
using mocks:
Screenshot 2024-12-05 at 10 29 30 AM

using boundary controller:
Screenshot 2024-12-05 at 10 04 58 AM

## How to Test

- create a static JSON credential with a nested object
- add the credential as a brokered credential to the target
- select the above target ID in the desktop app and click connect
- all the fields of the credential should show in a key/value format
## Checklist


ui/desktop/tests/acceptance/projects/sessions/session-test.js
```diff
@@ -148,13 +148,73 @@ module('Acceptance | projects | sessions | session', function (hooks) {
...
+
+    assert
+      .dom('.secret-container:nth-of-type(1)')
+      .hasText('username ■ ■ ■ ■ ■ ■ ■ ■ ■ ■ ■ ■');
+
+    await click('.secret-container:nth-of-type(1) .hds-icon');
+
+    assert
+      .dom('.secret-container:nth-of-type(1) .secret-content')
+      .hasText('test');
```
ZedLi, line 215: Shouldn't we also be checking/asserting the fields with escaped characters? Actually, did we also have a test for this for vault credential as well?

DhariniJeeva: testing this is quite tricky- dom assertions output contain text as `password\\withslash\\` (with esc chars) but the removal of esc happens when it is displayed on the screen, that's why I didn't include esc chars test. is there a way to test this? 🤔 

`assert.dom('.secret-container:nth-of-type(3) .secret-content').hasText("password\\withslash\\");` 
This is as same as the input we pass to this assert, which to me didn't seem to add any value. I also tried asserting innerHTML, which also shows the esc chars.

DhariniJeeva: > Actually, did we also have a test for this for vault credential as well?

this test is for vault credential, did you mean one for static? I added one for static now

ui/desktop/tests/acceptance/projects/sessions/session-test.js
```diff
@@ -148,13 +148,135 @@ module('Acceptance | projects | sessions | session', function (hooks) {
...
+    await click(TARGET_CONNECT_BUTTON);
+    assert.strictEqual(currentURL(), urls.session);
+
+    assert.dom('.credential-panel-header').exists();
+    assert.dom('.credential-panel-body').exists();
+    assert.dom('.credential-secret').exists();
+
+    assert
+      .dom('.secret-container:nth-of-type(1)')
+      .hasText('username ■ ■ ■ ■ ■ ■ ■ ■ ■ ■ ■ ■');
```
calcaide, line 209: None blocking comment: not sure what is the purpose of testing this.
As per my understanding, we are asserting the secret is by default "hidden". IMHO that's a component test that [we already do here](https://github.com/hashicorp/boundary-ui/blob/main/ui/desktop/tests/integration/components/hidden-secret-test.js#L40-L44).

ui/desktop/tests/acceptance/projects/sessions/session-test.js
```diff
@@ -148,13 +148,135 @@ module('Acceptance | projects | sessions | session', function (hooks) {
...
+
+    await click('.secret-container:nth-of-type(1) .hds-icon');
+
+    assert
```
calcaide, line 213: I think we are missing some assertions here, my thoughts:

The implementation put together here is to handle nested data from static json credentials, and if the given data is:

```json
data: {
  backslash: 'password\\withslash\\',
  backslash1: 'password\\fwithslash\\f',
  password: '123',
  username: 'test',
  email: {
    address: 'test.com',
  },
```

I will assert each one, specially the `email` one, since is the nested object within this data set.

Let me know if I am missing something.

calcaide (changes requested):
Nice work! Left some comments 😉

ui/desktop/tests/acceptance/projects/sessions/session-test.js
ZedLi: > this test is for vault credential, did you mean one for static? I added one for static now

I meant for the previous PR that handled the vault credential, did we also have a test for escape characters?

I would use template strings to check against (which will display it raw with no escaping) and compare that to what's being displayed. Using normal strings will always have to have escape characters.

This is also something we should add an E2E test for (not for this PR) but this is something we definitely should validate and getting in the habit of adding

DhariniJeeva: agreed that was my plan to handle this in e2e, added a [ticket](https://hashicorp.atlassian.net/browse/ICU-15986) for e2e test.

DhariniJeeva: I can add more details and link slack conversation to this ticket in a bit, can you give me a link to Admin UI where we show this credential panel? As far as I'm aware, this is solely a desktop change, so I'm wondering if I missed this 

this is a parent [ticket](https://hashicorp.atlassian.net/browse/ICU-13856) where the issue is reported

calcaide: We have no showing credentials panel within Admin, but following the [new ticket created](https://hashicorp.atlassian.net/browse/ICU-15986) what I mean is:
- The ticket created is under Platform Maintenance, wich is a none scoped parent. In other words, can apply to anything within the code base.
- The ticket created title is: "e2e test for credentials display" and has no description. So someone, can missunderstood it as add an e2e test for the credentials store screen within Admin.

I am not sure we discuss anything in slack, but within this PR comment.

DhariniJeeva: ah ok I thought you were talking about this PR's ticket. 

I updated the e2e ticket and added further context and linked parent issue

ui/desktop/tests/acceptance/projects/sessions/session-test.js
DhariniJeeva: I refactored the tests and included an assert for this

ui/desktop/tests/acceptance/projects/sessions/session-test.js
DhariniJeeva: yeah, the component handles that part, I refactored the test to only cover nested and esc chars which are covered in this PR. 

I added this originally as there were no tests covering a session with a credential. We only had one test case that covered "visiting session with no credential" but there was nothing to test the case, "visiting session with a credential". 

After some thinking, I refactored this to two new test cases that are addressed in this PR and previous PR related to this [work](https://hashicorp.atlassian.net/browse/FRB-192) 

1. visiting a session with a static credential (handle nested without esc chars)
2. visiting a session with a vault credential (handle nested without esc chars)

ui/desktop/tests/acceptance/projects/sessions/session-test.js
DhariniJeeva: @ZedLi I added tests for both static and vault that handle nested data and esc chars, thanks for the feedback on using template strings.

ui/desktop/tests/acceptance/projects/sessions/session-test.js
```diff
@@ -148,13 +148,126 @@ module('Acceptance | projects | sessions | session', function (hooks) {
...
+    assert.dom('.credential-panel-header').exists();
+    assert.dom('.credential-panel-body').exists();
+    assert.dom('.credential-secret').exists();
+
+    // Check if nested data is displayed in a key/value format without escape characters
+    assert
+      .dom('.secret-container:nth-of-type(1)')
+      .hasText('backslash ■ ■ ■ ■ ■ ■ ■ ■ ■ ■ ■ ■');
+    await click('.secret-container:nth-of-type(1) .hds-icon');
+    const expectedOutput = String.raw`password\withslash\ `;
```
DhariniJeeva, line 209: I added a space at the end to avoid linter error since I'm using a backslash and we want to test the UI displays this output with no additional esc chars.

ZedLi: Are you relying on `hasText` trimming whitespace? Wouldn't it be different output then?

What's the linter issue, is it because of the trailing backslash which confuses it? If that's the issue, why not have the backslashes just in the middle instead (and maybe include things like `\t`, `\0`, etc)

ui/desktop/tests/acceptance/projects/sessions/session-test.js
```diff
@@ -148,13 +148,126 @@ module('Acceptance | projects | sessions | session', function (hooks) {
...
+
+    assert.dom('.credential-panel-header').exists();
+    assert.dom('.credential-panel-body').exists();
+    assert.dom('.credential-secret').exists();
+
+    // Check if nested data is displayed in a key/value format without escape characters
+    const expectedOutput = String.raw`password\withslash\ `;
+    assert
+      .dom('.secret-container:nth-of-type(1)')
+      .hasText('nested_secret.complex_nest.blackslash ■ ■ ■ ■ ■ ■ ■ ■ ■ ■ ■ ■');
```
ZedLi, line 264: It seems weird to include the ■ characters, why not just use an `includesText` assert instead?

ZedLi (dismissed):
Looks good! Left a few more minor comments

ui/desktop/tests/acceptance/projects/sessions/session-test.js
DhariniJeeva: good point, I changed to includes

ui/desktop/tests/acceptance/projects/sessions/session-test.js
DhariniJeeva: I switched to `strictEqual` to be more accurate, the `assert.Dom` and `hasText` did not throw an error when there was an additional space at the end, I believe it ignores the white space

ZedLi: How come the new assert uses a `trim()` then? Does the text normally have trailing or leading whitespace?

[hasText](https://github.com/mainmatter/qunit-dom/blob/master/API.md#hastext) normally strips the whitespace

DhariniJeeva: I don't know if it normally has whitespaces, but I added trim() as a precaution in case of whitespaces

ZedLi: If that's the case, what's the difference between what you changed it to vs using `hasText`? Since it seems like we're stripping whitespace either way

DhariniJeeva: yeah, it's pretty much the same now,  I didn't want to rely on `hasText` to remove whitespaces and wanted to be more readable with trim and strictequal but yeah we could just use hasText too.

ZedLi: Yeah I think it'd be better to just stick with `assert.dom` and `hasText` then since it seems they do the same thing

ZedLi (dismissed):

ui/desktop/tests/acceptance/projects/sessions/session-test.js
```diff
@@ -148,13 +148,133 @@ module('Acceptance | projects | sessions | session', function (hooks) {
...
+        '.secret-container:nth-of-type(1) .secret-content',
+      ).textContent.trim(),
+      expectedOutput,
+    );
+    assert
+      .dom('.secret-container:nth-of-type(2)')
+      .includesText('email.address');
+    await click('.secret-container:nth-of-type(2) .hds-icon');
+
+    assert.strictEqual(
```
calcaide, line 219: Same comment [as here](https://github.com/hashicorp/boundary-ui/pull/2611/files#r1881073402)

ui/desktop/tests/acceptance/projects/sessions/session-test.js
```diff
@@ -148,13 +148,133 @@ module('Acceptance | projects | sessions | session', function (hooks) {
...
+
+    assert.dom('.credential-panel-header').exists();
+    assert.dom('.credential-panel-body').exists();
+    assert.dom('.credential-secret').exists();
+
+    // Check if nested data is displayed in a key/value format without escape characters
+    const expectedOutput = String.raw`password\with\tslash`;
+    assert
+      .dom('.secret-container:nth-of-type(1)')
+      .includesText('nested_secret.complex_nest.blackslash');
```
calcaide, line 268: I will recommend using `.hasText` rather than `.includesText`, its more rigid and looks for exact string, kindda strictEqual.

ui/desktop/tests/acceptance/projects/sessions/session-test.js
```diff
@@ -148,13 +148,133 @@ module('Acceptance | projects | sessions | session', function (hooks) {
...
+    await click('.secret-container:nth-of-type(1) .hds-icon');
+    assert.strictEqual(
```
calcaide, line 270: Same comment [as here](https://github.com/hashicorp/boundary-ui/pull/2611/files#r1881073402)

ui/desktop/tests/acceptance/projects/sessions/session-test.js
DhariniJeeva: I changed this based on this [feedback](https://github.com/hashicorp/boundary-ui/pull/2611#discussion_r1878882834) and we don't really need those secret hidden boxes and only care about the text

ui/desktop/tests/acceptance/projects/sessions/session-test.js
```diff
@@ -148,13 +148,133 @@ module('Acceptance | projects | sessions | session', function (hooks) {
...
+
+    assert.dom('.credential-panel-header').exists();
+    assert.dom('.credential-panel-body').exists();
+    assert.dom('.credential-secret').exists();
+
+    // Check if nested data is displayed in a key/value format without escape characters
+    assert.dom('.secret-container:nth-of-type(1)').includesText('backslash');
+    await click('.secret-container:nth-of-type(1) .hds-icon');
+    const expectedOutput = String.raw`password\with\tslash`;
+    assert.strictEqual(
```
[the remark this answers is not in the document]
DhariniJeeva: got it, updated!

ZedLi (approved):

calcaide (approved):
LGTM!! Thanks for the work 🙌

cameronperera (approved):

PR #2614, opened 2024-12-05
Llb/worker filter improvements

cameronperera:
# Description

This is all of the worker filter improvement changes ready to be merged into main. All of this work has been approved in smaller PRs into the llb but we wanted to give the whole team time to review these changes.

## How to Test

The changes updated how we add or edit worker filters for Targets, Vault Credential Stores, and Storage Buckets.
You should also see a "Filter Generator" now as well. It should be hidden by default and has a toggle to show it if needed. The generator does not persist any data and should be empty any time it is shown.

## Checklist


lisbet-alvarez (approved):
🥳

ui/admin/app/components/form/storage-bucket/aws/index.hbs
```diff
@@ -453,30 +453,41 @@
...
+      aria-describedby={{helpId}}
+      class='storage-bucket-worker-filter'
+      @isRequired={{true}}
+      as |F|
+    >
+      <F.Legend id={{labelId}}>{{t
+          'resources.storage-bucket.form.worker_filter.label'
+        }}</F.Legend>
+      <F.HelperText id={{helpId}}>
+        {{t 'resources.storage-bucket.form.worker_filter.help'}}
```
laurenolivia, line 468: nit: @cameronperera I'm wondering if the text translation for this line where it reads `Boolean expression.` could be a clearer, complete sentence?

ui/admin/app/templates/scopes/scope/targets/target/workers.hbs
```diff
@@ -51,9 +51,13 @@
                   'resources.target.workers.accordion-label.ingress-workers'
                 }}</span>
               <Hds::Link::Standalone
-                @icon='edit'
+                @icon={{if @model.ingress_worker_filter 'edit' 'plus-circle'}}
```
laurenolivia, line 54: nit: there's text right above this dropdown that feels kind of like it's floating between two sections, rather than helper text that sits on top of the ingress/egress accordions. Wdyt?
Screenshot 2024-12-09 at 11 44 49 AM

laurenolivia (commented):
Great job! ✨ 
Left a few comments.

ui/admin/app/components/worker-filter-generator/index.hbs
```diff
@@ -0,0 +1,196 @@
...
+      <F.HelperText id={{helpId}}>{{t
+          'worker-filter-generator.filter-generator.formatted-result.description'
+        }}</F.HelperText>
+      <F.Control>
+        <Hds::Text::Body @tag='p' @weight='semibold'>{{t
+            'worker-filter-generator.filter-generator.formatted-result.label'
+          }}</Hds::Text::Body>
+        <div class='generated-results-container'>
+          <Hds::Form::TextInput::Base
+            readonly
```
DhariniJeeva, line 185: don't we need to pass in the boolean value?

ui/admin/app/components/form/storage-bucket/aws/index.hbs
cameronperera: Great feedback. We pull these descriptions from the protobufs. The one for storage buckets can be found [here](https://github.com/hashicorp/boundary/blob/main/internal/proto/controller/api/resources/storagebuckets/v1/storage_bucket.proto#L89) and it actually looks like it was updated. I can update ours as well to match.

ui/admin/app/components/form/storage-bucket/aws/index.hbs
```diff
@@ -453,30 +453,41 @@
...
+      <F.Control>
+        {{#if form.disabled}}
+          <Hds::CodeBlock
+            @language='bash'
+            @value={{@model.worker_filter}}
+            @hasCopyButton={{true}}
+            data-test-worker-filter
+          />
+        {{else}}
+          <WorkerFilterGenerator
```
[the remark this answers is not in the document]
cameronperera: I am not sure I understand the question. Could you elaborate more please?

ui/admin/app/components/worker-filter-generator/index.hbs
cameronperera: Great question. This input will never be modified directly so the readonly should persist no matter what state the generator is in.

ui/admin/app/templates/scopes/scope/targets/target/workers.hbs
cameronperera: I see what you mean. We should reach out to design and see what they think about this.

ui/admin/app/components/form/storage-bucket/aws/index.hbs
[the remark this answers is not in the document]
laurenolivia: There is no limit on the number of expressions that a user can add to the generator so a page could, in theory, look like the screenshot below and I wonder if we want to allow that:
Screenshot 2024-12-09 at 6 42 05 PM

ui/admin/app/components/form/storage-bucket/aws/index.hbs
laurenolivia: Thank you! 👍

ui/admin/app/components/form/storage-bucket/aws/index.hbs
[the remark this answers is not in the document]
calcaide: I don't think we should limit the number of expressions a user can add. Also, not sure yet how users feel with UX, but in case we want to limit something, we can limit the high the code block uses, therefore, if the code content is big, it will be scrollable.
[HDS docs about limit height](https://helios.hashicorp.design/components/code-block?tab=code#limit-height)

ui/admin/app/templates/scopes/scope/targets/target/workers.hbs
calcaide: Not sure I am understanding, but this text, is part of the worker diagram component. Isn't this intendeed?

ui/admin/app/components/form/storage-bucket/aws/index.hbs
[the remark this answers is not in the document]
cameronperera: Oh I see. I would agree with @calcaide that we shouldn't limit the user's input and I like the idea of a max height for the code block component. I will a max height after reaching out to design about this and the other spacing question.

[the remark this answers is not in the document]
cameronperera: Added a max height for the code blocks.

ui/admin/app/templates/scopes/scope/targets/target/workers.hbs
cameronperera: Talked with design and we will keep this as is for now.

ui/admin/app/styles/app.scss
```diff
@@ -1056,3 +1056,36 @@
...
+
+// worker filter generator
+.show-filter-generator {
+  .hds-form-field--layout-flag {
+    margin-bottom: 0;
+  }
+}
+
+.filter-generator-selection {
+  .hds-form-group__control-field {
```
ZedLi, line 1068: It seems like we're overriding our own style, is this right? We have a top level class just to override our own custom style?

ui/admin/app/styles/app.scss
```diff
@@ -1056,3 +1056,36 @@
...
+.generated-results-container {
+  display: flex;
+  flex-direction: row;
+  gap: 0.5rem;
+  align-items: flex-start;
+}
+
+// storage buckets
+.storage-bucket-worker-filter {
+  .hds-form-field--layout-flag {
```
ZedLi, line 1088: Same with this style, looks like it's our own style being overridden?

addons/core/translations/en-us.yaml
```diff
@@ -133,3 +133,28 @@ settings:
...
+    tag:
+      label: Tag
+      helper: Format a worker tag
+    name:
+      label: Name
+      helper: Format a worker name
+    operator:
+      matches: matches
+      contains: contains
+      default: --Select an option--
```
ZedLi, line 160: We also have a very similar copy here https://github.com/hashicorp/boundary-ui/blob/b0271096ccc38c3bd3d09ed7efe72339cde689a2/addons/core/translations/en-us.yaml#L39

Do we need more than one?

cameronperera: Good catch. I will update that to the existing one so it is the same across the app.

cameronperera: Fixed.

ui/admin/app/components/worker-filter-generator/index.hbs
```diff
@@ -0,0 +1,196 @@
...
+            id='generated-value'
+            name='generated_value'
+            @value={{this.generatedResult}}
+            @width='320px'
+          />
+          <Hds::Copy::Button @text='Copy' @targetToCopy='#generated-value' />
```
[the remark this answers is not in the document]
cameronperera: Cannot think of why we went that route but happy to change it. I believe the copy fails because the generated results has no value to copy. When I changed it over to `@textToCopy` it doesn't fail but it does throw an error in the console.
```
index.js:153 Uncaught (in promise) Error: Assertion Failed: `hds-clipboard` modifier - either a `text` or a `target` argument is required
```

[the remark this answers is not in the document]
ZedLi: Should we disable the button if there's no text then?

ui/admin/app/components/form/storage-bucket/aws/index.hbs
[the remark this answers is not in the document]
laurenolivia: That idea works too! I just wanted to avoid what the UI looks such as the case provided above.

ui/admin/app/components/worker-filter-generator/index.hbs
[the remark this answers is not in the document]
cameronperera: After some thought and reviewing [HDS patterns](https://helios.hashicorp.design/patterns/disabled-patterns), I have decided to hide the button if there is no result to copy. Still happy to hear thoughts and suggestions for this though.

https://github.com/user-attachments/assets/86a8e91f-77b0-423f-a383-15d01df5db57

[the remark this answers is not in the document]
ZedLi: Looks great!

ui/admin/app/styles/app.scss
```diff
@@ -1056,3 +1056,36 @@
     padding: 0.25rem 0.75rem;
   }
 }
+
+// worker filter generator
+.show-filter-generator {
```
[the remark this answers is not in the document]
cameronperera: Good catch. I think it was left over from a refactor.

ui/admin/app/styles/app.scss
cameronperera: Yeah we have a global style that is applying margin-bottom and we needed remove it specifically for this one spot. We actually do it in a couple of places. So we have the `.filter-generator-selection` class to narrow the changes to just that spot.

ui/admin/app/styles/app.scss
cameronperera: This one we needed because the Hds::Separator adds top and bottom to itself and the margin-bottom from the global style made the spacing too much.

ui/admin/app/components/credential-stores/credential-store/actions/index.hbs
```diff
@@ -18,6 +18,12 @@
       {{t 'resources.credential.actions.create'}}
     </dd.Interactive>
   {{/if}}
+  {{#if (and (feature-flag 'vault-worker-filter') @model.isVault)}}
```
calcaide, line 21: None blocking comment: asking for learning purposes, did we introduce the feature-flag `vault-worker-filter` within this initiative? or it was already present?

ui/admin/app/styles/app.scss
ZedLi: I think if we're starting to do this we should refactor this. We shouldn't be overriding our own styles as this becomes a nightmare to change due to specificity in the future.

ui/admin/app/components/credential-stores/credential-store/actions/index.hbs
cameronperera: That flag has been present for a while. It is an enterprise specific capability.

ui/admin/app/styles/app.scss
ZedLi: Sounds good, to be clear if the vast majority of the use case is valid in the global style, we should keep it. I wasn't sure how commonly we've been overriding this style at this point but if this has been more than just a few places, at the minimum we should make a utility class so we're not cluttering our CSS files just to override this same style as we're doing here.

lisbet-alvarez (approved):
🥳

ui/admin/app/components/credential-stores/credential-store/actions/index.hbs
calcaide: @cameronperera 
Gotcha. I wasn't aware and bc I couldn't see it on the feature dropdown I was 🤔 is this new? But all clear now 😄

ui/admin/app/components/form/storage-bucket/aws/index.hbs
```diff
@@ -456,30 +456,41 @@
...
-    {{#if @model.errors.worker_filter}}
-      <F.Error data-test-worker-filter-error as |E|>
-        {{#each @model.errors.worker_filter as |error|}}
-          <E.Message>{{error.message}}</E.Message>
-        {{/each}}
-      </F.Error>
-    {{/if}}
-  </Hds::Form::TextInput::Field>
+  {{#let (unique-id) (unique-id) as |labelId helpId|}}
+    <Hds::Form::Fieldset
```
calcaide, line 460: Is this element going to repeat within the screen? or we will have just 1? I am wondering why we need a `unique-id` here.

My thinking:
I understand we need to define `aria-labelledby` and `aria-describedby` for a11y purposes, but if I am not mistaken, this same attributes are to provide a "human readable" id, while unique-id is more "computer" than "human" language.

Thoughts?

ui/admin/app/components/form/storage-bucket/minio/index.hbs
```diff
@@ -315,30 +315,42 @@
...
-    </F.HelperText>
-    {{#if @model.errors.worker_filter}}
-      <F.Error data-test-worker-filter-error as |E|>
-        {{#each @model.errors.worker_filter as |error|}}
-          <E.Message>{{error.message}}</E.Message>
-        {{/each}}
-      </F.Error>
-    {{/if}}
-  </Hds::Form::TextInput::Field>
+  {{#let (unique-id) (unique-id) as |labelId helpId|}}
```
calcaide, line 318: Waiting on response, but same [comment here](https://github.com/hashicorp/boundary-ui/pull/2614/files#r1881229832)

ui/admin/app/components/worker-filter-generator/index.hbs
```diff
@@ -0,0 +1,198 @@
...
+      <F.Label>{{t
+          'worker-filter-generator.filter-generator.name.label'
+        }}</F.Label>
+      <F.HelperText>{{t
+          'worker-filter-generator.filter-generator.name.helper'
+        }}</F.HelperText>
+    </G.RadioField>
+  </Hds::Form::Radio::Group>
+
+  {{#let (unique-id) (unique-id) as |labelId helpId|}}
```
calcaide, line 91: [Same comment here](https://github.com/hashicorp/boundary-ui/pull/2614/files#r1881229832)

ui/admin/app/components/worker-filter-generator/index.hbs
```diff
@@ -0,0 +1,198 @@
...
+        }}</F.Legend>
+      <F.HelperText id={{helpId}}>{{t
+          'worker-filter-generator.filter-generator.input-values.description'
+        }}</F.HelperText>
+      <F.Control>
+        {{#if (eq this.selectedGeneratorType this.generatorTagType)}}
+          <Hds::Form::TextInput::Field
+            @value={{this.key}}
+            name='tag_key'
+            @width='320px'
```
calcaide, line 109: None blocking comments: I see the selects all defining a `@width='320px'` totally fine with it, but wondering, since all of them are within `<F.Control>` adding a width to the container will bring same effect.

ui/admin/app/components/worker-filter-generator/index.hbs
```diff
@@ -0,0 +1,198 @@
...
+          {{on 'input' (set-from-event this 'value')}}
+          as |F|
+        >
+          <F.Label>{{t 'form.value.label'}}</F.Label>
+        </Hds::Form::TextInput::Field>
+      </F.Control>
+    </Hds::Form::Fieldset>
+  {{/let}}
+
+  {{#let (unique-id) (unique-id) as |labelId helpId|}}
```
calcaide, line 163: [Same comment here](https://github.com/hashicorp/boundary-ui/pull/2614/files#r1881229832)

ui/admin/app/components/form/storage-bucket/aws/index.hbs
ZedLi: This is true for `aria-label` but `aria-labeledby` and `aria-describedby` describe the _element_ that has the actual human readable text. In that case, it's perfectly fine to use a unique ID that's not very readable as that's now what is actually used by the assistive technology

ui/admin/app/components/worker-filter-generator/index.hbs
cameronperera: I played around with the suggestion but it seems `<F.Control>` does not like getting a `@width` attribute.

ui/admin/app/components/form/storage-bucket/aws/index.hbs
cameronperera: Oooh, so these are simply going to associate the label and help text to the fieldset but not the context used by the assistive technology?

calcaide: As per mdn, [seems about right what Zed mentions](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-describedby):
```
The aria-describedby attribute lists the [id](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/id)s of the elements that describe the object. It is used to establish a relationship between widgets or groups and the text that describes them.
```
I am fine either way, but when using those, if there is no repetition of the elements, I personally prefer a human readable string than a generated uuid that changes by every render loop/refresh.

ZedLi: > Oooh, so these are simply going to associate the label and help text to the fieldset but not the context used by the assistive technology?

It's the equivalent of a label's [for](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/for) in cases where you don't actually have a label element. 

> I am fine either way, but when using those, if there is no repetition of the elements, I personally prefer a human readable string than a generated uuid that changes by every render loop/refresh.

That's fair, I don't have a strong opinion on this so don't mind either way but just a static ID would also work in this case. However, I would keep the unique ID in the component as you don't exactly know how a consumer will use it.

calcaide: That is correct. Being a component, unique-id is the right way to go.

PR #2615, opened 2024-12-06
Icu 14778 add more sessions tests

ZedLi:
✅ Closes: https://hashicorp.atlassian.net/browse/ICU-14778

# Description
Added more session tests. Also added support for dotenv which lets us use .env files and removed usages of relative paths which makes it difficult to run if you're not in the root.

Let me know if you get any test failures, there were flaky tests at times due to sessions being canceled/terminated which can affect other test runs but it _should_ be consistent now.

## How to Test
Run `yarn desktop` in the e2e folder.

## Checklist

### DIFF
--- a/e2e-tests/desktop/fixtures/electronTest.js
+++ b/e2e-tests/desktop/fixtures/electronTest.js
@@ -16,6 +16,7 @@ import { fileURLToPath } from 'url';
 const getExecutablePath = (buildDirectory = 'out') => {
   // Using process.cwd() can change depending on where you run the tests so we use the current file location
   // __dirname isn't available in ES modules so we need to indirectly get the directory
+  // TODO: Switch to `import.meta.dirname` when we switch to Node.js 20+
   const __filename = fileURLToPath(import.meta.url);
   const __dirname = path.dirname(__filename);
   const rootDir = path.resolve(__dirname, '../../../ui/desktop/electron-app');
--- a/e2e-tests/desktop/fixtures/electronTest.js
+++ b/e2e-tests/desktop/fixtures/electronTest.js
@@ -6,6 +6,7 @@
 import { test } from '@playwright/test';
 import path from 'path';
 import fs from 'fs';
+import { fileURLToPath } from 'url';
 
 /**
  * Get the path to the Boundary Desktop executable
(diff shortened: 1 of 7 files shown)


e2e-tests/global-setup.js
```diff
@@ -42,7 +51,10 @@ const authenticateToBoundary = async () => {
 
 export default globalSetup;
 
-export const authenticatedState = './admin/artifacts/authenticated-state.json';
+export const authenticatedState = path.resolve(
```
ZedLi, line 54: I moved the authentication file to its own `.auth` folder, as it seems [suggested](https://playwright.dev/docs/auth#core-concepts) by playwright

e2e-tests/desktop/fixtures/electronTest.js
```diff
@@ -6,14 +6,21 @@
...
 /**
  * Get the path to the Boundary Desktop executable
  * @param buildDirectory
  * @return {string}
  */
 const getExecutablePath = (buildDirectory = 'out') => {
-  const rootDir = path.resolve('../ui/desktop/electron-app');
+  // Using process.cwd() can change depending on where you run the tests so we use the current file location
+  // __dirname isn't available in ES modules so we need to indirectly get the directory
+  // TODO: Switch to `import.meta.dirname` when we switch to Node.js 20+
```
calcaide, line 19: Since the decouple of e2e test with the UI's, which was a good call, I do think e2e can diverge on what node version they run. In other words, I do think e2e test should use node 22 or even 24.
The e2e tests are not part of any related UI automation or procedure. I think using latest version can benefit more than harm the e2e suite.
The only "con" I see, is the developer running them might need to change the node version within its environment, since we work most of our time on node20, but that's it.

What are your thoughts?

ZedLi: Yeah I could make it a requirement to use a more recent node. I wasn't sure what others were still running so I didn't want to have people switch node versions if people were still using <20 but I have no problems with making it a requirement and using newer stuff.

24 doesn't seem to be in LTS yet (or even current) so I think 22 would be the highest we should go. The property was available starting in node 20 so I could just require >= 20 since we're about to upgrade our node versions across our CI and already support it in our apps if that's the case, I don't mind changing this

calcaide: Yep, aiming for LTS is the right way. I mentioned v24 as an example of if running latest benefits the e2e suite, due to a new feature or something similar I am down for it 😉 .

The switching versions, in this case I think its totally fine, we are doing it for the benefit of the e2e, also as you mention, as soon we are unblock with the windows runner we will roll out node upgrade anyways.

e2e-tests/global-setup.js
```diff
@@ -6,6 +6,15 @@
...
 import { checkEnv } from './helpers/general.js';
 import { LoginPage } from './admin/pages/login.js';
+import { fileURLToPath } from 'url';
+import path from 'path';
+
+const __filename = fileURLToPath(import.meta.url);
+const __dirname = path.dirname(__filename);
+
+// Import environment variables from .env file if available
+import dotenv from 'dotenv';
```
calcaide, line 16: None blocking commnet: we usually have the `imports` block first thing on files, is there a reason for this one to be out of the import block?

e2e-tests/desktop/fixtures/electronTest.js
ZedLi: Updated to use 20 and removed the extra code for versions below node 20

cameronperera (approved):

PR #2617, opened 2024-12-06
fix: 🐛 update @name attribute for RadioCard Groups

cameronperera:
✅ Closes: https://hashicorp.atlassian.net/browse/ICU-15676

# Description

This fixes an a11y audit issue where the `name` attribute for our radio card groups did not match. In a few places, the name matched the value but the name should be the same for all radio options.
I opted to simply use `Type` as the name value in most cases, except `plugin_type` which I created a new translation entry for. If we feel we need to be more specific for each resource type I can make that refactor.

Before:
Screenshot 2024-12-06 at 4 10 22 PM
After:
Screenshot 2024-12-06 at 4 09 29 PM

## How to Test

To test you can use the vercel build and create a new credential store. When you inspect the input for the Radio Cards the `name` attribute should be the same for both now.

## Checklist

### DIFF
--- a/addons/core/translations/form/en-us.yaml
+++ b/addons/core/translations/form/en-us.yaml
@@ -18,6 +18,8 @@ description:
 type:
   label: Type
   help: The type of the resource, to help differentiate schemas
+plugin_type:
+  label: plugin_type
 login_name:
   label: Login Name
   help: The account login name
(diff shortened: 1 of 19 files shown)


ZedLi (approved):
Thanks for the cleanup!

PR #2619, opened 2024-12-10
Change validation field from name (optional) to address (required)

calcaide:
# Description

Change the field used for validation test. Previously, we were using `name`, but for host creation, `name` is an optional field, changed to `address` which is required field to make the test more realistic.

## How to Test
- Run the test suite successfull.


lisbet-alvarez (approved):
Nice catch!

ZedLi (approved):

cameronperera (approved):

PR #2623, opened 2024-12-11
[ICU-15647] Replace Rose::Dialog with Hds::Modal in admin ui only

laurenolivia:
# Description
Replace `Rose::Dialog` component with `Hds::Modal` component in admin ui.

This change caused 63 tests to fail, which have also been fixed in this PR.

:tickets: [Jira ticket](https://hashicorp.atlassian.net/browse/ICU-15647)

## Screenshots 📸 
Before 👇 
Screenshot 2024-12-11 at 2 05 33 PM

After 👇 
Screenshot 2024-12-20 at 2 03 45 PM

## Recording 📹

https://github.com/user-attachments/assets/e25f49ce-7820-4592-8014-c4af4734af65

## How to Test
1. Navigate to Targets
2. Enter any text in a required field
3. Navigate to another section in navbar to trigger modal

## Checklist

### DIFF
--- a/addons/rose/app/styles/rose/components/_index.scss
+++ b/addons/rose/app/styles/rose/components/_index.scss
@@ -3,6 +3,8 @@
  * SPDX-License-Identifier: BUSL-1.1
  */
 
+// This list is sorted in ascending order to improve readability
+
 @import 'button';
 @import 'card';
 @import 'cards';
--- a/addons/rose/app/styles/rose/components/_index.scss
+++ b/addons/rose/app/styles/rose/components/_index.scss
@@ -11,8 +11,8 @@
 @import 'dropdown';
 @import 'form';
 @import 'frame';
-@import 'icon';
 @import 'header';
+@import 'icon';
 @import 'layout';
 @import 'list';
 @import 'metadata-list';
--- a/addons/rose/app/styles/rose/components/_index.scss
+++ b/addons/rose/app/styles/rose/components/_index.scss
@@ -4,18 +4,18 @@
  */
 
 @import 'button';
+@import 'card';
+@import 'cards';
+@import 'code-editor';
 @import 'dialog';
 @import 'dropdown';
 @import 'form';
 @import 'frame';
+@import 'icon';
 @import 'header';
 @import 'layout';
-@import 'icon';
 @import 'list';
-@import 'card';
-@import 'cards';
-@import 'page-header';
-@import 'nav';
-@import 'toolbar';
-@import 'code-editor';
 @import 'metadata-list';
+@import 'nav';
+@import 'page-header';
+@import 'toolbar';
--- a/ui/admin/app/templates/application.hbs
+++ b/ui/admin/app/templates/application.hbs
@@ -137,7 +137,7 @@
       @dismissButtonText={{t 'actions.cancel'}}
       @modal={{true}}
       @dismiss={{deny}}
-      data-test-modal
+      dialog
     >
       {{if
         confirmation.options.title
@@ -151,7 +151,7 @@
       </Hds::Text::Body>
     </M.Body>
     <M.Footer>
-      <Hds::ButtonSet {{style width='15rem'}}>
+      <Hds::ButtonSet>
         <Hds::Button
           @text={{if
             confirmation.options.confirm
@@ -160,13 +160,11 @@
           }}
           @color='primary'
           {{on 'click' accept}}
-          data-test-modal-confirm-btn
         />
         <Hds::Button
           @text={{t 'actions.cancel'}}
           @color='secondary'
           {{on 'click' deny}}
-          data-test-modal-cancel-btn
         />
       </Hds::ButtonSet>
     </M.Footer>
--- a/ui/admin/app/templates/application.hbs
+++ b/ui/admin/app/templates/application.hbs
@@ -137,7 +137,6 @@
       @dismissButtonText={{t 'actions.cancel'}}
       @modal={{true}}
       @dismiss={{deny}}
-      dialog
     >
       {{if
         confirmation.options.title
--- a/ui/admin/app/templates/application.hbs
+++ b/ui/admin/app/templates/application.hbs
@@ -133,7 +133,7 @@
 
 <PendingConfirmations as |confirmation accept deny|>
   <Hds::Modal @color='warning' as |M|>
-    <M.Header @icon='alert-triangle'>
+    <M.Header @icon='alert-triangle' @onDismiss={{deny}}>
       {{if
         confirmation.options.title
         (t confirmation.options.title)
--- a/ui/admin/app/templates/application.hbs
+++ b/ui/admin/app/templates/application.hbs
@@ -137,6 +137,7 @@
       @dismissButtonText={{t 'actions.cancel'}}
       @modal={{true}}
       @dismiss={{deny}}
+      data-test-modal
     >
       {{if
         confirmation.options.title
@@ -159,11 +160,13 @@
           }}
           @color='primary'
           {{on 'click' accept}}
+          data-test-modal-confirm-btn
         />
         <Hds::Button
           @text={{t 'actions.cancel'}}
           @color='secondary'
           {{on 'click' deny}}
+          data-test-modal-cancel-btn
         />
       </Hds::ButtonSet>
     </M.Footer>
--- a/ui/admin/app/templates/application.hbs
+++ b/ui/admin/app/templates/application.hbs
@@ -132,33 +132,40 @@
 {{/if}}
 
 <PendingConfirmations as |confirmation accept deny|>
-  <Rose::Dialog
-    @heading={{if
-      confirmation.options.title
-      (t confirmation.options.title)
-      (t 'actions.confirm')
-    }}
-    @dismissButtonText={{t 'actions.cancel'}}
-    @icon='flight-icons/svg/alert-triangle-16'
-    @style='warning'
-    @modal={{true}}
-    @dismiss={{deny}}
-    as |dialog|
-  >
-    <dialog.body>
-      <p>{{confirmation.text}}</p>
-    </dialog.body>
-    <dialog.footer>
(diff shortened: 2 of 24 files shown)


ui/admin/tests/helpers/selectors.js
```diff
@@ -39,3 +39,7 @@ export const DIALOG_UNSAVED_CHANGES_DISCARD =
   '.rose-dialog-footer button:first-child';
 export const DIALOG_UNSAVED_CHANGES_CANCEL =
   '.rose-dialog-footer button:last-child';
+
+export const MODAL_WARNING = '[data-test-modal]';
```
laurenolivia, line 43: Will rename this to something more specific

addons/rose/app/styles/rose/components/_index.scss
```diff
@@ -4,18 +4,18 @@
  */
 
 @import 'button';
+@import 'card';
```
calcaide, line 7: None blocking comment: What is the reason behind the re-ordering?

laurenolivia: I think alphabetizing lists improves readability

ui/admin/app/templates/application.hbs
```diff
@@ -132,33 +132,43 @@
...
-          (t confirmation.options.confirm)
-          (t 'actions.ok')
-        }}
-      </Rose::Button>
-      <Rose::Button @style='secondary' {{on 'click' deny}}>
-        {{t 'actions.cancel'}}
-      </Rose::Button>
-    </dialog.footer>
-  </Rose::Dialog>
+  <Hds::Modal @color='warning' as |M|>
```
calcaide, line 135: For the alert dialog we can assume it will be just one dialog element on the DOM at a time, so I think we can use the standard selector `dialog`. I will suggest to not use `data-test-*` selector for this dialog use case.

But in case we wanted to unique identify this dialog, I will add the attribute to the `<HDS::Modal>` element above, since its the parent of the component.

ui/admin/app/templates/application.hbs
```diff
@@ -132,33 +132,43 @@
...
+        (t 'actions.confirm')
+      }}
+    </M.Header>
+    <M.Body>
+      <Hds::Text::Body @tag='p'>
+        {{confirmation.text}}
+      </Hds::Text::Body>
+    </M.Body>
+    <M.Footer>
+      <Hds::ButtonSet {{style width='15rem'}}>
```
calcaide, line 154: There is a reason why we are using inline styles here? I don't think its a pattern we use within the codebase.

ui/admin/tests/helpers/selectors.js
calcaide, line 43: As mentioned above, I will suggest to not use `data-test-*` selectors for this use case.

In this case, unless there is an specific need I am missing I will encourauge for use standard selector. I.e:

```
export const MODAL_WARNING = 'dialog';
export const MODAL_WARNING_CONFIRM_BTN = 'dialog .hds-modal__footer button:first-child';
export const MODAL_WARNING_CANCEL_BTN = 'dialog .hds-modal__footer button:last-child';
```

ui/admin/app/templates/application.hbs
laurenolivia: I based my refactor on the helios [docs](https://helios.hashicorp.design/components/button?tab=code#loading-state), but open to suggestions if the team has a preferred way
Screenshot 2024-12-13 at 5 00 56 AM

calcaide: Within our codebase, we have no inline styles, all our styles are defined within scss files. 
So unless there is an specific reason for it, or I will avoid them, or I will bring the topic to tech time to look for team consensous on introducing this approach.

ZedLi: We shouldn't use an inline style. If you check out the note above in your doc link, HDS actually even says to not add an explicit width to the button (also this was added to the button set, not button) and called it out for a special use case when loading indicators can cause the button to change size.

ui/admin/app/templates/application.hbs
laurenolivia: @calcaide Thanks for the feedback. I am wondering why we would continue using `dialog` as a test selector on an Hds::Modal component? Esp since we are retiring `Rose::Dialog` from the codebase.

ui/admin/app/templates/application.hbs
```diff
@@ -132,33 +132,41 @@
...
-        {{t 'actions.cancel'}}
-      </Rose::Button>
-    </dialog.footer>
-  </Rose::Dialog>
+  <Hds::Modal @color='warning' as |M|>
+    <M.Header
+      @dismissButtonText={{t 'actions.cancel'}}
+      @modal={{true}}
+      @dismiss={{deny}}
+      dialog
```
calcaide, line 140: What the reason of adding `dialog` here?

ui/admin/tests/helpers/selectors.js
```diff
@@ -39,3 +39,11 @@ export const DIALOG_UNSAVED_CHANGES_DISCARD =
   '.rose-dialog-footer button:first-child';
 export const DIALOG_UNSAVED_CHANGES_CANCEL =
   '.rose-dialog-footer button:last-child';
+
+export const MODAL_WARNING = 'dialog';
+export const MODAL_WARNING_CONFIRM_BTN =
```
calcaide, line 44: I see we have some selectors MODAL other DIALOG, aren't these aiming to select the same dialog element?

laurenolivia: You are correct, there are vars selecting the same elements because we are migrating to Hds so this file will need to be cleaned up either during/after the Hds migration.

ui/admin/app/templates/application.hbs
laurenolivia: This was added in err. Thanks.

addons/rose/app/styles/rose/components/_index.scss
laurenolivia: Great advice! ✨

ui/admin/app/templates/application.hbs
```diff
@@ -132,33 +132,40 @@
...
-        }}
-      </Rose::Button>
-      <Rose::Button @style='secondary' {{on 'click' deny}}>
-        {{t 'actions.cancel'}}
-      </Rose::Button>
-    </dialog.footer>
-  </Rose::Dialog>
+  <Hds::Modal @color='warning' as |M|>
+    <M.Header
+      @dismissButtonText={{t 'actions.cancel'}}
```
ZedLi, line 137: Same here, is `@dismissButtonText` needed?

ui/admin/app/templates/application.hbs
```diff
@@ -132,33 +132,40 @@
...
+      @modal={{true}}
+      @dismiss={{deny}}
```
ZedLi, line 139: It looks like it's called `@onDismiss` in [HDS](https://helios.hashicorp.design/utilities/dialog-primitive?tab=code#dialogprimitiveheader)

ui/admin/tests/helpers/selectors.js
calcaide: @laurenolivia I am still seeing `DIALOG` and `MODAL` selectors mixed

laurenolivia: @calcaide Can you clarify what you mean by this? Are you referring to the how the vars are named? Can you give an example of what you'd like to see different?

Are you referring to how I moved the following vars into this file, but did not rename them?
```
DIALOG_TITLE_SELECTOR
DIALOG_MESSAGE_SELECTOR
```

calcaide: I am meaning what I started this conversatio with: I see we have some selectors MODAL other DIALOG, aren't these aiming to select the same modal?

laurenolivia: Gotcha. The original question was pointing to [these vars](https://github.com/hashicorp/boundary-ui/blob/main/ui/admin/tests/helpers/selectors.js#L37-L41) currently in main that were left in the selectors file, but I have since removed them.

I believe the new request is to rename these vars, which I moved into this file from another test file. Is that correct?
```
DIALOG_TITLE_SELECTOR 
DIALOG_MESSAGE_SELECTOR
```
renamed to:

```
MODAL_WARNING_TITLE
MODAL_WARNING_MESSAGE
```

ui/admin/app/templates/application.hbs
```diff
@@ -132,33 +132,40 @@
...
-      </Rose::Button>
-      <Rose::Button @style='secondary' {{on 'click' deny}}>
-        {{t 'actions.cancel'}}
-      </Rose::Button>
-    </dialog.footer>
-  </Rose::Dialog>
+  <Hds::Modal @color='warning' as |M|>
+    <M.Header
+      @dismissButtonText={{t 'actions.cancel'}}
+      @modal={{true}}
```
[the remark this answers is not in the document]
laurenolivia: No, good catch. Removed 👍

ui/admin/app/templates/application.hbs
laurenolivia: No, this is handled in the footer. Thanks.

ui/admin/app/templates/application.hbs
laurenolivia: Thanks for pointing to the docs. Removed 👍

ZedLi: We should still keep it, since it's still valid functionality

laurenolivia: Even if removing does not change functionality?

ZedLi: The `deny` function won't get called when you hit the dismiss button if it's removed, so it likely still does the same thing right now as `deny` also just closes but technically the cancel button and dismiss button now do different things

laurenolivia: Thanks 👍

ZedLi (approved):
Looks good!

PR #2627, opened 2024-12-13
Chore: refactor host catalog model and serializer to accommodate GCP plugin

DhariniJeeva:
# Description
This PR refactors the model and serializer to accommodate GCP plugin support 

https://hashicorp.atlassian.net/browse/ICU-16030

note: Mirage support will be added along with `form` changes  

response from post call
```
 {
   "type": "plugin",
   "plugin_name": "gcp",
   "description": null,
   "scope_id": "p_1234567890",
   "attributes": {
       "project_id": "id",
       "disable_credential_rotation": true,
      "client_email": "redacted",
       "zone": "us-central1-a"
   },
   "secrets": {
       "private_key_id": "redacted",
       "private_key": "redacted"
   }
}

```

## Checklist


addons/api/addon/generated/models/host-catalog.js
```diff
@@ -68,13 +69,15 @@ export default class GeneratedHostCatalogModel extends BaseModel {
 
   @attr('string', {
     for: 'plugin',
+    compositeType: 'aws',
```
calcaide, line 72: Asking for learning purposes: what is the goal of adding `compositeType` within the model?

addons/api/addon/generated/models/host-catalog.js
```diff
@@ -83,13 +86,17 @@ export default class GeneratedHostCatalogModel extends BaseModel {
   // AWS static credentials
   @attr('string', {
     for: 'plugin',
+    credentialType: 'static-credential',
```
calcaide, line 89: Asking for learning puposes: what is the goal of adding `credentialType` within the model? I think we had before types without the need of adding this field, so I wanted to understand the reasons.

addons/api/addon/generated/models/host-catalog.js
```diff
@@ -55,10 +55,11 @@ export default class GeneratedHostCatalogModel extends BaseModel {
   })
   plugin;
 
-  // AWS & Azure
+  // AWS, Azure & GCP
 
   @attr('boolean', {
     for: 'plugin',
+    compositeType: ['aws', 'azure', 'gcp'],
```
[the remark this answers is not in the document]
DhariniJeeva: yeah you are right, I'm looking into refactoring the exisiting `for` in app serializer to handle this

addons/api/addon/generated/models/host-catalog.js
DhariniJeeva: AWS plugin supports two credential types- static and dynamic, the fields were listed in the serializer, but now I moved it up to the models, so we don't list out the fields in the serializer and can make use of the `for`

addons/api/addon/generated/models/host-catalog.js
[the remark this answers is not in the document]
DhariniJeeva: thanks for the feedback - I moved the compositeType logic to app serializer and handled the credentialType part in resource specific serializer since it is resource specific. Let me know what you think

addons/api/addon/serializers/application.js
```diff
@@ -74,18 +75,33 @@ export default class ApplicationSerializer extends RESTSerializer {
...
-      if (options.isNestedAttribute) {
-        delete json?.attributes?.[key];
-      } else {
-        delete json[key];
+
+    if (options?.for) {
+      // Check the typeof the `for` option to determine if it is a plugin or not
+      const isPlugin =
+        typeOf(options.for) === 'object' && options.for?.type === 'plugin';
+      const isNotPlugin = typeOf(options.for) === 'string';
```
ZedLi, line 85: It can actually still be both a string or an array right?

addons/api/addon/serializers/application.js
```diff
@@ -74,18 +75,33 @@ export default class ApplicationSerializer extends RESTSerializer {
...
+        } else {
+          delete json[key];
+        }
+      };
+
+      // For plugins, the `for` option is an object with a `name` key
+      // For non-plugins, the `for` option is a string
+      if (
+        (isPlugin && !options.for?.name.includes(compositeType)) ||
+        (isNotPlugin && !options.for?.includes(recordType))
```
ZedLi, line 102: This check still works but `for` can also be an array for non plugins, we should probably note that in the comment still.

The only edge case could be if `recordType` was a substring of `for` and we would have a bug but I'm not sure if we'd hit that. Maybe it's still worth being explicit and checking for string and `===` vs array and `includes`?

addons/api/addon/serializers/application.js
ZedLi, line 1: Can we add more unit tests for the changes in this file?

addons/api/addon/serializers/application.js
```diff
@@ -74,18 +75,33 @@ export default class ApplicationSerializer extends RESTSerializer {
...
+      const isPlugin =
+        typeOf(options.for) === 'object' && options.for?.type === 'plugin';
+      const isNotPlugin = typeOf(options.for) === 'string';
+
+      // Helper function to handle deleting attribute and secret keys
+      const deleteKey = (json) => {
+        if (options.isNestedAttribute) {
+          delete json?.attributes?.[key];
+        } else if (options.isNestedSecret && json?.secrets?.[key]) {
+          delete json?.secrets?.[key];
```
ZedLi, line 92: Do we have to handle the case where it's only one key and deleting it would leave an empty object? Also same question in the host-catalog serializer

addons/api/addon/serializers/host-catalog.js
```diff
@@ -4,115 +4,43 @@
...
-    if (isPlugin && options.isNestedSecret && json.secrets) {
-      if (!fields.includes(key)) {
+    const isSamePluginName = options?.for?.name?.includes(compositeType);
+
+    // Remove attributes that are not applicable to the current credential type for the same plugin
+    // For example, 'aws' dynamic plugin has two credential types: 'static' and 'dynamic'
+    // API expects the fields to be 'null' for the fields that are not from the current credential type (that's why we are not using the application serializer for this, where we delete attrs)
+    if (
+      isSamePluginName &&
+      typeOf(options.for) === 'object' &&
```
ZedLi, line 32: Isn't this implied if `isSamePluginName` returns true since it's looking for `options?.for?.name`?

addons/api/addon/serializers/host-catalog.js
```diff
@@ -4,115 +4,43 @@
...
+      options.for.credentialType &&
+      options.for.credentialType !== credentialType
+    ) {
+      if (options.isNestedAttribute) {
+        json.attributes[key] = null;
```
ZedLi, line 37: How come we have to set it to `null` if it's an attribute but delete it if it's a nested secret or a normal top level field?

addons/api/addon/serializers/application.js
DhariniJeeva: yeah you are right, I will include the array for this check and update the comment as well

addons/api/addon/serializers/host-catalog.js
DhariniJeeva: this was based on a discussion from the backend team that we need to set these to null, but I'm going to double check about secrets and the top level fields

addons/api/addon/serializers/application.js
DhariniJeeva: right now I didn't handle empty cases as I noticed we do send `attributes: {}` to the api [here](https://github.com/hashicorp/boundary-ui/blob/ca4aad0a2b381257e6910e22f11f626da8e4c2eb/addons/api/tests/unit/serializers/managed-group-test.js#L35). Also checking this with foundations team if we need to clean this up

addons/api/addon/serializers/host-catalog.js
DhariniJeeva: good catch, I removed the redudant check

addons/api/addon/serializers/application.js
DhariniJeeva: added!

addons/api/addon/serializers/application.js
DhariniJeeva: empty `attributes: {}` is expected, but we need to clean up empty secrets. I pushed up the change for this one.

addons/api/addon/serializers/host-catalog.js
DhariniJeeva: this is expected when the user makes an update from dynamic to static `credentialType`, we have to explicitly pass null for assume role fields which are nested attributes. But the secrets and top-level fields follow the usual pattern where we delete the fields that don't belong to the type. I asked for clarification here in [slack](https://hashicorp.slack.com/archives/C07E8C15JGL/p1735936709962049)

addons/api/addon/serializers/application.js
```diff
@@ -74,18 +75,38 @@ export default class ApplicationSerializer extends RESTSerializer {
...
-    ) {
-      if (options.isNestedAttribute) {
-        delete json?.attributes?.[key];
-      } else {
-        delete json[key];
+
+    if (options?.for) {
+      // Check the typeof the `for` option to determine if it is a plugin or not
+      const isPlugin =
+        typeOf(options.for) === 'object' && options.for?.type === 'plugin';
```
ZedLi, line 84: Can you just check for `options.for?.type === 'plugin'`? Since if this is true it's an object and if it's not, it won't match the string `plugin` since you're using optional chaining

addons/api/addon/serializers/application.js
```diff
@@ -74,18 +75,38 @@ export default class ApplicationSerializer extends RESTSerializer {
...
+      const isNotPlugin =
```
ZedLi, line 85: Do you think it's redundant to have both `isPlugin` and `isNotPlugin`? Are there scenarios where `!isPlugin` won't be the same as `isNotPlugin`?

addons/api/addon/serializers/application.js
```diff
@@ -74,18 +75,38 @@ export default class ApplicationSerializer extends RESTSerializer {
...
+          delete json[key];
+        }
+      };
+
+      // For plugins, the `for` option is an object with a `name` key
+      // For non-plugins, the `for` option is a string or an array
+      if (
+        (isPlugin && !options.for?.name.includes(compositeType)) ||
+        (isNotPlugin && !options.for?.includes(recordType)) ||
+        (isNotPlugin && !options.for === recordType)
```
ZedLi, line 107: If you include the check for `!options.for === recordType`, I think you'll need to make it something that checks the type:

```suggestion
        (isNotPlugin && !options.for?.includes(recordType) && typeOf(options.for) === 'array') ||
        (isNotPlugin && !options.for === recordType) && typeOf(options.for) === 'string')
```

otherwise you still run in the same issue as it was originally where a string could be checked for in `!options.for?.includes(recordType))`, e.g. `options.for` is `awsExample` and `recordType` is `aws`.

I also don't think this is that likely of an issue so if you want to just keep the `!options.for?.includes(recordType)` check only and just add a comment to this I'd also be fine with that

addons/api/addon/serializers/application.js
DhariniJeeva: hmm yeah, it does seem redundant, I will leave the comments for more clarity and use `!isPlugin`

addons/api/addon/serializers/application.js
DhariniJeeva: ah yes, it needs the `typeof` check. thanks for the catch!

ZedLi (approved):
Looks good!

addons/api/addon/generated/models/host-catalog.js
DhariniJeeva: answered for this one in the below comment

PR #2629, opened 2024-12-13
test: 💍 Refactor host list test selectors

calcaide:
# Description

Refactor host-catalog host list acceptance test selectors.

:tickets: [Jira ticket](https://hashicorp.atlassian.net/browse/ICU-15501)

## How to Test
- Run `admin` test suite successfully.

### DIFF
--- a/ui/admin/tests/acceptance/host-catalogs/hosts/list-test.js
+++ b/ui/admin/tests/acceptance/host-catalogs/hosts/list-test.js
@@ -75,8 +75,9 @@ module('Acceptance | host-catalogs | hosts | list', function (hooks) {
         'list',
       ),
     );
-    assert.dom(selectors.MANAGE_DROPDOWN_HOST_CATALOG).exists();
-    assert.dom(commonSelectors.HREF(urls.hosts)).exists();
+
+    assert.dom(selectors.MANAGE_DROPDOWN_HOST_CATALOG).isVisible();
+    assert.dom(commonSelectors.HREF(urls.hosts)).isVisible();
   });
 
   test('User cannot navigate to index without either list or create actions', async function (assert) {
@@ -95,7 +96,7 @@ module('Acceptance | host-catalogs | hosts | list', function (hooks) {
     instances.hostCatalog.authorized_collection_actions.hosts = ['create'];
     await visit(urls.hostCatalog);
 
-    assert.dom(commonSelectors.HREF(urls.hosts)).exists();
+    assert.dom(commonSelectors.HREF(urls.hosts)).isVisible();
     await click(selectors.MANAGE_DROPDOWN_HOST_CATALOG);
     await click(selectors.MANAGE_DROPDOWN_HOST_CATALOG_NEW_HOST);
     assert.strictEqual(currentURL(), urls.newHost);
--- a/ui/admin/tests/acceptance/host-catalogs/hosts/list-test.js
+++ b/ui/admin/tests/acceptance/host-catalogs/hosts/list-test.js
@@ -4,7 +4,7 @@
  */
 
 import { module, test } from 'qunit';
-import { visit, currentURL, click } from '@ember/test-helpers';
+import { visit } from '@ember/test-helpers';
 import { setupApplicationTest } from 'admin/tests/helpers';
 import setupMirage from 'ember-cli-mirage/test-support/setup-mirage';
 import { authenticateSession } from 'ember-simple-auth/test-support';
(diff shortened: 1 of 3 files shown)


ui/admin/tests/acceptance/host-catalogs/hosts/list-test.js
```diff
@@ -69,33 +66,41 @@ module('Acceptance | host-catalogs | hosts | list', function (hooks) {
...
+    assert.dom(commonSelectors.HREF(urls.hosts)).doesNotExist();
   });
 
   test('User can navigate to index with only create action', async function (assert) {
     instances.hostCatalog.authorized_collection_actions.hosts = ['create'];
     await visit(urls.hostCatalog);
-    assert.ok(find(`[href="${urls.hosts}"]`));
-    await click(MANAGE_DROPDOWN_SELECTOR);
-    assert.dom(NEW_HOST_SELECTOR).hasAttribute('href', urls.newHost);
+
```
calcaide, line 97: Enhanced this test.
We were not validating the user can actually click and navigate to the desired url, we were checking the HREF element was present with the correct url. But the test is about the user can navigate, so add that part.
Let me know yout thoughts.

lisbet-alvarez: I like the enhancement of checking the link on line 98. I'm wondering if the checks on lines 100 and 104 should be moved to [this test case instead](https://github.com/hashicorp/boundary-ui/blob/main/ui/admin/tests/acceptance/host-catalogs/hosts/create-test.js#L101) in create-test.js because that test case is specific to the being able to create a new host. Thoughts?

lisbet-alvarez (dismissed):
Looks great! 🙌

ui/admin/tests/acceptance/host-catalogs/hosts/list-test.js
cameronperera: I would agree with what @lisbet-alvarez suggested and I am with this the test change.

cameronperera (approved):

cameronperera (approved):

ui/admin/tests/acceptance/host-catalogs/hosts/list-test.js
calcaide: @cameronperera @lisbet-alvarez 
You guys got me thinking here. Now I do think this test shouldn't exists within this file, let me explain:

- As Lis mentioned, we already have a specific test file for creation where this case is covered. I will move the enhancement assertions to that create-test file.
- This test file is for listing, therefore, navigate to creation its out of the context of listing, as the way I understand it.

What do you guys think?

lisbet-alvarez: I think that the test case is still valid because it should be making sure that a user can navigate to the `hosts` list route with only the `create` action since a user can navigate to list route with either or both. While the other test case in the `create-test.js` file was making sure that when the `create` authorization is present a user can navigate to the new host route.

cameronperera (approved):

ui/admin/tests/acceptance/host-catalogs/hosts/list-test.js
```diff
@@ -69,33 +66,28 @@ module('Acceptance | host-catalogs | hosts | list', function (hooks) {
...
     assert.notOk(
       instances.hostCatalog.authorized_collection_actions.hosts.includes(
         'list',
       ),
     );
-    assert.notOk(find(`[href="${urls.hosts}"]`));
-  });
-
-  test('User can navigate to index with only create action', async function (assert) {
```
laurenolivia, line 94: non-blocking: wondering why we removed this test?

calcaide: Follow [this thread](https://github.com/hashicorp/boundary-ui/pull/2629#discussion_r1884265981)for th full context.

Spoilert alert, I will add changes tomorrow, or when I have the time, to this PR.

ui/admin/tests/acceptance/host-catalogs/hosts/list-test.js
calcaide: @lisbet-alvarez 
You are right, added back the test.

lisbet-alvarez (approved):
Thank you for the refactor!!

ui/admin/tests/acceptance/host-catalogs/hosts/list-test.js
```diff
@@ -69,33 +66,38 @@ module('Acceptance | host-catalogs | hosts | list', function (hooks) {
...
+
     assert.ok(
       instances.hostCatalog.authorized_collection_actions.hosts.includes(
         'list',
       ),
     );
-    assert.dom(MANAGE_DROPDOWN_SELECTOR);
-    assert.ok(find(`[href="${urls.hosts}"]`));
+    assert.dom(selectors.MANAGE_DROPDOWN_HOST_CATALOG).exists();
+    assert.dom(commonSelectors.HREF(urls.hosts)).exists();
```
ZedLi, line 79: Should these have used `isVisible()` instead of `exists()`?

ZedLi (dismissed):

cameronperera (approved):

laurenolivia (approved):

PR #2633, opened 2024-12-17
refactor: 💡 rename arialabel to aria-labelledby for key/value

cameronperera:
✅ Closes: https://hashicorp.atlassian.net/browse/ICU-15701

# Description

I noticed this slight naming issue while investigating one of our a11y audit issues. The main issue of no `aria-label` has already been taking care of but the naming of the attributes was a little off. I updated it to say `aria-labelledby` to reduce confusion on what is actually be used when rendering new inputs for the key/value component.

## Checklist

### DIFF
--- a/ui/admin/app/components/form/field/list-wrapper/key-value/index.hbs
+++ b/ui/admin/app/components/form/field/list-wrapper/key-value/index.hbs
@@ -35,14 +35,14 @@
                   'form/field/list-wrapper/key-value/text'
                   disabled=@disabled
                   value=option.key
-                  aria-labelledby=keyHeaderID
+                  ariaLabelledBy=keyHeaderID
                   setContext=(set-from-event option 'key')
                 )
                 select=(component
                   'form/field/list-wrapper/key-value/select'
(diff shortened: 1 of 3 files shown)


ui/admin/app/components/form/field/list-wrapper/key-value/index.hbs
```diff
@@ -35,14 +35,14 @@
                   'form/field/list-wrapper/key-value/text'
                   disabled=@disabled
                   value=option.key
-                  ariaLabel=keyHeaderID
+                  aria-labelledby=keyHeaderID
```
ZedLi, line 38: if we can't pass it down by `...attributes`, I would keep the standard camelcase and call it `ariaLabeledBy`

cameronperera: Oh yes. Good suggestion. I will fix that now.

ZedLi (approved):

lisbet-alvarez (approved):
Thanks for cleaning this up!

PR #2636, opened 2024-12-20
[ICU-15403] Trim whitespace on storage buckets and host sets json attrs serializers

laurenolivia:
# Description
This PR  adds the `trimWhitespace` property to **storage bucket** and **host set** model attrs, and trims whitespace on serializer values before sending to the backend.

Context:
A bug was reported on the storage bucket and host set forms. The fields that will err if the values are not trimmed are: `role_arn`, `region`, and `preferred_endpoints`. We are handling this at the application level to account for the possibility that there may be other form fields that need to be trimmed in the future.

:tickets: [Jira ticket](https://hashicorp.atlassian.net/browse/ICU-15403)

## Screenshots 📸 
Before: 👇 
Screenshot 2024-12-20 at 9 40 22 AM

After: 👇 
Screenshot 2024-12-20 at 9 41 53 AM

### Storage Buckets
1. Spin up an env using: `ENABLE_MIRAGE=false API_HOST=http://localhost:9200 yarn start`
2. Navigate to Storage Bucket in left side navbar
3. Create a new storage bucket and make sure to fill either `region` or `role_arn` fields and leave whitespace
4. Submit the form
5. Use browser Network tab and view the response body from the POST request
6. Values should not have whitespace and UI should not throw an error

Steps to connect a storage bucket to AWS found [here](https://github.com/hashicorp/boundary-aws-demo-stack)

### Host Set
1. Spin up an env using: `ENABLE_MIRAGE=false API_HOST=http://localhost:9200 yarn start`
2. Navigate to Host Set in left side navbar
3. Create a host set within a host catalog (type: AWS, Azure, or GCP) and make sure to fill `preferred_endpoints` field and leave whitespace
4. Submit the form
5. Use browser Network tab and view the response body from the POST request
6. Values should not have whitespace and UI should not throw an error

Note: you need a valid worker filter to test these changes against prod, for example:
```
"dev" in "/tags/type"
```

## Checklist

### DIFF
--- a/addons/api/addon/serializers/application.js
+++ b/addons/api/addon/serializers/application.js
@@ -109,18 +109,22 @@ export default class ApplicationSerializer extends RESTSerializer {
         deleteKey(json);
       }
     }
-    if (
-      options.trimWhitespace &&
-      json.attributes[attribute?.name] &&
-      options.isNestedAttribute &&
-      type === 'string'
-    ) {
-      console.log('str', json);
-      json.attributes[attribute.name] = json.attributes[attribute.name].trim();
-    } else if (options.trimWhitespace && type === 'array') {
-      console.log('arr', json);
-      // console.log(json.preferred_endpoints);
-      // json.preferred_endpoints = json.preferred_endpoints.map(str => str.trim());
+    if (options.trimWhitespace) {
+      if (
+        options.isNestedAttribute &&
+        json.attributes[attribute?.name] &&
+        type === 'string'
+      ) {
+        json.attributes[attribute.name] =
+          json.attributes[attribute.name].trim();
+      }
+      // Right now, we've only identified one string-array attribute that
+      // will err if not trimmed before sending to backend
+      // This fix may require a refactor in the future
+    } else if (type === 'string-array') {
+      json.preferred_endpoints = json.preferred_endpoints.map((str) =>
+        str.trim(),
+      );
     }
     return value;
   }
--- a/addons/api/addon/serializers/application.js
+++ b/addons/api/addon/serializers/application.js
@@ -60,7 +60,6 @@ export default class ApplicationSerializer extends RESTSerializer {
     // Do not serialize `disabled` fields.
     // TODO:  disabled is temporarily disabled
     if (key === 'disabled') delete json[key];
-    // Push nested attributes down into the attributes key
 
     // Before we transform our json
     // Let's trim the values in the json obj
@@ -73,6 +72,7 @@ export default class ApplicationSerializer extends RESTSerializer {
       }
     }
 
+    // Push nested attributes down into the attributes key
     if (options.isNestedAttribute && json[key] !== undefined) {
       if (!json.attributes) json.attributes = {};
       json.attributes[key] = json[key];
--- a/addons/api/addon/serializers/application.js
+++ b/addons/api/addon/serializers/application.js
@@ -112,9 +112,15 @@ export default class ApplicationSerializer extends RESTSerializer {
     if (
       options.trimWhitespace &&
       json.attributes[attribute?.name] &&
+      options.isNestedAttribute &&
       type === 'string'
     ) {
+      console.log('str', json);
       json.attributes[attribute.name] = json.attributes[attribute.name].trim();
+    } else if (options.trimWhitespace && type === 'array') {
+      console.log('arr', json);
+      // console.log(json.preferred_endpoints);
+      // json.preferred_endpoints = json.preferred_endpoints.map(str => str.trim());
     }
     return value;
   }
--- a/addons/api/addon/serializers/application.js
+++ b/addons/api/addon/serializers/application.js
@@ -109,6 +109,13 @@ export default class ApplicationSerializer extends RESTSerializer {
         deleteKey(json);
       }
     }
+    if (
+      options.trimWhitespace &&
+      json.attributes[attribute?.name] &&
+      type === 'string'
+    ) {
+      json.attributes[attribute.name] = json.attributes[attribute.name].trim();
+    }
     return value;
   }
 
--- a/addons/api/addon/serializers/application.js
+++ b/addons/api/addon/serializers/application.js
@@ -61,6 +61,18 @@ export default class ApplicationSerializer extends RESTSerializer {
     // TODO:  disabled is temporarily disabled
     if (key === 'disabled') delete json[key];
     // Push nested attributes down into the attributes key
+
+    // Before we transform our json
+    // Let's trim the values in the json obj
+    // Currently, json values will only ever be a string or an array
+    if (options.trimWhitespace && json[key]) {
+      if (typeOf(json[key]) === 'array') {
+        json[key] = json[key].map((str) => str.trim());
+      } else if (typeOf(json[key] === 'string')) {
+        json[key] = json[key].trim();
+      }
+    }
+
     if (options.isNestedAttribute && json[key] !== undefined) {
       if (!json.attributes) json.attributes = {};
       json.attributes[key] = json[key];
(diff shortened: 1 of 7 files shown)


addons/api/addon/serializers/storage-bucket.js
```diff
@@ -72,8 +72,17 @@ export default class StorageBucketSerializer extends ApplicationSerializer {
...
+
     // This deletes any fields that don't belong to the record's credential type
     if (options.isNestedAttribute && json.attributes) {
+      if (json.attributes?.region) {
+        json.attributes.region = json.attributes.region.trim();
+      }
+
+      if (credentialType == 'dynamic' && json.attributes?.role_arn) {
+        json.attributes.role_arn = json.attributes.role_arn.trim();
+      }
```
ZedLi, line 84: Rather than manually writing code to trim each field which might have to be done for multiple resources (this ticket also included host sets), I think we should instead make this a bit more generic and customize our application serializer to handle this since it looks like it will be a common scenario.

You could first add a new field to each attribute on models that need it which tells us we should trim the whitespace, something like `trimWhitespace: true`. Then customize the [serializeAttribute](https://api.emberjs.com/ember-data/4.12/classes/RESTSerializer/methods/serializeAttribute?anchor=serializeAttribute) in our application serializer to handle this new case and that way you only have to annotate the attributes on the model for this behavior rather than having to customize each individual serializer that might need whitespace trimmed.

addons/api/tests/unit/serializers/storage-bucket-test.js
```diff
@@ -359,4 +359,46 @@ module('Unit | Serializer | storage bucket', function (hooks) {
       },
     });
   });
+
+  test('it trims whitespace on region and role_arn fields', async function (assert) {
```
calcaide, line 363: I will suggest to not create a new test case for this specific. Perhaps we can embed this 2 fields are being serialized correctly in any test we already have within this test file? Preferebly, we can modify these 2 fields within all tests that are serializing to make sure we receive the expected behaviour. I.E: in one test add only 1 white space, in other test add 2 white spaces.

addons/api/addon/serializers/storage-bucket.js
laurenolivia: I moved this check to the application level to solve the bug for storage buckets. The host set attr `preferred_endpoints` is a string-array so it will need a separate solution. @lisbet-alvarez is going to be on-call and will add this to her queue, if time allows, she will finish the solution.

addons/api/tests/unit/serializers/storage-bucket-test.js
laurenolivia: Good idea

addons/api/addon/serializers/application.js
```diff
@@ -88,6 +88,13 @@ export default class ApplicationSerializer extends RESTSerializer {
         delete json[key];
       }
     }
+    if (
+      options.trimWhitespace &&
+      json.attributes[attribute?.name] &&
```
ZedLi, line 93: I know this is still a WIP but wanted to point out that this makes the assumption that the field is _only_ an attribute that needs to be trimmed but it could also be a field that is top level as well, such as `name`. We should aim to be able to handle it for any field on the model.

laurenolivia: Thanks. Will reconsider my solution.

ZedLi: Yeah we should, to be clear I meant this should be able to applicable for _any_ model or form, not just for this storage bucket form

addons/api/addon/serializers/application.js
```diff
@@ -109,6 +109,23 @@ export default class ApplicationSerializer extends RESTSerializer {
...
+          json.attributes[attribute.name].trim();
+      }
+      // Right now, we've only identified one string-array attribute that
+      // will err if not trimmed before sending to backend
+      // This fix may require a refactor in the future
+    } else if (type === 'string-array' && json?.preferred_endpoints) {
+      json.preferred_endpoints = json.preferred_endpoints.map((str) =>
+        str.trim(),
+      );
+    }
```
lisbet-alvarez, line 128: Since we added the `trimWhitespace` option to the `preferred_endpoints` field then I think we can move the logic for `string-array` attributes inside of if statement of line 112. In addition to making it more generic so that it works with any `string-array` attribute, not just `preferred_endpoints`.

Also we should also support adding `trimWhitespace` to attributes that are not nested.
For example, we could move that check outside the if statement on line 112 so that the rest of the logic is more readable.
ex:  `const jsonField = options.isNestedAttribute ? json.attributes : json;`

Additionally, we could make this more readable by using a switch statement for the  attribute types (e.g. 'string', 'string-array').

laurenolivia: Yea, organizing the logic within a switch statement is a good idea. Working on a refactor. 👍

ZedLi: I think we should take a step back and look at what the goal here is. I agree with @lisbet-alvarez that we should not make this specific to `preferred_endpoints` as this is a generic application serializer (otherwise this specific logic should have lived in the host sets serializer). But we should also think if all these `if` statements or a switch statement are even needed.

> Will make it easier in the future to modify the code as we discover more fields that need to be trimmed.

We want to avoid a situation where we have to refactor this in the future if there _are_ more/different fields that need to be trimmed. We should be able to handle them regardless.

What this means is we shouldn't be looking at the model field types, (e.g. `string-array`) since we don't actually care (especially as we would have to keep adding all possible new transformer types), we just care if it's a string or possibly an array. The logic is also being over complicated by considering `nestedAttributes` or `nestedSecrets`. If you consider what `nestedAttributes` are, they're not something that's native to ember, we're actually the ones creating it and transforming them (lines 64 and 70). We don't even need to check if they're nested if we just trim these fields before we actually transform the json and in that way avoid all the extra logic.

laurenolivia: Also, agree that the logic should be more generic.

The way I understand this fix is: how we trim fields depends on where the value is stored in the json obj (e.g. inside `attributes: {}`) and the value's type (e.g. I can call `.map()` on `preferred_endpoints` because it is type `string-array` ).

RE:
> We don't even need to check if they're nested if we just trim these fields before we actually transform the json and in that way avoid all the extra logic

Not sure I follow why we don't need to check if the attr is nested, since that impacts how we iterate over the keys. Can you elaborate on what you're suggesting?

ZedLi: Does this section make sense?
> If you consider what nestedAttributes are, they're not something that's native to ember, we're actually the ones creating it and transforming them (lines 64 and 70).

Regarding this:
> how we trim fields depends on where the value is stored in the json obj (e.g. inside attributes: {})

We're actually the ones actually doing the storing of the json obj, we have full control of it so in fact we just have to trim it _before_ we store it in the json obj

laurenolivia: Thanks, that solution is much cleaner 👍

addons/api/tests/unit/serializers/storage-bucket-test.js
```diff
@@ -26,7 +26,7 @@ module('Unit | Serializer | storage bucket', function (hooks) {
       bucket_name: 'bucketname',
       bucket_prefix: 'bucketprefix',
       worker_filter: 'workerfilter',
-      region: 'eu-west-1',
+      region: 'eu-west-1 ',
```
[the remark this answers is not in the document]
cameronperera: Good catch. I second that as we can verify that it isn't trimming the whitespace on accident.

This makes me think, should we default to trimming the whitespace and explicitly set it to false if we don't want to trim it?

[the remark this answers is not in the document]
ZedLi: No I think in general we shouldn't. Some fields like descriptions they might want that whitespace (new lines), I think we're really just providing this as a convenience for things we _know_ shouldn't have trailing or leading whitespace and would error out otherwise

ZedLi (approved):

PR #2639, opened 2024-12-21
feat: 🎸 Use snapshots or OCR for E2E tests when comparing terminal

ZedLi:
# Description
When reviewing this PR, please take a look at both commits and test both. The goal of this PR is to resolve this [TODO](https://github.com/hashicorp/boundary-ui/pull/2639/files#diff-52f8b49078352758f374411a5105b29c6683ddbd4a0dcf558ef64ee3925aad96L227-L228) as sometimes we want to interact with our integrated terminal. The main problem is that the terminal is built using a `canvas` element which makes it difficult to test with playwright as it's basically an image with no DOM elements to interact with. The particular problem we're trying to solve in this PR is knowing when the desktop client properly finishes making the SSH connection and is "done" when using the integrated terminal. 

The first commit adds a golden screenshot to compare to what we expect a properly connected SSH session to look like. This of course can be quite brittle due to many factors (size of window, if the EC2 instance has new updates, timestamps, etc) but allowing a little bit of difference seems to work still. It's hard to say how brittle it is as the target machines can change so I'm curious if it works well for others. I had to use `poll` as it turns out the documentation is a bit misleading and `toHaveScreenshot` doesn't actually wait until they match and just compares screenshots after they have been stabilized, which is usually before the SSH connection finishes. 

The second commit adds `tesseract.js` and takes a different approach by using an OCR package to interpret the screenshot of the canvas element that holds our terminal. I use a phrase that I expect to see every time we connect to an EC2 instance and have it keep checking to see if it appears. The main downside is the setup of the package adds a decent amount of time (~5s on my mac but I set it up so it only runs once per worker and should be reused) and a few more seconds to actually recognize the image in the test. In theory this should be less brittle but may also be unnecessary as it's decently slower.

## How to Test
Check out each commit and try running `yarn desktop` or `yarn desktop:dev`

## Checklist


moduli (commented):
Some points of consideration
- Terminal output may look different if using docker-based infrastructure vs. AWS-based infrastructure. Not sure if there are any test cases that need docker.
- If using snapshots, it's possible that some people's personal shells may affect terminal output

ZedLi:
> Terminal output may look different if using docker-based infrastructure vs. AWS-based infrastructure. Not sure if there are any test cases that need docker.

I haven't used docker for anything yet in DC tests but my assumption was we would only be testing against targets that were spun up somewhere which I assumed would be EC2 instances when using SSH targets. What do you generally connect to for the docker based tests, localhost? 

> If using snapshots, it's possible that some people's personal shells may affect terminal output

Yup, the assumption was that the output from logging into an EC2 instance would scroll past it, which may not be true in all cases.

moduli:
> > Terminal output may look different if using docker-based infrastructure vs. AWS-based infrastructure. Not sure if there are any test cases that need docker.
> 
> I haven't used docker for anything yet in DC tests but my assumption was we would only be testing against targets that were spun up somewhere which I assumed would be EC2 instances when using SSH targets. What do you generally connect to for the docker based tests, localhost?
> 
> > If using snapshots, it's possible that some people's personal shells may affect terminal output
> 
> Yup, the assumption was that the output from logging into an EC2 instance would scroll past it, which may not be true in all cases.

On docker-based tests, we spin up a separate container to use as a target (https://hub.docker.com/r/linuxserver/openssh-server).

laurenolivia:
@ZedLi interesting PR. Upon reading the PR description I'm wondering if a short video or screenshots would help review this one? I'm curious about the discovery that led to this work, is there a jira ticket for more context?

laurenolivia:
@ZedLi Thanks for adding context to the description.

RE: I had to use poll as it turns out the documentation is a bit misleading...

Can you link the docs you are referencing?

e2e-tests/desktop/fixtures/baseTest.js
```diff
@@ -7,7 +7,13 @@ import { mergeTests } from '@playwright/test';
 import { electronTest } from './electronTest.js';
 import { authenticateTest } from './authenticateTest.js';
 import { test as envTest } from '../../global-setup.js';
+import { tesseractTest } from './tesseractTest.js';
 
 // The last item in the list will override any previous ones in a fixture name collision
```
laurenolivia, line 12: Interesting work.

RE: override any previous ones

To clarify this commented line, any variables that could possibly exist in a test suite (e.g. envTest), but also exist in `tesseractTest`(being last) will be overridden?

ZedLi: Yeah, so if there's any `use` variables from these fixtures that have the same name, the one that appears later in this list will end up taking priority and override any previous ones.

ZedLi:
> Can you link the docs you are referencing?

I was just referencing the playwright docs for [toHaveScreenshot](https://playwright.dev/docs/api/class-pageassertions#page-assertions-to-have-screenshot-1) where they mention:

> This function will wait until two consecutive page screenshots yield the same result, and then compare the last screenshot with the expectation.

which is not quite accurate.

e2e-tests/desktop/fixtures/baseTest.js
laurenolivia: Gotcha. Thanks. I wonder if that comment could be a bit clearer? Wdyt?

calcaide (approved):
Thanks for the work!!!

I do personally don't like to rely on screenshots for testing, but as you already mention and explain, there isn't much of an option for this specific case.

e2e-tests/desktop/fixtures/baseTest.js
ZedLi: What do you think it should say? This was more of a general comment to remind people how merging works since

laurenolivia: I read a bit of the playwright docs to get more context, thanks for sharing those. I'd prob comment something like:

```
// Please be aware that defining like fixtures across different test envs 
// in Playwright can cause naming collisions
// As a resolution, any like declarations that exist in the last test in this list item
// will override all declarations, if those like declarations exist in previous test envs
```

laurenolivia: Non-blocking suggestion, though. Good work 👍

e2e-tests/desktop/fixtures/tesseractTest.js
```diff
@@ -0,0 +1,23 @@
...
+  },
+  tesseract: [
+    async ({}, use) => {
+      const worker = await createWorker('eng', 1, {
+        cachePath: './artifacts',
+      });
+      await use(worker);
+      await worker.terminate();
+    },
+    { scope: 'worker' },
```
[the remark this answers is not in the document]
ZedLi: Here's the [doc](https://playwright.dev/docs/test-fixtures#worker-scoped-fixtures) for worker scopes

e2e-tests/desktop/fixtures/baseTest.js
ZedLi: I'll update this comment to be a bit more descriptive to explain the merging behavior

hashicc (changes requested):
Testing request changes

e2e-tests/desktop/fixtures/tesseractTest.js
```diff
@@ -0,0 +1,23 @@
+/**
+ * Copyright (c) HashiCorp, Inc.
+ * SPDX-License-Identifier: BUSL-1.1
+ */
+
+import { test } from '@playwright/test';
+import { createWorker } from 'tesseract.js';
+
+export const tesseractTest = test.extend({
+  textToMatch: ({}, use) => {
```
hashicc, line 10: I'm reading this as providing a regex for `/To run a command as administrator|Welcome to OpenSSH Server/`. Since there's no state from here would it be easier to have it directly available at the top of the file rather than a playwright fixture? I'm still trying to figure out what becomes a fixture and what doesn't. My current basis is if something is stateful, has lifecycles or needs a common base setup and then exposed on a per-test basis, otherwise there's extra hops to see what the value.

hashicc (approved):
Looks good, one small non-blocking suggestion.

I thought of another option that isn't as "real" as this for testing but might work. Taking the last _x_ amount of characters and attaching it to a `data-testId` and asserting against that. This might not be performant in the case a lot of scrolling text in which case it could be debounced before updating the DOM attribute. But that wouldn't  actually test that it made it to the terminal itself.

I'm still leaning towards favoring the OCR option but thought I'd mention it

moduli (commented):
Tried running it locally. It failed one time due to the following

```
  1) tests/sessions.spec.js:213:3 › Filtering sessions tests › Filters by status ───────────────────

    Error: expect(received).toMatch(expected)

    Expected pattern: /To run a command as administrator|Welcome to OpenSSH Server/
    
    Received string:  "in vscode, nvm not work; use °load-nvm®·
    ssh 127.0.0.1 -p 57748 -o NoHostAuthenticationForLocalhost=yes·
    [oh-my-zsh] Would you like to update? [Y/n]·
    [oh-my-zsh] You can update manually by running “omz update”
    ET FE FPF
    ) sh 127.0.0.1 -p 57748 -o NoHostAuthenticationForLocalhost=yes·
    sh: 127.0.0.1: No such file or directory
    PT FE FPF
    >
    "
```

It looks like, due to my setup, the terminal prompted to update something, so the first `s` in the `ssh` command was used to respond to that prompt. It's just something specific to me, and that prompt doesn't appear every time, but just wanted to share. When that prompt doesn't appear, it looks like things are working great.

ZedLi:
> I thought of another option that isn't as "real" as this for testing but might work. Taking the last _x_ amount of characters and attaching it to a `data-testId` and asserting against that. This might not be performant in the case a lot of scrolling text in which case it could be debounced before updating the DOM attribute. But that wouldn't actually test that it made it to the terminal itself.

How would this work in practice?  That is, how are you envisioning the "taking the last _x_ amount of characters? Wouldn't this require OCR to be able to take the characters or am I misunderstanding?

e2e-tests/desktop/fixtures/tesseractTest.js
ZedLi: Yes you're right it's just providing a regex. The alternative in my mind was having an exported `const` somewhere (or even in this file) as it's realistically a shared regex. I didn't have any strong feelings about this and I think you're right as I'm leaning to just have it being exported.

However, I would argue that it's quite similar to how we pass our env variables in `envTest` or `authenticateTest`. If we didn't already have it setup from admin, I would've just manually destructured them in tests like this rather than passing them through a fixture. Do you view these as different?

```
test(`example test`, async ({ page }) => {
  const { username, password } = process.env;
  await page.getByLabel('User Name').fill(username);
  await page.getByLabel('Password').fill(password);
});
```

calcaide (approved):
Run the tests manually successful! Thanks for the work 🙌

e2e-tests/desktop/fixtures/tesseractTest.js
hashicc: The main difference I see for those env vars is that they're:
1. meant to be dynamic per test process
2. at the "environment" level it makes more sense to be globally available (and thus mixed in via a playwright fixture)

I also like the ability for the playwright fixture to centralize and enforce some expectation for required env vars within a playwright fixture (something that I wanted to do for fixtures that use an existing fixture-based env var but it comes in as `''` like some of the `LDAP` ones, and we know we should probably just throw).

Where `/To run a command as administrator|Welcome to OpenSSH Server/.` is a constant check with no external dependencies and is scoped within a few tests. It's a bit harder to see where `textToMatch` comes from or its context once destructured from a merged test but that could also be fixed with a name change (and maybe playwright's types keep the "go to source" in tact anyways).

But I'm still figuring out the right balance for playwright fixtures. In their docs they definitely like to lean on them.

ZedLi: Are they _actually_ dynamic per test process? I actually consider them static configurations we happened to have dynamically generated before our tests and don't expect them to change between tests. 

> at the "environment" level it makes more sense to be globally available (and thus mixed in via a playwright fixture)

If they're globally available (and not considered stateful to the fixture), why have them mixed in via a playwright fixture? I understand the desire to have them centralized (and don't disagree with that and I think is the main benefit), but that could have easily been a separate utility class/file as well. In fact, playwright in general [seems to encourage](https://playwright.dev/docs/test-parameterize#passing-environment-variables) you to just use the env variables directly.

> My current basis is if something is stateful, has lifecycles or needs a common base setup and then exposed on a per-test basis

I actually fully agree with your original statement here and am now getting the sense that these environment variables actually don't pass muster for this. I always felt it was slightly off having them in a fixture and I think I'm reconsidering how we handle env variables. 

Realistically I also could have made my variable not a regex and depend on the `BOUNDARY_ADDR` to see what kind of environment we're running in and return a string based on that, which would now bring it in line to being closer to one of our env variables. I'm going to still move this regex to just an exported `const` but I think this is making me rethink our env variables more than anything.

hashicc: > Are they actually dynamic per test process? I actually consider them static configurations we happened to have dynamically generated before our tests and don't expect them to change between tests.

I think what I meant was that they don't change between tests but possibly change between test runs. For instance, changed on the environment before running `yarn playwright ...`, eg: this one `bucketEndpointUrl: process.env.E2E_BUCKET_ENDPOINT_URL` could be different between test runs depending on enos config, making their values more opaque. In the case of the regexp it's truly static.

> I actually consider them static configurations
This is exactly it, many of these that are keyed directly off the root playwright fixture should probably be on a `config` key.

e2e-tests/desktop/fixtures/tesseractTest.js
```diff
@@ -0,0 +1,23 @@
...
+import { createWorker } from 'tesseract.js';
+
+export const textToMatch =
+  /To run a command as administrator|Welcome to OpenSSH Server/;
+
+export const tesseractTest = test.extend({
+  tesseract: [
+    async ({}, use) => {
+      const worker = await createWorker('eng', 1, {
+        cachePath: './artifacts',
```
lisbet-alvarez, line 16: question for learning purposes: On test failures what type of data is being stored in the artifacts folder? Based on the the [docs](https://github.com/naptha/tesseract.js/blob/master/docs/api.md#:~:text=common%20to%20modify-,cachePath,-path%20for%20the) it says `traineddata` but not exactly sure what that means in this context.

lisbet-alvarez (approved):
The OCR options def. seems to be less brittle on my end. I was not able to get test cases to pass using the snapshot comparison option.

e2e-tests/desktop/fixtures/tesseractTest.js
ZedLi: So for this `tesseract` package, nothing is generated from test failures that gets stored in `artifiacts`. `traineddata` is actually just the language files that gets downloaded when running for the first time (which is why it can be slow) that it uses for its engine to actually run. It's basically where we store the setup files that are needed for this package.

PR #2640, opened 2024-12-27
Fix memory leak issue when using fetch in oidc authentication

lisbet-alvarez:
# Description

When making a `fetch` request that returns a response body, we have to consume the body in order to avoid undefined behavior. If we do not consume response body then its up to the env. garbage collection to free up the memory. In linux this was causing a memory leak during pending oidc authentications. Therefore, it is [best practice](https://undici.nodejs.org/#/?id=garbage-collection) to always consume response body (with `.json`). This[ github issue](https://github.com/nodejs/node/issues/51162) was illustrates what we were also running into.

This memory leak issue for oidc authentication is visible in linux desktop client v2.2.0. But was not able to reproduce in mac. (haven't tried windows)
Before desktop client v2.2.0 this was not an issue because we were using `ember-fetch` but was introduced in [this commit](https://github.com/hashicorp/boundary-ui/commit/1947365c91d9a9d02885220b602084821d573d9b) since we migrated away from that package and started using `fetch` native to node.

Before:

https://github.com/user-attachments/assets/0fc686f2-b1b8-4495-82f7-776e73954b15

After:

https://github.com/user-attachments/assets/a44c7453-19ed-4c24-abb2-bb29d033dc33

## How to Test

1. Use amazon workspaces (linux env) test account & checkout this branch
2. Test oidc authentication workflow
3. Test restore workflow by using ctrl + R after signing in
4. Validate memory and CPU don't go up indefinitely, when clicking sign-in for oidc auth.

## Checklist

### DIFF
--- a/addons/auth/addon/authenticators/base.js
+++ b/addons/auth/addon/authenticators/base.js
@@ -105,10 +105,6 @@ export default class BaseAuthenticator extends SimpleAuthBaseAuthenticator {
     // The `attributes` field exists on the Go side for its convenience but is
     // unnecessary here.
     Object.assign(data, data.attributes);
-    console.log('NormalizeData function ---------------------');
-    console.log(data);
-    console.log(username);
-    console.log('---------------------------------------------');
     // Add booleans indicated the scope type
     data.isGlobal = data?.scope?.type === 'global';
     data.isOrg = data?.scope?.type === 'org';
@@ -127,7 +123,7 @@ export default class BaseAuthenticator extends SimpleAuthBaseAuthenticator {
   async invalidate(options) {
     const { token } = options;
     const deauthEndpointURL = this.buildDeauthEndpointURL(options);
-    const response = await waitForPromise(
+    await waitForPromise(
       fetch(deauthEndpointURL, {
         method: 'delete',
         headers: { Authorization: `Bearer ${token}` },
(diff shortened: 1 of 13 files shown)


ZedLi:
Great find!! I'm curious, do you know if this is reproducible in other environments and not just linux?

addons/auth/addon/authenticators/base.js
```diff
@@ -66,6 +66,12 @@ export default class BaseAuthenticator extends SimpleAuthBaseAuthenticator {
         headers: { Authorization: `Bearer ${token}` },
       }),
     );
+
+    // Note: Always consume response object in order to avoid memory leaks.
+    // visit https://undici.nodejs.org/#/?id=garbage-collection for more info.
```
ZedLi, line 71: Just want to point out we don't use `undici` since this is a node package, but is still relevant since it points out the fetch spec depends on [garbage collection](https://fetch.spec.whatwg.org/#garbage-collection) which means this behavior might not be well defined. Might be worth clarifying that in this comment?

Also, should we consider making a `HEAD` request instead as suggested in your link and [here](https://github.com/node-fetch/node-fetch/issues/83) if we don't need the body? I don't think there's much data returned so it might not matter too much either way

lisbet-alvarez: thank you for the feedback!! comment updated!
i tested out using the head method tho i get the "method not allowed" error in the console :(
image

ZedLi: Ah I guess we don't support `HEAD` requests 😢

cameronperera (approved):

DhariniJeeva (approved):
great fix!  thanks for the work!

ZedLi (approved):
Looks good! Tested the issue is fixed in my local ubuntu VM

PR #2512, opened 2024-10-07
fix: 🐛 cannot retry oidc authentication in desktop & admin

lisbet-alvarez:
✅ Closes: https://hashicorp.atlassian.net/browse/ICU-15250

# Description

retry was broken for both admin & desktop.
Admin retry was broken due to this [refactor](https://github.com/hashicorp/boundary-ui/pull/2385)
Desktop retry has been broken for quite some time already.

## Screenshots (if appropriate)

Uploading Screen Recording 2024-10-07 at 4.46.03 PM.mov…

## How to Test

1. spin up local dev instance
2. use existing oidc auth method to sign in.
3. retry link should cause new browser to open.

## Checklist

- [X] I have added before and after screenshots for UI changes
- [ ] I have added JSON response output for API changes
- [ ] I have added steps to reproduce and test for bug fixes in the description
- [ ] I have commented on my code, particularly in hard-to-understand areas
- [X] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works

PR #2641, opened 2024-12-27
fix: 🐛 oidc retry not working for secure links in desktop ui

lisbet-alvarez:
# Description

Looks like I missed a case when fixing the retry bug.
So the retry link was only working when using localhost env and this needs to be tested with secure oidc auth.

https://github.com/user-attachments/assets/b58786a9-eeda-4916-b1ca-0b5fc9cab3d4

## How to Test

1. Use hcpb cluster w/ oidc auth method configured
2. Click sign in
3. Authentication Pending should appear.
4. Clicking "Retry" link should open new window.

## Checklist

### DIFF
--- a/ui/admin/app/controllers/scopes/scope/authenticate/method/oidc.js
+++ b/ui/admin/app/controllers/scopes/scope/authenticate/method/oidc.js
@@ -3,21 +3,35 @@
  * SPDX-License-Identifier: BUSL-1.1
  */
 
-import Controller from '@ember/controller';
+import Controller, { inject as controller } from '@ember/controller';
 import { inject as service } from '@ember/service';
+import { action } from '@ember/object';
 
 export default class ScopesScopeAuthenticateMethodOidcController extends Controller {
+  @controller('scopes/scope/authenticate/method/index') authenticateMethod;
+
   // =services
 
   @service session;
 
-  // =attributes
+  // =actions
 
   /**
-   * Authentication URL for the pending OIDC flow, if any.
-   * @type {?string}
+   * Retry by starting a new OIDC authentication flow.
+   * @param {object} e
    */
-  get authURL() {
-    return this.session.data.pending.oidc.attributes.auth_url;
+  @action
+  async retryAuthentication(e) {
+    // TODO: These event handlers can be removed when rose dialog gets
+    // refactored to use hds.
+    e.preventDefault();
+    e.stopPropagation();
+
+    const scope = this.authMethod.scope;
+    const authenticatorName = `authenticator:${this.authMethod.type}`;
+    await this.authenticateMethod.startOIDCAuthentication(authenticatorName, {
+      scope,
+      authMethod: this.authMethod,
+    });
   }
 }
--- a/ui/admin/app/controllers/scopes/scope/authenticate/method/oidc.js
+++ b/ui/admin/app/controllers/scopes/scope/authenticate/method/oidc.js
@@ -22,10 +22,9 @@ export default class ScopesScopeAuthenticateMethodOidcController extends Control
    */
   @action
   async retryAuthentication(e) {
-    // TODO: These event handlers can be removed when rose dialog gets
+    // TODO: This event handler can be removed when rose dialog gets
     // refactored to use hds.
     e.preventDefault();
-    e.stopPropagation();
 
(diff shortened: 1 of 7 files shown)


ui/desktop/electron-app/src/index.js
```diff
@@ -156,16 +156,10 @@ const createWindow = (partition, closeWindowCB) => {
...
-  // and releases.hashicorp.com domain to download the desktop app or
-  // link to the release page for the desktop app.
+  // We allow secure links and localhost to open an external window.
   browserWindow.webContents.setWindowOpenHandler(({ url }) => {
-    if (
-      isLocalhost(url) ||
-      url.startsWith('https://developer.hashicorp.com/') ||
-      url.startsWith('https://releases.hashicorp.com/boundary-desktop/') ||
-      url.startsWith('https://support.hashicorp.com/hc/en-us')
-    ) {
```
lisbet-alvarez, line 168: is it bad practice to allow all secure links ? Should we look into an alternative solution ? the `openExternal` ipc handler allows all secure links, wondering why this handler was more specific...

cameronperera: I actually thought we were allowing all `https` and made exceptions for other urls. I will look to see if I can find what I am thinking of.

cameronperera: I was thinking of this and I was wrong 😆 

So we need to allow `https` because we won't know what url they are using for oidc correct?

lisbet-alvarez: yup that is correct!

ZedLi: I assume we wanted to whitelist the expected urls for opening up links, since there's always a risk that we might miss something. Wasn't the `openExternal` handler made specifically for OIDC workflows since that can open any url? Did we change this?

ZedLi: Actually I think I see, in this PR https://github.com/hashicorp/boundary-ui/pull/2512 it looks like we changed it to a link which is why we're using the default handler for opening up links.

Should we have changed it to assume the auth url is still valid? If we retry don't we need to restart the oidc workflow here?
https://github.com/hashicorp/boundary-ui/blob/8605a5422148f50da33059be7b76ff13ad5f530f/ui/desktop/app/controllers/scopes/scope/authenticate/method/index.js#L32

cameronperera (approved):

ui/admin/app/controllers/scopes/scope/authenticate/method/oidc.js
```diff
@@ -3,21 +3,34 @@
...
+   * Retry by starting a new OIDC authentication flow.
+   * @param {object} e
    */
-  get authURL() {
-    return this.session.data.pending.oidc.attributes.auth_url;
+  @action
+  async retryAuthentication(e) {
+    // TODO: This event handler can be removed when rose dialog gets
+    // refactored to use hds.
+    e.preventDefault();
```
ZedLi, line 27: Is this to prevent the link from being actually clicked so we can use this handler?

ui/desktop/app/templates/scopes/scope/authenticate/method/oidc.hbs
```diff
@@ -11,7 +11,7 @@
     <p>{{t 'resources.auth-method.messages.pending.description'}}</p>
     <p>
       {{t 'resources.auth-method.questions.no-see-window'}}
-      <a href={{this.authURL}} target='_blank' rel='noreferrer noopener'>
+      <a href='#' role='button' {{on 'click' this.retryAuthentication}}>
```
ZedLi, line 14: How come we have the role set to a `button`? Is it not considered a link?

ui/admin/app/controllers/scopes/scope/authenticate/method/oidc.js
lisbet-alvarez: correct!

ui/desktop/app/templates/scopes/scope/authenticate/method/oidc.hbs
lisbet-alvarez: ah okay thats a mistake, i thought it was needed for the 'click' handler but its not.

PR #2644, opened 2025-01-06
test(e2e): add e2e test for scope chooser with filtered targets

hashicc:
Closes: https://hashicorp.atlassian.net/browse/ICU-16121

# Description
Adds an e2e test to ensure that the scope choose in the desktop app correctly filters the displayed targets.

:tickets: [Jira ticket](https://hashicorp.atlassian.net/browse/ICU-16121)

## How to Test

With enos and existing e2e setup working, run this command from the `e2e-tests` folder:
```
yarn playwright test --config desktop/playwright.config.js --ui
```

## Checklist

### DIFF
--- a/e2e-tests/desktop/tests/scope.spec.js
+++ b/e2e-tests/desktop/tests/scope.spec.js
@@ -1,0 +1,85 @@
+import { expect, test } from '../fixtures/baseTest.js';
+import * as boundaryHttp from '../../helpers/boundary-http.js';
+
+test.describe('Scope tests', async () => {
+  let orgA;
+  let projectA;
+  let targetA;
+
+  let orgB;
+  let projectB;
+  let targetB;
+
+  test.beforeEach(async ({ request, targetAddress, targetPort }) => {
+    // Group A resources
+    orgA = await boundaryHttp.createOrg(request);
+    projectA = await boundaryHttp.createProject(request, orgA.id);
+    targetA = await boundaryHttp.createTarget(request, {
+      scopeId: projectA.id,
+      type: 'tcp',
+      port: targetPort,
+      address: targetAddress,
+    });
+
+    // Group B resources
+    orgB = await boundaryHttp.createOrg(request);
+    projectB = await boundaryHttp.createProject(request, orgB.id);
+    targetB = await boundaryHttp.createTarget(request, {
+      scopeId: projectB.id,
+      type: 'tcp',
+      port: targetPort,
+      address: targetAddress,
+    });
+  });
+
+  test.afterEach(async ({ request }) => {
+    if (orgA) {
+      await boundaryHttp.deleteOrg(request, orgA.id);
+    }
+
+    if (orgB) {
+      await boundaryHttp.deleteOrg(request, orgB.id);
+    }
+  });
+
+  test('Shows the filtered targets based on selected scope', async ({ authedPage }) => {
+    const headerNav = await authedPage.getByLabel('header-nav');
+    await expect(headerNav).toBeVisible();
+    await expect(headerNav).toContainText('Global');
+
+    await expect(
+      authedPage.getByRole('link', { name: targetA.name }),
+    ).toBeVisible();
+    await expect(
+      authedPage.getByRole('link', { name: targetB.name }),
+    ).toBeVisible();
+
+    await headerNav.click();
+    const orgAHeaderNavLink = await authedPage.getByRole('link', {
+      name: orgA.name,
+    });
+    await orgAHeaderNavLink.click();
+
+    await expect(headerNav).toContainText(orgA.name);
+    await expect(
+      authedPage.getByRole('link', { name: targetA.name }),
+    ).toBeVisible();
+    await expect(
+      authedPage.getByRole('link', { name: targetB.name }),
+    ).not.toBeVisible();
+
+    await headerNav.click();
+    const orgBHeaderNavLink = await authedPage.getByRole('link', {
+      name: orgB.name,
+    });
+    await orgBHeaderNavLink.click();
+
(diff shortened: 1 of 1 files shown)


e2e-tests/desktop/tests/scope.spec.js
```diff
@@ -0,0 +1,92 @@
...
+      await boundaryHttp.deleteOrg(request, orgB.id);
+    }
+  });
+
+  test('Shows the filtered targets based on selected scope', async ({
+    authedPage,
+  }) => {
+    const headerNav = await authedPage.getByLabel('header-nav');
+    await expect(headerNav).toBeVisible();
+    await expect(headerNav).toContainText('Global');
```
ZedLi, line 55: Will these `toContainText` check for the other dropdown options as they're nested in the dom? I'm wondering if these are a bit of a false positive since I assume it would pass for any of the dropdown text as all of the options would be returned from [toContainText](https://playwright.dev/docs/api/class-locatorassertions#locator-assertions-to-contain-text).

e2e-tests/desktop/tests/scope.spec.js
```diff
@@ -0,0 +1,125 @@
+/**
+ * Copyright (c) HashiCorp, Inc.
+ * SPDX-License-Identifier: BUSL-1.1
+ */
+
+import { expect, test } from '../fixtures/baseTest.js';
+import * as boundaryHttp from '../../helpers/boundary-http.js';
+
+test.describe('Scope tests', async () => {
```
calcaide, line 9: None blocking comment: This test file is using a `describe` block to wrap all the file test. Following the previous test files such as `targets.spec` or `sessions.spec` seems like the pattern is define firts variables we might need, then `beforeEach` `afterEach` statements to finish creating a `describe` block and wrapping test cases within.
Is any reason why this test file follows a different approach? I don't have strong opinions in a specific approach, but it will be good we keep consistency between them.
Thoughts?

calcaide (dismissed):
Nice addition!! definetly we were missing this one, thanks for taking care!

I left a none blocking comment 😉

e2e-tests/desktop/tests/scope.spec.js
ZedLi: Oof I'll be honest, I didn't realize we were using a `details` and `summary` elements for these which makes it a little weird. In the interest of simplifying things, it looks like we attach the text to a title in the `summary` and again under a span with a class of `rose-dropdown-trigger-text` [here](https://github.com/hashicorp/boundary-ui/blob/9b1f052a569fc1579befffda87de3a91ba5e417b/addons/rose/addon/components/rose/dropdown/index.hbs#L25).

I think we should just keep it simple and use a locator for either one of those (obviously it's not ideal to use specific implementation details but I think the alternative is harder to understand) to confirm that the navigator is selecting the right scope, especially as we're in the middle of migrating to a different scope picker/nav bar https://github.com/hashicorp/boundary-ui/pull/2646. This means the test likely would need to get updated again after that change is made regardless. (cc: @lisbet-alvarez)

I'm curious if `await expect(getByLabel('header-nav').locator('summary')).toHaveText('Global');` or even `await expect(getByLabel('header-nav').locator('.rose-dropdown-trigger-text')).toHaveText('Global');` works.

hashicc: Agreed, it was a bit tricky without relying on selectors and the overhead of understanding `toMatchAriaSnapshot` is quite a bit.

e2e-tests/desktop/tests/scope.spec.js
hashicc: I agree, having a consistent pattern makes things much more approachable and easier to change later. I'll make the change and then we can figure out later if we want to change our approach.

My thoughts behind having the variables within the `describe` block were:
1. I was thinking it'd be clearer for garbage collection. In this case the variables aren't being exported or referenced by anything that's exported so I'm not sure if it'd be an issue, it's still in the "global" scope of the module. When the variables exist only within the function scope their lifetime of usage (I think) is clearer for garbage collection.
2. Some variables that are unchanging can live up top as `const` but having a separation between test definitions/helpers and `let` variables might make it clearer to the ownership of the variable is "one way layer lower" belonging to the `describe` block and and the `beforeEach` that's changing the variable.

This is the only [example](https://playwright.dev/docs/test-fixtures#without-fixtures) I've been able to find where they share a variable between tests. I see in other examples where they have top-level `const` variables but for `beforeEach` examples they mostly don't have a shared variable at all or use fixtures 🤔. I don't see them _not_ saying to use a shared variable though.  

Going down the rabbit hole of playwright best practices and found this [eslint plugin](https://github.com/playwright-community/eslint-plugin-playwright) that might be a nice addition, but it doesn't have anything specifically with regards to what we're chatting about.

calcaide: Hey @hashicc  those are valid and nice points, specially 1 I agree with what you say.

100% fine with it, that's why I made it a none blocking comment ;) and I think the discussion of what pattern to follow is very welcome for learning purposes team wide. What do you think bringing it to FE tech time? so all the team can chime in :)

e2e-tests/desktop/tests/scope.spec.js
hashicc: Move to simpler scoped selector on `summary` in https://github.com/hashicorp/boundary-ui/pull/2644/commits/50790c5ec8ce4f75298e62de56e19ca368f24d0e

DhariniJeeva (approved):
looks great, thanks for adding this!

ZedLi (approved):
Looks great, awesome work!

calcaide (approved):
Great work!! 🙌

PR #2652, opened 2025-01-09
refactor: 💡 Update name radio option label & set toggle default to on

lisbet-alvarez:
# Description

Apply feedback from design sync

https://github.com/user-attachments/assets/756c5a7c-71ce-4d6e-b9e4-ea8780fb4ab2

## How to Test

1. go to target/vault cred. store/storage bucket
2. validate worker filter generator is toggled on by default.

## Checklist

### DIFF
--- a/addons/core/translations/en-us.yaml
+++ b/addons/core/translations/en-us.yaml
@@ -154,7 +154,7 @@ worker-filter-generator:
     helper: Format a worker tag
   name:
     label: Name
-    helper: Format a worker name
+    helper: Format a worker name using regex
   operator:
     matches: matches
     contains: contains
--- a/addons/core/translations/en-us.yaml
+++ b/addons/core/translations/en-us.yaml
@@ -154,7 +154,7 @@ worker-filter-generator:
     helper: Format a worker tag
   name:
     label: Name
-    helper: Format a worker name using regex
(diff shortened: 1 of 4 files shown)


cameronperera (dismissed):
Thanks for the quick update!

madeleinetle (approved):

addons/core/translations/en-us.yaml
```diff
@@ -154,7 +154,7 @@ worker-filter-generator:
     helper: Format a worker tag
   name:
     label: Name
-    helper: Format a worker name
+    helper: Format a worker name using regex
```
ZedLi, line 157: Isn't regex only valid if a user chooses the `matches` keyword though? Or am I mistaken?

lisbet-alvarez: yeah i see ur point, the wording could imply that regex can be used with all operators.

lisbet-alvarez: Yup adding the wording change in the dropdown works.
If the user chooses `matches (regex)` then it should still be `"/name" matches ""`
Also side-note: `includes` should actually be `contains`

DhariniJeeva (approved):
looks good!

cameronperera (approved):

PR #2660, opened 2025-01-22
test: 💍 add e2e tests for credentials display in desktop

DhariniJeeva:
# Description

https://hashicorp.atlassian.net/browse/ICU-15986

## How to Test

Run through the admin README to setup enos using enterprise boundary. At the e2e root folder, run `yarn desktop` 

## Checklist


e2e-tests/desktop/tests/credentials.spec.js
```diff
@@ -0,0 +1,285 @@
...
+        nested: {
+          key1: 'value1',
+          key2: 'value2',
+        },
+      },
+    });
+
+    // Create Vault token and apply policies
+    execSync(
+      `vault policy write ${boundaryPolicyName} ./admin/tests/fixtures/boundary-controller-policy.hcl`,
```
DhariniJeeva, line 88: I'm reusing the ones from admin, let me know if we should keep this separate and create policies for desktop folder

DhariniJeeva:
I pulled credentials out to a separate file, but I haven't removed static credential test from target, I will push those changes soon.

e2e-tests/desktop/tests/credentials.spec.js
ZedLi: I think it's fine and we can pull it out at some point but can we not hardcode the path since this assumes you're running it from the root?

DhariniJeeva: ah good point, I will look into this

e2e-tests/desktop/tests/targets.spec.js
```diff
@@ -177,34 +141,6 @@ test.describe('Targets tests', async () => {
         authedPage.getByRole('heading', { name: 'Sessions' }),
       ).toBeVisible();
 
-      await expect(authedPage.getByText(credential.name)).toBeVisible();
```
DhariniJeeva, line 180: moved these to credentials file

e2e-tests/desktop/tests/credentials.spec.js
```diff
@@ -0,0 +1,285 @@
...
+      .getByRole('listitem')
+      .filter({ hasText: 'password' })
+      .getByLabel('Toggle secret visibility')
+      .click();
+    await expect(
+      authedPage
+        .getByRole('listitem')
+        .filter({ hasText: 'password' })
+        .locator('pre'),
+    ).toHaveText(String.raw`pass\word`);
```
ZedLi, line 233: How come this is ``String.raw`pass\word` ``  since on [line 93](https://github.com/hashicorp/boundary-ui/pull/2660/files#diff-411688c020c9741d39b6c48029c502b7b4c831aa55f11a0f40dbfbe496bf7c77R93) it's ``String.raw`pass\\word` ``?

e2e-tests/desktop/tests/credentials.spec.js
```diff
@@ -0,0 +1,310 @@
...
+      target: targetWithInjectedCredentials,
+      hostSourceIds: [hostSet.id],
+    });
+    targetWithInjectedCredentials = await boundaryHttp.addInjectedCredentials(
+      request,
+      {
+        target: targetWithInjectedCredentials,
+        credentialIds: [sshPrivateKeyCredential.id],
+      },
+    );
```
ZedLi, line 198: Maybe I missed this but what is `targetWithInjectedCredentials` used for?

e2e-tests/helpers/boundary-http/credential-libraries.js
```diff
@@ -0,0 +1,32 @@
...
+import { checkResponse } from './responseHelper.js';
+
+/**
+ * Creates a vault generic credential library
+ * @param {import('@playwright/test').APIRequestContext} request
+ * @param {string} credentialStoreId ID of the credential store where the credential will be created
+ * @param {string} vaultPath Path to secret in vault
+ * @returns {Promise<Serializable>}
+ */
+export async function createVaultCredentialGenericSecret(
```
ZedLi, line 16: Should this be `createVaultGenericSecretsCredentialLibrary` instead?

e2e-tests/desktop/tests/credentials.spec.js
```diff
@@ -0,0 +1,310 @@
...
+
+    // Create a static credential store and json credential
+    jsonCredential = await boundaryHttp.createStaticCredentialJson(request, {
+      credentialStoreId: credentialStore.id,
+      data: {
+        nested: {
+          key1: 'value1',
+          key2: 'value2',
```
ZedLi, line 91: Can we also include characters that would normally be escaped here as well to confirm the behavior?

e2e-tests/desktop/tests/targets.spec.js
```diff
@@ -89,10 +85,6 @@ test.beforeEach(
       target: targetWithTwoHosts,
       hostSourceIds: [hostSetWithTwoHosts.id],
     });
-    targetWithTwoHosts = await boundaryHttp.addBrokeredCredentials(request, {
```
ZedLi, line 92: I think we'd still want to test the use case that has a target with two hosts, I think it's fine if one of them just happens to be a brokered credential

e2e-tests/desktop/tests/credentials.spec.js
```diff
@@ -0,0 +1,285 @@
...
+    targetAddress,
+    targetPort,
+    sshUser,
+    sshKeyPath,
+    vaultAddr,
+    page,
+  }) => {
+    execSync(`vault policy delete ${secretPolicyName}`);
+    execSync(`vault policy delete ${boundaryPolicyName}`);
+    execSync(`vault secrets disable ${secretsPath}`);
```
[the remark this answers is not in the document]
DhariniJeeva: yeah I was getting an error saying that the path `e2e-secrets` already exists when the test is run again and I noticed that we delete these first in beforeEach so I followed the same pattern here

e2e-tests/desktop/tests/credentials.spec.js
DhariniJeeva: without escaping the `\`, using single quotes around the password like '`String.raw`pass\word`'  seems to work fine when passed to vault but eventually it times out and crashes, so I tried escaping the backslash.

e2e-tests/desktop/tests/credentials.spec.js
DhariniJeeva: I originally moved this from targets, but I moved it back. But I will have to check if injected credentials are also displayed in the credential panel for ssh targets

e2e-tests/desktop/tests/targets.spec.js
DhariniJeeva: yeah that makes sense, I modified the tests name for the new tests

e2e-tests/desktop/tests/credentials.spec.js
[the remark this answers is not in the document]
hashicc: It looks like other tests are using the same `secretPolicyName`, `boundaryPolicyName`, `secretsPath`, and those are using `beforeEach` (here's an [example](https://github.com/hashicorp/boundary-ui/blob/1cbf3b227ebe6f8d0eb8d2ae8c6e43c606e87a34/e2e-tests/admin/tests/credential-store-vault-ent.spec.js#L28-L30)). I think that those test suites are leaving the global state in an unclean state so the other tests then need to also use `beforeEach`, including this test.

One thing that might help is adding a `uuid` suffix or some shorter random identifier to avoid collisions. If that works, then in a separate PR we could migrate the others to `afterEach` and/or random suffix.

e2e-tests/desktop/tests/credentials.spec.js
```diff
@@ -0,0 +1,294 @@
...
+      clientToken,
+    );
+    vaultGenericCredentialLibrary =
+      await boundaryHttp.createVaultGenericSecretsCredentialLibrary(request, {
+        credentialStoreId: vaultCredentialStore.id,
+        vaultPath: `${secretsPath}/data/${secretName}`,
+      });
+
+    // Create a tcp target with vault brokered credential Libraries
+    targetWithBrokeredVaultCredentials = await boundaryHttp.createTarget(
```
hashicc, line 139: It looks like `targetWithBrokeredVaultCredentials` is being re-assigned here from a few different `boundaryHttp` methods and then also in the arguments. Is it the same target just being returned from each operation `createTarget`, `addHostSource`, `addBrokeredCredentials`? It's difficult to tell by looking at the method signatures of the `boundaryHttp` client.

😄 I don't know if there's anything actionable here. It might be a case where the pipe operator ([still being worked out](https://github.com/tc39/proposal-pipeline-operator)) would nicely chain these and scope the variable from one call to the next.

e2e-tests/desktop/tests/credentials.spec.js
```diff
@@ -0,0 +1,294 @@
...
+        .getByRole('listitem')
+        .filter({ hasText: 'password' })
+        .locator('pre'),
+    ).toHaveText(String.raw`pass\word`);
+
+    await authedPage.getByRole('button', { name: 'End Session' }).click();
+    await expect(authedPage.getByText('Canceled successfully.')).toBeVisible();
+    await expect(
+      authedPage.getByRole('heading', { name: 'Targets' }),
+    ).toBeVisible();
```
hashicc, line 223: Is this being done for clean up? I wonder if we can get by with just relying on cleanup of the org.

e2e-tests/desktop/tests/credentials.spec.js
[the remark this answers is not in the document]
calcaide: I do have no strong opinion on what approach to take. What Chad suggest could be a possible solution, in case we plan into run these tests in parallel it will help to avoid collisions.

Personally, I would prefer the test that creates those `secretPolicyName`, `boundaryPolicyName` and `secretPath` within the `beforeEach` also takes responsibility on deleting them perhaps within an `afterEach`.

I am fine with the current approach within this PR, we can have consensus in what approach to take and refactor this specific part in a dedicated PR.

e2e-tests/desktop/tests/credentials.spec.js
DhariniJeeva: this is not needed, we only display brokered credentials in the cred panel in desktop, I have removed these from credentials test file.

e2e-tests/desktop/tests/credentials.spec.js
DhariniJeeva: ah yes it is the same target. After creation, we use the created target to addHostSource and broker credentials to that target and then in desktop, we connect to this target to view the brokered credentials and make sure the credentials are displayed in the expected format

e2e-tests/desktop/tests/credentials.spec.js
DhariniJeeva: yeah we end the session and make sure we cancel it successfully and then we handle org deletion in afterEach. How would we handle this org cleanup? 🤔  I think org deletion can still have sessions but I'm also not 100% sure, is there a way to check if there are abandoned sessions?

hashicc: That's a good point, maybe deleting the org wouldn't do all the cascading deletes necessary for resources it own. I think it's alright to leave this clean up here since the clean up is relevant only to this test and it's explicit. My comment was mostly centered around reading these lines and figuring out if it was related to the purpose of the test, or clean up, and then thought "maybe this test could be simpler if the deleting org cleaned up all this". Maybe a comment above these steps letting people know it's just for clean up would help future readers of this test case.

e2e-tests/desktop/tests/credentials.spec.js
ZedLi: I see, so this is how vault handles escape characters?

e2e-tests/desktop/tests/credentials.spec.js
ZedLi: I wanted to mention we also need to manually cancel these sessions as when we try and close the DC while there are active sessions, we have an electron native dialog pop up [here](https://github.com/hashicorp/boundary-ui/blob/371301b4bad27a5df529de6ec5e84380aabf45d7/ui/desktop/electron-app/src/index.js#L293) which I don't think we can click through with playwright (as far as I could tell quickly testing)

e2e-tests/desktop/tests/credentials.spec.js
DhariniJeeva: so I tested this using spawnSync and it didn't need an esc character, I'm thinking some weird issues with special characters in shell env and nothing to do with vault.

e2e-tests/desktop/tests/credentials.spec.js
DhariniJeeva: thanks for bringing this up, Chad. I added a comment for this.

e2e-tests/desktop/tests/credentials.spec.js
```diff
@@ -0,0 +1,285 @@
...
+    sshUser,
+    sshKeyPath,
+    vaultAddr,
+    page,
+  }) => {
+    execSync(`vault policy delete ${secretPolicyName}`);
+    execSync(`vault policy delete ${boundaryPolicyName}`);
+    execSync(`vault secrets disable ${secretsPath}`);
+
+    await page.goto('/');
```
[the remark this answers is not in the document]
DhariniJeeva: oops this was accidentally pasted and don't need it

e2e-tests/desktop/tests/credentials.spec.js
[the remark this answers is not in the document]
DhariniJeeva: thanks for looking into this. yeah, sounds like a good opportunity to clean these up. I can test the suffix suggestion and handle in a separate PR if that's something the team prefers and modify all the tests where we use these

calcaide (dismissed):
LGTM! Thanks for adding this e2e test!! 🙌

hashicc (approved):
🚀 Another e2e test! Nice to see the test suite growing

e2e-tests/desktop/tests/credentials.spec.js
```diff
@@ -0,0 +1,304 @@
...
+
+    // End session as active sessions will show a popup when trying to close the DC
+    await authedPage.getByRole('button', { name: 'End Session' }).click();
+    await expect(authedPage.getByText('Canceled successfully.')).toBeVisible();
+    await expect(
+      authedPage.getByRole('heading', { name: 'Targets' }),
+    ).toBeVisible();
+  });
+
+  test('Display JSON static credential in a key value format and handle special characters', async ({
```
ZedLi, line 236: How come we don't end session in this test? I also get the popup at the end of this test

DhariniJeeva: oh it should be added, I also received a popup at the end of the test and missed to add it here, thank you for testing!

ZedLi (approved):
Thanks for adding another test!

PR #2664, opened 2025-01-24
test(e2e/desktop): Address failures in session tests

moduli:
# Description
This PR addresses some test failures in the desktop client e2e test suite when I ran them against 0.19.0.

The first issue was the use of this locator: `await authedPage.getByLabel('Cancel').all()`. This was including the `Canceling` search filter, which caused some behavior problems in the test -- I believe we're just looking for the cancel session button here. This locator was modified to be more specific and only look for `Cancel` buttons in the session table.
Screenshot 2025-01-24 at 2 12 21 PM

The other issue was with canceling multiple sessions. When I was running it locally, it was never able to cancel all the sessions on the page (it would miss one every time). This new implementation does a few things
- works in a loop to cancel the number of cancel buttons it sees by clicking the first one each time
- dismisses the success alert after each cancel so that the test is more methodical

## How to Test
```
❯ yarn desktop:dev                       
yarn run v1.22.22
$ yarn playwright test --config desktop/playwright.config.js --reporter=html
$ /Users/mycow/Developer/boundary-ui/node_modules/.bin/playwright test --config desktop/playwright.config.js --reporter=html

Running 17 tests using 1 worker
  Consider splitting slow test files to speed up parallel execution
  1 skipped
  16 passed (2.7m)

To open last HTML report run:

  yarn playwright show-report

✨  Done in 163.77s.
```

## Checklist


e2e-tests/desktop/pages/sessionPage.js
```diff
@@ -0,0 +1,32 @@
...
+  constructor(page) {
+    this.page = page;
+  }
+
+  async cancelAllSessions() {
+    while (
+      (await this.page
+        .getByRole('row')
+        .filter({ hasNot: this.page.getByRole('columnheader') })
+        .getByLabel('Cancel')
```
ZedLi, line 18: I think this is good to be specific in the table as I totally forgot about the other `Canceling` text on the screen, but I wonder if we should have just done `getByLabel('Cancel', {exact: true})` instead? Maybe we should do it now either way?

e2e-tests/desktop/pages/sessionPage.js
```diff
@@ -0,0 +1,32 @@
+/**
+ * Copyright (c) HashiCorp, Inc.
+ * SPDX-License-Identifier: BUSL-1.1
+ */
+
+import GlobalPage from './globalPage.js';
+
+export default class SessionPage {
```
ZedLi, line 8: Should this have extended the `globalPage` rather than importing here? We also called it `basePage` in admin, should they be called the same?

e2e-tests/desktop/pages/sessionPage.js
```diff
@@ -0,0 +1,32 @@
...
+  constructor(page) {
+    this.page = page;
+  }
+
+  async cancelAllSessions() {
```
ZedLi, line 13: What do you think about doing something like this instead?

```
    await expect
      .poll(
        async () => {
          await this.page
            .getByRole('row')
            .filter({ hasNot: this.page.getByRole('columnheader') })
            .getByLabel('Cancel')
            .first()
            .click();

          const globalPage = new GlobalPage(this.page);
          await globalPage.dismissSuccessAlert();

          return await this.page
            .getByRole('row')
            .filter({ hasNot: this.page.getByRole('columnheader') })
            .getByLabel('Cancel')
            .count();
        },
        {
          message: 'Ensure all sessions are cancelled',
        },
      )
      .toBe(0);
```

I like having the built in expect clause for this rather than using just a while loop and it makes the error more obvious if this does fail

ZedLi (approved):
Left a few non blocking comments, thanks for fixing this!

ZedLi (approved):
Looks good!

e2e-tests/desktop/pages/sessionPage.js
ZedLi: Yeah sorry I should have been clearer, I like the filter but I do think it's better to have the `exact`

PR #2669, opened 2025-01-28
feat: 🎸 add showFilterGenerator to WorkerFilterGenerator comp

cameronperera:
✅ Closes: https://hashicorp.atlassian.net/browse/ICU-16282

# Description

This change makes it so we can determine if the "Show filter generator" is toggled on by default or not. When the generator is used in a longer form like creating a Vault credential store or storage bucket, we want to hide the generator. This keeps the form from being overly long. In places like the Worker Filter tab for Target for Vault credential store, we have a route dedicated to updating the worker filter where showing the generator by default makes more sense.

New Storage Bucket:

https://github.com/user-attachments/assets/629e6f7a-c9a5-465e-bd20-902035b4c422

New Vault credential store:

https://github.com/user-attachments/assets/f99bfc22-db29-425f-b6b5-f4cf60b2a3b0

Editing Vault credential store worker filter:

https://github.com/user-attachments/assets/b779dadd-5fba-44ac-a143-3e7af0a97747

Editing Target worker filter:

https://github.com/user-attachments/assets/10089452-f9e7-4b23-94c8-39b3a635caa7

## Checklist

### DIFF
--- a/ui/admin/app/components/form/worker-filter/index.hbs
+++ b/ui/admin/app/components/form/worker-filter/index.hbs
@@ -10,7 +10,11 @@
   @disabled={{@model.isSaving}}
   as |form|
 >
-  <WorkerFilterGenerator @model={{@model}} @name={{@name}} />
+  <WorkerFilterGenerator
+    @model={{@model}}
+    @name={{@name}}
+    @showFilterGenerator={{true}}
+  />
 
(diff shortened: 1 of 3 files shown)


lisbet-alvarez (approved):
Thanks for this improvement!!

DhariniJeeva (approved):
looks good

PR #2672, opened 2025-01-30
fix: 🐛 update hidden secrets and flatten-nested-object logic

cameronperera:
# Description

update hidden secrets and flatten-nested-object logic

Before:

https://github.com/user-attachments/assets/53fece78-8290-41c2-9cc7-062783db53a6

After:

https://github.com/user-attachments/assets/e67b1801-2d81-432f-bdba-53941c5b130f

## How to Test

Create a JSON credential with varying types of values like 0, false, true, null, and '' (empty string). Only null and '' values should be filtered out. All secrets are still displayed when viewing the API output.

## Checklist

### DIFF
--- a/addons/api/addon/utils/flatten-nested-object.js
+++ b/addons/api/addon/utils/flatten-nested-object.js
@@ -18,8 +18,8 @@ export const flattenObject = (obj, result = {}, parentKey = '') => {
     if (typeOf(obj[key]) === 'object' && obj[key] !== null) {
       // Recursively flatten the object
       flattenObject(obj[key], result, `${parentKey}${key}.`);
-    } else if (obj[key]) {
-      // Only add to result if the value is not null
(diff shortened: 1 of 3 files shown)


ZedLi (approved):
Looks good! Thanks for the fix, it might also be good to add some of these examples to the nested json [here](https://github.com/hashicorp/boundary-ui/blob/064375d4573bac0e160479213b8a65bd225c4072/e2e-tests/desktop/tests/credentials.spec.js#L86) and in the vault example to assert the correct display

lisbet-alvarez (approved):

laurenolivia:
@cameronperera Had a couple non-blocking questions, and didn't want to hold up your merge. Will chat offline.

DhariniJeeva:
great catch and thanks for the fix!

PR #2674, opened 2025-02-01
fix: 🐛 reapply storage policy errors out after a failed delete

lisbet-alvarez:
# Description

When a delete fails on a session-recording, the `isDeleted` attribute is set to `true` but `isDestroyed` is set to `false` so when we try to use `this.save` on model it sends another `DELETE` request instead of a `POST` request for the `reapply-storage-policy` method. Solution was to `rollbackAttributes` after a failed `DELETE`.

Additionally, hide delete button when storage policy is retain "forever" on a session recording.

BEFORE:

https://github.com/user-attachments/assets/5d841191-f87e-4690-8914-211765132526

AFTER:

https://github.com/user-attachments/assets/0fbd3f9b-b003-49e1-9b24-9c9d6ba1ef7f

hide delete button when retain policy is "Forever"

Uploading Screen Recording 2025-02-04 at 1.23.35 PM.mov…

## Checklist

### DIFF
--- a/ui/admin/app/abilities/session-recording.js
+++ b/ui/admin/app/abilities/session-recording.js
@@ -32,10 +32,12 @@ export default class OverrideSessionRecordingAbility extends SessionRecordingAbi
 
   /**
    * This override ensures that session recordings may be deleted only if the
-   * session-recording feature flag is enabled.
+   * session-recording feature flag is enabled and policy allows deletion.
    */
   get canDelete() {
-    return this.features.isEnabled('ssh-session-recording')
+    const retainForever =
+      this.model.retain_until?.toISOString() === '9999-12-31T23:23:23.999Z';
+    return this.features.isEnabled('ssh-session-recording') && !retainForever
       ? super.canDelete
       : false;
   }
--- a/ui/admin/app/abilities/session-recording.js
+++ b/ui/admin/app/abilities/session-recording.js
@@ -35,9 +35,8 @@ export default class OverrideSessionRecordingAbility extends SessionRecordingAbi
    * session-recording feature flag is enabled and policy allows deletion.
    */
   get canDelete() {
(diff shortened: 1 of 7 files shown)


ZedLi (approved):
Tested and works great, thanks for the fix!

cameronperera (approved):

cameronperera (dismissed):

ui/admin/app/abilities/session-recording.js
```diff
@@ -32,10 +32,12 @@ export default class OverrideSessionRecordingAbility extends SessionRecordingAbi
 
   /**
    * This override ensures that session recordings may be deleted only if the
-   * session-recording feature flag is enabled.
+   * session-recording feature flag is enabled and policy allows deletion.
    */
   get canDelete() {
-    return this.features.isEnabled('ssh-session-recording')
+    const retainForever =
+      this.model.retain_until?.toISOString() === '9999-12-31T23:23:23.999Z';
```
ZedLi, line 39: Should this have just been a getter on the model? It seems we would use this in more than one place

ZedLi (approved):
Looks good, left a non blocking comment

ui/admin/app/abilities/session-recording.js
lisbet-alvarez: ohh great point, ill update rn!

calcaide (dismissed):

DhariniJeeva (approved):
thanks for the fix, looks great!

ui/admin/app/abilities/session-recording.js
lisbet-alvarez: updated!

ZedLi (approved):
Thanks for the fix!

cameronperera (approved):

PR #2680, opened 2025-02-03
chore: add `eslint-plugin-playwright` rules for e2e tests and fix errors

hashicc:
# Description
This prs adds the [`eslint-plugin-playwright`](https://github.com/playwright-community/eslint-plugin-playwright) eslint rule with the recommended rules applied.

There are 110 new warnings reported, most of which are either:
* `Avoid having conditionals in tests  (playwright/no-conditional-in-test)`
* `Test has no assertions (playwright/expect-expect)`

We could:
1. leave these warnings
2. ignore them individually
3. ignore them globally within our eslint config

I think these are good warnings to have though so I'd recommend either option 1. or 2.

## How to Test
1. Add an `async` to a describe block callback
6. Error should be reported by `yarn lint` and/or within the editor

## Checklist

### DIFF
--- a/e2e-tests/admin/tests/auth-method-ldap.spec.js
+++ b/e2e-tests/admin/tests/auth-method-ldap.spec.js
@@ -218,14 +218,14 @@ test('Set up LDAP auth method @ce @ent @docker', async ({
       }
     }
 
-    expect(
+    await expect(
       page
         .getByRole('cell', { name: ldapAccountName })
         .locator('..')
         .getByRole('cell')
         .nth(fullNameIndex),
     ).toHaveText(ldapUserName);
-    expect(
+    await expect(
       page
         .getByRole('cell', { name: ldapAccountName })
         .locator('..')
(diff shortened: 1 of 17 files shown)


e2e-tests/admin/tests/auth-method-ldap.spec.js
```diff
@@ -218,14 +218,14 @@ test('Set up LDAP auth method @ce @ent @docker', async ({
       }
     }
 
-    expect(
+    await expect(
```
hashicc, line 221: This plugin catches certain expects that should be awaited.

moduli (approved):

DhariniJeeva (approved):
thanks for adding this!

calcaide:
None blocking comment:

When running `yarn lint` within `boundary-ui/e2e-tests/` I am getting errors within `e2e-tests/playwright-report`. Do we need to exclude the directory `playwrght-report/` from the linting?

PR #2685, opened 2025-02-12
refactor: 💅  Aliases: create-test

laurenolivia:
## Description
This PR refactors our aliases: create-test suite.

:tickets: [Jira ticket](https://hashicorp.atlassian.net/browse/ICU-15454)

## How to Test
Tests should pass ✅ 

## Checklist

### DIFF
--- a/ui/admin/app/components/form/alias/index.hbs
+++ b/ui/admin/app/components/form/alias/index.hbs
@@ -24,7 +24,7 @@
     {{#if @model.errors.name}}
       <F.Error as |E|>
         {{#each @model.errors.name as |error|}}
-          <E.Message>{{error.message}}</E.Message>
+          <E.Message data-test-error-message-name>{{error.message}}</E.Message>
         {{/each}}
       </F.Error>
     {{/if}}
--- a/ui/admin/tests/acceptance/aliases/create-test.js
+++ b/ui/admin/tests/acceptance/aliases/create-test.js
@@ -50,8 +50,8 @@ module('Acceptance | aliases | create', function (hooks) {
     await visit(urls.newAlias);
 
     await fillIn(commonSelectors.FIELD_NAME, commonSelectors.FIELD_NAME_VALUE);
-    await fillIn(selectors.DESTINATION_ID_SELECTOR, 'tcp_123');
-    await fillIn(selectors.HOST_ID_SELECTOR, 'h_123');
+    await fillIn(selectors.FIELD_DESTINATION_ID_SELECTOR, 'tcp_123');
+    await fillIn(selectors.FIELD_HOST_ID_SELECTOR, 'h_123');
     await click(commonSelectors.SAVE_BTN);
     const alias = this.server.schema.aliases.findBy({
       name: commonSelectors.FIELD_NAME_VALUE,
--- a/ui/admin/tests/acceptance/aliases/create-test.js
+++ b/ui/admin/tests/acceptance/aliases/create-test.js
@@ -14,8 +14,6 @@ import { setupIndexedDb } from 'api/test-support/helpers/indexed-db';
 import * as commonSelectors from 'admin/tests/helpers/selectors';
 import * as selectors from './selectors';
 
-const FIELD_ERROR_TEXT_SELECTOR = '.hds-form-error__message';
-
 module('Acceptance | aliases | create', function (hooks) {
   setupApplicationTest(hooks);
   setupMirage(hooks);
@@ -114,7 +112,7 @@ module('Acceptance | aliases | create', function (hooks) {
     assert
       .dom(commonSelectors.ALERT_TOAST_BODY)
       .hasText('The request was invalid.');
-    assert.dom(FIELD_ERROR_TEXT_SELECTOR).hasText('Name is required.');
+    assert.dom(commonSelectors.FIELD_TEXT_ERROR).hasText('Name is required.');
   });
 
   test('can navigate to new aliases route with proper authorization', async function (assert) {
--- a/ui/admin/tests/acceptance/aliases/create-test.js
+++ b/ui/admin/tests/acceptance/aliases/create-test.js
@@ -12,6 +12,9 @@ import { Response } from 'miragejs';
 import { authenticateSession } from 'ember-simple-auth/test-support';
 import { setupIndexedDb } from 'api/test-support/helpers/indexed-db';
 import * as commonSelectors from 'admin/tests/helpers/selectors';
+import * as selectors from './selectors';
+
+const FIELD_ERROR_TEXT_SELECTOR = '.hds-form-error__message';
 
 module('Acceptance | aliases | create', function (hooks) {
   setupApplicationTest(hooks);
@@ -20,14 +23,6 @@ module('Acceptance | aliases | create', function (hooks) {
 
   let features;
   let getAliasCount;
-
-  const SAVE_BTN_SELECTOR = '[type="submit"]';
-  const CANCEL_BTN_SELECTOR = '.rose-form-actions [type="button"]';
-  const NAME_FIELD_SELECTOR = '[name="name"]';
-  const DESTINATION_ID_SELECTOR = '[name="destination_id"]';
-  const HOST_ID_SELECTOR = '[name="authorize_session_arguments"]';
-  const FIELD_ERROR_TEXT_SELECTOR = '.hds-form-error__message';
-  const NAME_FIELD_TEXT = 'random string';
 
   const instances = {
     scopes: {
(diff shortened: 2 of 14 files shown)


ui/admin/tests/acceptance/aliases/selectors.js
```diff
@@ -0,0 +1,7 @@
+/**
+ * Copyright (c) HashiCorp, Inc.
+ * SPDX-License-Identifier: BUSL-1.1
+ */
+
+export const DESTINATION_ID_SELECTOR = '[name="destination_id"]';
```
calcaide, line 6: If these selectors refers to a field, the pattern to follow will be FIELD_[NAME_OF_THE_FILED] , as examples you can check [accounts selector file](https://github.com/hashicorp/boundary-ui/blob/main/ui/admin/tests/acceptance/accounts/selectors.js#L18-L28) or [auth-method seelctors file](https://github.com/hashicorp/boundary-ui/blob/main/ui/admin/tests/acceptance/auth-methods/selectors.js#L6-L97)

ui/admin/tests/acceptance/aliases/create-test.js
```diff
@@ -56,37 +49,37 @@ module('Acceptance | aliases | create', function (hooks) {
     const aliasCount = getAliasCount();
     await visit(urls.newAlias);
 
-    await fillIn(NAME_FIELD_SELECTOR, NAME_FIELD_TEXT);
-    await fillIn(DESTINATION_ID_SELECTOR, 'tcp_123');
-    await fillIn(HOST_ID_SELECTOR, 'h_123');
-    await click(SAVE_BTN_SELECTOR);
+    await fillIn(commonSelectors.FIELD_NAME, commonSelectors.FIELD_NAME_VALUE);
+    await fillIn(selectors.DESTINATION_ID_SELECTOR, 'tcp_123');
```
calcaide, line 53: If you feel the value we are fillingIn this field will be need again `tcp_123`, in example, to validate an assertion when updating the field, the pattern we follow is to create a const for it. For example a [field selector + field value on auth-methods domain](https://github.com/hashicorp/boundary-ui/blob/main/ui/admin/tests/acceptance/auth-methods/selectors.js#L10-L11) or [accounts selectors](https://github.com/hashicorp/boundary-ui/blob/main/ui/admin/tests/acceptance/accounts/selectors.js#L24-L25).
If you know or no need now to repeat this value, hardcoding is totally fine 😉

calcaide (changes requested):
Great work for selector refactor.

I will request to keep the scope of this PR only within aliases domain. Currently, it's including work on auth-methds, roles, scopes and targets domain.

ui/admin/tests/acceptance/aliases/create-test.js
laurenolivia: I verified in the aliases acceptance tests (create, read, update, list, delete) that those two fields are never reused. 👍

hashicc (approved):
💯 Looks good! Nice to see some of the `commonSelectors` fill in the blanks here

DhariniJeeva (approved):
looks good!

ui/admin/tests/acceptance/aliases/selectors.js
```diff
@@ -0,0 +1,7 @@
+/**
+ * Copyright (c) HashiCorp, Inc.
+ * SPDX-License-Identifier: BUSL-1.1
+ */
+
+export const FIELD_DESTINATION_ID_SELECTOR = '[name="destination_id"]';
+export const FIELD_HOST_ID_SELECTOR = '[name="authorize_session_arguments"]';
```
cameronperera, line 7: `_SELECTOR` can be removed from these selectors since it is implied when importing from a selectors file.

ui/admin/tests/helpers/selectors.js
```diff
@@ -10,6 +10,7 @@ export const EDIT_BTN = '.rose-form-actions [type=button]';
 export const FIELD_NAME = '[name=name]';
 export const FIELD_NAME_VALUE = 'Random name';
 export const FIELD_NAME_ERROR = '[data-test-error-message-name]';
+export const FIELD_TEXT_ERROR = '.hds-form-error__message';
```
cameronperera, line 13: I don't think we actually need this selector. The `FIELD_NAME_ERROR` should be the selector for finding the error. If that `data-test` attribute isn't in the template file, you can simply add it.

cameronperera (changes requested):

ui/admin/tests/acceptance/aliases/selectors.js
laurenolivia: Is the general rule that we append `_SELECTOR` if the const is not in a selectors file, otherwise we do not append?

cameronperera: Yes that is correct.

ui/admin/tests/helpers/selectors.js
cameronperera: It is supposed to be specifically related to name. Any error we need to act or assert against should have an explicit error selector. I explained this more in the other [PR](https://github.com/hashicorp/boundary-ui/pull/2693#issuecomment-2669582232).

calcaide: +1 to the last comment.

ui/admin/tests/acceptance/aliases/create-test.js
```diff
@@ -113,13 +106,13 @@ module('Acceptance | aliases | create', function (hooks) {
...
     await visit(urls.newAlias);
-    await click(SAVE_BTN_SELECTOR);
+    await click(commonSelectors.SAVE_BTN);
     await a11yAudit();
 
     assert
       .dom(commonSelectors.ALERT_TOAST_BODY)
       .hasText('The request was invalid.');
-    assert.dom(FIELD_ERROR_TEXT_SELECTOR).hasText('Name is required.');
+    assert.dom(commonSelectors.FIELD_TEXT_ERROR).hasText('Name is required.');
```
calcaide, line 115: I know the code was already there, as part of refactoring selectors, and one of the reasons to keep it one domain at a time, is to also "evaluate" our tests.

This test is mocking a field error and asserting the form behaves the way we expect. In this case, we are marking the field name as required, but the aliases creation form does not have the field name as required.
The aliases creation form, has as required alias value. I will suggest to modify this mock to mark as required the alias value field.
And as a selector for such error message [I will recomend to follow what has been mentioned here](https://github.com/hashicorp/boundary-ui/pull/2685#discussion_r1962277721).

calcaide (changes requested):
I left 2 comments

cameronperera (dismissed):
LGTM

ui/admin/app/components/form/alias/index.hbs
```diff
@@ -24,7 +24,7 @@
     {{#if @model.errors.name}}
       <F.Error as |E|>
         {{#each @model.errors.name as |error|}}
-          <E.Message>{{error.message}}</E.Message>
+          <E.Message data-test-error-message-name>{{error.message}}</E.Message>
```
calcaide, line 27: I see we add this `data-test-` selector, and this is all good and fine. But I do not see any selector using it. If we are not using this selector, do we need to add it?

calcaide (changes requested):
Left a comment on the data-test selector. Almost there, refactor looking good 😉

ui/admin/app/components/form/alias/index.hbs
laurenolivia: That error message is currently selecting on [an hds class](https://github.com/hashicorp/boundary-ui/blob/main/ui/admin/tests/acceptance/aliases/create-test.js#L29) and we agreed to move away from selecting on hds classes, so I removed `FIELD_ERROR_TEXT_SELECTOR` and replaced it with our common selector [FIELD_NAME_ERROR](https://github.com/hashicorp/boundary-ui/blob/750cc1ecae338621bdbfa3b0bac5dbac5d426d05/ui/admin/tests/helpers/selectors.js#L12), but I had to add the data test attr to that error for [this test](https://github.com/hashicorp/boundary-ui/blob/2e7c5716c2e1d712af9a17e94628d38e3ec534d6/ui/admin/tests/acceptance/aliases/create-test.js#L115).

calcaide: Gotcha. I miss that.

Recap. We are adding the data-test-* attribute to the template, so the common selector can be used within this creation form 😉

calcaide (approved):
Let's go!! looks good to me 😃

DhariniJeeva (approved):
looks good!

DhariniJeeva (approved):

lisbet-alvarez (approved):

PR #2688, opened 2025-02-13
refactor: 💡 use hds side nav w/ hds frame instead of rose

lisbet-alvarez:
# Description

Refactor to use `Hds::SideNav` and `Hds::AppFrame` components.
Note: There is an issue with the `Hds::SideNav::List` component so we are using the `Hds::SideNav::Portal` component in the meantime.

The side nav should not be visible on the `onboarding`, `change-password` , and `authenticate` routes.

## How to Test

1. Validate all links in the side nav route to expected pages.
2. Validate side nav is present in expected pages.
3. Validate dropdown options still function as expected.
    a. user menu options such as sign-out, change-password, light mode, dark mode, etc.
    b. dev tool toggle options such as feature editions and licensed features.

## Checklist


ui/admin/app/templates/scopes/scope.hbs
```diff
@@ -33,218 +33,314 @@
...
-            )
-          }}
-            <nav.link @route='scopes.scope.workers'>
-              <Rose::Icon @name='flight-icons/svg/cpu-16' @size='medium' />
-              {{t 'titles.workers'}}
-            </nav.link>
-          {{/if}}
-        </Rose::Nav::Sidebar>
+{{#if this.session.isAuthenticated}}
+  {{#in-element this.sideBarContainer}}
```
ZedLi, line 37: Why did we have to portal this in?

ui/admin/app/templates/scopes/scope.hbs
```diff
@@ -33,218 +33,314 @@
...
-              (feature-flag 'byow')
-            )
-          }}
-            <nav.link @route='scopes.scope.workers'>
-              <Rose::Icon @name='flight-icons/svg/cpu-16' @size='medium' />
-              {{t 'titles.workers'}}
-            </nav.link>
-          {{/if}}
-        </Rose::Nav::Sidebar>
+{{#if this.session.isAuthenticated}}
```
ZedLi, line 36: Is this check being done twice since it looks like there's one here too?
https://github.com/hashicorp/boundary-ui/blob/9de44d5f3a6bfb7383f602b1e2f7d77c015a1986/ui/admin/app/templates/application.hbs#L9

ui/admin/app/templates/scopes/scope.hbs
```diff
@@ -33,218 +33,314 @@
...
+          <:logo>
+            <Hds::SideNav::Header::HomeLink
+              @icon='boundary'
+              @color='var(--token-color-boundary-brand)'
+              @ariaLabel={{(app-name)}}
+              @route='index'
+            />
+          </:logo>
+          <:actions>
+            {{#if this.session.isAuthenticated}}
```
ZedLi, line 50: Is this check also needed? It looks you already guard against this in the container itself?

ui/admin/app/templates/scopes/scope.hbs
```diff
@@ -33,218 +33,314 @@
...
+                    {{t (concat 'themes.' theme.label)}}
+                  </dd.Radio>
+                {{/each}}
+              </Hds::Dropdown>
+            {{/if}}
+          </:actions>
+        </Hds::SideNav::Header>
+      </:header>
+      <:body>
+        <Hds::SideNav::Portal::Target />
```
ZedLi, line 128: Why did we portal this in when it's also in the same template?

Could we not have just used `Hds::SideNav::List`?

ZedLi (commented):
🤯 This looks incredible!

ui/admin/app/templates/scopes/scope.hbs
lisbet-alvarez: thats a good point, i think the same thing can be accomplished using the side nav list directly.

ui/admin/app/templates/scopes/scope.hbs
lisbet-alvarez: ah good catch

ui/admin/app/templates/scopes/scope.hbs
lisbet-alvarez: yup good point, ill take a look at using the list component directly.

ui/admin/app/templates/scopes/scope.hbs
lisbet-alvarez: great catch!

addons/rose/addon/styles/hds/overrides.scss
```diff
@@ -53,3 +53,20 @@ form {
     color: inherit;
   }
 }
+
+// TODO: This is a temporary fix that ensures changing the color theme won't alter the dropdown colors.
+.hds-app-frame__sidebar .hds-side-nav {
```
lisbet-alvarez, line 58: Hds team is helping us find a cleaner solution for this style override. So this fix is just temporary.

ui/admin/app/templates/scopes/scope.hbs
lisbet-alvarez: Edit: Just realized you are referring to the `in-element` helper usage. That is because we don't show the side nav on all routes but we do want the side nav to be inserted into the correct hds::frame  component.

ui/admin/app/templates/scopes/scope/targets/target/manage-alias.hbs
```diff
@@ -37,7 +37,7 @@
     <page.actions>
       {{#if (or perms.canUpdate perms.canDelete)}}
 
-        <Hds::Dropdown as |dd|>
+        <Hds::Dropdown data-test-manage-alias as |dd|>
```
cameronperera, line 40: Do we want to specify that this is the manage alias dropdown in the targets route?

ui/admin/tests/acceptance/aliases/delete-test.js
```diff
@@ -17,7 +17,8 @@ module('Acceptance | aliases | delete', function (hooks) {
 
   let aliasCount;
 
-  const DROPDOWN_BUTTON_SELECTOR = '.hds-dropdown-toggle-icon';
+  const DROPDOWN_BUTTON_SELECTOR =
+    'tbody tr td:last-child .hds-dropdown-toggle-icon';
```
cameronperera, line 21: asking for learning: Did this need to be updated since we have more dropdown (specifically in the new side nav) elements on the screen? Either way, I think this is a good change.

addons/rose/addon/styles/hds/overrides.scss
```diff
@@ -53,3 +53,20 @@ form {
     color: inherit;
   }
 }
+
+// TODO: This is a temporary fix that ensures changing the color theme won't alter the dropdown colors.
+.app-frame-sidebar-style-overrides .side-nav-style-overrides {
```
KristinLBradley, line 58: Do the style overrides fail if you use only one parent selector? For example what if you delete the `.app-frame-sidebar-style-overrides` class and use only `.side-nav-style-overrides` as the parent selector?

lisbet-alvarez: correct it does, due to specificity

ui/admin/tests/acceptance/aliases/delete-test.js
lisbet-alvarez: yup that is correct!

ui/admin/app/templates/scopes/scope/targets/target/manage-alias.hbs
lisbet-alvarez: ah great point!

addons/rose/addon/styles/hds/overrides.scss
KristinLBradley: Sorry just tested this in Vercel and it's looking like the 2 selectors are needed so far...

ui/admin/app/templates/application.hbs
```diff
@@ -5,113 +5,23 @@
 
 {{page-title (app-name)}}
 
-<Rose::Layout::Global as |layout|>
+<Hds::AppFrame @hasHeader={{false}} as |Frame|>
```
didoo, line 8: Not sure how it works, but if you bring the same changes to the "desktop" app and remove/replace the `<Rose::Layout::Global>` instance from [the playground.hbs file](https://github.com/hashicorp/boundary-ui/blob/main/addons/rose/tests/dummy/app/templates/playground.hbs), at that point you can remove entirely the code (hbs, css, js, etc) of the `Rose::Layout::Global` component 

_(not sure if that would be outside of the scope of this PR)_

didoo: Similarly for the `Rose::Layout::Sidebar` which is only left in the playground and in [this page](https://github.com/hashicorp/boundary-ui/blob/main/ui/desktop/app/templates/scopes/scope/projects.hbs)

ui/admin/app/templates/scopes/index.hbs
```diff
@@ -6,25 +6,10 @@
...
-      </main>
-    </sidebarLayout.body>
-    <sidebarLayout.footer>
-      <Hds::AppFooter @theme={{this.session.data.theme}} as |AF|>
-        <AF.Link @href={{(doc-url)}}>{{t 'titles.documentation'}}</AF.Link>
-        <AF.LegalLinks />
-      </Hds::AppFooter>
-    </sidebarLayout.footer>
-  </Rose::Layout::Sidebar>
+  <Rose::Layout::Centered>
```
didoo, line 9: Nice cleanup!

ui/admin/app/templates/scopes/scope.hbs
didoo: My 2 cents tip: add a comment in the `ui/admin/app/templates/application.hbs` file, above the `<Frame.Sidebar` declaration, to explain what is going on. I wasn't sure how you were adding the content to the sidebar, until I read this comment above.

didoo: I don't know if/how it would apply to your use case, but have a look at how Cloud UI has solved the problem of having some pages without a sidebar: https://github.com/hashicorp/cloud-ui/blob/main/addons/core/package/src/components/hcp-app-frame.gts#L148

(essentially, if you pass `@hasSidebar={{a false value}}` the sidebar is not rendered at all)

This may avoid doing double portalling.

ui/admin/app/templates/application.hbs
didoo: (the reason I know it, is because in those days I am doing an audit of all the page/layout components used in our products codebases, and I noticed in Boundary these Two rose components (`Rose::Layout::Global` and `Rose::Layout::Sidebar`) and under each one of them, in my audit document, I've written: _"Likely the combination of Global+Sidebar layout components can be replaced with a single Hds::AppFrame component"_ 🙂

didoo (commented):
@lisbet-alvarez I've left a few comments, I hope you don't mind

(as mentioned in one of the comments, I am currently working on auditing page/layout components across our products' codebases, hence some context about this PR)

ui/admin/app/templates/scopes/scope.hbs
lisbet-alvarez: interesting, thank you @didoo for the link + info! ill try this out since i think that will work for this use case as well!

ui/admin/app/templates/application.hbs
```diff
@@ -5,113 +5,108 @@
...
-                class='dropdown-radio-btn'
-                as |G|
-              >
-                {{#each this.themes as |theme|}}
-                  <G.RadioField
-                    @value={{theme.value}}
-                    checked={{eq this.session.data.theme theme.value}}
-                    {{on 'change' (fn this.toggleTheme theme.value)}}
-                    as |F|
+<Hds::AppFrame @hasHeader={{false}} @hasSidebar={{this.showSideNav}} as |Frame|>
```
didoo, line 8: 🎉

ui/admin/app/templates/account/change-password.hbs
ZedLi, line 1: Unrelated to this file but I wanted to point out it looks like the onboarding page isn't centered correctly now

image

Was there also a decision made about how the banner would be handled on pages that don't have the nav bar anymore either (like this page)? Are we just removing it now?

ZedLi (approved):
Looks good, just a few minor cosmetic comments I had!

ui/admin/app/templates/application.hbs
```diff
@@ -5,113 +5,108 @@
...
+          </:actions>
+        </Hds::SideNav::Header>
+      </:header>
+
+      <:body>
+        <Hds::SideNav::Portal::Target />
+      </:body>
+
+      <:footer>
+        <ScopePicker />
```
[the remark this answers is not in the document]
lisbet-alvarez: oh yup great suggestion 🔎  !

ui/admin/app/templates/account/change-password.hbs
lisbet-alvarez: good catch lemme look into that!
i need to touch base w/ design on that but either way planning to address that in another PR (too not add more changes to this one) along with clean up tasks such as fixing any broken e2e tests.

lisbet-alvarez:
> This looks amazing! I tested and it looks we don't display `settings` when oss is selected, seems to be an ent feature. But do we want still want to show the title `settings` for oss? Screenshot 2025-02-28 at 4 33 53 PM

good catch! fixed!

ui/admin/app/templates/scopes/index.hbs
```diff
@@ -6,25 +6,10 @@
 {{outlet}}
```
hashicc, line 6: I wasn't able to tell if this template is being used because [the route redirects](https://github.com/hashicorp/boundary-ui/blob/371301b4bad27a5df529de6ec5e84380aabf45d7/ui/admin/app/routes/scopes/index.js#L24-L34) to a global scope when none is set. If it's not being used it might be useful to check if we have test coverage for this behavior and then delete the template and controller.

hashicc (approved):
🎉 Fantastic work @lisbet-alvarez! This is such a nice upgrade to the menu experience. I had one question about whether the `ui/admin/app/templates/scopes/index.hbs` template and its corresponding controller are being used, otherwise everything looks good to me!

Things I learned from this PR:
* ✨ `@color='var(--token-color-boundary-brand)'`, css properties can be passed into color values within our components. That's pretty cool and not something I've thought about before.
* 😎 Hds has really thought through their menus and makes use of `<Hds::SideNav::Portal::Target />` (`application.hbs`) as a "portal" for allowing injectable content from the `<Hds::SideNav::Portal />` (`scopes/scope.hbs`) from any other corresponding template

cameronperera (dismissed):
🎉

didoo:
> * 😎 Hds has really thought through their menus and makes use of `<Hds::SideNav::Portal::Target />` (`application.hbs`) as a "portal" for allowing injectable content from the `<Hds::SideNav::Portal />` (`scopes/scope.hbs`) from any other corresponding template

@hashicc credits go to the initial developers of the SideNav in Cloud UI/TFC [we stand on the shoulders of giants] which _I think_ have been @meirish @cbfx and others. We (as HDS devs) ported their code and built on top of it, but most of the work was already done.

DhariniJeeva (approved):
great work!!

ui/admin/app/templates/scopes/index.hbs
lisbet-alvarez: great catch, updated!

ui/admin/app/controllers/application.js
```diff
@@ -39,6 +40,13 @@ export default class ApplicationController extends Controller {
     return THEMES;
   }
 
+  get showSideNav() {
+    return (
+      this.router.currentRouteName.startsWith('scopes.scope') &&
```
laurenolivia, line 45: Nice work. 

Just curious, I noticed your PR description mentions "The side nav should not be visible on the onboarding, change-password , and authenticate routes."

Can you share a bit of context around why the side nav should only appear on routes that start with `scopes.scope`?

ui/admin/app/routes/account.js
```diff
@@ -15,9 +15,14 @@ export default class AccountRoute extends Route {
...
-   * If arriving here unauthenticated, redirect to index for further processing.
+   * If arriving here unauthenticated, redirect to index for further processing
+   * else redirect to change-password route since account route is empty.
    */
   beforeModel() {
-    if (!this.session.isAuthenticated) this.router.transitionTo('index');
+    if (!this.session.isAuthenticated) {
+      this.router.transitionTo('index');
+    } else {
+      this.router.transitionTo('account.change-password');
```
laurenolivia, line 25: I'm curious about the rationale for adding this else statement? If the user has arrived to this route unauthenticated, is the assumption that they must be locked out if the account route is empty and thus they should reset their pw?

ui/admin/app/styles/app.scss
```diff
@@ -173,8 +173,6 @@
 
 // Onboarding styles
 .create-resources {
-  display: flex;
```
laurenolivia, line 176: Wondering why we removed `display: flex;`? Merely curious is all!

ui/admin/app/templates/account/change-password.hbs
```diff
@@ -5,32 +5,20 @@
 
 {{page-title (t 'actions.change-password')}}
 
-<Rose::Layout::Sidebar as |sidebarLayout|>
-  <sidebarLayout.body>
-    <main>
-      <Rose::Layout::Centered>
-        <Rose::Layout::Page as |page|>
+<Rose::Layout::Centered>
```
laurenolivia, line 8: Ah, I haven't seen a Rose component in a while!

ui/admin/app/controllers/application.js
lisbet-alvarez: Great question, this is to keep the functionality the same as before. Additionally, `scopes.scope` is the route where we load global scope and org scopes (& project scopes if in project level) and having those scopes loaded is necessary for displaying sub routes and the scope picker.

ui/admin/app/routes/account.js
lisbet-alvarez: we don't display anything in the 'account' route (template is empty) so if the user were to go there it would be an empty screen so its best to redirect to the change-password route.

ui/admin/app/styles/app.scss
lisbet-alvarez: oh yup, this was to fix display issues this was causing (the frame was off centered.

cameronperera (approved):
🎉

PR #2691, opened 2025-02-13
refactor: 💅  Aliases: read-test

laurenolivia:
## Description
This PR refactors our aliases: read-test suite.

:tickets: [Jira ticket](https://hashicorp.atlassian.net/browse/ICU-15455)

## How to Test
Tests should pass ✅ 

## Checklist

### DIFF
--- a/ui/admin/tests/helpers/selectors.js
+++ b/ui/admin/tests/helpers/selectors.js
@@ -15,7 +15,6 @@ export const FIELD_DESCRIPTION = '[name=description]';
 export const FIELD_DESCRIPTION_VALUE = 'description';
 
 export const TABLE_RESOURCE_LINK = (url) => `tbody [href="${url}"]`;
-export const TABLE_LINK_SELECTOR = '.hds-table__tbody tr:first-child a';
 
 export const RESOURCE_NOT_FOUND_SUBTITLE =
   '[data-test-error-application-state] .hds-application-state__error-code';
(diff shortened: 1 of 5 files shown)


ui/admin/tests/helpers/selectors.js
```diff
@@ -15,6 +15,7 @@ export const FIELD_DESCRIPTION = '[name=description]';
 export const FIELD_DESCRIPTION_VALUE = 'description';
 
 export const TABLE_RESOURCE_LINK = (url) => `tbody [href="${url}"]`;
+export const TABLE_LINK_SELECTOR = '.hds-table__tbody tr:first-child a';
```
laurenolivia, line 18: Moving `TABLE_LINK_SELECTOR` into the commonSelectors file since this const is also being used in [/acceptance/policy/read-test.js](https://github.com/hashicorp/boundary-ui/blob/f9411794c433409b25f3f28fc9eb217d34e1e87c/ui/admin/tests/acceptance/policy/read-test.js#L22), thus this will assist in code cleanup efforts.

DhariniJeeva (approved):

ui/admin/tests/helpers/selectors.js
cameronperera: We should use `TABLE_RESOURCE_LINK` here. We opted to use `TABLE_RESOURCE_LINK` since we can be very specific with this selector while still being flexible. We originally used a selector like `TABLE_LINK_SELECTOR` but we had to duplicate it if we wanted to click on a resource in the second or third row. By passing in the URL and using the href, it gives us a little more flexibility.

cameronperera (changes requested):

calcaide (changes requested):
+1 [to this comment](https://github.com/hashicorp/boundary-ui/pull/2691/files#r1962254294). Getting close! 🙌

calcaide (approved):
🙌 Thank you for the refactor!

lisbet-alvarez (approved):
nice cleanup!!!

DhariniJeeva (approved):

PR #2693, opened 2025-02-13
refactor: 💅 Add selector to test helpers, implement across test suite

laurenolivia:
# Description
This PR adds the `FIELD_TEXT_ERROR` to our test helpers file, and replaces all instances of this non-unique err where tests were either selecting `.hds-form-error__message` or exporting a similarly named constant.

:tickets: [Jira ticket](https://hashicorp.atlassian.net/browse/ICU-16379) 

## How to Test
Tests should pass ✅ 

## Checklist

### DIFF
--- a/ui/admin/tests/acceptance/aliases/create-test.js
+++ b/ui/admin/tests/acceptance/aliases/create-test.js
@@ -26,7 +26,6 @@ module('Acceptance | aliases | create', function (hooks) {
   const NAME_FIELD_SELECTOR = '[name="name"]';
   const DESTINATION_ID_SELECTOR = '[name="destination_id"]';
   const HOST_ID_SELECTOR = '[name="authorize_session_arguments"]';
-  const FIELD_ERROR_TEXT_SELECTOR = '.hds-form-error__message';
   const NAME_FIELD_TEXT = 'random string';
 
   const instances = {
@@ -119,7 +118,7 @@ module('Acceptance | aliases | create', function (hooks) {
     assert
       .dom(commonSelectors.ALERT_TOAST_BODY)
       .hasText('The request was invalid.');
-    assert.dom(FIELD_ERROR_TEXT_SELECTOR).hasText('Name is required.');
+    assert.dom(commonSelectors.FIELD_TEXT_ERROR).hasText('Name is required.');
   });
 
   test('can navigate to new aliases route with proper authorization', async function (assert) {
--- a/ui/admin/tests/acceptance/auth-methods/create-test.js
+++ b/ui/admin/tests/acceptance/auth-methods/create-test.js
@@ -544,7 +544,9 @@ module('Acceptance | auth-methods | create', function (hooks) {
     assert
       .dom(commonSelectors.ALERT_TOAST_BODY)
       .hasText('The request was invalid.');
-    assert.dom(selectors.FIELD_ERROR).hasText('At least one URL is required.');
+    assert
+      .dom(commonSelectors.FIELD_TEXT_ERROR)
+      .hasText('At least one URL is required.');
   });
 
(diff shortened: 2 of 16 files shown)


lisbet-alvarez (commented):
looks good!
im wondering, do we also want to updated other places that use a similar selector ([example](https://github.com/hashicorp/boundary-ui/blob/main/ui/admin/tests/acceptance/auth-methods/update-test.js#L451)).

laurenolivia:
> looks good! im wondering, do we also want to update other places that use a similar selector ([example](https://github.com/hashicorp/boundary-ui/blob/main/ui/admin/tests/acceptance/auth-methods/update-test.js#L451)).

Good eye, guru 👀 

Note: @lisbet-alvarez and I clarified her question offline.

Next: It may be worth discussing that my change means we now have these two selectors:
```
export const FIELD_NAME_ERROR = '[data-test-error-message-name]';
export const FIELD_TEXT_ERROR = '.hds-form-error__message';
```

And, while, these two are not the same, they are quite similar and may need to be resolved.

DhariniJeeva (approved):
looks good

cameronperera:
I [commented](https://github.com/hashicorp/boundary-ui/pull/2685#discussion_r1962246725) something similar in another PR but we likely don't need this PR. `export const FIELD_NAME_ERROR = '[data-test-error-message-name]';` is the selector that we should be using to select the error on the name field. If we use `export const FIELD_TEXT_ERROR = '.hds-form-error__message';` there is a chance we get a false negative if there are multiple errors on the same form. We also made a decision long ago (I understand this is context newer team members may not have) to avoid using HDS classes if possible in our test selectors. We now try to rely on the HTML elements, custom css classes (only if present for styling purposes), or data-test attributes. This is a case where data-test attributes come in handy.

calcaide:
+1 to @cameronperera [comment](https://github.com/hashicorp/boundary-ui/pull/2693#issuecomment-2669582232). There is just 1 usage of this type selector that needs to refactored within `auth-methods` I introduced that one at domain level and left a comment of the side effects, [here code](https://github.com/hashicorp/boundary-ui/blob/main/ui/admin/tests/acceptance/auth-methods/selectors.js#L97). That said, such selector is wrong and must be refactored on my end.

calcaide (changes requested):
I am not sure we need this PR.

- As [per this comment](https://github.com/hashicorp/boundary-ui/pull/2693#issuecomment-2669582232).
- This PR is aiming for several domains when we are prescribing one at a time.
- It changes integration tests, where right now, the team is working on acceptance test. Integration or other type of tests will come later on.

laurenolivia:
> +1 to @cameronperera [comment](https://github.com/hashicorp/boundary-ui/pull/2693#issuecomment-2669582232). There is just 1 usage of this type selector that needs to refactored within `auth-methods` I introduced that one at domain level and left a comment of the side effects, [here code](https://github.com/hashicorp/boundary-ui/blob/main/ui/admin/tests/acceptance/auth-methods/selectors.js#L97). That said, such selector is wrong and must be refactored on my end.

I created a ticket [here](https://hashicorp.atlassian.net/browse/ICU-16476) to fix the auth-methods selecting on an hds class. I will work on that fix next.

laurenolivia:
> I am not sure we need this PR.
> 
> * As [per this comment](https://github.com/hashicorp/boundary-ui/pull/2693#issuecomment-2669582232).
> * This PR is aiming for several domains when we are prescribing one at a time.
> * It changes integration tests, where right now, the team is working on acceptance test. Integration or other type of tests will come later on.

Yes, this PR is going to be closed, and the changes will be domain-specific moving fwd. 👍

laurenolivia:
Closing to focus on domain-specific changes, rather than sweeping changes. See additional comments above. 👍

PR #2694, opened 2025-02-18
feat: 🎸 add logging to notifyError decorator

cameronperera:
✅ Closes: https://hashicorp.atlassian.net/browse/ICU-16334

# Description

I took this approach to make adding error logs for desktop client easier. The plan is if we need add logs, we can simply add `log: true` to the options object of the `notifyError` decorator. 

The scope of this PR is just to add this logging. I am reaching out to product about which errors we show in the desktop compared to the logs. They likely should be the same now and if product is good with that I will update this specific instance to show the error to the user and record it to the log file.

## Checklist

### DIFF
--- a/addons/core/addon/decorators/notify.js
+++ b/addons/core/addon/decorators/notify.js
@@ -61,10 +61,12 @@ export function notifySuccess(notification) {
  *                              electron logging service
  * @param {string} options.log.origin - defaults to undefined, the origin that
  *                                     triggered the error
+ * @param {string} options.log.level - defaults to undefined, the log level that
+ *                                     our electron logging service should use
  */
 export function notifyError(
   notification,
-  options = { catch: false, sticky: true, log: undefined },
+  options = { catch: false, sticky: true },
 ) {
   return function (_target, _propertyKey, desc) {
     const method = desc.value;
@@ -77,11 +79,8 @@ export function notifyError(
         return await method.apply(this, arguments);
       } catch (error) {
         if (options.log) {
-          const { origin } = options.log;
-          __electronLog?.error(
-            `Error triggered in '${origin}':`,
-            error.message,
-          );
+          const { origin, level = 'error' } = options.log;
+          __electronLog?.[level](`${origin}:`, error.message);
         }
         const candidateKey =
           typeof notification === 'function'
--- a/addons/core/addon/decorators/notify.js
+++ b/addons/core/addon/decorators/notify.js
@@ -54,10 +54,12 @@ export function notifySuccess(notification) {
  *                                  the notification until user dismissal
  * @param {object} options.catch - defaults to false, whether or not to catch
  *                                 and squelch the error
+ * @param {object} options.log - defaults to false, whether or not to log
+ *                                 the error to our electron logging service
  */
 export function notifyError(
   notification,
-  options = { catch: false, sticky: true },
+  options = { catch: false, sticky: true, log: false },
 ) {
   return function (_target, _propertyKey, desc) {
     const method = desc.value;
@@ -69,6 +71,9 @@ export function notifyError(
       try {
         return await method.apply(this, arguments);
       } catch (error) {
+        if (options.log) {
+          __electronLog?.error('notifyError', error.message);
+        }
         const candidateKey =
           typeof notification === 'function'
             ? notification.apply(this, [error])
--- a/addons/core/addon/decorators/notify.js
+++ b/addons/core/addon/decorators/notify.js
@@ -5,6 +5,8 @@
 
 import { getOwner } from '@ember/application';
 
+const { __electronLog } = globalThis;
+
 /**
  * Decorates a method and, if the method does not error, shows a success
  * notification via the notify service.  The text of the notification is derived
--- a/addons/core/addon/decorators/notify.js
+++ b/addons/core/addon/decorators/notify.js
@@ -52,16 +52,19 @@ export function notifySuccess(notification) {
  *
  * @param {string|function} notification
  * @param {object} options
- * @param {object} options.sticky - defaults to true, whether or not to persist
+ * @param {boolean} options.sticky - defaults to true, whether or not to persist
  *                                  the notification until user dismissal
- * @param {object} options.catch - defaults to false, whether or not to catch
+ * @param {boolean} options.catch - defaults to false, whether or not to catch
  *                                 and squelch the error
- * @param {object} options.log - defaults to false, whether or not to log
- *                                 the error to our electron logging service
+ * @param {object} options.log - defaults to undefined, object with unique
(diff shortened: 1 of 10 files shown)


addons/core/addon/decorators/notify.js
```diff
@@ -69,6 +73,9 @@ export function notifyError(
       try {
         return await method.apply(this, arguments);
       } catch (error) {
+        if (options.log) {
+          __electronLog?.error('notifyError', error.message);
```
ZedLi, line 77: I'm not sure that `notifyError` is a very useful thing to add to our log statements as it doesn't help us know where it originates from. We should include an option for consumers to enrich this log statement by customizing what the message says so that we can know the origin.

Otherwise in the future it might be hard to figure out where a "notifyError" log statement is coming from. We _could_ do something like attach a stack trace but I'm not the biggest fan of that as it's a lot of noise and can sometimes have security implications (but maybe it's worth it?).

I also know we are sometimes losing some error information either by re-throwing or when we're serializing the error from electron through IPC which makes it less feasible currently (though I might tackle in a different PR now that I'm writing this out lol)

addons/core/addon/decorators/notify.js
```diff
@@ -54,10 +56,12 @@ export function notifySuccess(notification) {
  *                                  the notification until user dismissal
  * @param {object} options.catch - defaults to false, whether or not to catch
  *                                 and squelch the error
+ * @param {object} options.log - defaults to false, whether or not to log
+ *                                 the error to our electron logging service
  */
 export function notifyError(
   notification,
-  options = { catch: false, sticky: true },
+  options = { catch: false, sticky: true, log: false },
```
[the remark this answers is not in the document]
cameronperera: Right we discussed having it on by default. Let me look at the different `notifyError` uses in the desktop client. For admin UI, it just would do nothing so that should be an issue right?

addons/core/addon/decorators/notify.js
cameronperera: Chad and I were discussing the mock issue for electronLog. We talked about adding notifyError tests and creating the mock then. We could add that change then as well. I do agree with you that `notifyError` isn't super helpful.

addons/core/addon/decorators/notify.js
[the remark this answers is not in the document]
ZedLi: Yeah it should no-op in admin UI. I'm still waffling on whether we want to have it on by default, I'm guessing most are CRUDL actions which aren't _that_ useful 😐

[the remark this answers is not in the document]
cameronperera: We could leave it as is for now then do some investigation into this?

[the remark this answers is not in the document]
ZedLi: Yup I think that's fine

addons/core/addon/decorators/notify.js
```diff
@@ -50,14 +52,19 @@ export function notifySuccess(notification) {
...
+ * @param {object} options.log - defaults to undefined, object with unique
+ *                              information about the error to send to our
+ *                              electron logging service
+ * @param {string} options.log.origin - defaults to undefined, the origin that
+ *                                     triggered the error
  */
 export function notifyError(
   notification,
-  options = { catch: false, sticky: true },
+  options = { catch: false, sticky: true, log: undefined },
```
ZedLi, line 67: Is this `undefined` default necessary?

addons/core/addon/decorators/notify.js
```diff
@@ -69,6 +76,13 @@ export function notifyError(
       try {
         return await method.apply(this, arguments);
       } catch (error) {
+        if (options.log) {
+          const { origin } = options.log;
+          __electronLog?.error(
+            `Error triggered in '${origin}':`,
+            error.message,
```
ZedLi, line 83: Any reason we have to bake in this `origin` in the first place? Couldn't we just have the consumer say what to log since you have access to the error and have it take in the entire string?

It also might be be more succinct to just do `resetClusterUrl: ${message}`

addons/core/addon/decorators/notify.js
cameronperera: I don't feel like it is but I added it to keep with the pattern. Would you prefer I remove it, or all of them, especially since the defaults are explained in the signature?
Those aren't even true defaults. This is a comment lower in the decorator:

> // Default is `true`.  Since `options` is a new object instance
        // when passed, the values shown in the signature aren't
        // true defaults.  For example, if a value `{ catch: true }` were
        // passed, `sticky` would actually be falsy, even though the intended
        // default is `true`.  Hence this extra step.
        const sticky = options.sticky === undefined ? true : options.sticky;

ZedLi: I think the other two are fine since they're booleans and it's fine to default them to actual booleans, but `log` would still be `undefined` either way. I think they're not "true defaults" only in the sense of how JS works. Otherwise I think it's accurate

ZedLi (approved):
Looks great, thanks for updating this!

ui/desktop/app/routes/application.js
```diff
@@ -29,7 +29,10 @@ export default class ApplicationRoute extends Route {
    * reported by the main process.  If they differ, update the main process
    * clusterUrl so that the renderer's CSP can be rewritten to allow requests.
    */
-  @notifyError(({ message }) => message, { catch: true })
+  @notifyError(({ message }) => message, {
+    catch: true,
+    log: { origin: 'ApplicationRoute:beforeModel' },
```
DhariniJeeva, line 34: is this the route format we should follow when we log from a route's hook? I'm ok with this approach, I was just curious bc the above controller log has the action name

DhariniJeeva (approved):
looks good! I left a non blocking comment

ui/desktop/app/routes/application.js
cameronperera: I think this is a solid approach. Since it is a `beforeModel` hook, it isn't clear where it is coming from so I figured adding the class route name is a good approach.

addons/core/addon/decorators/notify.js
```diff
@@ -69,6 +78,10 @@ export function notifyError(
       try {
         return await method.apply(this, arguments);
       } catch (error) {
+        if (options.log) {
+          const { origin, level = 'error' } = options.log;
+          __electronLog?.[level](`${origin}:`, error.message);
```
hashicc, line 83: If I'm parsing this correctly when the `origin` is not set the log ends up being something like `: Some error message here`? Not a big deal. I think once we get a test for the decorator some of these cases will shake out

hashicc (approved):
Thanks @cameronperera! This is a great step to capturing more feedback

PR #2696, opened 2025-02-19
test(e2e): Add check for credential store worker filters

moduli:
# Description
This PR adds a check in the Admin UI e2e test suite for credential store worker filters. Note: This requires a change to be added to `boundary-enterprise`: https://github.com/hashicorp/boundary/pull/5553 in order to specify the private address for vault. 

https://hashicorp.atlassian.net/browse/ICU-16391

Screenshot 2025-02-19 at 10 53 21 AM

## How to Test
This test is enterprise only and works on both aws and docker based test infrastructure.

## Checklist


e2e-tests/global-setup.js
```diff
@@ -83,6 +83,7 @@ export const test = baseTest.extend({
   sshUser: process.env.E2E_SSH_USER,
   targetAddress: process.env.E2E_TARGET_ADDRESS,
   targetPort: process.env.E2E_TARGET_PORT,
-  vaultAddr: process.env.E2E_VAULT_ADDR,
+  vaultAddr: process.env.E2E_VAULT_ADDR_PUBLIC,
+  vaultAddrPrivate: process.env.E2E_VAULT_ADDR_PRIVATE,
```
ZedLi, line 87: When should we use which?

ZedLi (approved):

e2e-tests/global-setup.js
ZedLi: Could we also update the usage of the vault address in the desktop test as well in this PR then?

moduli: I don't believe any changes are needed in the desktop tests. They create credential stores without a worker filter, so the public address can be used.

PR #2697, opened 2025-02-21
chore: upgrade to yarn 4

DhariniJeeva:
# Description
[Jira](https://hashicorp.atlassian.net/browse/ICU-10601)

This PR upgrades yarn from 1 to 4. 

**NOTE**: This migration is done by committing the binary file instead of using corepack as committing the binary and accessing it via yarnPath seems like the safest/simplest option given the state of corepack support from node.(https://github.com/nodejs/node/pull/51981) 

## How to Test

- check out this repo locally
- run `npx npkill -D -x -y` [to remove all node_modules]
- run `yarn install` 

## Checklist


package.json
```diff
@@ -8,39 +8,37 @@
...
+    "test:addons-core": "yarn workspace core test",
+    "test:addons-rose": "yarn workspace rose test",
+    "test:ui-admin": "yarn workspace admin test",
+    "test:ui-desktop": "yarn workspace desktop test",
+    "compliance:licenses": "license-checker --onlyAllow 'Apache*;Apache License, Version 2.0;Apache-2.0;Apache 2.0;Artistic-2.0;BSD;BSD-3-Clause;CC-BY-3.0;CC-BY-4.0;CC0-1.0;ISC;MIT;MPL-2.0;BUSL-1.1;Public Domain;Python-2.0;Unicode-TOU;Unlicense;WTFPL;BlueOak-1.0.0' --excludePackages 'boundary-ui;admin@0.0.0;desktop@0.0.0;e2e-tests@0.0.0;cycle@1.0.3;cldr-core@43.1.0;make-plural@7.2.0;@hashicorp/ember-asciinema-player@0.0.0;path-scurry@1.10.1'",
     "doc:toc": "doctoc README.md",
     "prepare": "husky"
   },
   "devDependencies": {
+    "concurrently": "^9.1.0",
```
DhariniJeeva, line 41: this needed to be a devdependecy as I was getting error otherwise

package.json
```diff
@@ -49,20 +47,22 @@
     "lint-staged": "^15.2.10"
   },
   "resolutions": {
-    "**/nomnom/underscore": "^1.12.1",
-    "**/core/ember-inline-svg/**/nth-check": "^2.0.1",
-    "**/ember-cli/**/ansi-html": "^0.0.9",
-    "**/ember-cli/testem/fireworm/async": "^2.6.4",
-    "**/@hashicorp/design-system-components/ember-stargate": "^0.6.0",
-    "micromatch": "^4.0.8"
+    "nomnom/underscore": "^1.12.1",
```
DhariniJeeva, line 50: yarn 4 automatically removed all the glob resolutions but I still had to manually add these to avoid errors

ui/desktop/package.json
```diff
@@ -25,14 +25,13 @@
...
     "lint:sass:fix": "stylelint '**/*.scss' --fix",
-    "lint:electron": "electronegativity -r -i ./electron-app",
     "format": "concurrently \"npm:format:*\" --names \"format:\"",
     "format:hbs": "prettier --write 'app/**/*.hbs'",
     "format:js": "prettier --write '{app,config,stories,tests,electron-app/src}/**/*.js' *.js",
     "format:sass": "prettier --write app/**/*.scss",
     "start": "ember serve --port 4201",
     "prestart:desktop": "cd electron-app && yarn premake",
-    "start:desktop": "ember electron",
+    "start:desktop": "yarn prestart:desktop && ember electron",
```
DhariniJeeva, line 34: https://yarnpkg.com/migration/guide#explicitly-call-the-pre-and-post-scripts

package.json
```diff
@@ -8,39 +8,37 @@
...
   "scripts": {
     "build": "yarn build:ui:admin",
-    "build:ui:admin": "yarn --cwd ui/admin build",
-    "build:ui:admin:oss": "yarn --cwd ui/admin build:oss",
-    "build:ui:admin:enterprise": "yarn --cwd ui/admin build:enterprise",
-    "build:ui:admin:hcp": "yarn --cwd ui/admin build:hcp",
-    "build:ui:desktop": "yarn --cwd ui/desktop build",
-    "build:ui:desktop:production": "yarn --cwd ui/desktop build:production",
-    "build:ui:desktop:app": "yarn --cwd ui/desktop build:desktop",
+    "build:ui:admin": "yarn workspace admin build",
```
DhariniJeeva, line 15: switched these to workspace since we have workspaces support and it seemed like a better fit and supposedly faster than `--cwd`

ui/desktop/electron-app/config/forge.config.js
```diff
@@ -137,21 +137,5 @@ module.exports = {
         });
       });
     },
-    packageAfterPrune: async (_, buildPath) => {
```
DhariniJeeva, line 140: since we are pinning node-gyp v10, we don't need this workaround anymore, I tested it, but lmk if there's anything I missed

.yarn/releases/yarn-4.6.0.cjs
calcaide, line 1: Just wondering, is this yarn code being included? or what is this file for?

ui/desktop/package.json
```diff
@@ -15,8 +15,8 @@
     "build:production": "cross-env EMBER_CLI_ELECTRON=true ember build --environment=production",
     "build:development": "ember build",
     "prebuild:desktop": "shx rm -rf electron-app/ember-dist electron-app/out && shx cp -r dist/ electron-app/ember-dist/",
-    "build:desktop": "yarn build:desktop:install && yarn --cwd electron-app make",
-    "build:desktop:install": "yarn install --cwd electron-app",
+    "build:desktop": "yarn prestart:desktop && yarn build:desktop:install && cd electron-app && yarn make",
```
hashicc, line 18: This previously called `yarn:desktop:install` but now it calls `yarn prestart:desktop`, is the `build:desktop:install` not needed? I view these as grouped tasks with subtasks depth by `:`. If the `yarn prestart:desktop` is also needed what do you think about making a `build:desktop:premake` (there's probably a better name I'm missing) to keep these grouped?

Also I noticed that `yarn --cwd` is still used on line 19 but for line 18 it's been refactored to using `cd ... && yarn`, can this be changed back to using `--cwd`? I find the `--cwd` easier to track than the `cd`s

ui/desktop/electron-app/patches/lodash.template+4.5.0.patch
```diff
@@ -1,45 +0,0 @@
-diff --git a/node_modules/lodash.template/index.js b/node_modules/lodash.template/index.js
```
hashicc, line 1: ✂️ 🥳  Thanks for getting this cleaned up. So nice to not have to rely on patches.

hashicc (changes requested):
Nice work! I have a few comments/suggestions to consider. I'm looking forward to the nicer things we get with yarn 4. ⚡⚡⚡ Just as a rough benchmark it looks like it's twice as fast from a cold install

hashicc:
There's [two yarn links in the root `README.md`](https://github.com/hashicorp/boundary-ui/blob/fac7d2b128fc9ff09c29baa4d7b7528e5dc68364/README.md#L221-L224) that could be updated to the modern docs. A mention in the readme that we're using yarn 4 might be useful since my assumption is many javascript projects didn't make the jump so people might still be accustomed to yarn 1.

package.json
ZedLi: Unrelated but I gotta say I was a bit confused when I read it's only one level of depth and then see `@babel/core/@babel/generator` until I remembered packages can also have slashes 😮‍💨

ui/desktop/electron-app/package.json
ZedLi, line 1: The `make` script also needs to call `premake` manually now too

ui/desktop/package.json
```diff
@@ -41,17 +40,18 @@
     "posttest": "ember coverage-merge && yarn clean:coverage"
```
ZedLi, line 40: This also needs to be called by `test` manually now

ui/desktop/package.json
```diff
@@ -25,14 +25,13 @@
...
     "format:hbs": "prettier --write 'app/**/*.hbs'",
     "format:js": "prettier --write '{app,config,stories,tests,electron-app/src}/**/*.js' *.js",
     "format:sass": "prettier --write app/**/*.scss",
     "start": "ember serve --port 4201",
     "prestart:desktop": "cd electron-app && yarn premake",
-    "start:desktop": "ember electron",
+    "start:desktop": "yarn prestart:desktop && ember electron",
     "clean:coverage": "rm -rf coverage_*",
     "pretest": "yarn clean:coverage",
     "test": "concurrently \"npm:test:*\" --names \"test:\"",
```
ZedLi, line 37: This should also call `pretest` and `posttest` manually now too

.yarnrc.yml
```diff
@@ -0,0 +1,6 @@
+# Copyright (c) HashiCorp, Inc.
+# SPDX-License-Identifier: BUSL-1.1
+
+nodeLinker: node-modules
+
+yarnPath: .yarn/releases/yarn-4.6.0.cjs
```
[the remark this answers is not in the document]
laurenolivia: Hmm googling this feedback

.yarn/releases/yarn-4.6.0.cjs
calcaide: All good, I was asking for learning purposes 😃

package.json
calcaide: I agree with @hashicc comment. Taking `ansi-html` for example, the current resolution will force `ansi-html` specific version within the entire codebase, when the issue its only within `ember-cli`.

ui/admin/package.json
```diff
@@ -89,6 +90,7 @@
     "eslint-plugin-qunit": "^8.0.1",
     "js-bexpr": "hashicorp/js-bexpr#9b4a4b54d85eba67fdfc0990133d1518d890b1e1",
     "loader.js": "^4.7.0",
+    "postcss": "^8.4.31",
```
calcaide, line 93: None blocking comment: Why do we need to add postcss now as dependency?

package.json
DhariniJeeva: thanks for the feedback and the examples, makes sense to keep it relative.

ui/admin/package.json
DhariniJeeva: I get warning during installation if I don't provide this as a devDependency
`admin@workspace:ui/admin doesn't provide postcss (p389e8), requested by stylelint-config-standard-scss`

.yarnrc.yml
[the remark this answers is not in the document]
DhariniJeeva: I was thinking of doing manual updates, the local installation says whenever there is a new yarn update and that can be a good indicator to update yarn.  Are you thinking of other alternatives that we can use in place of corepack? I did not investigate other options but might be worth considering in the future

ui/desktop/package.json
DhariniJeeva: thanks for bringing this up, I renamed some of the scripts, let me know if those make sense

package.json
```diff
@@ -49,20 +47,22 @@
...
     "commitizen": {
-      "path": "./node_modules/cz-conventional-changelog"
+      "path": "git-cz"
     }
   },
   "engines": {
     "node": "20.* || 22.*"
   },
-  "packageManager": "yarn@1.22.19+sha1.4ba7fc5c6e704fce2066ecbfb0b0d8976fe62447"
+  "packageManager": "yarn@4.6.0"
```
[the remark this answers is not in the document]
DhariniJeeva: ah yes, I removed it!

package.json
```diff
@@ -8,39 +8,37 @@
       "addons/*",
       "ui/*",
       "e2e-tests"
-    ],
-    "nohoist": [
-      "**/electron**"
```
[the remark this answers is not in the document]
DhariniJeeva: it does not look like we need to do the equivalent of `nohoist` in yarn 4, I tested it with and without in both yarn 1 and yarn4 and did not see any issue so my guess is that it is not needed.

package.json
```diff
@@ -49,20 +47,20 @@
...
-    "**/ember-cli/**/ansi-html": "^0.0.9",
-    "**/ember-cli/testem/fireworm/async": "^2.6.4",
-    "**/@hashicorp/design-system-components/ember-stargate": "^0.6.0",
-    "micromatch": "^4.0.8"
+    "nomnom/underscore": "^1.12.1",
+    "@hashicorp/design-system-components/ember-stargate": "^0.6.0",
+    "micromatch": "^4.0.8",
+    "ember-inline-svg/nth-check": "^2.0.1",
+    "node-gyp": "^10.0.0",
+    "ember-cli/ansi-html": "^0.0.9",
```
lisbet-alvarez, line 55: Looks like `ansi-html` is resolving to `0.0.7` now. I think its because `ember-cli` doesn't have `ansi-html` as a dependency. Looks like its `broccoli-middleware` and `broccoli` (those are deps. of `ember-cli`) that use `ansi-html`.

DhariniJeeva: ooh good catch! I think we might not need this resolution anymore, since this didn't resolve correctly and didn't cause any issues

package.json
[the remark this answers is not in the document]
ZedLi: Makes sense, this looks like the commit https://github.com/hashicorp/boundary-ui/commit/64bb3536d32e109a4da9585235d0f5efe4a6264e for why we did it originally so I don't think it's really relevant anymore

ui/desktop/electron-app/package.json
```diff
@@ -43,11 +42,11 @@
     "@electron-forge/maker-squirrel": "^7.2.0",
     "@electron-forge/maker-zip": "^7.2.0",
     "electron": "31.0.1",
-    "unzipper": "^0.11.5",
     "patch-package": "^8.0.0",
```
ZedLi, line 45: Can we remove this package since it shouldn't be needed anymore?

ui/desktop/electron-app/package.json
```diff
@@ -12,11 +12,10 @@
     "build:config": "node -e \"require('./config/desktop.js').setup()\"",
     "start": "electron-forge start",
     "package": "electron-forge package",
-    "premake": "yarn build:cli && yarn build:config",
-    "make": "electron-forge make",
+    "configure": "yarn build:cli && yarn build:config",
```
ZedLi, line 15: I think this is fine but this could also just go under the other 2 `build` commands since it just calls both, it could just be `build:all`

package.json
lisbet-alvarez: I think this [dependabot alert](https://github.com/hashicorp/boundary-ui/security/dependabot/26) will reopen if there's no resolution since versions less than 0.0.8 are vulnerable. We can add two resolutions or just one global resolution
```
"broccoli/ansi-html": "^0.0.9",
"broccoli-middleware/ansi-html": "^0.0.9",
```
or i think this would work too:
```
"ansi-html": "^0.0.9",
```

DhariniJeeva: oh thank you for linking that!

.yarnrc.yml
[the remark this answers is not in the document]
hashicc: > I'm just wondering how do we would go about updating (and verifying) the vendored file in the future?

What do you think about documenting the process in our readme or putting a comment in this yaml file to instructions on how to update yarn? I'm thinking about the case where we have to maintain this but it isn't exactly clear where we would pull a new `yarn-x.x.x.cjs` from

ui/desktop/electron-app/package.json
DhariniJeeva: 💯  that makes more sense, I will rename and move it up

hashicc (approved):
So many great changes including yarn 4 itself. Thanks for all this 🥳

package.json
```diff
@@ -49,20 +47,21 @@
...
-    "**/nomnom/underscore": "^1.12.1",
-    "**/core/ember-inline-svg/**/nth-check": "^2.0.1",
-    "**/ember-cli/**/ansi-html": "^0.0.9",
-    "**/ember-cli/testem/fireworm/async": "^2.6.4",
-    "**/@hashicorp/design-system-components/ember-stargate": "^0.6.0",
-    "micromatch": "^4.0.8"
+    "nomnom/underscore": "^1.12.1",
+    "@hashicorp/design-system-components/ember-stargate": "^0.6.0",
+    "micromatch": "^4.0.8",
+    "ember-inline-svg/nth-check": "^2.0.1",
```
lisbet-alvarez, line 53: apologies im barely noticing that `nth-check` isn't a direct dependency of `ember-inline-svg`. im thinking `"css-select/nth-check": "^2.0.1",` should work

DhariniJeeva: oops good catch, yes it "css-select" when I do yarn why, changing it

lisbet-alvarez (approved):
Awesome work on this huge upgrade!! 🥳

ZedLi (approved):
Looks great, thank you for this work!!

ui/admin/package.json
```diff
@@ -76,7 +76,7 @@
     "ember-load-initializers": "^2.1.2",
     "ember-modifier": "^4.1.0",
     "ember-page-title": "^7.0.0",
-    "ember-pollster": "https://github.com/hashicorp/ember-pollster.git#718db797382b04aca281216319885c5ff625f539",
+    "ember-pollster": "https://github.com/hashicorp/ember-pollster.git#605d82e986d829b098667e4a5e221877a4ee4ac8",
```
laurenolivia, line 79: I wasn't aware of the ember-pollster repo until now 👍 Thanks

cameronperera (approved):
Thanks for getting this working!

package.json
```diff
@@ -49,20 +47,21 @@
...
-    "**/ember-cli/**/ansi-html": "^0.0.9",
-    "**/ember-cli/testem/fireworm/async": "^2.6.4",
-    "**/@hashicorp/design-system-components/ember-stargate": "^0.6.0",
-    "micromatch": "^4.0.8"
+    "nomnom/underscore": "^1.12.1",
+    "@hashicorp/design-system-components/ember-stargate": "^0.6.0",
+    "micromatch": "^4.0.8",
+    "css-select/nth-check": "^2.0.1",
+    "node-gyp": "^10.0.0",
+    "ansi-html": "^0.0.9",
```
calcaide, line 55: None blocking comment: it is possible to keep `ansi-html` relative?

calcaide (approved):
Amazing work and amazing win for the team!!! 🙌

package.json
DhariniJeeva: @calcaide  I originally had it under ember-cli, but ansi-html is not a direct dependency and they are used in broccoli-middleware and broccoli as pointed out by Lis [here](https://github.com/hashicorp/boundary-ui/pull/2697#discussion_r1972260084). Do you think we should use two relative paths for this dependency since they are used by both broccoli and broccoli-middleware? I'm open to changing this if that option fits us better than not specifying the relative path at all.

calcaide: Ahh gotcha.

I am fine leaving it like this, because with the ember upgrade I am sure we can review the `ansi-html` resolution 😉

PR #2703, opened 2025-02-27
fix: 🐛 Fix development build for desktop client

ZedLi:
# Description
Currently building the DC fails if you don't set the `RELEASE_COMMIT` env which usually happens when building in development. I just also set a dummy `0.0.0` for the raw release version variable.

## How to Test
Building DC works without setting any env variables. 

## Checklist


DhariniJeeva (approved):
thanks for the fix, I was running into this same issue right now!

PR #2705, opened 2025-03-04
refactor: 💡 remove usage of ember/runloop & ember-pollster

lisbet-alvarez:
:tickets: [Jira ticket](https://hashicorp.atlassian.net/browse/ICU-16533)

# Description

Refactor usage of ember runloop due to a new 

Remove usage of ember/runloop in:
- `Form::Authenticate` component
- `LoadingButton` component
- `ClockTick` service
- `From::Account` components
    - Usage here is not needed. From old commit history it seems, like usage was attempting to fix a bug. However, that bug behaviour is not reproducible anymore. 

Remove `ember-pollster`usage and replace with `ember-concurrency`. Some usage of `ember/runloop` was also replaced with `ember-concurrency`. 

NOTE: `ember/runloop` is still used in test files for our instance-initializers. By default `ember g instance-initializer` creates test files that have `ember/runloop` usage. For that usage, we will just disable the eslint rule. Additionally, this PR does not refactor usage in the rose::codeeditor::toolbar component since that will be replaced with HDS.

clock-tick:

https://github.com/user-attachments/assets/5e91e6c2-4f24-4bb1-9627-cb1f248ab0dc

transparent sessions:

https://github.com/user-attachments/assets/0aa2591e-cb6b-4f64-8ebf-bafc89c40730

## How to Test

1. Validate change-password & set-password functionality for password type accounts.
2. Validate authentication functionality (logging in to admin ui & desktop ui)
3. Validate OIDC authentication for both admin ui & desktop ui
4. Validate session connection using client agent.
    - Use `ENABLE_MIRAGE=false yarn start:desktop` & connect using bounday dev instance
    - Make sure client-agent is running by going to `settings` tab
    - in a terminal try `ssh ssh.boundary.dev` but make sure the `Generated localhost ssh target with an alias` target has at least one brokered credential.
    - In both mac & windows user should get a notification that when clicked takes them to established session page in DC.
    - Polling is used for retrieving the sessions from the client agent that's why this functionality needs to be tested.
5. Validate `Refresh` button shows loading indicator icon when user click on it for routes such as `targets` & `sessions`.

## Checklist


ui/admin/config/environment.js
```diff
@@ -195,6 +194,8 @@ module.exports = function (environment) {
 
     // Notification timeout should be 0 for fast tests
     ENV.flashMessageDefaults.timeout = 0;
+    // OIDC Authentication timeout should be 0 for fast tests
+    ENV.oidcPollingTimeoutSeconds = 0;
```
hashicc, line 198: Awesome! Thanks for making this configurable

ui/desktop/app/routes/scopes/scope/authenticate/method/oidc.js
```diff
@@ -19,6 +19,16 @@ export default class ScopesScopeAuthenticateMethodOidcRoute extends Route {
   @service session;
   @service router;
 
+  // =attributes
+
+  poller = task({ drop: true }, async () => {
+    while (!this.session.isAuthenticated) {
+      console.log('test');
```
hashicc, line 26: ✂️ `console.log('test');`

ui/desktop/app/components/settings-card/client-agent/index.js
```diff
@@ -63,6 +63,8 @@ export default class SettingsCardClientAgentComponent extends Component {
     } else {
       await this.ipc.invoke('resumeClientAgent');
     }
-    await this.router.refresh('scopes.scope.projects.settings');
+    // Refresh projects page in order to stop or start
+    // polling when client agent status changes.
+    await this.router.refresh('scopes.scope.projects');
```
hashicc, line 68: I'm trying to understand how this was working before with `scopes.scope.projects.settings` vs the new `scopes.scope.projects`. The ember docs for the `Router#refresh` for its argument:
> routeName String 
the route to refresh (along with all child routes)

> If a route name is provided and refers to a currently active route, it will refresh only that route and its descendents. 

So I'm not sure how `scopes.scope.projects.settings` would have been refreshed before to restart the poller because it's not the route that contains the poller, `scopes.scope.projects` is 🤔. Maybe, I'm missing something but in any case your change looks correct!

hashicc (commented):
Everything looks great so far. I want to take a closer look at `ui/desktop/app/routes/scopes/scope/projects.js` tomorrow but so far so good!

addons/core/addon/components/loading-button/index.js
```diff
@@ -4,22 +4,13 @@
...
-  @tracked isLoading = false;
+  // =attributes
 
-  @action
-  async toggleRefresh() {
-    this.isLoading = true;
-    try {
-      await this.args.onClick();
-    } catch (e) {
-      console.error('Error while loading data', e);
```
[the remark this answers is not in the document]
lisbet-alvarez: I was going off of this [docs section](https://ember-concurrency.com/docs/tutorial/refactor#:~:text=Version%205%3A%20Handle%20Promise%20Rejection/Async%20Exceptions%20(with%20Tasks)). But this is a good reminder to test this (by throwing an error and making sure its logged).

addons/core/addon/services/clock-tick.js
```diff
@@ -75,9 +74,9 @@ export default class ClockTickService extends Service {
   }
 
   /**
-   * Updates the value of `now` within the next runloop.
+   * Updates the value of `now`.
    */
   _tick() {
-    next(() => (this.now = Date.now()));
+    this.now = Date.now();
```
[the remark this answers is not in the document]
lisbet-alvarez: these are good points, ill refactor this that usage as well, thank you!

ui/desktop/app/routes/scopes/scope/authenticate/method/oidc.js
lisbet-alvarez: woops thanks for catching this!

addons/core/addon/components/loading-button/index.hbs
```diff
@@ -6,7 +6,7 @@
 <Hds::Button
   @text={{t 'actions.refresh'}}
   @color='secondary'
-  @icon={{if this.isLoading 'loading' 'reload'}}
+  @icon={{if this.toggleRefresh.isRunning 'loading' 'reload'}}
```
DhariniJeeva, line 9: where do we get this `isRunning` from?

lisbet-alvarez: oh good question, that is [exposed by the task](https://ember-concurrency.com/docs/tutorial/refactor#:~:text=Version%202%3A%20Add%20a%20Loading%20Spinner%20(with%20Tasks))

DhariniJeeva: oh I see that we get this state from `ember-concurrency`

ui/desktop/app/components/settings-card/client-agent/index.js
lisbet-alvarez: yeah, im not sure of how it was originally getting refreshed either since i couldn't find any thing else that could trigger resetting the `isClientAgentRunning` value in `projects.js`

addons/core/ember-cli-build.js
```diff
@@ -14,6 +14,13 @@ module.exports = function (defaults) {
     'ember-simple-auth': {
       useSessionSetupMethod: true,
     },
+    babel: {
+      plugins: [
+        // ... any other plugins
+        require.resolve('ember-concurrency/async-arrow-task-transform'),
+        // NOTE: put any code coverage plugins last, after the transform.
```
DhariniJeeva, line 21: are these comments needed?

ui/admin/app/components/form/account/password/change-password/index.js
```diff
@@ -25,23 +23,21 @@ export default class FormAccountPasswordChangePasswordIndexComponent extends Com
   /**
    * @type {boolean}
    */
-  @computed('currentPassword', 'newPassword')
```
DhariniJeeva, line 28: nice cleanup!

ui/admin/app/routes/scopes/scope/authenticate/method/oidc.js
```diff
@@ -43,16 +52,19 @@ export default class ScopesScopeAuthenticateMethodOidcRoute extends Route {
...
+   */
+  activate() {
+    this.poller.perform();
   }
 
   /**
    * When this route is deactivated (exited), stop polling for changes and close
    * any windows opened via the window manager service.
    */
   deactivate() {
```
DhariniJeeva, line 66: where are we using these methods?

DhariniJeeva: this and activate

addons/core/addon/components/loading-button/index.js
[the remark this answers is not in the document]
hashicc: >  bubble the exception to the console (so that error reporters can catch it)

I was looking but didn't see it, thanks for the link.

ui/admin/app/routes/scopes/scope/authenticate/method/oidc.js
hashicc: These are hooks on ember `Route` when enter/leaving routes:
* [`activate`](https://api.emberjs.com/ember/6.2/classes/Route/methods/activate?anchor=activate)
* [`deactivate`](https://api.emberjs.com/ember/6.2/classes/Route/methods/activate?anchor=deactivate)

addons/core/ember-cli-build.js
lisbet-alvarez: nope not really, removed them, thank u!

addons/core/addon/components/loading-button/index.hbs
calcaide: Pretty neat!

ui/admin/app/components/form/account/password/change-password/index.js
calcaide: +1 to Dharini's comment, well done 🙌

ui/admin/config/environment.js
calcaide: Indeed!!!

calcaide (dismissed):
Terrific job @lisbet-alvarez !!! very well done 🙌

ui/admin/config/environment.js
```diff
@@ -52,7 +52,6 @@ module.exports = function (environment) {
     appName: APP_NAME,
     companyName: 'HashiCorp',
 
-    sessionPollingTimeoutSeconds: 300,
```
priya-patel04, line 55: For my own understanding, why is this no longer needed ?

priya-patel04 (dismissed):
Looks great!! Just one non-blocking comment

addons/core/addon/services/clock-tick.js
```diff
@@ -25,9 +25,15 @@ export default class ClockTickService extends Service {
   frequency = 1000;
 
   /**
-   * The timer interval ID
+   * Updates the value of `now`every `frequency` milliseconds.
    */
-  #timer;
+  _tick = dropTask(async () => {
```
ZedLi, line 30: Would we ever see a scenario where this task is getting dropped?

ui/desktop/app/components/settings-card/client-agent/index.js
ZedLi: I think this is a bug since pausing the client agent doesn't actually prevent you from calling the sessions route so likely if we paused, it was still polling unnecessarily and works again if you resume. It likely would have been an issue if it was already paused and then tried to resume 😬. Great catch!!

ui/desktop/app/routes/scopes/scope/projects.js
```diff
@@ -76,41 +70,45 @@ export default class ScopesScopeProjectsRoute extends Route {
...
-    super.willDestroy();
+    super.willDestroy(...arguments);
+    this.poller.cancelAll();
   }
 
   /**
    * Poll for new sessions with credentials. Sends a notification for each new session that has a credential.
    */
-  async poller() {
+  poller = restartableTask(async () => {
```
ZedLi, line 80: Shouldn't this task be a droppable one rather than restarted? Theoretically if the request took longer than 2 seconds, it would never finish each time since it would restart.

ui/desktop/app/routes/scopes/scope/projects.js
```diff
@@ -76,41 +70,45 @@ export default class ScopesScopeProjectsRoute extends Route {
   }
 
   willDestroy() {
```
ZedLi, line 72: I'm wondering if we should be using `activate` and `deactivate` like in the OIDC route?

ui/desktop/app/routes/scopes/scope/projects.js
```diff
@@ -76,41 +70,45 @@ export default class ScopesScopeProjectsRoute extends Route {
...
+          const auth_token_id = sessionData?.id;
+          const token = sessionData?.token;
+
+          try {
+            await this.ipc.invoke('addTokenToDaemons', {
+              tokenId: auth_token_id,
+              token,
+            });
+
+            this.poller.perform();
```
ZedLi, line 102: Can you reference this method recursively to start the task?

addons/core/addon/services/clock-tick.js
lisbet-alvarez: hmm ur right since these aren't tasks triggered by a user action

ui/admin/config/environment.js
lisbet-alvarez: great question, seems to me that it was leftover from a refactor since there were no references to this anywhere else.

priya-patel04: Ahh got it. Thanks. Then should we remove `sessionPollingTimeoutSeconds` from desktop `config/environment` file as well ?

lisbet-alvarez: in that case no because for desktop i recycled the name for the polling in the the `routes/scopes/scope/projects.js` route so it is being used.

ui/desktop/app/routes/scopes/scope/projects.js
lisbet-alvarez: hm i don't think this will be an issue since `perform` is only called from within this task function and also from within the model hook  and in those cases previous instances of the task should be canceled. Thoughts?

ui/desktop/app/routes/scopes/scope/projects.js
lisbet-alvarez: yeah thats a good point on using `deactivate` instead since its called when the route is completed exited and aligns with this use case. on the point of using activate, i think keeping the logic in the model hook works best since `activate` hook is not called on refreshes.

ui/desktop/app/routes/scopes/scope/projects.js
lisbet-alvarez: yup, i tested this by throwing an error so it gets to this block. the currently running task will get canceled and new task will start running from the beginning. Also just realized that the log call should be moved before this perform call. thank u

ui/desktop/app/routes/scopes/scope/projects.js
ZedLi: Ah yeah I think I forgot each poll isn't doing a `perform` and was thinking each request would be a new task, this makes sense

ZedLi (approved):
Great work! We finally got off ember-pollster and come back full circle 😆

cameronperera (approved):

PR #2710, opened 2025-03-06
Initial E2E DC setup

ZedLi:
# Description
Initial setup of e2e tests for DC. This right now just runs on a linux runner and uses docker for enos setup


.github/workflows/test-e2e-desktop.yaml
```diff
@@ -0,0 +1,109 @@
+name: Test E2E Desktop Client
+
+on:
+  pull_request:
+    types: [labeled]
+  workflow_dispatch:
```
moduli, line 6: - define what cadence we want this to run (every push? on PR?)
- we probably want to keep the `workflow_dispatch` option
- perhaps only run this when there are changes to the desktop client code base

.github/workflows/test-e2e-desktop.yaml
```diff
@@ -0,0 +1,109 @@
...
+          ssh -V
+
+          echo "127.0.0.1 localhost boundary" | sudo tee -a /etc/hosts
+          echo "127.0.0.1 localhost worker" | sudo tee -a /etc/hosts
+          echo "127.0.0.1 localhost vault" | sudo tee -a /etc/hosts
+
+      - name: Get Boundary docker image
+        run: |
+          docker pull hashicorp/boundary-enterprise:latest
+          docker save hashicorp/boundary-enterprise:latest | gzip > /tmp/boundary-enterprise.zip
```
moduli, line 67: We'll need to figure out what version of boundary to run against. If `main`, we'll need to figure out how get the latest docker image

(we'll also need to update the environment variables set in the `set up test infra` step

.github/workflows/test-e2e-desktop.yaml
```diff
@@ -0,0 +1,109 @@
...
+
+      - name: Set up Enos
+        uses: hashicorp/action-setup-enos@b9fa53484a1e8fdcc7b02a118bcf01d65b9414c9
+        with:
+          github-token: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Install Vault CLI
+        run: |
+          wget https://releases.hashicorp.com/vault/1.12.2/vault_1.12.2_linux_amd64.zip -O /tmp/vault.zip
+          unzip /tmp/vault.zip -d /usr/local/bin
```
moduli, line 43: We'll need to see if we want to figure out how to keep the version of vault in sync with the version of vault used in the docker container versus just hard-coding the version here

.github/workflows/test-e2e-desktop.yaml
moduli: Additionally, there are two version of boundary being specified here
- one for the docker image
- one for the code base we checkout that contains the enos/infra setup code

Do we want to keep those in sync or have them be defined separately? We may also want to allow a user to specify, at a minimum, the version of the docker image (could be useful, say, to run against an LLB for a feature or for release testing).

.github/workflows/test-e2e-desktop.yaml
```diff
@@ -0,0 +1,109 @@
...
+      SETUP_CLI: true
+    steps:
+      - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
+        with:
+          fetch-depth: 0
+
+      - name: Set up Node
+        uses: actions/setup-node@39370e3970a6d050c480ffad4ff0ed4d3fdee5af # v4.1.0
+        with:
+          node-version: 20
```
moduli, line 26: Not sure if there's a better way to keep this node version in sync with other places

ZedLi: I think this will just be like our other actions and we'll have to keep them in sync ourselves

.github/workflows/test-e2e-desktop.yaml
ZedLi: Yeah I think we should keep them in sync. I was actually thinking maybe we should just combine them by checking out the repo for enos and then also building the image ourselves at that version rather than pulling from anywhere.

moduli: 👍 Worth trying

.github/workflows/test-e2e-desktop.yaml
ZedLi: Yup, right now I just have it set to when there's a `desktop` label added which happens automatically if anything under `ui/desktop` is changed but someone could also add the label manually if for example `auth` or `core` was changed only.

I was also going to eventually add an input to the `workflow_dispatch` to take in a version for the boundary enterprise

.github/workflows/test-e2e-desktop.yaml
ZedLi: It seems building the ent image also will require some adjustments so for now I added a todo and just keep using `latest`

calcaide (approved):
LGTM!! Thanks for this 🙌

hashicc (approved):
🏗️ Awesome, this was really clear to step through
Here's the branch's [CI run for this action](https://github.com/hashicorp/boundary-ui/actions/runs/13708739314/job/38340275828?pr=2710) for anyone looking

DhariniJeeva (approved):
looks great!!

.github/workflows/test-e2e-desktop.yaml
```diff
@@ -0,0 +1,110 @@
...
+        id: infra
+        run: |
+          echo "${{ secrets.BOUNDARY_ENT_LICENSE }}" > ./support/src/boundary-enterprise/enos/license.hclic
+          ssh-keygen -N '' -t ed25519 -f ~/.ssh/github_enos
+          mkdir -p ./enos/terraform-plugin-cache
+          export ENOS_VAR_enos_user=$GITHUB_ACTOR && \
+          export ENOS_VAR_aws_ssh_private_key_path=~/.ssh/github_enos && \
+          export ENOS_VAR_boundary_docker_image_name=hashicorp/boundary-enterprise:latest && \
+          export ENOS_VAR_boundary_docker_image_file=/tmp/boundary-enterprise.zip && \
+          enos scenario launch --timeout 60m0s --chdir ./support/src/boundary-enterprise/enos e2e_ui_docker_ent builder:crt
```
lisbet-alvarez, line 81: til that `--timeout` and `--chdir` flags were available for launching an enos scenario!

.github/workflows/test-e2e-desktop.yaml
```diff
@@ -0,0 +1,110 @@
...
+      - name: Install Vault CLI
+        run: |
+          wget https://releases.hashicorp.com/vault/1.12.2/vault_1.12.2_linux_amd64.zip -O /tmp/vault.zip
+          unzip /tmp/vault.zip -d /usr/local/bin
+
+      - name: Checkout boundary-enterprise
+        uses: actions/checkout@692973e3d937129bcbf40652eb9f2f61becf3332 # v4.1.7
+        with:
+          repository: hashicorp/boundary-enterprise
+          #          ref: ${{ github.event.inputs.boundary-enterprise-branch }}
```
lisbet-alvarez, line 49: is this comment still needed?

lisbet-alvarez (approved):
wow getting e2e tests in CI is a big improvement!!

.github/workflows/test-e2e-desktop.yaml
ZedLi: Heh I also have never used these flags, all credit to @moduli

.github/workflows/test-e2e-desktop.yaml
```diff
@@ -0,0 +1,110 @@
...
+      - name: Get Boundary docker image
+        run: |
+          docker pull hashicorp/boundary-enterprise:latest
+          docker save hashicorp/boundary-enterprise:latest | gzip > /tmp/boundary-enterprise.zip
+
+        # TODO: Set up license in a better way? Use directly rather than writing to file and pull from secrets manager
+      - name: Set up test infra
+        id: infra
+        run: |
+          echo "${{ secrets.BOUNDARY_ENT_LICENSE }}" > ./support/src/boundary-enterprise/enos/license.hclic
```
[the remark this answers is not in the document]
ZedLi: Ah I didn't think about this but the variable is actually exported here:
https://github.com/hashicorp/boundary-ui/blob/93ca423d7b92fadd405fe5bd53b0b2cdad57c52f/.github/workflows/test-e2e-desktop.yaml#L16

It seems we have variables being declared in two different places so this might be something to cleanup, though we might be able to not need this echo and directly use the env variable in the future

.github/workflows/test-e2e-desktop.yaml
ZedLi: Hmm the plan is to eventually take an input so I don't think it hurts to leave it there for now

.github/workflows/test-e2e-desktop.yaml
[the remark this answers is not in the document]
lisbet-alvarez: ah ok that makes sense, thanks for linking where the var is exported. I agree this can be left as is and if possible cleaned up later 👍

PR #2711, opened 2025-03-07
Icu 16534 refactor usage of tracked using this args

lisbet-alvarez:
:tickets: [Jira ticket](https://hashicorp.atlassian.net/browse/ICU-16534)

# Description

This refactor is in preparation for ember upgrade that will update eslint which comes with some new rules. One of them is disallowing usage of `this.args` with `@tracked`. Our usage was limited to initializing tracked variables so that was refactored to use getters instead to initialize.

## How to Test

1. Use vercel deployments or checkout branch locally
2. Verify [aws host catalog form](https://boundary-9bbz3zq83-hashicorp.vercel.app/scopes/p_yie834txcy/host-catalogs/new?type=aws)(credential type radio buttons) work as expected.
3. Verify [aws storage bucket form](https://boundary-9bbz3zq83-hashicorp.vercel.app/scopes/global/storage-buckets/new?compositeType=aws) work as expected.
4. Verify [worker filter generator toggle](https://boundary-9bbz3zq83-hashicorp.vercel.app/scopes/p_yie834txcy/targets/t_rj4sz2z4gl/edit-ingress-worker-filter) works as expected.
## Checklist

### DIFF
--- a/ui/admin/app/controllers/scopes/scope/storage-buckets/index.js
+++ b/ui/admin/app/controllers/scopes/scope/storage-buckets/index.js
@@ -4,7 +4,7 @@
  */
 
 import Controller from '@ember/controller';
-import { inject as service } from '@ember/service';
+import { service } from '@ember/service';
 import { action } from '@ember/object';
 import { loading } from 'ember-loading';
 import { confirm } from 'core/decorators/confirm';
--- a/.github/workflows/test-e2e-desktop.yaml
+++ b/.github/workflows/test-e2e-desktop.yaml
@@ -1,0 +1,110 @@
+name: Test E2E Desktop Client
+
+on:
+  pull_request:
+    types: [labeled, opened, synchronize, reopened]
+  workflow_dispatch:
+
+jobs:
+  test:
+    if: ${{ github.repository == 'hashicorp/boundary-ui' && contains(github.event.pull_request.labels.*.name, 'desktop')}}
+    name: Run Desktop E2E Tests
+    runs-on: ubuntu-22.04
+    env:
+      ENOS_VAR_boundary_edition: "ent"
+      ENOS_VAR_e2e_debug_no_run: true
+      ENOS_VAR_boundary_license_path: "./license.hclic"
+      SETUP_CLI: true
+    steps:
+      - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
+        with:
+          fetch-depth: 0
+
+      - name: Set up Node
+        uses: actions/setup-node@39370e3970a6d050c480ffad4ff0ed4d3fdee5af # v4.1.0
+        with:
+          node-version: 20
+          cache: "yarn"
+          cache-dependency-path: |
+            yarn.lock
+            ui/desktop/electron-app/yarn.lock
+
+      - name: Set up Terraform
+        uses: hashicorp/setup-terraform@a1502cd9e758c50496cc9ac5308c4843bcd56d36
+
+      - name: Set up Enos
+        uses: hashicorp/action-setup-enos@b9fa53484a1e8fdcc7b02a118bcf01d65b9414c9
+        with:
+          github-token: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Install Vault CLI
+        run: |
+          wget https://releases.hashicorp.com/vault/1.12.2/vault_1.12.2_linux_amd64.zip -O /tmp/vault.zip
+          unzip /tmp/vault.zip -d /usr/local/bin
+
+      - name: Checkout boundary-enterprise
+        uses: actions/checkout@692973e3d937129bcbf40652eb9f2f61becf3332 # v4.1.7
+        with:
+          repository: hashicorp/boundary-enterprise
+          #          ref: ${{ github.event.inputs.boundary-enterprise-branch }}
+          path: support/src/boundary-enterprise
+          token: ${{ secrets.ELEVATED_GITHUB_TOKEN }}
+
+      - name: GH fix for localhost resolution
+        run: |
+          cat /etc/hosts && echo "-----------"
+          sudo sed -i 's/::1 *localhost ip6-localhost ip6-loopback/::1 ip6 -localhost ip6-loopback/g' /etc/hosts
+          cat /etc/hosts
+          ssh -V
+
+          echo "127.0.0.1 localhost boundary" | sudo tee -a /etc/hosts
+          echo "127.0.0.1 localhost worker" | sudo tee -a /etc/hosts
+          echo "127.0.0.1 localhost vault" | sudo tee -a /etc/hosts
+
+      # TODO: Build image ourselves in case we don't want to use latest?
+      - name: Get Boundary docker image
+        run: |
+          docker pull hashicorp/boundary-enterprise:latest
+          docker save hashicorp/boundary-enterprise:latest | gzip > /tmp/boundary-enterprise.zip
+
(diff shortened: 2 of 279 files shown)


ui/admin/app/components/form/host-catalog/aws/index.js
```diff
@@ -13,8 +13,8 @@ import { tracked } from '@glimmer/tracking';
 
 export default class FormHostCatalogAwsComponent extends Component {
   // =attributes
-  @tracked selectedCredentialType = this.args.model.credentialType;
 
+  @tracked selectedCredentialType = this.defaultCredentialType;
```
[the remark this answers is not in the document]
lisbet-alvarez: hmm i also did initially attempt just making `selectedCredentialType` a getter but that didn't work like I expected it to since the UI didn't refresh on user input. Im thinking its because `credentialType` is not an actual attribute of the model?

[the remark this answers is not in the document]
ZedLi: Ohh I didn't realize that, it looks like it's a private field? I think that private field itself would then need to be tracked unfortunately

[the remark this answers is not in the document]
lisbet-alvarez: ohh yup that makes sense, ill try that! thank u!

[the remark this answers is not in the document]
lisbet-alvarez: actually nvm labeling the `credentialType` property from the model as tracked causes an error :(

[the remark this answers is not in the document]
ZedLi: Is it because it's on a private field or something else?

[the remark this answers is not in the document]
lisbet-alvarez: nope so i had to make the field public in order to use `tracked` (also had to change var name so it doesn't conflict with getter & setter names) and just doing that is causing a runtime error:
Screenshot 2025-03-10 at 12 34 07 AM

addons/api/addon/models/host-catalog.js
```diff
@@ -40,7 +41,8 @@ export const STATIC_CREDENTIAL_FIELDS = ['access_key_id, secret_access_key'];
 export default class HostCatalogModel extends GeneratedHostCatalogModel {
   // =attributes
 
-  #credentialType;
+  @tracked credentialTypeValue;
```
ZedLi, line 44: Traditionally you would call this `_credentialType` before JS added private fields since it's not meant to be accessed directly

addons/api/addon/serializers/host-catalog.js
```diff
@@ -34,7 +34,7 @@ export default class HostCatalogSerializer extends ApplicationSerializer {
     ) {
       if (options.isNestedAttribute) {
         json.attributes[key] = null;
-      } else if (options.isNestedSecret && json.secrets[key]) {
+      } else if (options.isNestedSecret && json.secrets && json.secrets[key]) {
```
ZedLi, line 37: Would this have worked?

```suggestion
      } else if (options.isNestedSecret && json?.secrets[key]) {
```

addons/api/addon/models/host-catalog.js
```diff
@@ -62,22 +64,22 @@ export default class HostCatalogModel extends GeneratedHostCatalogModel {
    * @type {string}
    */
   get credentialType() {
-    if (!this.#credentialType) {
+    if (!this.credentialTypeValue) {
       if (DYNAMIC_CREDENTIAL_FIELDS.some((field) => this[field])) {
-        this.#credentialType = TYPE_CREDENTIAL_DYNAMIC;
+        return TYPE_CREDENTIAL_DYNAMIC;
```
calcaide, line 69: None blocking comment: 
The previous implementation checks if `this.credentialType` exists.
If it doesn't, it sets a value.
At the end, it always returns `this.credentialType`

The new implementation checks if `this.credentialType` exists.
If it doesnt', returns a value. BUT, and here is my questions, never sets the `credentialType` @tracked property. I have nothing against, I would like to learn Why we change the approach on this?
If it does, returns the `this.credentialType`

ui/admin/app/components/form/storage-bucket/aws/index.hbs
```diff
@@ -221,9 +221,9 @@
     {{#each this.credentials as |credentialType|}}
       <G.RadioCard
         @value={{credentialType}}
-        @checked={{eq credentialType this.selectedCredentialType}}
+        @checked={{eq credentialType @model.credentialType}}
         @disabled={{form.disabled}}
-        {{on 'input' (fn this.updateCredentialTypeSelection credentialType)}}
+        {{on 'input' (fn @changeCredentialType credentialType)}}
```
calcaide, line 226: Neet! I forgot we have `changeCredentialType` within the controller, nice clean up of the component deleting `updateCredentialTypeSelection`!!

calcaide:
Looking great so far!! left a couple of comments and will wait on other feedback too 😉

addons/api/addon/models/host-catalog.js
lisbet-alvarez: oh thats a good idea! thank u!

addons/api/addon/serializers/host-catalog.js
lisbet-alvarez: thank u for the suggestion, this made me realize that something similar works.

ui/admin/app/controllers/scopes/scope/storage-buckets/index.js
```diff
@@ -55,6 +56,10 @@ export default class ScopesScopeStorageBucketsIndexController extends Controller
   @notifyError(({ message }) => message)
   @notifySuccess('notifications.save-success')
   async save(storageBucket) {
+    // If the role_arn is empty, then the credential type should be static
+    if (!storageBucket.role_arn) {
+      storageBucket.credentialType = TYPE_CREDENTIAL_STATIC;
+    }
```
lisbet-alvarez, line 62: if user has an existing storage bucket with static credential type and edits type to dynamic credential without entering a `role_arn` value then on save it will show type change to 'dynamic' even though its still static. So this change was added so that type stays as 'static'. We already do the same for host-catalogs.

addons/api/addon/models/host-catalog.js
lisbet-alvarez: Great question, this getter shouldn't set the value of `credentialType` once it was changed to be `tracked` because we get [this error](https://github.com/hashicorp/boundary-ui/pull/2711#discussion_r1987705022). It also doesn't need to set the value for `credentialType` because the getter takes care of returning the correct type on initial load (when the private variable `credentialType` is undefined).

ui/admin/app/controllers/scopes/scope/storage-buckets/index.js
ZedLi: Good catch!

addons/api/addon/serializers/host-catalog.js
ZedLi: Woops I had it on the wrong field 😅

ZedLi (approved):
Looks great!

DhariniJeeva (approved):
Thank you for the refactor! 🎉

addons/api/addon/models/host-catalog.js
calcaide: @lisbet-alvarez thanks for your detailed answer! makes sens 😃

PR #2712, opened 2025-03-07
test: 💍 fix false positive test

cameronperera:
# Description

Updated use of `data-test` selector and fix false positive test.
This test case is an example of a time when creating an external file called `selectors` does not make sense. The selector is only used 2 times and only in this file. Perfect example of when going against the new norm is ok.

## Checklist


lisbet-alvarez (approved):
great catch!

calcaide (approved):
Indeed with your comment, this test is at the root and without domain, so one file test = no need for selectors file 😉 

LGTM!! 🙌 Thanks for the work!

PR #2713, opened 2025-03-07
test: 💍 use test selectors for groups

DhariniJeeva:
# Description
This pr uses test selectors for delete and update groups 

[update](https://hashicorp.atlassian.net/browse/ICU-15482)
[delete](https://hashicorp.atlassian.net/browse/ICU-15483)

## Checklist

- [ ] I have added before and after screenshots for UI changes
- [ ] I have added JSON response output for API changes
- [ ] I have added steps to reproduce and test for bug fixes in the description
- [ ] I have commented on my code, particularly in hard-to-understand areas
- [ ] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works

PR #2718, opened 2025-03-10
Refactor: 💡 User acceptance tests - create, delete and read

priya-patel04:
# Description
Test selector refactor for users tests:
[create](https://hashicorp.atlassian.net/browse/ICU-16441)
[read](https://hashicorp.atlassian.net/browse/ICU-16442)
[delete](https://hashicorp.atlassian.net/browse/ICU-16444)

## Checklist

### DIFF
--- a/ui/admin/tests/acceptance/users/create-test.js
+++ b/ui/admin/tests/acceptance/users/create-test.js
@@ -57,9 +57,9 @@ module('Acceptance | users | create', function (hooks) {
     const usersCount = getUsersCount();
     await visit(urls.users);
 
-    await click(`[href="${urls.newUser}"]`);
-    await fillIn('[name="name"]', 'User name');
-    await click('[type="submit"]');
+    await click(commonSelectors.HREF(urls.newUser));
+    await fillIn(commonSelectors.FIELD_NAME, commonSelectors.FIELD_NAME_VALUE);
+    await click(commonSelectors.SAVE_BTN);
 
     assert.strictEqual(getUsersCount(), usersCount + 1);
   });
@@ -67,14 +67,14 @@ module('Acceptance | users | create', function (hooks) {
   test('users can navigate to new users route with proper authorization', async function (assert) {
     await visit(urls.orgScope);
 
-    await click(`[href="${urls.users}"]`);
+    await click(commonSelectors.HREF(urls.users));
 
     assert.true(
       instances.scopes.org.authorized_collection_actions.users.includes(
         'create',
       ),
     );
-    assert.dom(`[href="${urls.newUser}"]`).exists();
+    assert.dom(commonSelectors.HREF(urls.newUser)).exists();
   });
 
   test('users cannot navigate to new users route without proper authorization', async function (assert) {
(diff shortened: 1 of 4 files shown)


ui/admin/tests/acceptance/users/create-test.js
```diff
@@ -130,15 +130,18 @@ module('Acceptance | users | create', function (hooks) {
       );
     });
 
-    await click(`[href="${urls.newUser}"]`);
-    await fillIn('[name="description"]', 'test');
-    await click('[type="submit"]');
+    await click(commonSelectors.HREF(urls.newUser));
```
calcaide, line 133: None blocking comment: I know the code was there and a bit out of the scope of the selectors refactor. That said, this test is mocking a request where marks the field name as required, but in reality, for users name is not a required field. That means we are testing a none realistic case, but we need to test the failed API response. In these cases were the domain has no required fields, Cameron recently posted this example.
Let me know your thoughts  😉 

I know this is something out of the scope of this PR where we are refactoring test selectors. I am fine if you want to fix this on a separate PR .

ui/admin/tests/acceptance/users/delete-test.js
```diff
@@ -46,63 +44,63 @@ module('Acceptance | users | delete', function (hooks) {
     urls.orgScope = `/scopes/${instances.scopes.org.id}`;
     urls.users = `${urls.orgScope}/users`;
     urls.user = `${urls.users}/${instances.user.id}`;
+    this.getUsersCount = () => this.server.schema.users.all().models.length;
```
calcaide, line 47: I will suggest to not use `this.` for this approach. We have a pattern set for this cases [an example here](https://github.com/hashicorp/boundary-ui/pull/2713/files#diff-96c9d815b19821b19a64954dd93955a4c25cdc9ce2c8cc221504d626ddad48fcR49).

Let me know your throughs 😉

calcaide (changes requested):
Looking nice, Great work!! I left a couple of comments 😉

ui/admin/tests/acceptance/users/delete-test.js
priya-patel04: The goal with using `this` was to make it reusable for multiple tests. If we already have a set pattern for this cases, makes sense to follow them  😄

calcaide (approved):
Looks great!!! thanks for the refactor 🙌

ui/admin/tests/acceptance/users/create-test.js
```diff
@@ -84,31 +84,33 @@ module('Acceptance | users | create', function (hooks) {
...
+    await click(commonSelectors.HREF(urls.users));
 
     assert.false(
       instances.scopes.org.authorized_collection_actions.users.includes(
         'create',
       ),
     );
 
-    assert.dom('.rose-button-primary').doesNotExist();
+    assert.dom(commonSelectors.HREF(urls.newUser)).doesNotExist();
```
cameronperera, line 95: Nice catch. I don't remember what `.rose-button-primay` is but now I know exactly what this test is asserting.

cameronperera (approved):
LGTM

lisbet-alvarez (approved):
Nice cleanup!!

PR #2721, opened 2025-03-11
Fix locationType deprecations.

calcaide:
# Description

Within Desktop Client, we never deprecated auto location, [more info about this deprecation here](https://deprecations.emberjs.com/id/deprecate-auto-location). Therefore, when running the DC in the browser we were seeing 2 deprecations messages:
1. Using `auto` as value has been deprecated. So only `hash` or `history` are accepted.
2. Using detection code at `locationType` setting.

To solve 1, we decide to change `auto` to use `history`. [As per documentation](https://guides.emberjs.com/release/configuring-ember/specifying-url-type/), `auto` will uses `history` if supported by the browsers and fallback to `hash` if not. Since we are pretty sure the browser used by electron supports `history`, is safe to directly uses `history`.

To solve 2, we decide to set a const with the value before we are setting the environment config, therefore at setting time the variable has already a value.

## Screenshots:

Before, showing deprecation messages:
Screenshot 2025-03-11 at 1 56 19 PM

After, showing no deprecation messages
Screenshot 2025-03-11 at 1 57 52 PM

### DIFF
--- a/addons/rose/tests/dummy/config/environment.js
+++ b/addons/rose/tests/dummy/config/environment.js
@@ -10,7 +10,7 @@ module.exports = function (environment) {
     modulePrefix: 'dummy',
     environment,
     rootURL: '/',
-    locationType: 'auto',
+    locationType: 'history',
     EmberENV: {
       EXTEND_PROTOTYPES: false,
       FEATURES: {
--- a/ui/desktop/config/environment.js
+++ b/ui/desktop/config/environment.js
@@ -6,13 +6,14 @@
 'use strict';
 
 const APP_NAME = process.env.APP_NAME || 'Boundary';
+const locationType = process.env.EMBER_CLI_ELECTRON ? 'hash' : 'history';
 
 module.exports = function (environment) {
   const ENV = {
     modulePrefix: 'desktop',
     environment,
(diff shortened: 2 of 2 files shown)


lisbet-alvarez (commented):
Thank you for cleaning this up, great catch!!
this doesn't really affect anything but we should probably still clean-up [this 'auto' usage ](https://github.com/hashicorp/boundary-ui/blob/main/addons/rose/tests/dummy/config/environment.js#L13)too

lisbet-alvarez (approved):
Thank u for fixing this!

calcaide:
> Thank you for cleaning this up, great catch!! this doesn't really affect anything but we should probably still clean-up [this 'auto' usage ](https://github.com/hashicorp/boundary-ui/blob/main/addons/rose/tests/dummy/config/environment.js#L13)too

Great catch!!

Not that I am looking to put effort in rose, since we know eventually we want to deprecate it, but this one I agree with you takes no time to fix it, so I pushed the change! thanks for the catch @lisbet-alvarez

lisbet-alvarez (approved):
Looks great!!

DhariniJeeva (approved):
good catch! thanks for the cleaning this up

ZedLi (approved):

PR #2723, opened 2025-03-12
fix: 🐛 Fix search field race condition in DC targets

ZedLi:
✅ Closes: [ICU-14775](https://hashicorp.atlassian.net/browse/ICU-14775)

# Description
This is an example of one way we could use `ember-concurrency` to solve the search field dropping updates when typing. Take a look at the ticket and before and after videos to see what the issue is.

When we originally add searching, I did some spikes to see if there's any easy ways to solve this and had actually tried `ember-concurrency`. The main issue why it didn't work for me originally was I tried it in two different ways which neither worked:
1. I added the task to the `handleSearchInput` handler which logically is where I think debouncing should be happening
2. I returned a task from the route's `model` _but_ still did debouncing on the `handleSearchInput` handler

Neither of these really work as it seems like what happens is even if we add debouncing to the event handler itself, the `model` hook can still fire off more than once and when one of these hooks actually finishes, it refreshes the page and the query params to what the search was that finished (which is usually an older one) and if you're still typing, you lose the newer search text. 

From @lisbet-alvarez [PR](https://github.com/hashicorp/boundary-ui/pull/2705) I realized the proper way is to actually use `ember-concurrency`'s await inside the task which actually prevents the model hook from finishing before it's fully settled.

This comes with two main annoyances:
1. We're technically adding a straight 250ms delay if there's ever a search. I tried to limit when it would happen by only doing it when there's actually a search parameter but if for they searched then filtered down after, we would now be adding an unnecessary (albeit small) delay.
2. Any model hook that needs this will need to return a task which is just wrapping the logic in a separate function which is not great ergonomically.

I'd love to see if anyone has any better solutions as I see 4 current paths forward:
1. Do this for any model hook that needs it
2. Do what I outlined in the comments in the JIRA task which is remove auto searching form the field and add an enter button so the search only fires when a user explicitly hits enter or clicks the button.
3. Do a much larger refactor that pulls out data fetching to a component that wraps the searching/filtering. We initially didn't design it this way because it seemed better to keep the convention and consistency of all data fetching happening in the routes but the need for debouncing makes it much more difficult. It also seems this [use case](https://discuss.emberjs.com/t/readers-questions-is-it-bad-to-load-data-in-components/14521#highly-interactive-loading-9) is actually one that is called out that could make sense to break out of the route convention.
4. Don't debounce, probably not great but also right now probably less worse than the current UX. 

Before:

https://github.com/user-attachments/assets/5a35ffdd-488a-43d2-84b5-74aa0f423519

After:

https://github.com/user-attachments/assets/21cbe379-1e8c-4ec0-9608-6f279516e040

## How to Test
Start up the DC and try typing at different paces to see if some keyboard events are lost. You can use this [cluster](https://27425825-515c-4d8c-beb6-aa91cbc20c5a.boundary.hcp.to) for more targets if you'd like as the frequency it happens changes depending on how long requests take as well as how fast you type.

## Checklist

- [x My changes generate no new warnings


ui/desktop/app/routes/scopes/scope/projects/targets/index.js
```diff
@@ -82,78 +83,93 @@ export default class ScopesScopeProjectsTargetsIndexRoute extends Route {
...
+  }
 
-    const orgScope = this.modelFor('scopes.scope');
-    const projects = this.modelFor('scopes.scope.projects');
+  retrieveData = restartableTask(
+    async ({ search, scopes, availableSessions, types, page, pageSize }) => {
+      // Only add a debounce timeout if there's a search field
+      // Not perfect if they search first then filter as it then adds extra time
+      if (search) {
+        await timeout(250);
```
hashicc, line 95: > We're technically adding a straight 250ms delay if there's ever a search. I tried to limit when it would happen by only doing it when there's actually a search parameter but if they searched then filtered down after, we would now be adding an unnecessary (albeit small) delay when filtering.

I have something that appears to work only when the search param changes. Ideally, you would be able to use `this.retrieveData?.lastCanceled?.args?[0].search` either directly in the task or in the model hook and not have to stash the previous `search` manually like I've done, but it was always `null`. A restartable task *should* cancel the previous task instance but I couldn't find it on the task. Even when I did a `this.retrieveData.cancelAll()` before performing the next task I didn't see a `lastCanceled` task instance. The other `task*` properties were either `null` or referencing the currently running task instance which makes sense but doesn't help.

```diff
diff --git a/ui/desktop/app/routes/scopes/scope/projects/targets/index.js b/ui/desktop/app/routes/scopes/scope/projects/targets/index.js
index d0811ba37..8ba41eeab 100644
--- a/ui/desktop/app/routes/scopes/scope/projects/targets/index.js
+++ b/ui/desktop/app/routes/scopes/scope/projects/targets/index.js
@@ -75,6 +75,8 @@ export default class ScopesScopeProjectsTargetsIndexRoute extends Route {
     if (!this.session.isAuthenticated) this.router.transitionTo('index');
   }
 
+  lastSearch = null;
+
   /**
    * Loads queried targets, the number of targets under current scope, and
    * projects for filtering options.
@@ -84,15 +86,28 @@ export default class ScopesScopeProjectsTargetsIndexRoute extends Route {
    * isCacheDaemonRunning: boolean, isLoadIncomplete: boolean, isCacheRefreshing: boolean}>}
    */
   async model(params) {
-    return this.retrieveData.perform(params);
+    const useDebounce =
+      this.lastSearch !== null && this.lastSearch !== params.search;
+    this.lastSearch = params.search;
+    return this.retrieveData.perform({ ...params, useDebounce });
   }
 
   retrieveData = restartableTask(
-    async ({ search, scopes, availableSessions, types, page, pageSize }) => {
-      // Only add a debounce timeout if there's a search field
-      // Not perfect if they search first then filter as it then adds extra time
-      if (search) {
-        await timeout(250);
+    async ({
+      useDebounce,
+      search,
+      scopes,
+      availableSessions,
+      types,
+      page,
+      pageSize,
+    }) => {
+      if (useDebounce) {
+        console.log('⏳ waiting because search is different');
+        await timeout(1000);
+        console.log('⏳⏳⏳⏳ waiting is done');
+      } else {
+        console.log('💨💨💨 not waiting!');
       }
 
       const isCacheRunningPromise = this.ipc.invoke('isCacheDaemonRunning');
```

ZedLi: Great call! I originally tried to see what the previous search was to check if it was different by accessing the `transition.from?.queryParams` but had noticed it wasn't always immediately up to date so I just kept it simple but didn't think to just keep track of the param manually as it works exactly as as it should.

hashicc: Awesome, glad that works!

😢 I'm still really bummed that `this.retrieveData.lastCanceled` is breaking my mental model since the [UI docs](https://ember-concurrency.com/docs/task-concurrency/#restartable) ([and tests](https://github.com/machty/ember-concurrency/blob/49c5d1d78fd9441878af9ba5c2b41d6e949aeba0/packages/test-app/tests/unit/task-test.js#L216)) clearly show that a previously running task is cancelled in a restartable task. It would be so much cleaner to just use that derived state than have to this manual bookkeeping.

hashicc (dismissed):
🚀 Nice fix! It worked well in my tests.

Between the options forward:
Between 1. (e-c) and 2. (add button, with enter), I defer to whatever folks think the best UX is. I think there's cases for both.

For 3., I am curious what a sharable pattern would look like. I like this PR as a step forward.

✨ Thanks for considering all the options and the deep dive!

ui/desktop/app/routes/scopes/scope/projects/targets/index.js
lisbet-alvarez: perhaps `this.retrieveData?.last?.args?.[0].search` could work because its getting the previous call to the `retrieveData` task? Also saw that `lastPerformed` is an option but to me looks the same as `last`.

lisbet-alvarez (approved):
This is a huge improvement for users!! Thank you for the detailed description & for this refactor! 🥳

cameronperera (approved):
Great work! Do we have an epic to track getting this changed in our other routes that can search/filter?

ui/desktop/app/routes/scopes/scope/projects/targets/index.js
ZedLi: Yeah I played around with it as well and used `lastPerformed` which worked for me, I assumed `last` would include any that were incomplete, errored, canceled, etc but I also am not sure why there's no canceled task 🤔

ZedLi:
> Great work! Do we have an epic to track getting this changed in our other routes that can search/filter?

I was going to include the admin UI ones in this PR as well since they're the same change across the board, would you prefer I separate them out?

cameronperera:
> > Great work! Do we have an epic to track getting this changed in our other routes that can search/filter?
> 
> I was going to include the admin UI ones in this PR as well since they're the same change across the board, would you prefer I separate them out?

However you planned to tackle this is fine with me. I wasn't sure if that was mentioned during our team meeting so I wanted to confirm. Only reason to split them up would be keeping the desktop CHANGELOG entry purely desktop related changes but that is a minor detail.

ui/desktop/app/routes/scopes/scope/projects/targets/index.js
lisbet-alvarez: Thanks for investigating, those are great finds! and `lastPerformed` does make more sense.

DhariniJeeva (approved):
Thanks for exploring different options and explaining it, this is a huge UX improvement!

cameronperera (approved):

ui/desktop/app/routes/scopes/scope/projects/targets/index.js
priya-patel04: I was looking into `lastSuccessful` and thinking if we should skip debounce if last search was same AND successful ?  Because with `lastPerformed` 
 debounce will be skipped even if last search failed due to some reason

priya-patel04 (approved):
Thanks for the detailed explanation and the fix. Left one non-blocking comment.

ui/desktop/app/routes/scopes/scope/projects/targets/index.js
ZedLi: If the search failed, we would still want the most recent call to know the last search which `lastPerformed` gives. If we used `lastSuccessful` and each subsequent task failed, we wouldn't get the subsequent updates and just the last one that succeeded. 

>  Because with lastPerformed debounce will be skipped even if last search failed due to some reason

Can you give an example of what are you referring to here? Since it seems like even if it failed, it should still get debounced

priya-patel04: Thanks for the explanation. And yes you are right even if it fails, it will be debounced using `lastPerformed`. Apologies for the confusion 😅

calcaide:
Sorry I am late for the review.

As discussed within tech-time I am fine with this approach, and do this for any model hook that needs it is understandable thanks for fixing it and putting it together.

Regarding the paths you mention, I like 2, but obviously that has a different UX and I understand the reasons why we will prefer go with this approach.
About approach 3, I might need to think more about it, I have encounter feelings of wrapping these in a component, as you mentioned at tech time feels going against the convention. Which I think, it can translate to more complexity.

Thanks for the work

PR #2724, opened 2025-03-12
chore: 🤖 Change to self hosted runners for test jobs

ZedLi:
# Description
Change the jobs to use self hosted runners for test jobs that have a time limit. See this [thread](https://hashicorp.slack.com/archives/C012GTH1C4E/p1741802158660339)


hashicc (dismissed):
🚀

calcaide (dismissed):
Thanks for taking care!!

calcaide:
None blocking comment: I do see the latest Validate Monorepo workflow was cancelled, do we have a successful run where it uses the desired runners?

.github/workflows/build-admin-ui.yaml
```diff
@@ -9,7 +9,7 @@ on:
 jobs:
   dependencies:
     name: Install dependencies
-    runs-on: ubuntu-latest
+    runs-on: ${{ fromJSON(vars.RUNNER) }}
```
cameronperera, line 12: For my learning sake: Where are we defining `vars.RUNNER`

ZedLi: It's in the repo settings
image

DhariniJeeva: this is neat, I didn't know you could use vars and switch runners based on repos

DhariniJeeva (approved):
looks good!

lisbet-alvarez (approved):
Nice work!!

.github/workflows/build-admin-ui.yaml
calcaide: I think I might be missing repo permissions, I am not seeing the screenshot you posted from repo settings, are we talking [about these settings](https://github.com/hashicorp/boundary-ui/settings)?

ZedLi: Yes, you'll need admin access which I don't think you have, I just request it for a few hours or a day to change these

calcaide (approved):
Thanks for the work, I left a none blocing comment/question 🙌

PR #2731, opened 2025-03-14
chore: 🤖 Add chrome for self hosted runners when running tests

ZedLi:
# Description
Our self hosted runner [images](https://github.com/hashicorp/crt-github-runners/releases/tag/ubuntu-22.04-x86_64-78abdae-13654489443) don't include chrome by default so we'll need to manually include it for our tests.

### DIFF
--- a/.github/workflows/monorepo-validate.yaml
+++ b/.github/workflows/monorepo-validate.yaml
@@ -62,6 +62,9 @@ jobs:
     runs-on: ${{ fromJSON(vars.RUNNER) }}
     timeout-minutes: 15
     steps:
+      - uses: browser-actions/setup-chrome@c785b87e244131f27c9f19c1a33e2ead956ab7ce # v1.7.3
+        with:
+          install-dependencies: true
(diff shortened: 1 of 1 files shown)


.github/workflows/monorepo-validate.yaml
```diff
@@ -62,6 +62,9 @@ jobs:
     runs-on: ${{ fromJSON(vars.RUNNER) }}
     timeout-minutes: 15
     steps:
+      - uses: browser-actions/setup-chrome@c785b87e244131f27c9f19c1a33e2ead956ab7ce # v1.7.3
+        with:
+          install-dependencies: true
```
ZedLi, line 67: Technically I don't think we need this as it looks like they [include dependencies](https://github.com/hashicorp/crt-github-runners/blob/16d07c4353b8e1c2e87e9843e5696149c96d399e/packer/scripts/installers/chrome-dep.sh) for us but it doesn't hurt I imagine

cameronperera (approved):

hashicc (approved):
🏄‍♂️

calcaide (approved):
Let's gooo!!!

PR #2727, opened 2025-03-13
chore: 🤖 update session list tests with selectors

cameronperera:
# Description

I have a few `tbody tr` selectors with TODOs for when this [PR](https://github.com/hashicorp/boundary-ui/pull/2719/files) is merged in. I will keep this PR as a draft until that is ready.

:tickets: [Jira ticket](https://hashicorp.atlassian.net/browse/[ICU-16426)

## Checklist

- [ ] I have added before and after screenshots for UI changes
- [ ] I have added JSON response output for API changes
- [ ] I have added steps to reproduce and test for bug fixes in the description
- [ ] I have commented on my code, particularly in hard-to-understand areas
- [ ] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works

PR #2733, opened 2025-03-14
Small test selector updates

cameronperera:
# Description

These are 2 small changes that I mentioned in this [PR](https://github.com/hashicorp/boundary-ui/pull/2727).
Adds a more specific name to the original `SIDEBAR_NAV_LINK` selector and moves `TABLE_FIRST_ROW_ACTION_DROPDOWN` into the common selectors.

## Checklist

### DIFF
--- a/ui/admin/tests/acceptance/auth-methods/selectors.js
+++ b/ui/admin/tests/acceptance/auth-methods/selectors.js
@@ -102,8 +102,6 @@ export const MANAGE_DROPDOWN_MAKE_PRIMARY =
 export const MANAGE_DROPDOWN_DELETE =
   '[data-test-manage-auth-method] ul li:last-child button';
 
-export const TABLE_FIRST_ROW_ACTION_DROPDOWN =
-  'tbody tr:first-child td:last-child button';
 export const TABLE_FIRST_ROW_ACTION_FIRST_ITEM =
   'tbody tr:first-child td:last-child ul li:first-child button';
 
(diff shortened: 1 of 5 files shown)


ui/admin/tests/acceptance/auth-methods/selectors.js
```diff
@@ -102,8 +102,6 @@ export const MANAGE_DROPDOWN_MAKE_PRIMARY =
 export const MANAGE_DROPDOWN_DELETE =
   '[data-test-manage-auth-method] ul li:last-child button';
 
-export const TABLE_FIRST_ROW_ACTION_DROPDOWN =
-  'tbody tr:first-child td:last-child button';
 export const TABLE_FIRST_ROW_ACTION_FIRST_ITEM =
```
cameronperera, line 105: You might wonder why I didn't move `TABLE_FIRST_ROW_ACTION_FIRST_ITEM` either since it has the same name and functionality as the selector in storage-buckets?
```
export const TABLE_FIRST_ROW_ACTION_FIRST_ITEM =
  'tbody tr:first-child td:last-child ul li:first-child a';
```

This one is using a `button` while the one in storage-buckets is using an `a` tag. We may need to refactor that dropdown item but potentially not either. I decided to not include that in this PR to prevent scope creep and introduce changes outside of tests.

calcaide: Ahhh that's a great catch!! and a good solution 😉 +1 to refactor that, also totally fine with a separate PR for it. Great work 🙌

calcaide (approved):
Great work!! thanks for the refactor 🙌

priya-patel04 (approved):
Lgtm!

PR #2735, opened 2025-03-14
Fix Searchbar debouncing in admin UI

ZedLi:
# Description
This is a follow up to https://github.com/hashicorp/boundary-ui/pull/2723 to also use tasks in admin UI to fix the search field debounce issue.

## How to Test
Try typing in the search field for various resources at different speeds.

## Checklist

### DIFF
--- a/ui/admin/app/controllers/scopes/scope/aliases/index.js
+++ b/ui/admin/app/controllers/scopes/scope/aliases/index.js
@@ -8,7 +8,6 @@ import { service } from '@ember/service';
 import { action } from '@ember/object';
 import { tracked } from '@glimmer/tracking';
 import { loading } from 'ember-loading';
-import { debounce } from 'core/decorators/debounce';
 import { notifySuccess, notifyError } from 'core/decorators/notify';
 import { confirm } from 'core/decorators/confirm';
 
@@ -120,7 +119,6 @@ export default class ScopesScopeAliasesIndexController extends Controller {
    * @param {object} event
    */
   @action
-  @debounce(250)
   handleSearchInput(event) {
     const { value } = event.target;
     this.search = value;
--- a/ui/admin/app/controllers/scopes/scope/groups/index.js
+++ b/ui/admin/app/controllers/scopes/scope/groups/index.js
@@ -10,7 +10,6 @@ import { action } from '@ember/object';
 import { loading } from 'ember-loading';
 import { confirm } from 'core/decorators/confirm';
 import { notifySuccess, notifyError } from 'core/decorators/notify';
-import { debounce } from 'core/decorators/debounce';
 
 export default class ScopesScopeGroupsIndexController extends Controller {
   // =services
@@ -23,7 +22,7 @@ export default class ScopesScopeGroupsIndexController extends Controller {
 
   queryParams = ['search', 'page', 'pageSize'];
 
-  @tracked search = '';
+  @tracked search;
   @tracked page = 1;
   @tracked pageSize = 10;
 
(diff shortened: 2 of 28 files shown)


ui/admin/app/routes/scopes/scope/roles/role/scopes.js
```diff
@@ -40,49 +41,61 @@ export default class ScopesScopeRolesRoleScopesRoute extends Route {
...
 
-    let allGrantScopes = [];
-    let totalItems = grantScopes.length;
-    if (filters.id.length) {
-      parentScopes.forEach((id) => filters.scope_id.push({ equals: id }));
-      types.forEach((type) => filters.type.push({ equals: type }));
+      let grantScopes = [];
+      if (!parentScopes.length && !types.length) {
+        role.grantScopeKeywords.forEach((keyword) => {
+          if (keyword.includes(search) || !search) {
```
ZedLi, line 71: There was a slight change in logic here, I added the `|| !search` because I was surprised to see that changing the initial state of `search` from empty string to undefined broke tests.

It turns out we were depending on `keyword.includes(search)` when search was `''` to include all the keywords by default which seemed unexpected so I made this more explicit.

ui/admin/app/controllers/scopes/scope/groups/index.js
```diff
@@ -23,7 +22,7 @@ export default class ScopesScopeGroupsIndexController extends Controller {
 
   queryParams = ['search', 'page', 'pageSize'];
 
-  @tracked search = '';
```
ZedLi, line 26: I made all the initial `search` states to `undefined` instead of empty string because it shouldn't affect our initial search logic and to make sure that we don't add an extra debounce on an initial load as `this.retrieveData?.lastPerformed?.args?.[0].search !== params.search;` could return true as the param would be an empty string while the last performed task would be `undefined`.

I didn't think it was necessary to make the check against all falsy values such as `null`, `undefined`, `''` but let me know if you think otherwise.

cameronperera (approved):
LGTM

ui/admin/app/controllers/scopes/scope/aliases/index.js
```diff
@@ -8,7 +8,6 @@ import { service } from '@ember/service';
 import { action } from '@ember/object';
 import { tracked } from '@glimmer/tracking';
 import { loading } from 'ember-loading';
-import { debounce } from 'core/decorators/debounce';
```
calcaide, line 11: None blocking comment:
Out of the PR context but I notice that after this fix, the only place where we still using the debounce decorator is on the dropdown component within core, [here code](https://github.com/hashicorp/boundary-ui/blob/main/addons/core/addon/components/dropdown/index.js#L42), should we create a Jira task to try refactor that one with ember-concurrency as well? 
I think if we do that, we can get rid of the debounce decorator. What are your thoughts?

calcaide (approved):
Thank you for this amazing Fix/refactor 🙌 this is a UX win for real!

ui/admin/app/controllers/scopes/scope/aliases/index.js
ZedLi: I had the same question but adding a task here didn't look like it would really get us much since that one was just dealing with filtering out the dropdown all in memory (and tests seemed fine). I could just change it to a task tho if we feel like we should get rid of the decorator and we make the decision to always use them from now on in these scenarios. 

I'm fine with either way though, not sure what others think.

hashicc (approved):
💯 Great to see the pattern hold across the other use cases

DhariniJeeva (approved):
looks great, thanks for the fix! 🎉

priya-patel04 (approved):
Thanks for the refactor! 🚀

lisbet-alvarez (approved):
Wow thank you for this huge refactor!! 🥳

PR #2732, opened 2025-03-14
Refactor: Policy tests - create, read and update

priya-patel04:
# Description
Test selector refactor for Policy tests:
[create](https://hashicorp.atlassian.net/browse/ICU-16400)
[update](https://hashicorp.atlassian.net/browse/ICU-16402)
[read](https://hashicorp.atlassian.net/browse/ICU-16401)

## Checklist

- [ ] I have added before and after screenshots for UI changes
- [ ] I have added JSON response output for API changes
- [ ] I have added steps to reproduce and test for bug fixes in the description
- [ ] I have commented on my code, particularly in hard-to-understand areas
- [ ] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works

PR #2736, opened 2025-03-17
Refactor: Policy tests - list and delete

priya-patel04:
# Description
Test selector refactor for Policy tests:
[list](https://hashicorp.atlassian.net/browse/ICU-16403)
[delete](https://hashicorp.atlassian.net/browse/ICU-16404)

## Checklist

### DIFF
--- a/ui/admin/tests/acceptance/policy/list-test.js
+++ b/ui/admin/tests/acceptance/policy/list-test.js
@@ -4,7 +4,7 @@
  */
 
 import { module, test } from 'qunit';
-import { visit, click, currentURL, find } from '@ember/test-helpers';
+import { visit, click, currentURL } from '@ember/test-helpers';
 import { setupApplicationTest } from 'admin/tests/helpers';
 import setupMirage from 'ember-cli-mirage/test-support/setup-mirage';
 import { setupIndexedDb } from 'api/test-support/helpers/indexed-db';
@@ -18,10 +18,6 @@ module('Acceptance | policies | list', function (hooks) {
 
   let featuresService;
   let intl;
-
-  const STORAGE_POLICY_TITLE = 'Storage Policies';
-  const DROPDOWN_BUTTON_SELECTOR = '.hds-dropdown-toggle-icon';
-  const DROPDOWN_ITEM_SELECTOR = '.hds-dropdown-list-item a';
 
   const instances = {
     scopes: {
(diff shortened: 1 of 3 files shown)


DhariniJeeva (approved):
looks good!

ui/admin/tests/acceptance/policy/list-test.js
```diff
@@ -127,13 +123,13 @@ module('Acceptance | policies | list', function (hooks) {
...
-    assert.dom(`[href="${urls.policies}"]`).exists();
+    assert.dom(commonSelectors.HREF(urls.policies)).exists();
 
-    await click(`[href="${urls.policies}"]`);
+    await click(commonSelectors.HREF(urls.policies));
 
     assert.dom(commonSelectors.PAGE_MESSAGE_DESCRIPTION).hasText(
       intl.t('descriptions.create-but-not-list', {
-        resource: STORAGE_POLICY_TITLE,
+        resource: 'Storage Policies',
```
DhariniJeeva, line 132: non blocker- should we move this to the customer selector that you created [here](https://github.com/hashicorp/boundary-ui/pull/2732/files)

laurenolivia (approved):
🎀

ui/admin/tests/acceptance/policy/list-test.js
cameronperera: Just want to share my thought on this. It looks like this string is only used twice in this file and potentially only in this file. I would lean towards not creating a selector for it. If it was used more often and in other files, for sure create a selector. This definitely falls in the gray (or grey) area. Either direction I am fine with but thought I would share my reasoning if I was doing this work.

ui/admin/tests/acceptance/policy/list-test.js
```diff
@@ -185,12 +181,16 @@ module('Acceptance | policies | list', function (hooks) {
...
-    await click(DROPDOWN_BUTTON_SELECTOR);
+    await click(commonSelectors.HREF(urls.policies));
+    await click(commonSelectors.TABLE_FIRST_ROW_ACTION_DROPDOWN);
+
+    assert
+      .dom(commonSelectors.TABLE_FIRST_ROW_ACTION_DROPDOWN_ITEM_LINK)
+      .exists()
+      .hasText('Edit');
+
+    await click(commonSelectors.TABLE_FIRST_ROW_ACTION_DROPDOWN_ITEM_LINK);
```
cameronperera, line 192: ❤️ This reads so much better now. I can actually tell what we are acting on and asserting against.

ui/admin/tests/acceptance/policy/list-test.js
priya-patel04: Yeah I agree with @cameronperera and it make sense to not create a selector for this string.

calcaide (approved):
Awesome work!! thanks for the refactor 🙌

PR #2739, opened 2025-03-20
Upgrade electron forge from 7.5.0 to 7.7.0

calcaide:
# Description

Upgrade electron-forge to latest in preparation for electron upgrade.

[Pipeline test run](https://github.com/hashicorp/boundary-desktop-releases/actions/runs/13975391991)

PR #2740, opened 2025-03-20
Upgrade ember-electron from 4.2.2 to 6.0.0

calcaide:
# Description

Upgrade ember-electron from `4.2.2` to `6.0.0`. In preparation for electron upgrade

[As per changelog](https://github.com/adopted-ember-addons/ember-electron/releases), it is safe to upgrade. 

:tickets: [Jira ticket](https://hashicorp.atlassian.net/browse/ICU-15805)

**UPDATE:** We change the goal from `ember-electron` v7 to v6. After reproducing some issues and debugging, `ember-electron` v7 does not support ember < 4.8 and we are using ember 4.12. So for pragmatism we decided to upgrade to 6. Once the ember upgrade is complete, we will review this dependency.

## How to Test

- [Pipeline test run](https://github.com/hashicorp/boundary-desktop-releases/actions/runs/13976693141).
- Run Desktop e2e suite.

### DIFF
--- a/.husky/prepare-commit-msg
+++ b/.husky/prepare-commit-msg
@@ -21,4 +21,3 @@ else
   # This is to allow interaction with the terminal for commitizen
   exec < /dev/tty && npx git-cz --hook || true
 fi
-
--- a/package.json
+++ b/package.json
@@ -39,7 +39,6 @@
   },
   "devDependencies": {
     "concurrently": "^9.1.0",
-    "cz-conventional-changelog": "^3.3.0",
     "doctoc": "^2.2.0",
     "git-cz": "^4.9.0",
     "husky": "^9.1.6",
--- a/ui/admin/app/components/form/role/index.hbs
+++ b/ui/admin/app/components/form/role/index.hbs
@@ -26,7 +26,7 @@
     {{#if @model.errors.name}}
       <F.Error as |E|>
         {{#each @model.errors.name as |error|}}
-          <E.Message>{{error.message}}</E.Message>
+          <E.Message data-test-error-message-name>{{error.message}}</E.Message>
         {{/each}}
       </F.Error>
     {{/if}}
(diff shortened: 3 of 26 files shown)


hashicc (dismissed):
Looking at the `ember-electron` changelog I agree it looks safe. Also I think the other `@electron-forge/*` have already been merged [here](https://github.com/hashicorp/boundary-ui/pull/2739), so this PR is really just `ember-electron`.

ZedLi (approved):
> After reproducing some issues and debugging

What were the issues? Considering we were already on 4.12, I'm curious what the problems were

calcaide:
@ZedLi 

My apologie, rather than answering your comment I click "Edit" and answer within your own comment.

I corrected that and now a propper answer:

The test finishes but not the process that runs it. In local test, test process keeps running infinetely, until you manually stops it. In CI it times out. I didn't digg that much to indentify the cause of the issue tho.

My findings are:
- We use a testem-electron file that loads the test runner from `ember-electron`. [Here file](https://github.com/hashicorp/boundary-ui/blob/main/ui/desktop/testem-electron.js)
- We use a test wrapper within electron-app/ that call helpers [from test-support](https://github.com/hashicorp/boundary-ui/blob/main/ui/desktop/electron-app/tests/index.js#L16).
- Initially I though the issue must be related to these 2 files mentioned above, checked ember-electron changes on those between v6 and v7 and found no differences, so I start thinking it can be related to 3rd party deps those processes uses such as `electron-devtools-installer`. I digg a bit on those but didn't find a good breadcrumb to follow.
- I reproduced the same behaviour on Mac and Linux, so thinking its not OS/env related.
- I also find out in v7 they dropped support for < ember 4.8. After this I thought best to keep the ember upgrade and other deps rolling, and attempt ember-electron upgrade after Ember + Electron upgrade.

Open to discuss this further if need.

DhariniJeeva (approved):
looks good! thanks for the upgrade!

cameronperera (approved):

PR #2744, opened 2025-03-24
Refactor: Scopes tests

priya-patel04:
# Description
Scopes tests refactor:
[add-storage-policy](https://hashicorp.atlassian.net/browse/ICU-16422)
[list](https://hashicorp.atlassian.net/browse/ICU-16421)
[update](https://hashicorp.atlassian.net/browse/ICU-16419)
[delete](https://hashicorp.atlassian.net/browse/ICU-16420)
[read](https://hashicorp.atlassian.net/browse/ICU-16415)

## Checklist


ui/admin/tests/acceptance/scopes/selectors.js
```diff
@@ -0,0 +1,20 @@
+/**
+ * Copyright (c) HashiCorp, Inc.
+ * SPDX-License-Identifier: BUSL-1.1
+ */
+
+export const SCOPE_FORM = 'main .rose-form';
```
calcaide, line 6: None blocking comment: I will suggest to use the html attribute `form` or `main form` as selector for this case.

My thinking is next:
- There is only 1 form visible, so its safe to use the `form` selector.
- Using `.rose-form` we are coupling the selector to a rose element class. Isn't a big deal, but if possible is best to avoid.

Also I would like to mention, If you think having a selector for the form element is useful in other domains we can even add this FORM selector to the commonSelectors file. I did not see a form element selector on the tests I refactor but wondering if others use it 😉 

Let me know your thoughts 🙌

ui/admin/tests/acceptance/scopes/selectors.js
```diff
@@ -0,0 +1,20 @@
...
+export const FIELD_STORAGE_POLICY_DROPDOWN = '[name=policy_id]';
```
calcaide, line 7: None blocking comment: I think, since this is a form field [using a select](https://helios.hashicorp.design/components/form/select), [rather than a dropdown](https://helios.hashicorp.design/components/dropdown) component, using `_SELECT` rather than `_DROPDOWN` could be more appropriate.

What are your thoughts?

ui/admin/tests/acceptance/scopes/read-test.js
```diff
@@ -79,40 +74,40 @@ module('Acceptance | scopes | read', function (hooks) {
...
+    await click(commonSelectors.HREF(urls.globalScopeEdit));
     await a11yAudit();
 
     assert.strictEqual(currentURL(), urls.globalScopeEdit);
-    assert.dom(FORM_SELECTOR).exists();
-    assert.dom(MANAGE_DROPDOWN_SELECTOR).doesNotExist();
-    assert.dom(DELETE_ACTION_SELECTOR).doesNotExist();
+    assert.dom(selectors.SCOPE_FORM).exists();
+    assert.dom(selectors.MANAGE_PROJECTS_DROPDOWN).doesNotExist();
+    assert.dom(selectors.MANAGE_PROJECTS_DROPDOWN_DELETE).doesNotExist();
```
calcaide, line 83: None blocking comment: I know it was like this before and I am not saying needs to be deleted, but it feels to me redundant that we check `MANAGE_PROJECTS_DROPDOWN` and `MANAGE_PROJECTS_DROPDOWN_DELTE` not existing, I think should be sufficient with the `MANAGE_PROJECTS_DROPDOWN`.

ui/admin/tests/helpers/selectors.js
```diff
@@ -62,3 +62,5 @@ export const FILTER_DROPDOWN = (filterName) =>
 export const FILTER_DROPDOWN_ITEM = (itemList) => `input[value="${itemList}"]`;
 export const FILTER_DROPDOWN_ITEM_APPLY_BTN = (filterName) =>
   `.search-filtering [name="${filterName}"] .hds-dropdown__footer button`;
+
+export const LINK_LIST_ITEM_TEXT = '.link-list-item__text';
```
calcaide, line 66: None blocking question: Do you think this selector could be use in other test files?

calcaide (approved):
Amazing work!! Very happy to see despite this refactor has slightly different pattern than other domains you solved with ease and style 💃 
I left some none blocking comments, thanks for the refactor!! 🙌

ui/admin/tests/helpers/selectors.js
priya-patel04: Yes, it can be used for `targets` test files

ui/admin/tests/acceptance/scopes/read-test.js
priya-patel04: Yes agreed, no need for redundant check. Thanks for catching it. I will remove it.

ui/admin/tests/acceptance/scopes/selectors.js
priya-patel04: Sure, I can use html attribute instead. And I don't see this form element selector used in other tests as of now, so not moving it to the commonSelectors file

ui/admin/tests/acceptance/scopes/add-storage-policy-test.js
```diff
@@ -98,36 +94,43 @@ module('Acceptance | scope | add storage policy', function (hooks) {
...
+    assert.dom(selectors.FIELD_STORAGE_POLICY_SELECT).hasValue();
 
+    await click(commonSelectors.SAVE_BTN);
+
+    assert.strictEqual(currentURL(), urls.orgScopeEdit);
     assert.strictEqual(
-      find(LINK_LIST_SELECTOR_ITEM_TEXT).textContent.trim(),
+      find(commonSelectors.LINK_LIST_ITEM_TEXT).textContent.trim(),
       policyTwo.name,
     );
```
cameronperera, line 135: Can we update this assert as well to use `assert.dom().hasText()`?

ui/admin/tests/acceptance/scopes/add-storage-policy-test.js
```diff
@@ -139,17 +142,18 @@ module('Acceptance | scope | add storage policy', function (hooks) {
...
 
-    await click(CANCEL_BTN_SELECTOR);
-    assert.strictEqual(currentURL(), urls.orgScopeEdit);
+    await click(commonSelectors.CANCEL_BTN);
 
+    assert.strictEqual(currentURL(), urls.orgScopeEdit);
     assert.strictEqual(
-      find(LINK_LIST_SELECTOR_ITEM_TEXT).textContent.trim(),
+      find(commonSelectors.LINK_LIST_ITEM_TEXT).textContent.trim(),
       policyOne.name,
```
cameronperera, line 157: Same here for the assert.

ui/admin/tests/acceptance/scopes/list-test.js
```diff
@@ -61,60 +58,97 @@ module('Acceptance | scopes | list', function (hooks) {
   test('user can search for a specifc org scope by id', async function (assert) {
     await visit(urls.globalScope);
 
-    assert.dom(SCOPE_LINK_SELECTOR(instances.scopes.org1.id)).exists();
-    assert.dom(SCOPE_LINK_SELECTOR(instances.scopes.org2.id)).exists();
+    assert
+      .dom(selectors.TABLE_ROW_SCOPE_LINK(instances.scopes.org1.id))
+      .exists();
```
cameronperera, line 63: Could we update the `.exists()` to `.isVisible()`?

cameronperera (changes requested):
Looks. Just a few comments. For the `isVisible` assert comment, that is for all the `exists` asserts.

ui/admin/tests/acceptance/scopes/list-test.js
priya-patel04: Yes, updated!

ui/admin/tests/helpers/selectors.js
calcaide: Awesome!! thank you for the addition 🙌

ui/admin/tests/acceptance/scopes/selectors.js
```diff
@@ -0,0 +1,20 @@
+/**
+ * Copyright (c) HashiCorp, Inc.
+ * SPDX-License-Identifier: BUSL-1.1
+ */
+
+export const SCOPE_FORM = 'main form';
```
calcaide, line 6: None blocking question: I wonder if we are using this type of selector, selecting the form, in other places. If so, this is a very good candidate to be part of commonSelectors. Good catch! I will keep it in my mind when reviewing other test selector refactors 😉

ui/admin/tests/acceptance/scopes/read-test.js
calcaide: Thank you 😃

calcaide (dismissed):
Re-approving, thank you for the work 🙌

ui/admin/tests/acceptance/scopes/selectors.js
priya-patel04: Ahh looks like it is being used in this [PR](https://github.com/hashicorp/boundary-ui/pull/2749) too. I can move it to the commonSelector file.

cameronperera: I will keep my selector in place until this PR is merged.

cameronperera (approved):

PR #2745, opened 2025-03-25
chore: 🤖 upgrade ember-data to v5.3.11

lisbet-alvarez:
:tickets: [Jira ticket](https://hashicorp.atlassian.net/browse/ICU-16544)

# Description

* Use of `.getSchemaDefinitionService()`is deprecated and replaced with `schema`
Pasted Graphic 3
* Use of `.attributesDefinitionFor()` is deprecated and replaced with `fields` method. As a result we get back the attribute definitions as a Map type instead of an object type.
Pasted Graphic 4
* Bolean, Date, Number, and String transforms are no longer automatically installed for consumers, we have to specify those instead. Note that warnings still appear in **tests only** for the **api addon** due to the ember-data exports taking precedence in the dummy app but we can ignore these since we are correctly re exporting the correct transforms, and stores. Additionally these warnings won't appear anymore in v6.0 since ember-data will officially remove their exports and ours will take precedence. Also shoutout to @hashicc for helping debug this and figuring out that the ember-data transforms were taking precedence instead of our exports 🙌 
Pasted Graphic 5
* Importing `Store` from `ember-data/store` is valid in order to import default configuration. Ember eslint was updated in v12.2.1 to allow this usage. Therefore, once we update eslint we can remove the line that disables the rule.
* Attempting to call createRecord with an id that was already used in `findRecord` triggers the error below due to [requirement of unique ID](https://deprecations.emberjs.com/id/ember-data-deprecate-non-uniq-collected-payloads). We were using createRecord to make a temporary record that could be used when retrieving the specified scope record failed due to unauthenticated or permission denied errors.
Pasted Graphic 7
* Blueprints for custom transforms were updated so also updated our transforms to adhere to the [new syntax](https://api.emberjs.com/ember-data/release/classes/transform)
* Added extra config in admin and desktop ember-cli-build.js files to opt into new behavior in order to silence deprecation warnings for `DEPRECATE_STORE_EXTENDS_EMBER_OBJECT`
* Updated ember-auto-import since it was causing build errors due to resolution conflicts.
* Ember-data now lists ember/test-waiters as a peer dependency so that was added to devDependencies in api add-on.
* @warp-drive/build-config was added as a devDependency for admin and desktop apps since its used to import `setConfig`. Due to the `ember-data` key in `ember-cli-build.js` being deprecated
image

## How to Test

Make sure admin & desktop UI build locally.
Ran admin e2e tests:
image
image
[boundary-desktop-releases test run](https://github.com/hashicorp/boundary-desktop-releases/actions/runs/14184042154)

## Checklist


addons/api/app/transforms/number.js
```diff
@@ -0,0 +1 @@
+export { NumberTransform as default } from '@ember-data/serializer/transform';
```
lisbet-alvarez, line 6: The boolean, date, number and string transforms are no longer auto exported for us. So we have to explicitly export.

addons/api/package.json
```diff
@@ -23,18 +23,19 @@
...
     "ember-cli-htmlbars": "^6.2.0",
     "ember-cli-mirage": "^3.0.3",
-    "ember-data": "~4.12.0",
+    "ember-data": "~5.3.12",
     "miragejs": "^0.1.48",
     "tracked-built-ins": "^3.1.1"
   },
   "devDependencies": {
-    "@babel/core": "^7.25.2",
+    "@babel/core": "^7.26.10",
```
lisbet-alvarez, line 38: This is not related to ember-data upgrade, just noticed it was already resolving to 7.26.10.

addons/api/package.json
```diff
@@ -23,18 +23,19 @@
     "doc:toc": "doctoc README.md"
   },
   "dependencies": {
+    "@ember/test-waiters": "^4.1.0",
     "dexie": "^3.2.4",
-    "ember-auto-import": "^2.8.1",
+    "ember-auto-import": "^2.10.0",
```
lisbet-alvarez, line 28: ember-auto-import needs to be updated or else causes console error.

addons/api/package.json
```diff
@@ -57,7 +58,7 @@
     "ember-page-title": "^7.0.0",
     "ember-qunit": "^8.1.0",
     "ember-resolver": "^10.0.0",
-    "ember-source": "~4.12.0",
+    "ember-source": "~5.12.0",
```
lisbet-alvarez, line 61: We will need to clear peer dep. warnings (as a follow-up item) due to this update.

addons/api/addon/services/store.js
```diff
@@ -3,13 +3,36 @@
  * SPDX-License-Identifier: BUSL-1.1
  */
 
-import Store, { CacheHandler } from '@ember-data/store';
+// Additional configuration is needed when importing from @ember-data/store.
```
calcaide, line 6: Do you mind poiting to a deprecaiton guide or a link to where explains a bit more why these changes?
Nothing wrong with them but I want to understand better why now we need to add methods like `createCache`, `createSchemaService` and such.

As per my understanding of [the eslint rule](https://github.com/ember-cli/eslint-plugin-ember/blob/master/docs/rules/use-ember-data-rfc-395-imports.md) what changes are the imports, I am missing why we need to set defaults

ui/admin/app/routes/scopes/scope.js
```diff
@@ -29,14 +28,7 @@ export default class ScopesScopeRoute extends Route {
...
     // from ID because global has a fixed ID.
-    const type = id === 'global' ? TYPE_SCOPE_GLOBAL : TYPE_SCOPE_ORG;
-    return this.store.findRecord('scope', id).catch(() => {
-      const maybeExistingScope = this.store.peekRecord('scope', id);
-      const scopeOptions = { id, type };
-      return (
-        maybeExistingScope || this.store.createRecord('scope', scopeOptions)
-      );
-    });
+    return this.store.findRecord('scope', id);
```
calcaide, line 31: None blocking comment: The previous implementation returned a "maybeExistingScope" or it creates a new record for scope. I do understand this changes validates/assumes the scope record its ALWAYS present?

ui/admin/ember-cli-build.js
```diff
@@ -36,10 +38,16 @@ module.exports = function (defaults) {
         },
       },
     },
-    // TODO: Update to 4.12 when deprecations are resolved
-    //   as multiple things break when forcing compatibility on > 4.6
-    emberData: {
-      compatWith: '4.6',
+  });
+
+  // TODO: This is config can be removed in ember-data 6.0.
```
calcaide, line 43: None blocking comment: should we delete the "is"? so This config can be removed....

ui/desktop/app/services/store.js
```diff
@@ -3,7 +3,10 @@
  * SPDX-License-Identifier: BUSL-1.1
  */
 
-import Store, { CacheHandler } from '@ember-data/store';
+// This an example of using Store from ember-data/store to get default configuration automatically.
+// eslint-disable-next-line ember/use-ember-data-rfc-395-imports
+import Store from 'ember-data/store';
```
calcaide, line 8: Comparing with the `addons/api/addon/services/store` file, I see this one is importing store from `ember-data/store` and the other is `@ember-data/store`.

My understanding is that when importing from @ember-data we need to set defaults per ourselves?

I also left a comment/question within the `api/addon` store file.

calcaide (approved):
Looking veeery good!!!

I left comments and questions to understand better some changes, amazing work and digging into ember-data internals 🙌 💪

addons/api/addon/serializers/application.js
```diff
@@ -244,12 +244,10 @@ export default class ApplicationSerializer extends RESTSerializer {
    * @return {object}
    */
   normalizeMissingArrays(store, primaryModelClass, payload) {
-    const attrDefs = store
-      .getSchemaDefinitionService()
-      .attributesDefinitionFor({ type: primaryModelClass.modelName });
+    const attrDefs = store.schema.fields({ type: primaryModelClass.modelName });
```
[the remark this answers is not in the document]
lisbet-alvarez: Great question! The deprecation warnings actually were very detailed for this usage. Below i attached the warning for `store.getSchemaDefinitionService()` and warning for `attributesDefinitionFor`

Pasted Graphic 3

Pasted Graphic 4

addons/api/addon/services/store.js
lisbet-alvarez: Yup great questions! For more context, this ember-data upgrade changes what `store` configuration is provided by default when importing the `Store` class from `@ember-data/store`. [This doc](https://api.emberjs.com/ember-data/5.3/modules/@ember-data%2Fjson-api#:~:text=Note%20When%20using%20ember%2Ddata%20the%20below%20configuration%20is%20handled%20for%20you%20automatically.) is updated to show what configuration hooks need to be called by the consumer when creating a customizing the store service. That doc also mentions that to get the default configuration we can import `Store` from `ember-data` instead. This other deprecation guide also [mentions it here](https://deprecations.emberjs.com/id/ember-data-deprecate-legacy-imports#:~:text=you%20should%20import%20from%20ember%2Ddata/store%20instead%20of%20%40ember%2Ddata/store%20in%20order%20to%20receive%20the%20appropriate%20configuration%20of%20defaults.)
I also used the [Store implementation details](https://github.com/emberjs/data/blob/main/packages/-ember-data/src/store.ts) as reference for what the default config is.

ui/admin/ember-cli-build.js
lisbet-alvarez: oh yup great catch!

ui/admin/ember-cli-build.js
```diff
@@ -36,10 +38,16 @@ module.exports = function (defaults) {
...
+  // This silences ember-data deprecate warnings by setting to false to
+  // strip the deprecated code (thereby opting into the new behavior).
+  setConfig(app, __dirname, {
+    //compatWith: '5.3',
```
lisbet-alvarez, line 47: Before we had this `compatWith` field set to `4.6` (in admin ui only). Im wondering if we want to continue doing the same for `5.3`?

addons/api/addon/services/store.js
lisbet-alvarez: Resolving this since it is recommended to import from `ember-data/store` instead for default config

ui/desktop/app/services/store.js
```diff
@@ -3,7 +3,11 @@
  * SPDX-License-Identifier: BUSL-1.1
  */
 
-import Store, { CacheHandler } from '@ember-data/store';
+// Importing 'Store' from 'ember-data/store' to get default configuration automatically.
+// TODO: Remove eslint-disable after upgrading to eslint-plugin-ember >= v12.2.1
```
calcaide, line 7: none blocking note: when we perform the ember upgrade within desktop, we will update `eslint-plugin-ember` to latest and came back here for the fix 😉

calcaide (dismissed):
AMAZING work!! tested manually and I couldn't see any depreaction messages 🥳

ui/desktop/app/services/store.js
lisbet-alvarez: awesome! that works out then!

ui/admin/ember-cli-build.js
DhariniJeeva: I'm leaning towards removing it since we explicitly call out the deprecation warnings on L49 and L50.  I don't see a need for `compatWith` at this point.

DhariniJeeva (approved):
works great locally, didn't notice any deprecation warning! thanks for the upgrade ✨

ui/admin/ember-cli-build.js
lisbet-alvarez: good point! thank you! ill remove it!

calcaide (approved):

ZedLi (approved):
Great work!!

addons/api/addon/services/store.js
```diff
@@ -3,7 +3,11 @@
  * SPDX-License-Identifier: BUSL-1.1
  */
 
-import Store, { CacheHandler } from '@ember-data/store';
+// Importing 'Store' from 'ember-data/store' to get default configuration automatically.
```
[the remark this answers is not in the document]
lisbet-alvarez: Great question, default config includes the setup of hooks such as `createCache`, `createSchemaService`, `teardownRecord`, basically what is defined in [this file](https://github.com/emberjs/data/blob/main/packages/-ember-data/src/store.ts). This [commit here](https://github.com/hashicorp/boundary-ui/pull/2745/commits/22b0b8930be940b749bd67a2e206e68ebc0b9023) is an example though that was before  knowing that it's recommended to import form `ember-data/store`.

addons/api/addon/transforms/array.js
```diff
@@ -16,4 +15,8 @@ export default class ArrayTransform extends Transform {
     const arr = deserialized || [];
     return arr;
   }
+
+  static create() {
```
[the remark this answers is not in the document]
lisbet-alvarez: The previous format didn't cause any errors and this wasn't explicitly called out in the deprecation guides, it was documented in [these docs](https://api.emberjs.com/ember-data/5.3/classes/Transform). The new blueprints also conform to this new format (when using `ember g transfroms`)

addons/api/app/transforms/number.js
lisbet-alvarez: These were def. deprecation warnings. Below is an example:

Pasted Graphic 6

ui/admin/ember-cli-build.js
```diff
@@ -7,7 +7,9 @@
 
 const EmberApp = require('ember-cli/lib/broccoli/ember-app');
 
-module.exports = function (defaults) {
+module.exports = async function (defaults) {
+  const { setConfig } = await import('@warp-drive/build-config');
```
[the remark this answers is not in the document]
lisbet-alvarez: Yup using `setConfig` is definitely released. Also using the `emberData` key for configuration causes a deprecation warning to print when the app is building. So users are already being encouraged to use `setConfig` instead of the `emberData` key.

addons/api/addon/services/store.js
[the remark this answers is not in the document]
ZedLi: Gotcha, thanks for the info!

ui/admin/app/routes/scopes/scope.js
```diff
@@ -29,12 +29,12 @@ export default class ScopesScopeRoute extends Route {
...
     // from ID because global has a fixed ID.
-    const type = id === 'global' ? TYPE_SCOPE_GLOBAL : TYPE_SCOPE_ORG;
     return this.store.findRecord('scope', id).catch(() => {
+      const type = id === 'global' ? TYPE_SCOPE_GLOBAL : TYPE_SCOPE_ORG;
       const maybeExistingScope = this.store.peekRecord('scope', id);
-      const scopeOptions = { id, type };
       return (
-        maybeExistingScope || this.store.createRecord('scope', scopeOptions)
+        maybeExistingScope ||
+        this.store.push({ data: { id, type: 'scope', attributes: { type } } })
```
hashicc, line 37: Is this related to this comment:
> Attempting to call createRecord with an id that was already used in findRecord triggers the error below due to [requirement of unique ID](https://deprecations.emberjs.com/id/ember-data-deprecate-non-uniq-collected-payloads). We were using createRecord to make a temporary record that could be used when retrieving the specified scope record failed due to unauthenticated or permission denied errors.

I would have thought the previous way would work:
```js
this.store.createRecord('scope', scopeOptions)
```
Since there's already a check with `maybeExistingScope`. I was reviewing this bit of the docs [_Store createRecord() vs. push() vs. pushPayload()_](https://api.emberjs.com/ember-data/3.24.0/classes/Store) and it seems like the difference is the initial state of the record (`loaded.saved`). Maybe that's more accurate for what is going on here.

hashicc (approved):
🌟🌟🌟
thanks for documenting the specific changes and making this PR really easy to understand. I learned a interesting things about the direction ember data is taking. I have one non-blocking question around the `createRecord` vs `store.push`, for my own curiosity

ui/admin/app/routes/scopes/scope.js
lisbet-alvarez: Yup this change is related to the comment u pointed out.
Thank you for linking those docs!! 
To check my understanding, is it correct to say that perhaps `createRecord` was being called multiple times with the same id since the `maybeExistingScope` check wasn't catching that the temp record was already created since its not in the store. And this is what was causing the uniqueID error?

PR #2746, opened 2025-03-25
chore: 🤖 Fix git-cz usage for auto populating issues in hook

ZedLi:
✅ Closes: https://hashicorp.atlassian.net/browse/ICU-16777

# Description
Fixes an issue after the yarn 4 upgrade PR https://github.com/hashicorp/boundary-ui/pull/2697 where the commitizen hook to automatically populate the JIRA issue link when committing was broken. They were the same versions but for whatever reason, the `git-cz` binary was pointing to commitizen's CLI instead of `git-cz`. It seemed this bin's [executable](https://github.com/hashicorp/boundary-ui/blob/main/yarn.lock#L9808) was being installed over this [one](https://github.com/hashicorp/boundary-ui/blob/main/yarn.lock#L14451) (which should be the one we're calling). 

You can check where the `node_module/.bin` exectuable's symbolic link was pointing to:

After the yarn 4 upgrade (currently):
image

Before the yarn 4 upgrade and after this PR:
image
 
I could've just called `gitcz` instead but it didn't look like we needed `cz-conventional-changelog` as a dev dependency which also brings in `commitizen`. Because we're just calling with `npx git-cz` it doesn't look like we need to bring in `commitizen` as we're never calling it directly in the terminal and instead just depending on it during the commit hook. Removing the extra dependency gets rid of the `git-cz` executable conflict.

## How to Test
Try committing in this branch and it should auto populate the issue.

### DIFF
--- a/.husky/prepare-commit-msg
+++ b/.husky/prepare-commit-msg
@@ -21,4 +21,3 @@ else
   # This is to allow interaction with the terminal for commitizen
   exec < /dev/tty && npx git-cz --hook || true
 fi
-
--- a/package.json
+++ b/package.json
@@ -39,7 +39,6 @@
   },
   "devDependencies": {
     "concurrently": "^9.1.0",
-    "cz-conventional-changelog": "^3.3.0",
     "doctoc": "^2.2.0",
     "git-cz": "^4.9.0",
     "husky": "^9.1.6",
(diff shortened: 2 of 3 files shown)


hashicc (approved):
🚀

calcaide (approved):

DhariniJeeva (approved):
good catch! and thanks the fix

lisbet-alvarez (approved):
Thanks for fixing this!

PR #2747, opened 2025-03-25
Icu 16409 roles list & global-scope

cameronperera:
# Description

Test selector refactor for roles list & global-scope
Since global-scope was such a big file, I opted to just have these 2 files in this PR

:tickets: 
[List](https://hashicorp.atlassian.net/browse/ICU-16409)
[Global-Scope](https://hashicorp.atlassian.net/browse/ICU-16410)
## Checklist

### DIFF
--- a/addons/rose/addon/components/rose/pagination/index.hbs
+++ b/addons/rose/addon/components/rose/pagination/index.hbs
@@ -4,6 +4,7 @@
 }}
 
 <Hds::Pagination::Numbered
+  data-test-pagination
   @totalItems={{@totalItems}}
   @currentPage={{@currentPage}}
   @currentPageSize={{@currentPageSize}}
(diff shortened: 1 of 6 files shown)


addons/rose/addon/components/rose/pagination/index.hbs
```diff
@@ -4,6 +4,7 @@
 }}
 
 <Hds::Pagination::Numbered
+  data-test-pagination
```
cameronperera, line 7: I am open for a less generic name for this. I opted for a generic name since we shouldn't have 2 of these on the same page but never say never. Thoughts?

lisbet-alvarez (approved):
This was a big one! Thanks for the refactor

addons/rose/addon/components/rose/pagination/index.hbs
calcaide: I think its a good name. I agree, we shouldn't have more than 1 pagination component per page 😉

calcaide (approved):
Thanks for the great work!! 🙌

ZedLi (approved):

PR #2752, opened 2025-03-27
refactor: 💡 Batch writes to indexedDB

ZedLi:
✅ Closes: https://hashicorp.atlassian.net/browse/ICU-16795
Relevant [thread](https://hashicorp.slack.com/archives/C012GTH1C4E/p1742506671923969)

# Description
This PR aims to improve the loading experience for large amounts of resources in admin UI as well as reduce the peak memory usage. Currently, we have a sub-optimal experience when loading large amounts of resource (>100k ish) and what happens is first we retrieve _all_ the resources and keep it in memory before flushing it all to indexedDB. 

The issue with that is this requires us to hold the entire data set in memory and then a large portion of time spent writing to indexedDB which causes the browser to hang/stop responding which is a bad experience. On my mac, it takes about 45s to a min to retrieve all the records and another 45s to 1min to write it to disk. The time spent writing to disk causes the browser to hang entirely which may cause a user to assume it has frozen. This change reduces the time it takes to load everything about 40% for me. 

Improving the actual search experience is not in scope for this PR, this is only improving the initial loading of the data and any search improvements will be a followup. 

Take note of the JS heap size (blue line and blue graph under screenshots), listeners (yellow line), cpu usage (yellow bar at top), and network (small blue bars) in some of these profiles:

Before:
Screenshot 2025-03-27 at 2 03 26 PM
Note the large CPU time spent writing to disk right after all the network requests.

After:
Screenshot 2025-03-27 at 1 58 09 PM
Note the constant release of heap memory and it doesn't continuously climb up as we see in the previous screenshot.
It's also interesting to note that we're able to now both write to disk and hit the API so we don't completely block while waiting for another, which is ideal for reducing the bottleneck. Ideally, we'd like to see the network requests not stop to wait for writing or vice versa but this can vary depending on your machine specs and network speed. 

In the future if this is still an issue we can try and completely offload this work to a worker.

## How to Test
Connect to this [cluster](http://boundary-alb-wssconrz-1383983136.us-east-1.elb.amazonaws.com:9200) if you'd like to test (DM me for PW) and goto the aliases page. Make sure to clear out your indexedDB storage beforehand.  

## Checklist


DhariniJeeva (approved):
tested using the cluster and works perfectly well without any issues! I was able to view the results in a min or so. Thanks for this huge improvement and digging into this issue!

hashicc (approved):
🎉 This is great refactor! It looks like a lot but it's actually a very precise change with a few lines to add batch loading of resources cleanly

calcaide (approved):
Huge improvement!! divide and conquer!! thanks for the great work 🙌

cameronperera (approved):

lisbet-alvarez (approved):
Works as expected for me locally! 🥳 Thank you for this improvement!!

PR #2753, opened 2025-03-27
Chore icu 15394 upgrade electron

calcaide:
# Description

**BE AWARE:** this upgrade depends [on Deprecate Ubuntu 20](https://github.com/hashicorp/boundary-desktop-releases/pull/130) to be able to build. For the breaking change mentioned below, Ubuntu 20 does not have C++20 but Ubuntu 22 does.

Upgrade electron to latest version.

This upgrade has been done incremental following [breaking changes guide](https://www.electronjs.org/docs/latest/breaking-changes). Previously we try upgrade directly to latest reproducing issues at build time.

As a side effect of [Native modules now require C++20](https://www.electronjs.org/docs/latest/breaking-changes#behavior-changed-native-modules-now-require-c20) we had to upgrade `node-pty`. Be aware we are using a `beta` version of node-pty, using a beta version was discussed within tech time March 20 2025. We will perform e2e testing in different OS with binaries produced for our pipeline to validate everything works and nothing is broken.

:tickets: [Jira ticket](https://hashicorp.atlassian.net/browse/ICU-15394)

## How to Test
- boundary-desktop-releases builds successfully. [Test build here](https://github.com/hashicorp/boundary-desktop-releases/actions/runs/14118111002)
- Run desktop e2e test suite in different operating systems. Using binaries from build mentioned above.


ZedLi (approved):
Thanks for the update, built fine locally!

DhariniJeeva (approved):
tested locally and able to build and run desktop with no issues! 
thanks for the update

cameronperera (approved):

lisbet-alvarez (approved):
Great work!!! Thanks for taking care of this!

PR #2700, opened 2025-02-27
test(e2e): Enable Admin UI e2e tests to run against a local branch

moduli:
# Description
This PR updates the Admin UI e2e test suite to support running against a local version of the Admin UI independent of the UI that's bundled with a Boundary binary. This required a decoupling of the address used by the browser vs the Boundary CLI. To accomplish this, this PR adds a new environment variable `BOUNDARY_ADDR_BRANCH`. If this is set, the test suite will use that address for the browser. That way, we can run 
```
ember serve 
BOUNDARY_ADDR_BRANCH=http://localhost:4200 yarn run ...
```

The README was updated to include instructions for how to use it.

https://hashicorp.atlassian.net/browse/ICU-16563

## Screenshots (if appropriate)
Screenshot 2025-02-27 at 12 52 19 PM

## How to Test
Follow instructions outlined in README updates
```
❯ export BOUNDARY_ADDR_BRANCH="http://localhost:4200" # This is the URL provided by the first terminal
❯ yarn admin:ent:docker tests/target-ent.spec.js  --project=chromium  
yarn run v1.22.22
$ yarn run admin --grep "(?=.*@ent)(?=.*@docker)" tests/target-ent.spec.js --project=chromium
$ yarn playwright test --config admin/playwright.config.js --grep '(?=.*@ent)(?=.*@docker)' tests/target-ent.spec.js --project=chromium
$ /Users/mycow/Developer/boundary-ui/node_modules/.bin/playwright test --config admin/playwright.config.js --grep '(?=.*@ent)(?=.*@docker)' tests/target-ent.spec.js --project=chromium

Running 3 tests using 1 worker

  ✓  1 [chromium] › tests/target-ent.spec.js:21:1 › Verify session created for TCP target @ent @aws @docker (10.2s)

  ✓  2 [chromium] › tests/target-ent.spec.js:88:1 › Verify session created for SSH target @ent @aws @docker (14.1s)

  ✓  3 [chromium] › tests/target-ent.spec.js:162:1 › SSH target with host sources @ent @aws @docker (23.3s)

  Slow test file: [chromium] › tests/target-ent.spec.js (47.5s)
  Consider splitting slow test files to speed up parallel execution
  3 passed (50.6s)
✨  Done in 52.06s.
```

## Checklist

- [ ] I have added before and after screenshots for UI changes
- [ ] I have added JSON response output for API changes
- [ ] I have added steps to reproduce and test for bug fixes in the description
[...]

PR #2755, opened 2025-03-28
refactor: 💡 Force requests in playwright to use controller addr

ZedLi:
# Description
As a followup to PR https://github.com/hashicorp/boundary-ui/pull/2700 and this [thread](https://hashicorp.slack.com/archives/C012GTH1C4E/p1743185621713639), we should always be using the controller URL when making HTTP requests.

This PR overrides the `request` fixture with our own custom one that forces the baseURL to be the boundary controller URL, which fixes the issue in cases when testing against a local admin branch as the baseURL for requests would then be pointing incorrectly against the locally running server. 

## How to Test
Try running the tests against a local branch for admin UI.

### DIFF
--- a/e2e-tests/global-setup.js
+++ b/e2e-tests/global-setup.js
@@ -27,7 +27,8 @@ async function globalSetup() {
 const authenticateToBoundary = async () => {
   // Log in and save the authenticated state to reuse in tests
   const browser = await chromium.launch();
-  const page = await browser.newPage();
+  const browserContext = await browser.newContext();
+  const page = await browserContext.newPage();
   await page.goto(baseUrl);
 
   const loginPage = new LoginPage(page);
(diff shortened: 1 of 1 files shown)


e2e-tests/global-setup.js
```diff
@@ -27,23 +27,25 @@ async function globalSetup() {
 const authenticateToBoundary = async () => {
   // Log in and save the authenticated state to reuse in tests
   const browser = await chromium.launch();
-  const page = await browser.newPage();
+  const browserContext = await browser.newContext();
```
ZedLi, line 30: This is an unrelated change but I saw it's [recommended](https://playwright.dev/docs/api/class-browser#browser-new-page) to create your own [context](https://playwright.dev/docs/api/class-browser#browser-new-context) and close it directly. In this case, I don't think it matters that much but I saw we needed the context to grab the `storageState` on line 39 anyways so we might as well manually control it.

lisbet-alvarez (approved):
Thank you for fixing this! Works great!

cameronperera (approved):

calcaide (approved):
Thanks for the fix 🙌

PR #2756, opened 2025-03-28
Refactor test selectors for credential-store/credentials create test

calcaide:
# Description

Refactor test selectors for credential-store/credentials create test

:tickets: [Jira ticket](https://hashicorp.atlassian.net/browse/ICU-15475) 

## How to Test
- Run test successfully.


ui/admin/tests/acceptance/credential-store/credentials/selectors.js
```diff
@@ -0,0 +1,16 @@
+/**
+ * Copyright (c) HashiCorp, Inc.
+ * SPDX-License-Identifier: BUSL-1.1
+ */
+
+export const FIELD_TYPE_SSH = '[value=ssh_private_key]';
+export const FIELD_TYPE_USERNAME_PASSWORD = '[value=username_password]';
+export const FIELD_TYPE_JSON = '[value=json]';
+export const EDITOR = '[data-test-code-editor-field-editor]';
```
calcaide, line 9: Open to change the name of this selector 😉 . Initially I though on calling it EDITOR_CONTAINER but beign a domain specific selector file end up with this.

ui/admin/tests/acceptance/credential-store/credentials/create-test.js
```diff
@@ -300,32 +315,33 @@ module(
...
+      await click(commonSelectors.HREF(urls.newCredential));
+
       assert.false(featuresService.isEnabled('json-credentials'));
       assert.true(
         instances.staticCredentialStore.authorized_collection_actions.credentials.includes(
           'create',
         ),
       );
-      assert.dom('.hds-form-radio-card').exists({ count: 2 });
       assert
```
calcaide, line 336: Refactored these assertions to not use hds specific classes and reusing the existing selectors.

- We don't check anymore there are 2 `hds-form-radio-card`. Instead we check username & password and SSH key are present but JSON being not present.
- We kept the `hasAttribute` assertion, but I personally think can be redundant since the selector itself its using `value` i.e: `export const FIELD_TYPE_USERNAME_PASSWORD = '[value=username_password]';`

laurenolivia (commented):
Left one question

ui/admin/tests/acceptance/credential-store/credentials/create-test.js
```diff
@@ -198,17 +214,20 @@ module(
...
         instances.staticCredentialStore.authorized_collection_actions.credentials.includes(
           'create',
         ),
       );
-      assert.dom('.rose-layout-page-actions a').doesNotExist();
+      assert.dom(selectors.MANAGE_DROPDOWN_NEW_CREDENTIAL).doesNotExist();
     });
 
     test('saving a new username & password credential with invalid fields displays error messages', async function (assert) {
+      const errorMessage = 'Error in provided request.';
```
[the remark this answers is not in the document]
calcaide: Just want to clarify, we do not define the string 3xs in the test, we define 3xs within the test file, be aware each test within this test file does not share context, they run isolated from each other.

The selectors file is mean to host selectors, therefore I don't think this belongs to a selectors file, since it's not a selector, but a string we re-use per test context.

This is a pattern we follow in our tests when mocking data. The goal of this pattern is to write a string only once per test and reuse it, to avoid duplication. I.e: you write it once to generate the "fake" data, and re-write it at least once, per the test assertion.

Usually, the error message or error description are slightly different per test, so that's why we kept this pattern per test and not per test file.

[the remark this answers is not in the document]
laurenolivia: Was initially thinking something like [this one](https://github.com/hashicorp/boundary-ui/blob/main/ui/admin/tests/acceptance/auth-methods/selectors.js#L13) currently in main, storing a string value in a selectors file and reusing it in [subtests](https://github.com/hashicorp/boundary-ui/blob/main/ui/admin/tests/acceptance/auth-methods/create-test.js#L104).

DhariniJeeva (approved):
looks good! thanks for the refactor

ui/admin/tests/acceptance/credential-store/credentials/selectors.js
DhariniJeeva: I'm ok with this selector name!

PR #2757, opened 2025-03-31
feat: 🎸 Populate subject for OIDC account name displays

ZedLi:
✅ Closes: https://hashicorp.atlassian.net/browse/ICU-16634

# Description
OIDC accounts are guaranteed to have a subject claim which we should display. Currently we only look at email and full_name which are all optional for OIDC claims. In many cases, if there were no claims for these optional claims, OIDC accounts under users would show nothing which is not a good experience. 

I also added links to the fields which goes directly to the corresponding auth method and account as we only display accounts under an auth method. 

Before:
image

After:
image

## How to Test
Start up boundary dev and navigate to accounts under users. OIDC should display a name and they all should be able to goto the account directly. 

## Checklist


DhariniJeeva (approved):
looks good to me!

priya-patel04 (approved):
Lgtm!

lisbet-alvarez (approved):
Thx for fixing this!

calcaide (approved):
Thanks for this!! looks much better now 😉

PR #2759, opened 2025-04-01
Add tests for indexed db handler

hashicc:
# Description

To support the upcoming sorting work this adds a few tests cases to test the baseline expectations for the indexedb handler. To avoid creating multiples of 10,000 records to test batch loading, it also makes the batch size configurable by the ember environment config.

## Checklist

### DIFF
--- a/addons/api/addon/handlers/indexed-db-handler.js
+++ b/addons/api/addon/handlers/indexed-db-handler.js
@@ -10,8 +10,6 @@ import { queryIndexedDb } from '../utils/indexed-db-query';
 import { paginateResults } from '../utils/paginate-results';
 import { hashCode } from '../utils/hash-code';
 
-const BATCH_LIMIT = 10000;
-
 /**
  * Handler to sit in front of the API request layer
  * so we can handle any caching with indexedDb first
@@ -24,6 +22,10 @@ export default class IndexedDbHandler {
   }
 
   async request(context, next) {
+    const environmentConfig =
+      getOwner(this).resolveRegistration('config:environment');
+    const batchLimit = environmentConfig.api?.batchLimit ?? 10000;
+
     switch (context.request.op) {
       case 'query': {
         const { store, data } = context.request;
@@ -68,7 +70,7 @@ export default class IndexedDbHandler {
               payload = await adapter.query(store, schema, {
                 ...remainingQuery,
                 list_token: listToken,
-                batchLimit: BATCH_LIMIT,
+                batchLimit: batchLimit,
               });
 
               // await the previous writeToIndexedDbPromise before writing to indexedDb again
--- a/addons/api/addon/handlers/indexed-db-handler.js
+++ b/addons/api/addon/handlers/indexed-db-handler.js
@@ -17,14 +17,14 @@ import { hashCode } from '../utils/hash-code';
 export default class IndexedDbHandler {
   @service indexedDb;
 
+  batchLimit = 10_000;
+
   constructor(context) {
     setOwner(this, getOwner(context));
   }
 
   async request(context, next) {
-    const environmentConfig =
-      getOwner(this).resolveRegistration('config:environment');
-    const batchLimit = environmentConfig.api?.batchLimit ?? 10000;
+    const { batchLimit } = this;
 
     switch (context.request.op) {
       case 'query': {
--- a/addons/api/package.json
+++ b/addons/api/package.json
@@ -33,6 +33,9 @@
     "miragejs": "^0.1.48",
     "tracked-built-ins": "^3.1.1"
   },
+  "peerDependencies": {
+    "sinon": "^19.0.2"
+  },
   "devDependencies": {
     "@babel/core": "^7.26.10",
     "@babel/eslint-parser": "^7.21.3",
@@ -71,6 +74,7 @@
     "prettier": "^3.0.0",
     "qunit": "^2.22.0",
     "qunit-dom": "^3.2.1",
+    "sinon": "^19.0.2",
     "webpack": "^5.94.0"
   },
   "engines": {
--- a/addons/api/tests/unit/handlers/indexed-db-handler-test.js
+++ b/addons/api/tests/unit/handlers/indexed-db-handler-test.js
@@ -1,3 +1,8 @@
+/**
+ * Copyright (c) HashiCorp, Inc.
+ * SPDX-License-Identifier: BUSL-1.1
+ */
+
 import { module, test } from 'qunit';
 import { setupTest } from 'ember-qunit';
 import IndexedDbHandler from 'api/handlers/indexed-db-handler';
--- a/addons/api/tests/unit/handlers/indexed-db-handler-test.js
+++ b/addons/api/tests/unit/handlers/indexed-db-handler-test.js
@@ -85,14 +85,11 @@ module('Unit | Handler | indexed-db-handler', function (hooks) {
     // and these multiples can be asserted in the tests individually
     const testBatchLimit = 10;
     const serverResultPageLimit = 5;
-
-    hooks.beforeEach(function () {
-      handler.batchLimit = testBatchLimit;
-    });
-
     let aliases, aliasResponseHandlerSpy;
 
     hooks.beforeEach(function setupMirageData() {
+      handler.batchLimit = testBatchLimit;
+
       aliases = this.server.createList('alias', 50);
 
       const aliasResponseHandler = createPaginatedResponseHandler(
--- a/addons/api/tests/unit/handlers/indexed-db-handler-test.js
+++ b/addons/api/tests/unit/handlers/indexed-db-handler-test.js
@@ -35,7 +35,7 @@ function createPaginatedResponseHandler(mirageRecords, { pageSize }) {
       ({ list_token }) => list_token === listToken,
     );
     const nextResult = index === -1 ? null : results[index + 1];
-    return nextResult ?? null;
+    return nextResult;
   }
 
   // mirage handler compatible with paginated responses
@@ -46,7 +46,15 @@ function createPaginatedResponseHandler(mirageRecords, { pageSize }) {
     }
 
     const result = nextResult(listToken);
-    assert('requested paginated result should exist', result);
+
+    if (!result) {
+      return new Response(
+        404,
+        {},
+        `No results found for list token "${listToken}"`,
+      );
+    }
+
     return result;
   };
 }
--- a/addons/api/tests/unit/handlers/indexed-db-handler-test.js
+++ b/addons/api/tests/unit/handlers/indexed-db-handler-test.js
@@ -8,7 +8,7 @@ import sinon from 'sinon';
 import { assert } from '@ember/debug';
 import { setupIndexedDb } from 'api/test-support/helpers/indexed-db';
 
-function createPaginatedResponses(mirageRecords, { pageSize }) {
+function createPaginatedResponseHandler(mirageRecords, { pageSize }) {
   assert('pageSize is required', pageSize);
   const pagesCount = mirageRecords.length / pageSize;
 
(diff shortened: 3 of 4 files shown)


addons/api/package.json
```diff
@@ -33,6 +33,9 @@
     "miragejs": "^0.1.48",
     "tracked-built-ins": "^3.1.1"
   },
+  "peerDependencies": {
+    "sinon": "^19.0.2"
```
hashicc, line 37: I tried to use sinon in my test but it wasn't actually installed within the addon, despite having it used in addon-support `api/test-support/handlers/cache-daemon-search.js`.

I first tried to keep it only as a `devDependency` but got this build error:
```
api tried to import "sinon" in "api/test-support/handlers/cache-daemon-search.js" from addon code, but "sinon" is a devDependency. You may need to move it into dependencies
```
and this error would go away if I included it in `dependencies` but I don't think we'd want our host apps to bundle sinon. I found [this issue](https://github.com/embroider-build/ember-auto-import/issues/425) which is along the same lines of what I'm facing. I believe to make this work correctly it should be a `peerDependency`, and then in our host app we would need `sinon` as a `peerDependency`(which we already have for both desktop and admin).

addons/api/tests/unit/handlers/indexed-db-handler-test.js
```diff
@@ -0,0 +1,147 @@
...
+    applicationAdapter = store.adapterFor('application');
+
+    // spy on all handler and application adapter methods
+    sinon.spy(handler);
+    sinon.spy(applicationAdapter);
+
+    store.requestManager = manager;
+    manager.use([handler]);
+    await indexedDBService.setup(
+      `test-db-${faker.string.nanoid({ min: 5, max: 5 })}`,
```
hashicc, line 78: Is worth deleting these test dbs after each test? They could add up I suppose and hit some sort of browser maximum.

ZedLi: Yeah most likely, I originally made a [convenience function](https://github.com/hashicorp/boundary-ui/blob/20c4bbcee2dc452fd329e9329db5dc3ddca4e533/addons/api/addon-test-support/helpers/indexed-db.js) to setup indexedDB for tests, otherwise they'll just sit around forever

hashicc: 🕺 easy, so much better!

addons/api/tests/unit/handlers/indexed-db-handler-test.js
```diff
@@ -0,0 +1,151 @@
...
+  };
+}
+
+module('Unit | Handler | indexed-db-handler', function (hooks) {
+  setupTest(hooks);
+  setupMirage(hooks);
+
+  const testBatchLimit = 10;
+  let previousBatchLimit;
+  hooks.beforeEach(function setBatchLimitConfig() {
```
hashicc, line 51: Since there's so much setup I tried to break them up into `beforeEach`(/`afterEach`) sections.

They could also be made into their own functions on top that take a `hooks` argument:
```js
setupBatchLimitConfig(hooks);
setupIndexHandler(hooks);
setupTestMirageData(hooks);
```
But going this route I suppose I would have to stash the test variables on `this` to be used within each test which is a bit less explicit and harder to follow. Right now you can command-click on a variable and see it defined as a `let` above the corresponding `hooks.beforeEach`...

Let me know if anyone has other ideas or strong opinions

addons/api/package.json
calcaide: This change makes sense 😉

addons/api/addon/handlers/indexed-db-handler.js
```diff
@@ -24,6 +22,10 @@ export default class IndexedDbHandler {
   }
 
   async request(context, next) {
+    const environmentConfig =
```
calcaide, line 25: Asking for learning purposes: I understand this as reading the app  configuration file, correct me if I am wrong.
But I don't see adding `batchLimit` to Admin or Desktop config files.

Is the idea to add `batchLimit` to the UI's config file later? or is this a mechanism to could change `batchLimit` for testing purposes?

addons/api/tests/unit/handlers/indexed-db-handler-test.js
ZedLi: I personally think this is fine, it's still quite easy to understand. I like the inline named functions, I realize I don't usually do that and end up using a comment with an anonymous function but this obviates the need for them

ZedLi (approved):
Looks great! Thanks for laying down the base for these handler tests!

addons/api/addon/handlers/indexed-db-handler.js
hashicc: That's a really good question. Yes, optionally these could be provided by the host apps. Otherwise it defaults to 10000. That wasn't the goal here though since we only have two apps and hardcoding the batch size would work for both our consuming apps. 

The _main_ reason I wanted to make it configurable was so that it could be modified in tests. I didn't want to have to make 20,000 mirage models to have full 2 batches. But now that I'm seeing your question and seeing how I'm testing this I could make this a property on the class and override it for testing purposes on the instance.

In this file:

```js
export default class IndexedDbHandler {
 // ...
 batchSize = 10_000
 // ...
}
```

In test [here](https://github.com/hashicorp/boundary-ui/pull/2759/files#diff-d8094e4c17b4f2f5c6f307b7ba92c627013f8d1f6d9765ae73880dd86725cc62R69):
```js
handler = new IndexedDbHandler(store);
hander.batchSize = 10
```

I think I'm going to make that change because making something globally configurable is sometimes the best way (eg: controlling the duration of a flash message, if you have to wait 5 seconds for every flash message in every acceptance test it's a lot of wasted time), but in this case I think we're testing at the unit level and can change this one property (or make it part of the constructor)

addons/api/addon/handlers/indexed-db-handler.js
```diff
@@ -10,20 +10,22 @@ import { queryIndexedDb } from '../utils/indexed-db-query';
...
-const BATCH_LIMIT = 10000;
-
 /**
  * Handler to sit in front of the API request layer
  * so we can handle any caching with indexedDb first
  */
 export default class IndexedDbHandler {
   @service indexedDb;
 
+  batchLimit = 10_000;
```
hashicc, line 20: This allows it to be tested more easily. Thanks @calcaide for prompting the suggestion, we don't really need to make this globally configurable for host apps right now.

addons/api/tests/unit/handlers/indexed-db-handler-test.js
```diff
@@ -0,0 +1,193 @@
...
+      const aliasResponseHandler = createPaginatedResponseHandler(
+        this.server.schema.aliases.all().models,
+        { pageSize: serverResultPageLimit },
+      );
+
+      aliasResponseHandlerSpy = sinon.spy(aliasResponseHandler);
+      this.server.get('aliases', aliasResponseHandlerSpy);
+    });
+
+    test('it returns all data for a resource from the server', async function (assert) {
```
calcaide, line 107: Asking for learning purposes:
- `results` is a collection already sorted by `created_time`.
- Taking aliases, we sorted by created_time.
- We deep compare the sorted collection and `results`, they are the same.
- What we are really testing here? I am not saying the test is unnecessary bc I think it is, but I want to understand better what's on your head here 😃

addons/api/tests/unit/handlers/indexed-db-handler-test.js
```diff
@@ -0,0 +1,185 @@
...
+      assert.strictEqual(
+        aliasResponseHandlerSpy.callCount,
+        aliases.length / serverResultPageLimit,
+        'server handler should be called once for each page',
+      );
+    });
+  });
+
+  module('sorting', function () {
+    test('it sorts by descending `created_time` by default', async function (assert) {
```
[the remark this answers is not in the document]
calcaide: I do totally agree with you here, I was really happy to see this test and right away though its was a really good pattern to follow.

addons/api/tests/unit/handlers/indexed-db-handler-test.js
```diff
@@ -0,0 +1,193 @@
...
+        results.map(({ name }) => name),
+        '`name` property matches between the returned results and the mirage records sorted by created_time',
+      );
+    });
+
+    test('supports batching requests and writes to indexedDB', async function (assert) {
+      await store.query('alias', {});
+
+      assert.strictEqual(
+        handler.writeToIndexedDb.callCount,
```
calcaide, line 131: Asking for learning purposes: where the `callCount` comes from? is this an internal property?

addons/api/tests/unit/handlers/indexed-db-handler-test.js
calcaide: I do thin its fine too. If the test becomes more complex or we want to reuse these in other tests files I will be towards make their own functions, but for now its totally fine 😉  Thanks for the explanation

calcaide (dismissed):
I used this PR to understand better the indexed-db-handler and the "moving parts" (pagination, batch limits, etc) of this domain that we need to have in mind when working on pagination.

I left some questions ONLY for learning purposes 😉 
Thanks for the amazing work!

addons/api/tests/unit/handlers/indexed-db-handler-test.js
hashicc: That's a good question, and it's way too magical and I'm glad you pointed it out. I can make a change to make this more explicit. 

[`sinon.spy(handler);`](https://github.com/hashicorp/boundary-ui/pull/2759/files#diff-d8094e4c17b4f2f5c6f307b7ba92c627013f8d1f6d9765ae73880dd86725cc62R75) actually ends up spying on all methods on the `handler` instance. Calling the method works as normal but because it's been "spied" on, the function you're calling is actually an installed wrapper that tracks everything including `callCount`, `sinon` also puts `callCount` on the wrapped method.

Instead what I should do to do more explicit is what I did in other places in this test: `let handlerSpy = sinon.spy(handler);`, and then it's clear that this is a spy and `handlerSpy.callCount` would be very clear. If this was a typescript project you'd see the type of `handlerSpy` would be explicit but typescript would probably complain about `handler.writeToIndexedDb.callCount` because it wouldn't know how `callCount` got added (that's my guess at least).

addons/api/tests/unit/handlers/indexed-db-handler-test.js
hashicc: You're right this wouldn't be a great assertion to test sorting. What this is really testing though that is that all 50 `aliases` that are mirage models (the source) made it across the `store.query` as ember data model `results`.

Checking that 50 results made it across is one thing (`.length === .length`), but I'm ensuring that _every_ model made it across by using names. The `results` are sorted by `created_time` so to make the assertion easier I need to also sort the source mirage models by name, and then it's an easy array comparison. If I see every name from every 50 models from the source `aliases` match the 50 model names from the `results` then I'm assuming the result set is complete

addons/api/tests/unit/handlers/indexed-db-handler-test.js
```diff
@@ -0,0 +1,193 @@
...
+  });
+
+  module('fetching and batch loading', function (hooks) {
+    // `testBatchLimit` and `serverResultPageLimit` are intentionally difference
+    // so that multiple calls to the server are required to fetch a single batch,
+    // and these multiples can be asserted in the tests individually
+    const testBatchLimit = 10;
+    const serverResultPageLimit = 5;
+
+    hooks.beforeEach(function () {
```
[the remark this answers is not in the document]
hashicc: Now that I have separate tests into modules, this `beforeEach` could probably be merged with the one below `setupMirageData` and the function renamed, it doesn't really make sense to have it separated for just one line.

But yes, everything you said is 💯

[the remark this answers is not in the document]
hashicc: Collapsed `beforeEach` @ https://github.com/hashicorp/boundary-ui/pull/2759/commits/e5c38a7d709189269140cb7a38da0086e2fa4688

addons/api/tests/unit/handlers/indexed-db-handler-test.js
hashicc: Explicit spies used @ https://github.com/hashicorp/boundary-ui/pull/2759/commits/c253ce574c55aaf1e1e2b92ae9da98785d507d70

addons/api/tests/unit/handlers/indexed-db-handler-test.js
calcaide: Ah! gotcha that makes sense. I initially though is has something to do with the spy and methods the spy might be adding. I am glad to read the full explanation it makes sense on my head now 😉

calcaide (approved):

PR #2763, opened 2025-04-03
Llb/side nav refactor

lisbet-alvarez:
# Description

Replaces rose side nav usage in admin-ui with hds side nav.
Note following commits were added to fix merge conflicts when regularly updating branch with `main`:
[a61aecc](https://github.com/hashicorp/boundary-ui/pull/2763/commits/a61aecc05fc828df290f99e72eb63fc0430ef8ee)
[d66c24d](https://github.com/hashicorp/boundary-ui/pull/2763/commits/d66c24de06576b766e0b6fec9aef2bacf47565f6)
[5739bd1](https://github.com/hashicorp/boundary-ui/pull/2763/commits/5739bd19119284d58ada8eb79d73d2bd467537a1)
[4857808](https://github.com/hashicorp/boundary-ui/pull/2763/commits/48578089a55075228e4cd7bc105bb94b096dcc34)

:tickets: [Jira ticket](https://hashicorp.atlassian.net/browse/ICU-15685)

## Checklist

- [ ] I have added before and after screenshots for UI changes
- [ ] I have added JSON response output for API changes
- [ ] I have added steps to reproduce and test for bug fixes in the description
- [ ] I have commented on my code, particularly in hard-to-understand areas
- [ ] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works

PR #2765, opened 2025-04-04
refactor: 💡 update all header tags to use HDS::Text::Display

cameronperera:
✅ Closes: https://hashicorp.atlassian.net/browse/ICU-16863

# Description

This PR only changes `<h[1-6]>` tags. I did my best to match existing format and sizing but ultimately decided to stick as close to HDS standard styling. The main thing I added was a new `header-spacing` class since HDS leaves that to the consumer to manage. Happy to adjust that style if needed.

Examples of existing header tags compared to the Hds header.
Screenshot 2025-04-04 at 9 41 38 AM
Screenshot 2025-04-04 at 9 47 59 AM

## How to Test

Navigate to ALL the screens!

## Checklist


addons/rose/addon/components/rose/dialog/index.hbs
```diff
@@ -23,7 +23,10 @@
         <Rose::Icon @name={{@icon}} @size='medium' />
       {{/if}}
 
-      <h2 class='rose-dialog-heading'>{{@heading}}</h2>
+      <Hds::Text::Display
+        @tag='h2'
+        class='rose-dialog-heading'
```
calcaide, line 28: None blocking comment: with the change to Hds::Text, is this class need?

addons/rose/addon/components/rose/header/brand/index.hbs
```diff
@@ -15,7 +15,10 @@
   {{/if}}
 
   {{#if @text}}
-    <h1 class='rose-header-brand-text'>{{@text}}</h1>
+    <Hds::Text::Display
+      @tag='h1'
+      class='rose-header-brand-text'
```
calcaide, line 20: None blocking comment: with the change to Hds::Text, is this class need?

calcaide (dismissed):
I manually browse around the UI's and looks neat!
I run both, Admin and Desktop e2e suite with success!!

Left a couple of none blocking comments

Terrific work!!! 💪 🕺 🙌

addons/rose/addon/components/rose/dialog/index.hbs
hashicc: With #2766 I don't think we'll even have any `Rose::Dialog` usages left except for in tests, the only thing left will be `Rose::Dialog::Cover` being used for overlays

ui/desktop/app/styles/app.scss
```diff
@@ -653,3 +654,17 @@ $chevron-down: url("data:image/svg+xml;utf-8,<svg viewBox='0 0 24 24' fill='%231
...
+.header-spacing {
+  margin-top: 1rem;
+  margin-bottom: 0.5rem;
+  display: flex;
+
+  .doc-link {
+    margin-left: 0.25rem;
+
+    svg {
+      vertical-align: middle;
```
hashicc, line 667: It looks like the `display: flex` is fighting the text flow, if you remove `display: flex;` then you also don't need the `margin-left: 0.25rem;`. The `vertical-align: middle;` can be moved up to the `.doc-link` to simplify the selector nesting.

If you do want to go the flex route, I think using the `gap` property is better and then instead of `vertical-align` finding the corresponding alignment property for flexbox. But I think the regular text flow is fine for this case

hashicc (approved):
Browsed around and the small changes in heading sizes I preferred the HDS version! Looks great. The `header-spacing` I think is okay but could be simplified a bit

addons/rose/addon/components/rose/dialog/index.hbs
cameronperera: What @hashicc said is correct. This will be removed but I figured at the very least, we can update it while it still exist. So I opted to essentially leave it as-is since it will eventually be removed completely.

addons/rose/addon/components/rose/header/brand/index.hbs
cameronperera: It will not be needed shortly. @lisbet-alvarez's [PR](https://github.com/hashicorp/boundary-ui/pull/2763) will remove this component but I didn't want to not change it just in case this is merged first or anything like that.

ui/desktop/app/styles/app.scss
cameronperera: Wowzers. That is much nicer to read and flow. Thanks for the solid explanation.

hashicc (dismissed):
🚀 Well done! Love to see more HDS! 💪 💜

addons/rose/addon/components/rose/dialog/index.hbs
calcaide: Thanks for the explanation 😃

addons/rose/addon/components/rose/header/brand/index.hbs
calcaide: Thanks! that makes sense to me 😉

calcaide (approved):
Looks great!!! let's goo 🚀 with HDS 🙌

priya-patel04 (approved):
Nicely done! 🙌🏻

DhariniJeeva (dismissed):
thanks for this huge update!

hashicc (approved):
🚀 🚀 🚀

calcaide (approved):
🚀

priya-patel04 (approved):

laurenolivia (approved):
✨ nice

PR #2767, opened 2025-04-07
Chore icu 16553 deprecate a like array

calcaide:
# Description

Deprecate no A with array like, [more infor here](https://deprecations.emberjs.com/id/ember-data-no-a-with-array-like/)

:tickets: [Jira ticket](https://hashicorp.atlassian.net/browse/ICU-16553)

## How to Test
- Run Admin and Desktop unit test successfully.
- Run Admin and Desktop e2e tests successfully.

## Screenshots

Desktop e2e test:
Screenshot 2025-04-07 at 9 41 09 AM

Admin e2e test:
Be aware the OIDC failing test is not related to these changes and its present on branch `main`.
Screenshot 2025-04-07 at 2 32 30 PM


ui/admin/app/components/form/worker/create-worker-led/index.js
```diff
@@ -7,7 +7,7 @@ import Component from '@glimmer/component';
 import { tracked } from '@glimmer/tracking';
 import { service } from '@ember/service';
 import { action } from '@ember/object';
-import { A } from '@ember/array';
+import { TrackedArray } from 'tracked-built-ins';
```
calcaide, line 10: Why using TrackedArray rather than native JS array? 

[This component implementation is based on trackable array](https://guides.emberjs.com/release/in-depth-topics/autotracking-in-depth/#toc_arrays). When the user click the button to add a worker tag, the `addWorkerTag` function is call, such function pushed the new tag into the array.
`workerConfigText()` is call when the array is updated (because is trackable), then it follows workerConfigText implementation which will call other functions like `getTagConfigString`.

The above is my assumption of this component as specifically `workerTags`. If you think this assumption is wrong, please let me know and I will review this implementation.

ui/admin/app/routes/scopes.js
```diff
@@ -35,6 +34,6 @@ export default class ScopesRoute extends Route {
         scope_id: 'global',
         query: { filters: { scope_id: [{ equals: 'global' }] } },
       })
-      .catch(() => A([]));
+      .catch(() => []);
```
hashicc, line 37: I think it makes sense that a model would return a read-only array, that it shouldn't be mutated afterwards. Is that why this is not using a tracked array wrapper? It seems odd that before it was using the Ember Array wrapper 🤔

hashicc (dismissed):
💯 One small question about the returning empty `[]` instead of `new TrackedArray([])` when the previous was `A([])` but otherwise looks good to me!

ui/admin/app/routes/scopes.js
calcaide: You are right. After our zoom chat I agree this has been historically an Ember array, which is trackable. For the sake of not breaking things, I will set it as a trackable array, so if there are changes within the array it will cause re-render.

In our zoom chat you mention to wrap it in a ArrayProxy to follow up if its been call. I quickly did that, run the Admin UI and couldn't triggered. That said, this doesn't mean it's all good to leave it as read-only, as per my understanding this case will just hit when there is an error on the promise return by the store and I want to think, if such error occurs some part of the UI might be tracking changes to this array.

Thanks for the feedback and the zoom chat! 🙌

priya-patel04 (approved):
Nice!! 🙌🏻

cameronperera (approved):

DhariniJeeva (approved):
looks good! thanks for the update!

hashicc (approved):
💯 Thanks for making those changes!
They're safer, or at least in line keeping the reactivity in place

PR #2770, opened 2025-04-10
Replace Rose::Dialog::Cover with HDS::ApplicationState

priya-patel04:
# Description
Replace Rose::Dialog::Cover with HDS::ApplicationState

:tickets: [Jira ticket](https://hashicorp.atlassian.net/browse/ICU-16045)

#### Before:

Admin

https://github.com/user-attachments/assets/a409d4af-ffd2-4037-a5a1-cb380ad9fc49

Desktop

https://github.com/user-attachments/assets/572d875e-1dfb-43d0-80d9-8cd78c7c07b2

#### After:

Admin

https://github.com/user-attachments/assets/aa2e4e99-24c9-4c98-aa6b-042bc19c4749

Desktop

https://github.com/user-attachments/assets/178b5358-88d7-4796-8341-0bd81fe7a776

## How to Test
1. Go to sign-in page
2. Choose oidc auth method to sign in
3. Verify HDS auth pending component is rendered

## Checklist


ui/admin/config/content-security-policy.js
```diff
@@ -27,6 +27,7 @@ module.exports = function (environment) {
   if (environment === 'development') {
     policy['script-src'].push("'unsafe-eval'");
     policy['style-src'].push("'unsafe-inline'");
+    policy['style-src-attr'].push("'unsafe-inline'");
```
priya-patel04, line 30: Added this to fix csp issue:

Screenshot 2025-04-09 at 11 46 32 PM

addons/core/translations/resources/en-us.yaml
```diff
@@ -72,6 +72,7 @@ auth-method:
     delete: Delete Auth Method
     make-primary: Make Primary
     remove-as-primary: Remove as Primary
+    dont-see-new-window-retry: Don't see the new window? Retry
```
laurenolivia, line 75: nit: Wondering what you think about shortening this to
```
no-new-window-retry: Don't see the new window? Retry
```

ui/admin/config/content-security-policy.js
laurenolivia, line 30: Can you share more context about this fix? Asking for learning.

ui/desktop/app/controllers/scopes/scope/authenticate/method/oidc.js
```diff
@@ -13,24 +13,28 @@ export default class ScopesScopeAuthenticateMethodOidcController extends Control
...
     await this.authenticateMethod.startOIDCAuthentication(authenticatorName, {
       scope,
       authMethod: this.authMethod,
     });
   }
+
+  /**
+   * Route back to sign-in page.
+   */
+  @action
```
calcaide, line 36: Nice 😉

calcaide (dismissed):
Thank you for the work!! the application state with the illustrations looks fresh 💪 🙌

cameronperera (dismissed):
LGTM

DhariniJeeva:
Screenshot 2025-04-14 at 10 39 32 AM

I'm seeing some layout on the background which doesn't look it is supposed to be part of the icon. Can we double check if this background layout is intentional?

ui/admin/config/content-security-policy.js
DhariniJeeva: curios - did you try a prod build to see if this happens on it as well?

ui/desktop/app/templates/scopes/scope/authenticate.hbs
```diff
@@ -6,67 +6,71 @@
...
-      <br />
-      <LinkTo @route='cluster-url'>
-        {{t 'actions.change-cluster-url'}}
-      </LinkTo>
-    </p>
-  </BrandedCard>
+      {{#if @model.authMethods}}
+        {{outlet}}
+      {{/if}}
+      {{#unless @model.authMethods}}
```
hashicc, line 53: Unrelated to a change in your PR but could this be refactored to an `{{#if}}{{else}}{{/if}}` for the `@model.authMethods` case

ui/desktop/app/templates/scopes/scope/authenticate/method/oidc.hbs
```diff
@@ -3,22 +3,22 @@
...
+  </A.Media>
+  <A.Header @title={{t 'resources.auth-method.messages.pending.title'}} />
+  <A.Body @text={{t 'resources.auth-method.messages.pending.description'}} />
+  <A.Footer as |F|>
+    <F.Button
+      @text={{t 'resources.auth-method.actions.dont-see-new-window-retry'}}
+      @color='primary'
+      {{on 'click' this.retryAuthentication}}
+    />
+    <F.Button
```
hashicc, line 18: If you wanted to preserve the existing behavior instead of needing the `this.cancelAuthentication` you could use the `@route='scopes.scope.authenticate.method'` since [hds buttons supports links as buttons](https://helios.hashicorp.design/components/button#links-as-buttons).

hashicc (dismissed):
Nice work! The UI looks much better using the application state pattern.

My learning takeaway is that `Rose::Dialog::Cover` and `Hds::ApplicationState` aren't entirely interchangeable but overlap a bit. It can be looked at as two different changes:
1. Instead of using `Rose::Dialog::Cover` that showed the existing rendered content under, this removes it and in the parent template conditionally renders the `{{outlet}}` for OIDC login route so it's no longer like a "floated UI" (see before/after videos in pr description).
2. With the `Rose::Dialog::Cover` removed, instead a new UI layout pattern using `ApplicationState` for the OIDC page is used

I have one comment about how the cancel buttons could use the existing `@route` argument and be links like before. The slight difference is that users could inspect the link before clicking on it, but I don't see that being helpful in this case. They could also open the cancel link in a new tab which also doesn't seem valuable 🤔 The main benefit I suppose would be less lines of code to jump through to the action to see where it's pointing to

priya-patel04:
> Screenshot 2025-04-14 at 10 39 32 AM
> I'm seeing some layout on the background which doesn't look it is supposed to be part of the icon. Can we double check if this background layout is intentional?

The layout in the background is intentional and it is actually part of the icon.

ui/desktop/app/templates/scopes/scope/authenticate/method/oidc.hbs
priya-patel04: Initially I was avoiding to add the link to the button due to a11y concerns. But for inverse group button we can add links to the button. Updated!

ui/desktop/app/templates/scopes/scope/authenticate.hbs
priya-patel04: Yup, updated

hashicc (dismissed):
🚀 `Hds::ApplicationState` is a nice ui pattern and big improvement, thanks for this refactor!

DhariniJeeva:
> > Screenshot 2025-04-14 at 10 39 32 AM
> > I'm seeing some layout on the background which doesn't look it is supposed to be part of the icon. Can we double check if this background layout is intentional?
> 
> The layout in the background is intentional and it is actually part of the icon. In the new svg file added if you look closely there is faint background

can we double check with design if this should be part of the icon? none of the other icons have a background layout as far as I know

ui/admin/config/content-security-policy.js
laurenolivia: Oh, no worries. Thank you 👍

priya-patel04: I was able to resolve it using `` as a workaround. Ideally I think we would need to add 'unsafe-inline' for **style-src-attr** in [boundary](https://github.com/hashicorp/boundary/blob/4304ec0c1c491d1f872189e4c85a7c3bfbd825b1/internal/cmd/config/config.go#L210). But since this illustration is only used once for admin-ui, decided to not make any changes in boundary.

priya-patel04:
> > > Screenshot 2025-04-14 at 10 39 32 AM
> > > I'm seeing some layout on the background which doesn't look it is supposed to be part of the icon. Can we double check if this background layout is intentional?
> > 
> > 
> > The layout in the background is intentional and it is actually part of the icon. In the new svg file added if you look closely there is faint background
> 
> can we double check with design if this should be part of the icon? none of the other icons have a background layout as far as I know

Sure! I have confirmed with design, this is the right illustration to use for this pending state component.

DhariniJeeva (approved):
looks good!

cameronperera (approved):

PR #2772, opened 2025-04-11
feat: Clear flash messages during login

sohail-hashicorp:
# Description
Display of incorrect UI error message when the user tries to login after unsuccessful attempt.

## Short Summary of the Issue
A [ticket](https://hashicorp.atlassian.net/browse/ICU-16227) has been raised with the issue where the user encounter an error message after attempting to log in to Boundary using OpenID Connect (OIDC) with Vault as the authentication provider. While the login flow works as expected when the token is valid, user experienced unexpected behavior once the token expires - despite re-authenticating successfully, an error is displayed.

## RCA Details  
**1. Handling Expiry of Tokens**

When a user's OIDC token (from Vault) expires, they are redirected to the Vault login page for re-authentication. However, Vault fails to return a valid token back to Boundary after this process. As a result, Boundary receives an invalid token and responds with a 500 error.

- **Root Cause:** Vault does not return a fresh or valid token ID to Boundary after re-authentication. This causes Boundary to treat the session as invalid.
- **Note:** This issue has been [confirmed](https://hashicorp.slack.com/archives/C04L34UGZ/p1743579285190629) and is [tracked](https://hashicorp.atlassian.net/browse/VAULT-32519) separately

**2. Sticky Error Message on UI**

Once the 500 error is received due to the invalid token, an error message is shown in the UI. If the user subsequently logs in successfully (e.g., by manually authenticating through Vault in another tab), the previous error message still persists.

- **Root Cause:** Error messages are not automatically cleared when the login button is clicked again, resulting in stale messages being shown post-successful login.

Solution provided for Issue 2 (Sticky Error Message)
To resolve the sticky error message, considered the below solution:

**Message Reset on Login**

Clear all existing success or error messages when the user initiates a new login attempt. This ensures that only the most recent status message is shown.

:tickets: [Jira ticket](https://hashicorp.atlassian.net/browse/ICU-16227) 

Before:
image

After:
image

## How to Test

1. **Set Up Vault as an OIDC Provider**
To get started, create a Vault as an OIDC provider for the login, you can refer the [documentation](https://developer.hashicorp.com/vault/tutorials/auth-methods/oidc-identity-provider) for the setup process and use "HCP Vault Dedicated" because token has expiry time here.
Note: Use HCP Vault Dedicated, as tokens here have an expiration time, which is essential for testing.

2. **Run the UI with Mirage Disabled**
Ensure that ENABLE_MIRAGE flag is set to false and run the below command: 
`ENABLE_MIRAGE=false yarn start:desktop`

3. **Login via Vault**
Select the Vault as the login provide and login. If the token is valid, it should take you to the home page of the boundary.

4. **Test Invalid Token Scenario**
If the token is invalid, you will be redirected to Vault's login UI page to re-authenticate. Currently the vault's UI has a known issue of not passing the correct token and the [ticket](https://hashicorp.atlassian.net/browse/VAULT-32519) is open. After few minutes the API server will return a 500 error indicating an invalid token, and an error message will be displayed to the user.
image
image

5. **Workaround for Vault UI Issue**
We can bypass the above issue from Vault's UI by re-authenticating the vault from the browser by navigating to the HCP Vault Dedicated's page from the [HCP portal](https://portal.cloud.hashicorp.com/) - (configured in the step 1 above)

6. **Verify Successful Re-authentication**
After successful attempt, repeat the step 3 to login the error message should no longer be displayed.

7. **Reference Video**
A video demonstrating both the issue and the fix is attached for reference.

Before Fix:
https://github.com/user-attachments/assets/b73238f3-d118-41eb-91a1-c549ff181df9

After Fix:
https://github.com/user-attachments/assets/a04a8fd2-772d-4a34-9597-49bc00fc87c1

## Checklist

- [ ] ~~I have added JSON response output for API changes~~

### DIFF
--- a/ui/desktop/app/routes/scopes/scope/authenticate/method/oidc.js
+++ b/ui/desktop/app/routes/scopes/scope/authenticate/method/oidc.js
@@ -18,8 +18,6 @@ export default class ScopesScopeAuthenticateMethodOidcRoute extends Route {
 
   @service session;
   @service router;
-  @service flashMessages;
-  @service intl;
 
   // =attributes
 
--- a/ui/desktop/app/routes/scopes/scope/authenticate/method/oidc.js
+++ b/ui/desktop/app/routes/scopes/scope/authenticate/method/oidc.js
@@ -19,6 +19,7 @@ export default class ScopesScopeAuthenticateMethodOidcRoute extends Route {
   @service session;
   @service router;
   @service flashMessages;
+  @service intl;
 
   // =attributes
 
@@ -55,7 +56,11 @@ export default class ScopesScopeAuthenticateMethodOidcRoute extends Route {
    * When this route is activated (entered), start polling for authentication.
    */
   activate() {
-    this.flashMessages.clearMessages(); // clear any ui messages during login
+    const prevAuthFailedMessage = this.flashMessages.queue.find(
+      (flash) =>
+        flash.message === this.intl.t('errors.authentication-failed.title'),
+    );
+    prevAuthFailedMessage?.destroyMessage();
     this.poller.perform();
   }
 
--- a/ui/desktop/app/routes/scopes/scope/authenticate/method/oidc.js
+++ b/ui/desktop/app/routes/scopes/scope/authenticate/method/oidc.js
@@ -56,11 +56,6 @@ export default class ScopesScopeAuthenticateMethodOidcRoute extends Route {
    * When this route is activated (entered), start polling for authentication.
    */
   activate() {
-    const prevAuthFailedMessage = this.flashMessages.queue.find(
-      (flash) =>
-        flash.message === this.intl.t('errors.authentication-failed.title'),
-    );
-    prevAuthFailedMessage?.destroyMessage();
     this.poller.perform();
   }
 
@@ -77,7 +72,10 @@ export default class ScopesScopeAuthenticateMethodOidcRoute extends Route {
    * notified and returned to the index.
    */
   @action
-  @notifyError(() => 'errors.authentication-failed.title', { catch: true })
+  @notifyError(() => 'errors.authentication-failed.title', {
+    catch: true,
+    sticky: false,
+  })
   error(e) {
     this.router.transitionTo('scopes.scope.authenticate.method.index');
     // rethrow the error to activate the notifyError decorator
--- a/ui/desktop/app/routes/scopes/scope/authenticate/method/oidc.js
+++ b/ui/desktop/app/routes/scopes/scope/authenticate/method/oidc.js
@@ -18,6 +18,7 @@ export default class ScopesScopeAuthenticateMethodOidcRoute extends Route {
 
   @service session;
   @service router;
+  @service flashMessages;
 
   // =attributes
 
@@ -46,6 +47,7 @@ export default class ScopesScopeAuthenticateMethodOidcRoute extends Route {
    */
   setupController(controller) {
     super.setupController(...arguments);
+    this.flashMessages.clearMessages(); // clear any ui messages during login
     const authMethod = this.modelFor('scopes.scope.authenticate.method');
     controller.authMethod = authMethod;
   }
--- a/ui/desktop/app/routes/scopes/scope/authenticate/method/oidc.js
+++ b/ui/desktop/app/routes/scopes/scope/authenticate/method/oidc.js
@@ -47,7 +47,6 @@ export default class ScopesScopeAuthenticateMethodOidcRoute extends Route {
    */
   setupController(controller) {
     super.setupController(...arguments);
-    this.flashMessages.clearMessages(); // clear any ui messages during login
     const authMethod = this.modelFor('scopes.scope.authenticate.method');
     controller.authMethod = authMethod;
   }
@@ -56,6 +55,7 @@ export default class ScopesScopeAuthenticateMethodOidcRoute extends Route {
    * When this route is activated (entered), start polling for authentication.
    */
   activate() {
+    this.flashMessages.clearMessages(); // clear any ui messages during login
     this.poller.perform();
   }
 
--- a/ui/desktop/tests/unit/routes/scopes/scope/authenticate/method/oidc-test.js
+++ b/ui/desktop/tests/unit/routes/scopes/scope/authenticate/method/oidc-test.js
@@ -1,3 +1,8 @@
+/**
+ * Copyright (c) HashiCorp, Inc.
+ * SPDX-License-Identifier: BUSL-1.1
+ */
+
 import { module, test } from 'qunit';
 import { setupTest } from 'ember-qunit';
 import sinon from 'sinon';
--- a/ui/desktop/tests/unit/routes/scopes/scope/authenticate/method/oidc-test.js
+++ b/ui/desktop/tests/unit/routes/scopes/scope/authenticate/method/oidc-test.js
@@ -5,129 +5,17 @@
 
 import { module, test } from 'qunit';
 import { setupTest } from 'ember-qunit';
-import sinon from 'sinon';
 
 module(
   'Unit | Route | scopes/scope/authenticate/method/oidc',
   function (hooks) {
     setupTest(hooks);
 
-    let route, flashMessagesService, intlService;
-    let originalQueue;
-
-    function createFlashMessage(message, type = 'danger') {
-      return {
-        message,
-        type,
-        // Add destroyMessage method to the flash message object
-        // This is needed because we're using a simplified mock queue in tests rather than
-        // the actual flash messages implementation. In production, flash messages would have
-        // this method natively, but we need to add it manually for testing purposes.
-        destroyMessage() {
-          const index = flashMessagesService.queue.indexOf(this);
-          if (index !== -1) {
-            flashMessagesService.queue.splice(index, 1);
-          }
-        },
-      };
-    }
-
-    hooks.beforeEach(function () {
-      route = this.owner.lookup('route:scopes/scope/authenticate/method/oidc');
-      flashMessagesService = this.owner.lookup('service:flashMessages');
-      intlService = this.owner.lookup('service:intl');
-
-      sinon
-        .stub(intlService, 't')
-        .withArgs('errors.authentication-failed.title')
-        .returns('Authentication Failed');
-
-      // Need to stub the exists method to avoid "Cannot read properties of undefined (reading 'some')" error
-      sinon.stub(intlService, 'exists').returns(true);
-
-      // Store original queue and create a mock queue
-      originalQueue = flashMessagesService.queue;
-      flashMessagesService.queue = [];
-      flashMessagesService.queue.clear = function () {
-        this.length = 0;
-      };
-
-      sinon
-        .stub(flashMessagesService, 'danger')
-        .callsFake(function (message, opts) {
-          const flashMessage = createFlashMessage(message);
-          this.queue.push({ ...flashMessage, ...opts });
-        });
-
-      // Stub refresh method that is called by the poller task
-      // Without this stub, the route.activate() call would trigger the poller which attempts
-      // to call refresh(), causing "Cannot read properties of undefined" errors since the
-      // actual refresh implementation relies on router and model hooks not available in unit tests
-      route.refresh = () => {};
-    });
-
-    hooks.afterEach(function () {
-      flashMessagesService.queue = originalQueue;
-      sinon.restore();
-    });
-
-    test('it shows and clears flash message on error and activate', async function (assert) {
-      assert.expect(4);
-
-      const error = new Error('Authentication Failed');
-      const expectedErrorMessage = 'Authentication Failed';
-
-      route.router = { transitionTo: sinon.stub() };
-
-      await route.error(error);
-
-      assert.ok(
-        route.router.transitionTo.calledWith(
-          'scopes.scope.authenticate.method.index',
-        ),
-        'Router transition was triggered',
+    test('it exists', function (assert) {
+      let route = this.owner.lookup(
+        'route:scopes/scope/authenticate/method/oidc',
       );
-
-      assert.strictEqual(
-        flashMessagesService.queue.length,
-        1,
-        'Flash message was added by notifyError',
-      );
-
-      assert.strictEqual(
-        flashMessagesService.queue[0].message,
-        expectedErrorMessage,
-        'Correct error message is displayed',
-      );
-
-      route.activate();
-
-      assert.strictEqual(
-        flashMessagesService.queue.length,
-        0,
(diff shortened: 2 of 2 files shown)


calcaide (approved):
Nice work fixing this issue, and clever test to validate the work!! 🙌 💪

ui/desktop/app/routes/scopes/scope/authenticate/method/oidc.js
```diff
@@ -46,6 +47,7 @@ export default class ScopesScopeAuthenticateMethodOidcRoute extends Route {
    */
   setupController(controller) {
     super.setupController(...arguments);
+    this.flashMessages.clearMessages(); // clear any ui messages during login
```
laurenolivia, line 50: Ah, good idea!

laurenolivia (dismissed):
Nice work. The PR description and testing plan were very well written!

Also, welcome to Boundary!

ui/desktop/tests/unit/routes/scopes/scope/authenticate/method/oidc-test.js
```diff
@@ -17,5 +20,72 @@ module(
...
+        controller.authMethod,
+        fakeAuthMethod,
+        'controller.authMethod was set correctly',
+      );
+    });
+
+    test('setupController clears flash messages', function (assert) {
+      assert.expect(2);
+
+      // Get the route instance
```
cameronperera, line 55: nitpick: Do these comments provide extra context that the test itself does not? While comments can be helpful, in this scenario, the test is well written and formatted that it is easy to read and follow along. I would lean towards not including them here.

cameronperera (commented):
LGTM. I left a small nitpick in one of the tests.

ui/desktop/tests/unit/routes/scopes/scope/authenticate/method/oidc-test.js
```diff
@@ -17,5 +20,72 @@ module(
...
+    test('setupController clears UI messages and sets authMethod on the controller', function (assert) {
+      assert.expect(2);
+
+      class FlashMessagesStub extends Service {
+        clearMessages() {
+          assert.ok(true, 'flashMessages.clearMessages was called');
+        }
+      }
+
+      this.owner.register('service:flash-messages', FlashMessagesStub);
```
hashicc, line 33: Can you spy on the the existing service?

```js
// in test
const flashMessagesService = this.owner.lookup("service:flash-messages");
const clearMessagesSpy = sinon.spy(flashMessagesService, "clearMessages");

// after action
assert.ok(clearMessages.calledOnce);
```

ui/desktop/tests/unit/routes/scopes/scope/authenticate/method/oidc-test.js
```diff
@@ -17,5 +20,72 @@ module(
...
+      // Get the route instance
+      const route = this.owner.lookup(
+        'route:scopes/scope/authenticate/method/oidc',
+      );
+
+      // Mock the flashMessages service
+      const flashMessages = {
+        clearMessages: sinon.spy(),
+      };
+      route.flashMessages = flashMessages;
```
hashicc, line 64: Similar to my other comment can you spy on the service as is? It would be great if the mocking was setup the same between these two tests.

ui/desktop/tests/unit/routes/scopes/scope/authenticate/method/oidc-test.js
```diff
@@ -17,5 +20,72 @@ module(
...
+
+      // Mock the controller and model
+      const controller = {};
+      const authMethod = { id: 'test-auth-method' };
+      sinon
```
hashicc, line 69: Similar consistency for stubbing the `modelFor` would be good. It might be useful to set these up in a `beforeEach` hook

hashicc (changes requested):
A few requests around using sinon on the existing flash messages service and also setting up the two tests in the same way would help with readability. Thanks for the deep dive into this. I'll try and give it a manual test tomorrow

sohail-hashicorp:
Thanks @cameronperera and @hashicc for the comments. I have updated the test case, felt like one test case was redundant and have removed that. Here are the latest changes.

ui/desktop/tests/unit/routes/scopes/scope/authenticate/method/oidc-test.js
```diff
@@ -5,17 +5,43 @@
...
-      let route = this.owner.lookup(
-        'route:scopes/scope/authenticate/method/oidc',
-      );
       assert.ok(route);
     });
+
+    test('setupController clears UI messages and sets authMethod on the controller', function (assert) {
+      assert.expect(2);
+
+      let flashMessagesService, clearMessagesSpy;
```
DhariniJeeva, line 27: can these be consts?  since it does not look like we are reassigning

ui/desktop/tests/unit/routes/scopes/scope/authenticate/method/oidc-test.js
```diff
@@ -5,17 +5,43 @@
...
+      const fakeAuthMethod = { name: 'OIDC', type: 'oidc' };
+
+      flashMessagesService = this.owner.lookup('service:flash-messages');
+      clearMessagesSpy = sinon.spy(flashMessagesService, 'clearMessages');
+      route.flashMessages = flashMessagesService;
+
+      sinon.stub(route, 'modelFor').returns(fakeAuthMethod);
```
hashicc, line 34:
```js
sinon.stub(route, 'modelFor').withArgs('scopes.scope.authenticate.method').returns(fakeAuthMethod);
```
The `withArgs` makes it a bit more explicit and let's us know which model we are retrieving from in the test

ui/desktop/tests/unit/routes/scopes/scope/authenticate/method/oidc-test.js
```diff
@@ -5,17 +5,43 @@
...
 import { setupTest } from 'ember-qunit';
+import Controller from '@ember/controller';
+import sinon from 'sinon';
 
 module(
   'Unit | Route | scopes/scope/authenticate/method/oidc',
   function (hooks) {
     setupTest(hooks);
 
+    let route = this.owner.lookup(
```
hashicc, line 16: I tried to run this test locally and it wasn't listed and all tests were passing. As a quick test if you replaced this line with `throw new Error("hello");` then all the tests will still pass, but this module won't be listed. In both cases you will see the error in the console though which is what tipped me off: 
```
wrap-require.js:15 TypeError: Cannot read properties of undefined (reading 'lookup')
    at Object.<anonymous> (oidc-test.js:16:28)
```

In order to capture this type of error it needs to be within a test or hook:
```js
let route;

// the function passed to the hooks also needs to be a function that is not an arrow function so that `this` can be set on it
hooks.beforeEach(function () {
  route = this.owner.lookup('route:scopes/scope/authenticate/method/oidc');
});

File: ui/desktop/tests/unit/routes/scopes/scope/authenticate/method/oidc-test.js [hunk 9]
```diff
@@ -5,17 +5,43 @@
...
+      let flashMessagesService, clearMessagesSpy;
+      const fakeAuthMethod = { name: 'OIDC', type: 'oidc' };
+
+      flashMessagesService = this.owner.lookup('service:flash-messages');
+      clearMessagesSpy = sinon.spy(flashMessagesService, 'clearMessages');
+      route.flashMessages = flashMessagesService;
+
+      sinon.stub(route, 'modelFor').returns(fakeAuthMethod);
+
+      const controller = new (class extends Controller {})();
```
hashicc, line 36: Instead of creating an anonymous controller you can lookup the actual controller:
```js
const controller = this.owner.lookup('controller:scopes/scope/authenticate/method/oidc');
```

ui/desktop/app/routes/scopes/scope/authenticate/method/oidc.js
hashicc, line 50: I'm not sure `setupController` is the right hook on the route:
> A hook you can use to setup the controller for the current route.

If the aim is to clear the flash messages when entering the route then `activate` is probably better:
> This hook is executed when the router enters the route. It is not executed when the model for the route changes.

hashicc (changes requested):
Thanks for testing this bug and figuring it out. It seems quite involved to get everything setup to properly exercise the bug! I left some feedback on getting the test fixed so that it's running and possibly a better hook for clearing the messages.

Also can you help me with getting this set up, I wanted to test the bug locally. I was able to setup the vault cluster with hashicorp cloud and set it up as an oidc provider and log in successfully with a local boundary dev. I'm not sure how to get the token to a point where it fails and I see the error flash message.

ui/desktop/app/routes/scopes/scope/authenticate/method/oidc.js
sohail-hashicorp: Aha! You are correct, we can add the clearing of the flash message in the activate method. Have updated the code.

sohail-hashicorp:
@hashicc, if you'd like to test token expiration behavior, you can adjust the token's TTL (Time To Live) by updating the configuration in the Vault UI. Here's how you can do it:

1. Go to: [HCP portal](https://portal.cloud.hashicorp.com)
2. Select **Vault Dedicated**.
3. Choose your Vault cluster.
4. Click on **Launch Web UI** in the top right corner.
5. In the Web UI, navigate to the left menu and click **Access**.
6. Under Authentication Methods, click on **userpass/** — this is the method configured via our docs.
7. Select the **end-user** entry (this name might vary if you’ve used a custom username).
8. Click **Edit User** (top right corner).
9. Scroll down, expand the **Tokens** section.
10. Update the **Generated Token’s Initial TTL** field to a value like `2`, `5`, or `10` minutes — this sets how long the token will remain valid.

I've also recorded a short video which showcase the above steps:

https://github.com/user-attachments/assets/898a6063-9541-4c15-83b2-eb8ba4358346

ui/desktop/app/routes/scopes/scope/authenticate/method/oidc.js
hashicc: @sohail-hashicorp I saw that this was marked resolved but the `activate` method wasn't used. Does that still work as a hook to use for fixing this bug?

hashicc:
Thanks for steps, that was really helpful! I was able to go in and make the changes for the token.


I was digging into this a bit further and was thinking of ways that we can clear just the specific message in case other errors were encountered on the way to the oidc route and we wouldn't want to clear those as well. It looks like this message is [triggered here](https://github.com/hashicorp/boundary-ui/blob/e8fc6d163bbcc665e80cd9e6377df36618aa78dd/ui/desktop/app/routes/scopes/scope/authenticate/method/oidc.js#L73).  And the `notifyError` decorator dispatches the toast message [here](https://github.com/hashicorp/boundary-ui/blob/e8fc6d163bbcc665e80cd9e6377df36618aa78dd/addons/core/addon/decorators/notify.js#L101-L105). With this information could we find find the specific message:
```js
// with injected intl service, find the matching message that matches the same key used dispatch
const previousAuthenticationFailedFlashMessage = this.flashMessages.queue.find(flash => flash.message === this.intl.t("errors.authentication-failed.title");
// if it's not-nullish call `destroyMessage` on the message
previousAuthenticationFailedFlashMessage?.destroyMessage();
```

This is similar to how the `clearMessages` works but instead of finding a specific message to clear, [it clears all of them in a `forEach`](https://github.com/adopted-ember-addons/ember-cli-flash/blob/master/ember-cli-flash/src/services/flash-messages.js#L53).

I think in the test you might also be able to:
1. call the `ScopesScopeAuthenticateMethodOidcRoute.error(e)` method instead which would trigger `@notifyError`
2. Assert that the flash message service has one message and the name of that message matches the translated key of `"errors.authentication-failed.title"`
3. Then call `activate` to simulate re-entering the route
4. Assert the flash message service has zero messages

The benefit of this is the route ever gets refactored this test will likely fail (if either the message gets changed, or the notify error is removed) and will also need to be refactored. This might be doable as an acceptance test too.

ui/desktop/app/routes/scopes/scope/authenticate/method/oidc.js
sohail-hashicorp: My bad, I forgot to push the commit with this change.

sohail-hashicorp:
> Thanks for steps, that was really helpful! I was able to go in and make the changes for the token.
> 
> I was digging into this a bit further and was thinking of ways that we can clear just the specific message in case other errors were encountered on the way to the oidc route and we wouldn't want to clear those as well. It looks like this message is [triggered here](https://github.com/hashicorp/boundary-ui/blob/e8fc6d163bbcc665e80cd9e6377df36618aa78dd/ui/desktop/app/routes/scopes/scope/authenticate/method/oidc.js#L73). And the `notifyError` decorator dispatches the toast message [here](https://github.com/hashicorp/boundary-ui/blob/e8fc6d163bbcc665e80cd9e6377df36618aa78dd/addons/core/addon/decorators/notify.js#L101-L105). With this information could we find find the specific message:
> 
> ```js
> // with injected intl service, find the matching message that matches the same key used dispatch
> const previousAuthenticationFailedFlashMessage = this.flashMessages.queue.find(flash => flash.message === this.intl.t("errors.authentication-failed.title");
> // if it's not-nullish call `destroyMessage` on the message
> previousAuthenticationFailedFlashMessage?.destroyMessage();
> ```
> 
> This is similar to how the `clearMessages` works but instead of finding a specific message to clear, [it clears all of them in a `forEach`](https://github.com/adopted-ember-addons/ember-cli-flash/blob/master/ember-cli-flash/src/services/flash-messages.js#L53).
> 
> I think in the test you might also be able to:
> 
> 1. call the `ScopesScopeAuthenticateMethodOidcRoute.error(e)` method instead which would trigger `@notifyError`
> 2. Assert that the flash message service has one message and the name of that message matches the translated key of `"errors.authentication-failed.title"`
> 3. Then call `activate` to simulate re-entering the route
> 4. Assert the flash message service has zero messages
> 
> The benefit of this is the route ever gets refactored this test will likely fail (if either the message gets changed, or the notify error is removed) and will also need to be refactored. This might be doable as an acceptance test too.

@hashicc Thanks for the suggestion, that is a nice way to do it.

I have updated the code and the test cases with the above suggestion, can you have a look again. Thank you.

ui/desktop/app/routes/scopes/scope/authenticate/method/oidc.js
sohail-hashicorp: Have updated the PR to just be non-sticky in nature.

hashicc (approved):
🎉 I know there was a bit of back and forth on this one. I appreciate you working on it and getting it to this state. Thank you!

PR #2773, opened 2025-04-11
Add sorting foundation as util and using it within indexed-db-handler

calcaide:
# Description

This pull request adds the ability to specify sorting options when querying the store:
```js
this.store.query('target', { sort: { attribute: 'id', direction: 'desc' }});
```

:tickets: [Jira ticket](https://hashicorp.atlassian.net/browse/ICU-16778)

## How to Test
- Run tests within `addons/api`


addons/api/addon/utils/indexed-db-query.js
```diff
@@ -70,14 +70,7 @@ export async function queryIndexedDb(indexedDb, resource, query) {
...
 
-  // OrderBy has to be used on where clauses while if it's a
-  // collection already we have to use sortBy. We'll also sort by
-  // descending order based on created time to match the API as any
-  // "or" clause can change the order
-  if (filterCollection.orderBy) {
-    return filterCollection.orderBy(getKey('created_time')).reverse().toArray();
-  }
-  return filterCollection.reverse().sortBy(getKey('created_time'));
+  return filterCollection.toArray();
```
hashicc, line 73: I think the existing comment was checking `orderBy` if it was a table or `sortBy` if it was a collection. Luckily both have a `toArray`. Do we capture anywhere that this is a collection or table? I'm wondering if there's a way we can make that more apparent for future work in this area. No worries if not, was just a learning that would be useful to capture in some way.

addons/api/tests/unit/handlers/indexed-db-handler-test.js
```diff
@@ -191,5 +191,73 @@ module('Unit | Handler | indexed-db-handler', function (hooks) {
...
+        query: { sort: { attribute: 'name', direction: 'asc' } },
+      });
+      assert.deepEqual(
+        resultsDesc.map(({ name }) => name),
+        ['Alias 004', 'Alias 003', 'Alias 002', 'Alias 001'],
+      );
+      assert.deepEqual(
+        resultsAsc.map(({ name }) => name),
+        ['Alias 001', 'Alias 002', 'Alias 003', 'Alias 004'],
+      );
```
hashicc, line 260: very clear! ✨

addons/api/addon/utils/sort-api-records.js
```diff
@@ -0,0 +1,51 @@
...
+  string: (a, b) => String(a).localeCompare(String(b)),
+  date: (a, b) => new Date(a).getTime() - new Date(b).getTime(),
+};
+
+export const sortApiRecords = (records, { querySort, schema }) => {
+  // The consumer is already doing this, not sure is necessary
+  querySort = querySort ?? {};
+
+  // Why the spike uses an assertion to check the attribute exists? - Ask Chad about it.
+  // Perhaps what we should do is throw an error if the attribute does not exists?
```
hashicc, line 22: `assert` from `@ember/debug` does throw an error if it doesn't exist (if that's what the assertion tests).

```js
assert("assertion text to show in error, truthyCaseToAssert);
```

is roughly the same as
```js
if (!truthyCase) {
  throw new Error("error text");
}
```

You get one special ability with `assert` from `@ember/debug` thought. It gets stripped out in production builds. This means in dev and tests you'll get the feedback but in production you won't which is a good thing if you don't know handle the error but also don't want to crash

addons/api/addon/utils/sort-api-records.js
```diff
@@ -0,0 +1,51 @@
...
+const SORT_DIRECTION_DESCENDING = 'desc';
+
+const sortFunctions = {
+  string: (a, b) => String(a).localeCompare(String(b)),
+  date: (a, b) => new Date(a).getTime() - new Date(b).getTime(),
+};
+
+export const sortApiRecords = (records, { querySort, schema }) => {
+  // The consumer is already doing this, not sure is necessary
+  querySort = querySort ?? {};
```
hashicc, line 19: This could be taken out if you think it's unnecessary

addons/api/tests/unit/handlers/indexed-db-handler-test.js
```diff
@@ -191,5 +191,73 @@ module('Unit | Handler | indexed-db-handler', function (hooks) {
         ['Target 2023', 'Target 2022', 'Target 2021', 'Target 2020'],
       );
     });
+
+    test('it sorts by ascending (default) `name` using string sorting', async function (assert) {
```
hashicc, line 195: I like how the default and explicit cases are separate

addons/api/addon/utils/sort-api-records.js
```diff
@@ -0,0 +1,51 @@
+/**
+ * Copyright (c) HashiCorp, Inc.
+ * SPDX-License-Identifier: BUSL-1.1
+ */
+
+// created_time is not an attribute defined in models perse, but present in
```
hashicc, line 6:
```suggestion
// created_time is not an attribute defined in models per se, but present in
```

hashicc (commented):
🕺 This looks great, a few minor comments, I think continuing this pattern for the other types with tests is all that's really left

addons/api/addon/utils/indexed-db-query.js
calcaide: @hashicc 
Not sure if I am understanding your comment here or I am missing something.

My thinking here is, why we need to capture if is a table or a collection? the previous implementation was sorting and depending on the type, collection or table, should sort in a way or the other.

The current implementation has no sorting involved, sorting its been done at handler level, therefore we just need to return the collection or table as an array. Since [collection have a toArray()](https://dexie.org/docs/Collection/Collection.toArray()) and [table as well](https://dexie.org/docs/Table/Table.toArray()), I wonder what could be the need of keeping track of the type.

That said, I think its totally fair and need to add a comment here specify in this could be a collection or a table.
I added a simple comment and I am totally open to keep this discussion to leave this in better shape possible for maintenance, let me know your thoughts or if I a missing something.

addons/api/addon/utils/sort-api-records.js
calcaide: @hashicc 
I add this comment to remind myself to discuss it. The consumer is already checking on this so it's not necessary... BUT for the cost of it I think its a fair safety net, so I will leave it. No one got hurt by checking :P unless you think otherwise

addons/api/addon/utils/sort-api-records.js
calcaide: @hashicc 

Very interesting, I didn't though about the assert being stripped out in production code.

So my initial though here was let's throw an error, but your comment got me thinking. If we throw an error, are we going to deal with the error or just throw the error? And if so, do we inform the user this error is happening? are we going to log this error anywhere?

The answers to those questions, right now, is NO. We are not informing the user, neither handling the error. Therefore, the usage of assert will throw the error while testing but not in production.

Unless you think we should be doing something with this kind of errors, I think going with assertion is the way to go. What are your thoughts?

addons/api/addon/utils/indexed-db-query.js
hashicc: My comment is more about that we're calling it a collection `filterCollection` and could be a dexie `Table` and to re-learn that it could be either is a bit difficult without typescript. It's funny because the first reference is `let filterCollection = indexedDb[resource];` and at this point it's a table not a collection, you can see that in this bit:
```js
      buildInitialWhereClause({
        key: firstKey,
        table: indexedDb[resource],     // this is a table initially
        filterArrayOrObject: firstFilterArrayOrObject,
      }),
```

In the reduce it may be turned into a collection but it could also still be a table 🤪. Without typescript we don't know and we might write code that assumes it's a collection (like we had to before with `sortBy` and `orderBy`). So maybe a comment saying as such or renaming the variable `filterTableOrCollection`? Or maybe there's some other way that we can add clarity for the next time we're in this file.

addons/api/addon/utils/sort-api-records.js
hashicc: Sounds good to me! I think for this since it's optional by specifying the obejct it saves you from having to do optional chaining everywhere.

addons/api/addon/utils/sort-results.js
```diff
@@ -0,0 +1,48 @@
...
+  // Perhaps what we should do is throw an error if the direction is not correct?
+
+  const sortAttributeDataType = schema.attributes.get(sortAttribute)?.type;
+  const sortFunction =
+    sortFunctions[sortAttributeDataType] ?? sortFunctions.string;
+
+  return results.sort((a, b) => {
+    // Extract the values to sort and sort them.
+    const sortValueA = a.attributes[sortAttribute];
+    const sortValueB = b.attributes[sortAttribute];
```
ZedLi, line 41: Will we ever need to worry about fields that can be nested and aren't top level? e.g. `attributes.plugin.name`

hashicc: Is this a case where a model has an attribute that's an object? We're focused only on primitive `attr` types for the initial pass. That also brings up the case of `id` though since I think that exists outside of the `attributes` object. We should add a test to make sure that this works for `id`s since we know that's a case we need to support.

Now I'm wondering if falling back to a string is worth it if we don't recognize `sortAttributeDataType` as a valid type to sort on. Maybe it's better to throw and that's a case we'd rely on a custom sort function (not in scope for this work)

addons/api/addon/utils/sort-results.js
```diff
@@ -0,0 +1,48 @@
...
+      : SORT_DIRECTION_ASCENDING;
+  const sortDirection = querySort.direction || defaultSortDirection;
+
+  // ToDo: Check sortDirection is valid with assert
+  const sortAttributeDataType = schema.attributes.get(sortAttribute)?.type;
+
+  const sortFunction =
+    sortFunctions[sortAttributeDataType] ?? sortFunctions.string;
+
+  return results.toSorted((a, b) => {
```
calcaide, line 38: Changed from `sort` to `toSorted`. [toSorted](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/toSorted) returns a new array, so when sortResults is call twice rather than return the reference of the results, which will be sortered with the querySort of the last call we always return a new array, guarantee isolation of refernces between calls.

I found out this necessary for the tests where I am calling first with ascending then descending, but if you think this is not need let me know.

addons/api/tests/unit/utils/sort-results-test.js
```diff
@@ -0,0 +1,290 @@
+/**
```
calcaide, line 1: Added this unitary test for sort-results. As its name mentions, unitary testing is focusing solely on the sortResults util function without having to worry about where is used (the indexed-db-handler for the current context).
Despite seems redundant, I think is a must to have, please, let me know your thoughts.

addons/api/addon/utils/sort-results.js
calcaide: As @hashicc mentions, for the first pass we are focusing on primitives and we do not expect nested objects as per sorting resources list, but if you think so or see I missed something, let me know.

Regarding the ID case, I am not sure if I completely understand, isn't the resource ID part of the attributes? default to string I think is the way to go in this case, since ID's are alphanumeric. Please, let me know your thoughts

addons/api/tests/unit/utils/sort-results-test.js
hashicc: I agree, I think focusing on this unit for testing makes sense and can focus on any specific behaviors

addons/api/addon/utils/sort-results.js
hashicc: I tried to write a failing test but it turns out that the results from dexie are already sorted by `id` from the results by `queryIndexedDb`. But I was able to inspect that sorting isn't working on the `id` because it's not within `attributes`:
image

Which is [similar](https://github.com/hashicorp/boundary-ui/blob/9f6168439d54d122546023454b35c87dd56dda34/addons/api/addon/utils/indexed-db-query.js#L206-L219) to how we have to map from filters the json api structure.

I think it'd be valuable to have a test to ensure we don't regress on `id` if we swap out dexie:
```js
  test('it sorts ids', async function (assert) {
    const alias001 = this.server.create('alias', { id: 'alt_1' });
    const alias002 = this.server.create('alias', { id: 'alt_2' });
    const alias003 = this.server.create('alias', { id: 'alt_3' });
    const alias004 = this.server.create('alias', { id: 'alt_4' });

    const shuffledAliases = faker.helpers.shuffle([
      alias001,
      alias002,
      alias003,
      alias004,
    ]);

    this.server.get(
      'aliases',
      createPaginatedResponseHandler(shuffledAliases, {
        pageSize: 1,
      }),
    );

    const results = await store.query('alias', {
      query: { sort: { attribute: 'id' } },
    });

    assert.deepEqual(
      results.map(({ id }) => id),
      ['alt_1', 'alt_2', 'alt_3', 'alt_4'],
    );
  });
```

> Maybe it's better to throw

It might be good to have some guards here otherwise if they sort on attribute `doesNotExist`:
```js
    const results = await store.query('alias', {
      query: { sort: { attribute: 'doesNotExist' } },
    });
```
The results will still be sorted by `id` and that might be surprising since the attribute isn't known and we should have complete knowledge of what is known since we're using the ember-data schema.

addons/api/addon/utils/sort-results.js
hashicc: I like this new api! Yes, let's use `toSorted` 💜

addons/api/addon/utils/sort-results.js
```diff
@@ -0,0 +1,61 @@
...
+const SORT_DIRECTION_DESCENDING = 'desc';
+
+const sortFunctions = {
+  string: (a, b) => String(a).localeCompare(String(b)),
+  date: (a, b) => new Date(a).getTime() - new Date(b).getTime(),
+  number: (a, b) => Number(a) - Number(b),
+  boolean: (a, b) => Number(a) - Number(b),
+};
+
+const getSortValue = (sortValue, sortAttribute) => {
```
calcaide, line 19: @hashicc this is the first iteration I come with to support `id` sorting when not being within attributes.
Be aware the error message can be improved 😅 also not sure how to intl the message within a util yet.

addons/api/addon/utils/sort-results.js
calcaide: @hashicc I updated the unitary test of the sort-results to account for id sorting being not part of the `attributes` and also add a test case to assert we throw an error when attempting to sort by an attribute that is not present.
If that work looks good to you I will proceed to add this test to the indexed-db-handler test file 😉

addons/api/tests/unit/utils/sort-results-test.js
```diff
@@ -0,0 +1,328 @@
...
+      assert.deepEqual(
+        sortedResultsAsc.map(({ attributes: { active } }) => active),
+        [false, false, true, true],
+      );
+
+      assert.deepEqual(
+        sortedResultsDesc.map(({ attributes: { active } }) => active),
+        [true, true, false, false],
+      );
+    });
```
lisbet-alvarez, line 326: Perhaps we could also have a test case where it sorts correctly an attr that is outside the attributes object, even when attributes is defined. Let me know ur thoughts on that.

lisbet-alvarez (approved):
Looks great!! Just left a couple of suggestions

addons/api/addon/utils/sort-results.js
hashicc: Thanks, I expanded on these but this was the type of test I was looking for. Thanks for adding this!

addons/api/addon/utils/sort-results.js
```diff
@@ -0,0 +1,61 @@
+/**
+ * Copyright (c) HashiCorp, Inc.
+ * SPDX-License-Identifier: BUSL-1.1
+ */
+
+// created_time is not an attribute defined in models per se, but present in
```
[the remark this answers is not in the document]
hashicc: That's a good catch. So I think there are two where they are not defined (and maybe should be!)
* [channel-recording](https://github.com/hashicorp/boundary-ui/blob/9b1f052a569fc1579befffda87de3a91ba5e417b/addons/api/addon/generated/models/channel-recording.js#L59-L60) ([rest api](https://github.com/hashicorp/boundary/blob/main/internal/gen/controller.swagger.json#L7220-L7224))
* [connection-recording](https://github.com/hashicorp/boundary-ui/blob/9b1f052a569fc1579befffda87de3a91ba5e417b/addons/api/addon/generated/models/connection-recording.js#L66-L67) ([rest api](https://github.com/hashicorp/boundary/blob/main/internal/gen/controller.swagger.json#L7169-L7173))

I think `created_time` is a special case that we want to handle because it _should_ exist for all records and it's our default sort type so even if a model doesn't have it as part of its definition we still want to support it. I updated this comment too so let met know if that helps or maybe there's a better way this can be phrased

addons/api/tests/unit/utils/sort-results-test.js
hashicc: I [updated one of the tests](https://github.com/hashicorp/boundary-ui/pull/2773/files#diff-dd9c34a3b4e243fe71251d07b995021988553b0926cef5ce6581682f79abc52aR23-R46) that I think handles this case now. Let me know if I'm understanding correctly. So the test has something that exists in the `attributes` json api record but doesn't exist on the schema's `attributes` definition. In those cases (aside from `id` and `created_time`) we will throw because we I'm envisioning that we shouldn't allow "magic sorting" on attributes that are on the record but haven't been well defined on the model.

addons/api/addon/utils/sort-results.js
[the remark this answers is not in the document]
lisbet-alvarez: The new phrasing is very clear, thanks for updating it!

lisbet-alvarez (approved):
Looks great!! Awesome work!

Just left one question.

addons/api/addon/utils/sort-results.js
```diff
@@ -0,0 +1,73 @@
...
+    return record.attributes[attribute];
+  }
+
+  const attributes = Array.from(schema.attributes.keys());
+  throw new Error(
+    `The attribute "${attribute}" does not map to the model definition of type "${record.type}". Supported sortable attributes are 'id', 'created_time', or ${attributes.map((attr) => `'${attr}'`).join(', ')}`,
+  );
+}
+
+export const sortResults = (results, { querySort, schema }) => {
```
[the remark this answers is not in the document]
hashicc: Good catch, that's an interesting case and I don't think we've covered it. If I'm understanding correctly it would be like a multi-sort? Sort first on name and then when name has equal value (`undefined === undefined` in this case) then sort on id? That might be a good candidate for a custom sort. Either that or we just choose one attribute to sort on (`name` and let `undefined` float to the bottom or `id` 🤔).

Do you know which resource we'd run into this case? I could do a quick spike. Let's chat with product about it, too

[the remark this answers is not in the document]
lisbet-alvarez: Yeah custom sort could be the way to go.
Some examples would be roles, groups, scopes. Basically any resource where `name` is optional (we use a `displayName` getter to display either `name` or `id`.

hashicc (approved):
Although I picked up the PR for some tweaks I am going to approve it as is. The need for a custom sorting or multi-sort might be necessary in a follow-up PR

ZedLi (approved):
Looks good!

PR #2774, opened 2025-04-11
Delete electron-squirrel-startup from electron-app

calcaide:
# Description

Delete electron-squirrel-startup from `/electron-app` workspace.

Why we are deleting this dependency?
We are not using this dependency in our codebase. As per electron-squirrel-startup documentation, to use it you need to import it and we aren't doing it.

I also follow the breadcrumbs of the dependency with `yarn why electron-squirrel-startup` and no other dependency is using it.

:tickets: [Jira ticket](https://hashicorp.atlassian.net/browse/ICU-16871)

## Screenshots

Output of `yarn why electron-squirrel-startup`

Screenshot 2025-04-11 at 2 30 26 PM

## How to Test
- Run Desktop e2e test suite successfully.


cameronperera (approved):

hashicc (approved):
This is the [first commit](https://github.com/hashicorp/boundary-ui/commit/efbaed31bed771499a348e96e44b0273135c52d2#diff-a538f59393a12f20501f4890d66b67a0656e571a79d6823f3fc50340e69ce31bR52) I could find where the dependency was introduced and it seems to go back to the inception of the desktop electron app. Digging further it appears that it's installed as part of the electron forge [init process](https://github.com/electron/forge/blob/6b2d547a7216c30fde1e1fddd1118eee5d872945/packages/api/core/src/api/init-scripts/init-npm.ts#L17) and gets used in the [project templates](https://github.com/electron/forge/blob/6b2d547a7216c30fde1e1fddd1118eee5d872945/packages/template/base/tmpl/index.js#L5-L7) that scaffold out the project. You can see in their starter template that we did continue to use the `const createWindow = () => {` pattern and other parts of the template have been expanded as part of [our index.js](https://github.com/hashicorp/boundary-ui/blob/371301b4bad27a5df529de6ec5e84380aabf45d7/ui/desktop/electron-app/src/index.js#L88) but we didn't keep the part:
```js
// Handle creating/removing shortcuts on Windows when installing/uninstalling.
if (require('electron-squirrel-startup')) {
  app.quit();
}
```
And like you mentioned @calcaide, if it's not used anywhere, we likely don't need to include it as a dependency.

calcaide:
@hashicc Thank you for the deep down explanation 🙌 !!

PR #2776, opened 2025-04-14
Icu 16541 upgrade ember addons auth

lisbet-alvarez:
# Description

1. Ran `ember-cli-update --to 5.12`
2. Ran `ember init --no-welcome --yarn`

:tickets: [Jira ticket](https://hashicorp.atlassian.net/browse/ICU-16541)

## How to Test

Tests pass:
image
image

[Test Desktop Workflow passes
](https://github.com/hashicorp/boundary-desktop-releases/actions/runs/14483104698)

## Checklist


addons/auth/.template-lintrc.js
```diff
@@ -6,5 +6,5 @@
 'use strict';
 
 module.exports = {
-  extends: ['recommended', 'stylistic'],
+  extends: 'recommended',
```
calcaide, line 9: None blocking comment: are we deleting `stylistic` in purpose? in my case, when working on core, it deleted the `'ember-template-lint-plugin-prettier:recommended'` after the update comand, but I restore it since its something we still use/need, wondering if this could be same case.

lisbet-alvarez: ah yes it was one of the changes i kept from running `ember init` initially i was thinking its not needed but after [looking at the readme](https://github.com/ember-template-lint/ember-template-lint?tab=readme-ov-file#presets) it is needed since this preset is added when `ember-template-lint-plugin-prettier` isn't being used yet. good catch!!

addons/auth/.prettierrc.js
```diff
@@ -3,6 +3,8 @@
  * SPDX-License-Identifier: BUSL-1.1
  */
 
+'use strict';
```
DhariniJeeva, line 6: did this change anything at all? curious since this a config file that exports an object and strict mode doesn't have any effect. Totally fine to keep it for consistency though.

DhariniJeeva (approved):
left a couple of non blocking questions! great job on getting this updated 🎉

addons/auth/package.json
```diff
@@ -8,64 +8,72 @@
...
-    "format:js": "prettier --write '{addon,app,config,stories,tests}/**/*.js' *.js",
     "start": "ember serve",
-    "test": "ember test --test-port 0",
+    "test": "ember test",
     "test:ember-compatibility": "ember try:each",
     "precommit": "yarn lint && yarn format && yarn test",
     "doc:toc": "doctoc README.md"
   },
   "dependencies": {
+    "@babel/core": "^7.26.10",
```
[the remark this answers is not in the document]
lisbet-alvarez: This was one of the automatic updates when running `ember init` for the new blueprint changes.

addons/auth/.prettierrc.js
lisbet-alvarez: This was also one of the automatic updates when running `ember init` for the new blueprint changes. And I agree its good to keep for consistency.

calcaide (approved):
Run Admin UI and Desktop e2e successfully!! This is awesome!! thanks for the nice upgrade 🚀 💪 🙌

PR #2777, opened 2025-04-15
Ember upgrade for addons/core

calcaide:
# Description

Ember upgrade for addons/core

:tickets: [Jira ticket](https://hashicorp.atlassian.net/browse/ICU-16543)

## How to Test:
- Run tests from root of the project successfully.
- Run Admin UI e2e tests successfully.
- Run Desktop e2e test successfully.


addons/core/tests/integration/helpers/app-name-test.js
```diff
@@ -12,7 +12,7 @@ module('Integration | Helper | app-name', function (hooks) {
   setupRenderingTest(hooks);
 
   test('it renders', async function (assert) {
-    await render(hbs`{{app-name}}`);
+    await render(hbs`<app-name />`);
```
calcaide, line 15: This and the 2 other test files below where [changed due to this eslint rule](https://github.com/ember-template-lint/ember-template-lint/blob/master/docs/rule/no-implicit-this.md). The recommendation from the codemod is to tuse angle bracket when its a helper with no arguments.

calcaide: @aklkv 
You are totally right. After commiting this I realize I miss understood the recomendation, recomendation to use angle bracket was for component but this is a helper with no arguments, therefor make sense to use them with classic invocation.
I opted for adding [the eslint rule exempion ](https://github.com/hashicorp/boundary-ui/pull/2777/files#diff-543650cb72a7d2bc2e89a5d88c4b4a346c39869569c9e4fd52b08c8598a57d78R15), thanks for the catch!! 🙌

addons/core/package.json
```diff
@@ -50,45 +51,45 @@
   },
   "devDependencies": {
     "@babel/core": "^7.26.10",
```
lisbet-alvarez, line 53: looks like the ember upgrade added `@babel/core` to the `dependencies` section as well

addons/core/package.json
```diff
@@ -50,45 +51,45 @@
...
+    "@ember/optional-features": "^2.1.0",
+    "@ember/test-helpers": "^3.3.1",
+    "@embroider/test-setup": "^4.0.0",
     "@faker-js/faker": "^8.0.2",
     "@glimmer/component": "^1.1.2",
     "@glimmer/tracking": "^1.1.2",
     "broccoli-asset-rev": "^3.0.0",
     "concurrently": "^9.1.0",
     "doctoc": "^2.2.0",
     "ember-cli": "^5.12.0",
```
lisbet-alvarez, line 65: when running the cmd to update ember it also suggested to change `"ember-cli": "^5.12.0",` to `"ember-cli": "~5.12.0",`. Should we go with using `~` instead to follow the pattern suggested by ember?

aklkv: true https://github.com/ember-cli/ember-addon-output/blob/v5.12.0/package.json#L46

addons/core/package.json
aklkv: this part is important 👍🏻 https://github.com/ember-cli/ember-addon-output/blob/v5.12.0/package.json#L31

calcaide: Interesting! it add it to dependencies (v7.25.2), but skipped to me that was also present in devDependencies with a newer version v7.26.10.
Since your PR for addons/auth is using the v7.26.10, for consistency I will keep only the v7.26.10 within dependencies.
Thank you for the catch!!

addons/core/package.json
calcaide: 100% this was my bad, it actually changed to `~` but on my pursue of solving the issue with `ember-test-helpers` I had yesterday, I revert that change and forgot to clean it up.
Thank you for the catch!! 💪

addons/core/.eslintignore
```diff
@@ -1,25 +1,17 @@
...
-/vendor/
 
 # compiled output
+/declarations/
 /dist/
-/tmp/
 
 # dependencies
-/bower_components/
 /node_modules/
```
lisbet-alvarez, line 9: i think this `node_modules` specification might not needed anymore since according to the docs node_modules are [ignored by default](https://eslint.org/docs/v8.x/use/configure/ignore#:~:text=In%20addition%20to%20any%20patterns%20in%20the%20.eslintignore%20file%2C%20ESLint%20always%20follows%20a%20couple%20of%20implicit%20ignore%20rules%20even%20if%20the%20%2D%2Dno%2Dignore%20flag%20is%20passed.%20The%20implicit%20rules%20are%20as%20follows%3A). Same thing goes for the `.prettierignore` file ([docs here](https://prettier.io/docs/cli#:~:text=glob%20module.-,Prettier%20CLI%20will%20ignore%20files%20located%20in%20node_modules%20directory,-.%20To%20opt%20out)).

calcaide: Ahhh I was doubting about this one tbh. I understood deleting `bower_components` made sense since bower is long gone, but reading the ignored by default link you provided makes sense to delete it! thanks!

lisbet-alvarez (dismissed):
Just left a small suggestion. Other than that, this looks great!! 🙌

lisbet-alvarez (dismissed):
🥳

DhariniJeeva (dismissed):
left a non blocking question! Thanks for cleaning up the docs and updating this addon! 🎉

addons/core/.watchmanconfig
```diff
@@ -1,3 +1,3 @@
 {
-  "ignore_dirs": ["tmp", "dist"]
+  "ignore_dirs": ["dist"]
```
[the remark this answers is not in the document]
calcaide: It was one of the changes that we use the blueprints (automatic). My thought is that `tmp` was used for internal procedures and maybe they do not use it anymore. I tested building, running test and such and I don't see the creation of `tmp` but I do see `dist`. So despite is an automated change I think it makes sense.

lisbet-alvarez (approved):
🥳

cameronperera (approved):

PR #2783, opened 2025-04-18
[Replace] Rose::Form::Label and <input.field> elements

laurenolivia:
# Description
This PR replaces instances of `<input.label>` with `Hds::Form::Label`, and those instances only exist on our role/grants form.

Update:
Usage of `input.field` cause a11y violations, so we agreed to expand the scope of this ticket to also refactor the input field.

:tickets: [Jira ticket](https://hashicorp.atlassian.net/browse/ICU-16831)

Before 👇 
Screenshot 2025-04-18 at 12 43 13 PM

After 👇 
Screenshot 2025-04-29 at 2 45 20 PM

## Checklist

### DIFF
--- a/ui/admin/app/components/form/role/grants/index.hbs
+++ b/ui/admin/app/components/form/role/grants/index.hbs
@@ -15,15 +15,12 @@
       <list.item as |item|>
         <form.input @contextual={{true}}>
           <item.key>
-            <Hds::Form::Label @controlId='add-grant'>
+            <Hds::Form::Label>
               {{t 'resources.role.grant.actions.create'}}
             </Hds::Form::Label>
           </item.key>
           <item.cell>
-            <Input
-              class='rose-form-input-field'
-              id='add-grant'
-              @type='text'
+            <Hds::Form::TextInput::Field
               name='grant'
               @value={{this.newGrantString}}
             />
--- a/ui/admin/app/components/form/role/grants/index.hbs
+++ b/ui/admin/app/components/form/role/grants/index.hbs
@@ -15,17 +15,18 @@
       <list.item as |item|>
         <form.input @contextual={{true}}>
           <item.key>
-            <Hds::Form::Label>
+            <Hds::Form::Label @controlId='add-grant'>
               {{t 'resources.role.grant.actions.create'}}
             </Hds::Form::Label>
           </item.key>
           <item.cell>
             <Hds::Form::TextInput::Field
-              aria-describedby={{t 'form.grant.help'}}
-              @type='text'
-              name='grant'
-              @value={{this.newGrantString}}
               {{on 'input' this.onInputFieldChanged}}
+              name='grant'
+              @id='add-grant'
+              @type='text'
+              @value={{this.newGrantString}}
+              @extraAriaDescribedBy={{t 'form.grant.help'}}
             />
           </item.cell>
           <item.cell>
@@ -50,22 +51,24 @@
 >
 
   <Rose::List::KeyValue as |list|>
-    {{#each this.grants as |grant|}}
+    {{#each this.grants as |grant index|}}
       <list.item as |item|>
         <form.input @contextual={{true}}>
           <item.key>
-            <Hds::Form::Label>
+            <Hds::Form::Label @controlId='update-grant-{{index}}'>
               {{t 'form.grant.label'}}
             </Hds::Form::Label>
           </item.key>
           <item.cell>
             <Hds::Form::TextInput::Field
-              aria-describedby={{t 'form.grant.help'}}
-              @type='text'
+              {{on 'input' (fn this.onUpdateGrant grant)}}
               name='grant'
+              disabled={{if (can 'setGrants role' @model) false true}}
+              @id='update-grant-{{index}}'
+              @type='text'
               @value={{grant.value}}
-              @disabled={{if (can 'setGrants role' @model) false true}}
               @error={{@model.errors.grant_strings}}
+              @extraAriaDescribedBy={{t 'form.grant.help'}}
             />
           </item.cell>
         </form.input>
--- a/ui/admin/app/components/form/role/grants/index.hbs
+++ b/ui/admin/app/components/form/role/grants/index.hbs
@@ -8,36 +8,32 @@
     @onSubmit={{fn this.createGrant @addGrant}}
     @cancel={{@cancel}}
     @disabled={{@model.isSaving}}
-    as |form|
   >
     <Rose::List::KeyValue as |list|>
       {{! Add grant field (only one per form) }}
       <list.item as |item|>
-        <form.input @contextual={{true}}>
-          <item.key>
-            <Hds::Form::Label @controlId='add-grant'>
-              {{t 'resources.role.grant.actions.create'}}
-            </Hds::Form::Label>
-          </item.key>
-          <item.cell>
-            <Hds::Form::TextInput::Field
-              {{on 'input' (set-from-event this 'newGrantString')}}
-              name='grant'
-              @id='add-grant'
-              @type='text'
-              @value={{this.newGrantString}}
-              @extraAriaDescribedBy={{t 'form.grant.help'}}
-            />
-          </item.cell>
-          <item.cell>
-            <Hds::Button
-              disabled={{this.cannotSave}}
-              type='submit'
-              @color='secondary'
-              @text={{t 'actions.add'}}
-            />
-          </item.cell>
-        </form.input>
+        <item.key>
+          <Hds::Form::Label @controlId='add-grant'>
+            {{t 'resources.role.grant.actions.create'}}
+          </Hds::Form::Label>
+        </item.key>
+        <item.cell>
+          <Hds::Form::TextInput::Field
+            {{on 'input' (set-from-event this 'newGrantString')}}
+            name='grant'
+            @id='add-grant'
+            @type='text'
+            @value={{this.newGrantString}}
+          />
+        </item.cell>
+        <item.cell>
+          <Hds::Button
+            disabled={{this.cannotSave}}
+            type='submit'
+            @color='secondary'
+            @text={{t 'actions.add'}}
+          />
+        </item.cell>
       </list.item>
     </Rose::List::KeyValue>
   </Rose::Form>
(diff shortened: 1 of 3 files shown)


hashicc (commented):
With migrating just the label (`Rose::Form::Label`) and not the input (`Rose::Form::Input`) that yields the label (`input.label`) it looks like the a11y wiring between the label and the input is broken: a quick test for this is to click the label, _Grant_ or _New Grant_ and it should focus on the input. Instead of trying to restore the label linkage it might make more sense to migrate the `Rose::Form::Input` to `Hds::Form::Input`.

laurenolivia:
> > With migrating just the label (`Rose::Form::Label`) and not the input (`Rose::Form::Input`) that yields the label (`input.label`) it looks like the a11y wiring between the label and the input is broken: a quick test for this is to click the label, _Grant_ or _New Grant_ and it should focus on the input. Instead of trying to restore the label linkage it might make more sense to migrate the `Rose::Form::Input` to `Hds::Form::Input`.
> 
> @hashicc thanks for the feedback!
> 
> cc: @DhariniJeeva - do you mind if we increase the scope of this ticket?

Oh, actually, I just checked our epics. It doesn't appear that we have a ticket for replacing `Rose::Form::Input` (we do have a [ticket](https://hashicorp.atlassian.net/browse/ICU-15716) to retire the component, though), so if it's ok with you, I will just make that change within this PR and update the ticket to expand the scope?

DhariniJeeva:
> > > With migrating just the label (`Rose::Form::Label`) and not the input (`Rose::Form::Input`) that yields the label (`input.label`) it looks like the a11y wiring between the label and the input is broken: a quick test for this is to click the label, _Grant_ or _New Grant_ and it should focus on the input. Instead of trying to restore the label linkage it might make more sense to migrate the `Rose::Form::Input` to `Hds::Form::Input`.
> > 
> > 
> > @hashicc thanks for the feedback!
> > cc: @DhariniJeeva - do you mind if we increase the scope of this ticket?
> 
> Oh, actually, I just checked our epics. It doesn't appear that we have a ticket for replacing `Rose::Form::Input` (we do have a [ticket](https://hashicorp.atlassian.net/browse/ICU-15716) to retire the component, though), so if it's ok with you, I will just make that change within this PR and update the ticket to expand the scope?

that's a good point, I'm totally ok with migrating the input in this PR

ui/admin/app/components/form/role/grants/index.hbs
```diff
@@ -13,20 +13,18 @@
...
+        <form.input @contextual={{true}} as |input|>
           <item.key>
             <Hds::Form::Label>
               {{t 'resources.role.grant.actions.create'}}
             </Hds::Form::Label>
           </item.key>
           <item.cell>
-            <input.field title={{t 'form.grant.help'}} />
+            <input.field
+              type='text'
```
laurenolivia, line 24: Note: `<item.cell>` will not accept an Hds form field component. It either breaks the form, or breaks the `Add` button. I wonder if we should refactor this file and remove `Rose::List::KeyValue` because it does not seem to respond well to mixing Hds components within. Open to suggestions. Still WIP.

ui/admin/app/components/form/role/grants/index.hbs
```diff
@@ -13,20 +13,20 @@
...
-          @type='text'
-          @value={{this.newGrantString}}
-          as |input|
-        >
+        <form.input @contextual={{true}}>
           <item.key>
-            <input.label>{{t
-                'resources.role.grant.actions.create'
-              }}</input.label>
+            <Hds::Form::Label @controlId='add-grant'>
```
laurenolivia, line 18: Note: In order to honor the form layout of `Rose::List::KeyValue` I kept the `Hds::Form::Label` component inside of `<item.key>` and used `@controllerId` to connect the label to the `Input` component. This changed the form field specs, so I added `class='rose-form-input-field'` to `Input` to ensure the text fields have identical specs

ui/admin/tests/acceptance/accounts/delete-test.js
```diff
@@ -63,14 +63,14 @@ module('Acceptance | accounts | delete', function (hooks) {
   });
 
   test('can delete an account', async function (assert) {
-    const accountsCount = this.server.schema.accounts.all().models.length;
+    const accountsCount = this.server.db.accounts.length;
```
laurenolivia, line 66: Tests failed on `cannot read properties of 'schema'` so I tested a fix to replace with `server.db` and this works. Please note that there is a ticket to cleanup our usage of `server.db` - and there was also a discussion about using schema vs db [here](https://hashicorp.slack.com/archives/C07EMPECGFN/p1739372121647269?thread_ts=1739371374.513759&cid=C07EMPECGFN). TLDR "...we should be using schema unless you need the db for a specific reason"

ui/admin/app/components/form/role/grants/index.hbs
```diff
@@ -13,20 +13,21 @@
...
       {{! Add grant field (only one per form) }}
       <list.item as |item|>
-        <form.input
-          @contextual={{true}}
-          @name='grant'
-          @type='text'
-          @value={{this.newGrantString}}
-          as |input|
-        >
+        <form.input @contextual={{true}}>
```
hashicc, line 16: Does `<form.input @contextual={{true}}>` do anything any more since it's not yielding anything of use and `@contextual` is `true`?

ui/admin/app/components/form/role/grants/index.hbs
```diff
@@ -13,20 +13,21 @@
...
-                'resources.role.grant.actions.create'
-              }}</input.label>
+            <Hds::Form::Label @controlId='add-grant'>
+              {{t 'resources.role.grant.actions.create'}}
+            </Hds::Form::Label>
           </item.key>
           <item.cell>
-            <input.field title={{t 'form.grant.help'}} />
+            <Hds::Form::TextInput::Field
+              {{on 'input' this.onInputFieldChanged}}
```
priya-patel04, line 24: We can use `set-from-event` helper  here. It basically extracts the value from the event and assigns it to the specified property - `newGrantString` in this case.

`{{on 'input' (set-from-event this 'newGrantString')}}
`

ui/admin/app/components/form/role/grants/index.hbs
```diff
@@ -49,33 +50,26 @@
...
           <item.key>
-            <input.label>{{t 'form.grant.label'}}</input.label>
+            <Hds::Form::Label @controlId='update-grant-{{index}}'>
+              {{t 'form.grant.label'}}
+            </Hds::Form::Label>
           </item.key>
           <item.cell>
-            <input.field title={{t 'form.grant.help'}} />
+            <Hds::Form::TextInput::Field
+              {{on 'input' (fn this.onUpdateGrant grant)}}
```
priya-patel04, line 64: Same here, We can leverage `set-from-event` helper and remove `onUpdateGrant` function.

` {{on 'input' (set-from-event grant 'value')}}
`

ui/admin/app/components/form/role/grants/index.hbs
laurenolivia: Good catch

laurenolivia: I moved args into the Hds component down below (and tested if they broke the form) but missed testing the behavior on this one (when removed).

ui/admin/app/components/form/role/grants/index.hbs
laurenolivia: TIL

Should we err on the side of using this convention? Is this recommended bc it reduces code?

ui/admin/app/components/form/role/grants/index.hbs
laurenolivia: Jus tested removal. Looks like it still yields the form layout. Here is what the form does upon removal.
Screenshot 2025-04-30 at 12 24 17 PM

ui/admin/app/components/form/role/grants/index.hbs
DhariniJeeva: since the fn `onInputFieldChanged ` doesn't really do anything other than attaching the event, we should use our existing helper `set-from-event` to perform this action inline vs creating a new fn for this

laurenolivia: Oh I think that's an outdated version @DhariniJeeva

ui/admin/app/components/form/role/grants/index.hbs
ZedLi: I think Chad actually meant removing `<form.input @contextual={{true}}>` itself and not just the `@contextual` argument.

Take a look at what the component is doing:
https://github.com/hashicorp/boundary-ui/blob/93a57f26ee75e073427a1074fd893bf94ecf58e6/addons/rose/addon/components/rose/form/input/index.hbs#L6-L28

Are we still really using it at all and yielding anything?

ui/admin/app/components/form/role/grants/index.hbs
```diff
@@ -13,25 +13,26 @@
...
           </item.key>
           <item.cell>
-            <input.field title={{t 'form.grant.help'}} />
+            <Hds::Form::TextInput::Field
+              {{on 'input' (set-from-event this 'newGrantString')}}
+              name='grant'
+              @id='add-grant'
+              @type='text'
+              @value={{this.newGrantString}}
+              @extraAriaDescribedBy={{t 'form.grant.help'}}
```
ZedLi, line 29: What are these `@extraAriaDescribedBy` for? These attach extra IDs for [aria-describedBy](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Attributes/aria-describedby) which take in IDs, not internationalized text.

ZedLi (changes requested):

ui/admin/app/components/form/role/grants/index.hbs
laurenolivia: Note: we chatted offline 👍

ui/admin/app/components/form/role/grants/index.hbs
ZedLi: In retrospect, it looks like we had a title because there was no where to add a description for the input, so I think we should keep the title. It'll just be a description when you hover the input which we normally don't do but it's better than nothing for these input fields

laurenolivia: That's fair

ZedLi (approved):
Looks good!

priya-patel04 (approved):
Thanks for making the changes!

PR #2789, opened 2025-04-22
🛑 spike: do not merge: convert api addon to v2 format

hashicc:
Followed [v2 migration guide](https://github.com/embroider-build/embroider/blob/4574bb78d25a3d2bffc602b1484e089b0f5c4af1/docs/porting-addons-to-v2.md)

Notes:
* This pr has all but 2 tests passing in `addons/api-test-app`
* Need to figure out the convention for test apps for specific addons (eg: `addons/api` addon, this pr has the dummy app split into the test app at `addons/api-test-app`)
* following the `package.json` directions closely is important (this example pr likely has unnecessary dependencies in `api` and `api-test-app`
* the rollup build will add a key `ember-addon.app-js` with exports based on the `rollup.config.js` (these should probably be committed but I didn't see it mentioned in the guide)
* Using `rollup.config.js` globs for exports seemed to work okay (the instructions make it sound like maybe they have be listed explicitly)


hashicc:
This pr and notes can serve as a reference for the process to convert to an ember v2 addon

PR #2790, opened 2025-04-22
fix: 🐛 credential library serializer for SSH certificate

cameronperera:
✅ Closes: https://hashicorp.atlassian.net/browse/ICU-16880

# Description

A user reported a bug where changing the `key_type` of a Vault SSH Certificate credential library, the update was failing. This was due to the `key_bits` value not being set to null when changing from `rsa` or `ecdsa` to `ed25519`. When using `ed25519` the `key_bits` value needs to be cleared out (set to null).

https://github.com/user-attachments/assets/33222aee-4105-4d42-a69a-0946e6c45204

## How to Test

Create a Vault Credential Store then a SSH Certificates Credential Library (CL). 
First create the CL with a `key_type` of `ecdsa` and define the `key_bits` to `521`.
Next, change the `key_type` to `ed25519` and click Save. It should successfully update the CL.

## Checklist


DhariniJeeva (approved):
thanks for the fix!

lisbet-alvarez (approved):
Thanks for fixing this bug!! Works great!

PR #2793, opened 2025-04-23
Icu 16546 upgrade ember for admin UI

lisbet-alvarez:
# Description

1. Update admin ui to ember v5.12
2. Run ember init & codemods
3. Run linter + formatter

image

## How to Test

[Desktop Releases Test Run](https://github.com/hashicorp/boundary-desktop-releases/actions/runs/14624322298)

## Checklist


ui/admin/tests/unit/instance-initializers/feature-edition-test.js
```diff
@@ -32,7 +32,9 @@ module('Unit | Instance Initializer | feature-edition', function (hooks) {
   });
 
   hooks.afterEach(function () {
+    // eslint-disable-next-line ember/no-runloop
     run(this.instance, 'destroy');
+    // eslint-disable-next-line ember/no-runloop
     run(this.application, 'destroy');
   });
```
hashicc, line 40: This test _does_ test that the initializer succeeds but doesn't assert anything else. I'm guessing this was just the blueprint test. As a team do we generally commit and keep these `"it works"` tests?

hashicc (approved):
🕺 One question but no blockers at all. Nice work, thanks for the upgrade. I can't wait to use the new toys. I noticed `@ember/test-helpers` was downgraded but I have #2797 that I want to rebase after all the ember upgrade work that'll get all `@ember/test-helpers` in the monorepo to v5

ui/admin/tests/unit/instance-initializers/feature-edition-test.js
lisbet-alvarez: Yes, that is correct its the blueprint tests and, yup at the very minimum we keep the test file with the generated test case. The same thing can be observed for controller, and route unit tests. But ideally, we should be adding more test cases where ever possible.

lisbet-alvarez:
> I noticed `@ember/test-helpers` was downgraded but I have #2797 that I want to rebase after all the ember upgrade work that'll get all `@ember/test-helpers` in the monorepo to v5

Thats awesome!! Yeah only reason I had to downgrade (& match new blueprints) is because v4 was causing an error in the test env. Funny enough, once upgrading the api addon to v5.12, I can update `@ember/test-helpers` to v4 with no errors.
image

ui/admin/config/optional-features.json
```diff
@@ -2,5 +2,6 @@
   "application-template-wrapper": false,
   "default-async-observers": true,
   "jquery-integration": false,
-  "template-only-glimmer-components": true
+  "template-only-glimmer-components": true,
+  "no-implicit-route-model": true
```
cameronperera, line 6: learning: We need this because we previously relied on the fact that if you don't define a model hook for a route, it will use the model from the parent route correct?

cameronperera (approved):
🎉

ui/admin/package.json
```diff
@@ -16,8 +16,8 @@
     "build:oss": "EDITION=oss ember build --environment=production",
     "build:enterprise": "EDITION=enterprise ember build --environment=production",
     "build:hcp": "EDITION=hcp ember build --environment=production",
-    "lint": "concurrently \"npm:lint:*(!fix)\" --names \"lint:\"",
-    "lint:fix": "concurrently \"npm:lint:*:fix\" --names \"fix:\"",
+    "lint": "concurrently \"yarn:lint:*(!fix)\" --names \"lint:\"",
```
DhariniJeeva, line 19: good catch!

DhariniJeeva (approved):
Thanks for the work! looks good to me

PR #2780, opened 2025-04-16
Icu 16722 interaction components replace rose dropdown with hds dropdown

DhariniJeeva:
# Description

NOTE: this PR covers only a portion of the dropdown migration, I will be handling the worker tags and the desktop in a separate PR to keep this one moderately simple. 

what is covered: 
1. authenticate page both desktop and admin
2. add principals
3. add members

what's not covered: 
1. worker tags 
2. scopes selection desktop
3. header nav desktop

 https://hashicorp.atlassian.net/browse/ICU-16722

## Screenshots (if appropriate)

**before**:
Screenshot 2025-04-16 at 12 10 42 PM

**after**:
Screenshot 2025-04-16 at 12 45 29 AM

Screenshot 2025-04-16 at 12 07 07 PM

**members**:
**before**:
Screenshot 2025-04-16 at 12 11 28 PM

Screenshot 2025-04-16 at 12 11 25 PM

**after**:
Screenshot 2025-04-16 at 12 41 44 AM

**principals**:

**before**:
 
Screenshot 2025-04-16 at 12 14 20 PM

Screenshot 2025-04-16 at 12 14 24 PM

**after**:
Screenshot 2025-04-16 at 12 08 06 PM

Screenshot 2025-04-16 at 12 07 54 PM

## How to Test

- go to the routes mentioned above and click around, should not see any errors/warnings

## Checklist

- [x] I have added before and after screenshots for UI changes
- ~[ ] I have added JSON response output for API changes~
- ~[ ] I have added steps to reproduce and test for bug fixes in the description~
- [x] I have commented on my code, particularly in hard-to-understand areas
- [x] My changes generate no new warnings
- [x] I have added tests that prove my fix is effective or that my feature works

PR #2762, opened 2025-04-03
Replace Rose::Icon with HDS::Icon 

priya-patel04:
# Description
Replace Rose::Icon with HDS::Icon 

 :tickets: [Jira ticket](https://hashicorp.atlassian.net/browse/ICU-16709)

## How to Test
Use Vercel to see the changes for `admin-ui` and `desktop`

## Checklist

- [ ] I have added before and after screenshots for UI changes
- [ ] I have added JSON response output for API changes
- [ ] I have added steps to reproduce and test for bug fixes in the description
- [ ] I have commented on my code, particularly in hard-to-understand areas
- [ ] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works

PR #2794, opened 2025-04-24
Worker dropdown migration

DhariniJeeva:
# Description
This is the final portion of the dropdown changes mentioned in this [PR](https://github.com/hashicorp/boundary-ui/pull/2780)
 

before:
Screenshot 2025-04-23 at 4 42 47 PM

after:
 
 
Screenshot 2025-05-09 at 2 25 28 PM

 
 

## How to Test

- go to workers
- click on the dropdown toggle `tags`
- select filters and the workers should be filtered based on the tag selection and the route should reflect the tag selected
- clicking on clear filter should remove the route update from above step and clear all filters

## Checklist


ui/admin/app/templates/scopes/scope/workers/index.hbs
```diff
@@ -34,67 +34,18 @@
...
-        </toolbar.secondary>
+      <Dropdown
+        name='tags'
+        @listPosition='bottom-left'
+        @toggleText={{t 'form.tags.label'}}
+        @itemOptions={{this.groupedTags}}
+        @checkedItems={{this.tags}}
+        @applyFilter={{fn this.applyFilter 'tags'}}
+        @isGrouped={{true}}
+      />
```
lisbet-alvarez, line 45: There seems to be an extra border space around the dropdown content as observed in the image below.
image

lisbet-alvarez (commented):
Looking good so far!! Just left a couple of questions.

ui/admin/app/templates/scopes/scope/workers/index.hbs
```diff
@@ -34,67 +34,18 @@
...
-                @icon='x'
-                @text={{t 'actions.clear-all-filters'}}
-                {{on 'click' this.callClearAllFilters}}
-              />
-            {{/if}}
-          </Rose::Form>
-        </toolbar.primary>
-
-        <toolbar.secondary>
-          <ToolbarRefresher @onClick={{this.refresh}} />
```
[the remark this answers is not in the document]
DhariniJeeva: ah good catch, I didn't mean to remove this one, I add this back in

ui/admin/app/templates/scopes/scope/workers/index.hbs
DhariniJeeva: Where do you see this? I don't see this on the Vercel preview or locally

lisbet-alvarez: hmm im seeing it when using [vercel here](https://boundary-ui-git-worker-and-desktop-dropdown-migration-hashicorp.vercel.app/scopes/global/workers)

addons/core/addon/components/dropdown/index.hbs
```diff
@@ -35,16 +35,36 @@
...
-      <DD.Checkbox
-        @value={{itemOption.id}}
-        data-test-checkbox={{itemOption.id}}
-        checked={{includes itemOption.id @checkedItems}}
-        {{on 'change' this.selectItem}}
-      >
-        {{if itemOption.name itemOption.name itemOption.id}}
-      </DD.Checkbox>
-    {{/each}}
+    {{#if @isGrouped}}
```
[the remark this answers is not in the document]
DhariniJeeva: thanks for the link! I changed it similar to grouped scopes

ui/admin/app/styles/app.scss
```diff
@@ -1050,6 +1049,11 @@
     padding: 0.25rem 0.75rem;
   }
 }
+.dropdown-section {
+  padding: 0.5rem;
+  text-transform: uppercase;
+  color: var(--ui-gray);
```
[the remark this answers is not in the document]
DhariniJeeva: removed this, decided to follow the grouped scope styles

ui/admin/app/templates/scopes/scope/workers/index.hbs
DhariniJeeva: this is strange- I'm not seeing this extra scroll bar. Leaving this thread open in case others encounter it too.

priya-patel04: I am not seeing extra scroll bar both locally and on vercel

priya-patel04: I did notice that the `tags` icon are slightly misaligned. I remember in my Rose::Icon migration [PR](https://github.com/hashicorp/boundary-ui/pull/2762), I added style [here](https://github.com/hashicorp/boundary-ui/blob/185af6f00e8221394af741530f8dcc943ea675c0/addons/rose/app/styles/rose/components/dropdown/_section.scss#L23) to align them properly.

Expected:
Screenshot 2025-04-30 at 10 16 43 AM

DhariniJeeva: Thanks for the catch, pushed up a fix to align the icons!

ZedLi: image

I also notice the extra bars both locally and in vercel, it looks like it's from an extra `overflow: scroll` css?

DhariniJeeva: I noticed we were adding a [scroll](https://github.com/hashicorp/boundary-ui/blob/4295966b7f31c065faf1d1e3ec8af53a26716560/ui/admin/app/styles/app.scss#L991) property for rose which I had changed it to HDS, but HDS dropdown has it own scroll property. Hence the double scroll 😅  
Mocks didn't add enough items in my dropdown so I didn't catch this behaviour, clearing indexed DB weirdly added more items and I could see the double scroll. 
 
fixed this, thank you everyone for testing this!

addons/core/addon/helpers/is-tag-checked.js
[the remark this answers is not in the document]
DhariniJeeva: good point, I will make this a local helper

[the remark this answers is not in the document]
ZedLi: I have a feeling that this helper might not be needed if you're able to hash the tag object since it's really just trying to check both key and value for uniqueness

addons/core/addon/components/dropdown/index.js
```diff
@@ -50,12 +50,18 @@ export default class DropdownComponent extends Component {
...
+      this.selectedItems = [...this.selectedItems, item];
     } else {
-      this.selectedItems = this.selectedItems.filter((item) => item !== value);
+      if (this.args.isGrouped && item?.key && item?.value) {
+        this.selectedItems = this.selectedItems.filter(
+          (i) => !(i.key === item.key && i.value === item.value),
+        );
+      } else {
+        this.selectedItems = this.selectedItems.filter((i) => i !== item);
+      }
```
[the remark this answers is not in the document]
DhariniJeeva: Thanks for the feedback, I was able to send a base64 key/value for workers and decode that to display on the consumer side and update queryparams. Let me know how this looks

ui/admin/app/templates/scopes/scope/workers/index.hbs
```diff
@@ -34,64 +34,37 @@
...
+      <div class='workers-filtering'>
+        <Dropdown
+          name='tags'
+          @listPosition='bottom-left'
+          @toggleText={{t 'form.tags.label'}}
+          @itemOptions={{this.groupedTags}}
+          @checkedItems={{this.tags}}
+          @applyFilter={{fn this.applyFilter 'tags'}}
+          @isGrouped={{true}}
+          as |FD selectItem itemOptions|
```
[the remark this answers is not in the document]
DhariniJeeva: hmm I decided to call it `FD` since this is also a filter dropdown.

ui/admin/app/templates/scopes/scope/workers/index.hbs
```diff
@@ -34,64 +34,37 @@
...
+        >
+          {{#each-in (group-by 'key' itemOptions) as |key groupItems|}}
+            <FD.Generic>
+              <Hds::Icon @name='tag' @isInline={{true}} />
+              <Hds::Text::Display @tag='h5'>{{key}}</Hds::Text::Display>
+            </FD.Generic>
+            {{#each groupItems as |item|}}
+              <FD.Checkbox
+                @value={{item}}
```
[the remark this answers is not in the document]
DhariniJeeva: the refactor should fixed this one too. ty!

ui/admin/app/controllers/scopes/scope/workers/index.js
```diff
@@ -65,17 +119,22 @@ export default class ScopesScopeWorkersIndexController extends Controller {
...
+   * Sets a query param to the value of selectedItems
+   * @param {string} paramKey
+   * @param {[string]} selectedItems
    */
-  isEqual(firstTag, secondTag) {
-    return firstTag.key === secondTag.key && firstTag.value === secondTag.value;
+  @action
+  applyFilter(paramKey, selectedItems) {
+    // Decoding the selectedItems
+    const decodedItems = selectedItems.map((item) => {
```
ZedLi, line 131: Why do we decode back into an object instead of keeping it encoded for the query string and not use `encodedTags`? And just handle decoding back in the model hook

ui/admin/app/controllers/scopes/scope/workers/index.js
```diff
@@ -21,6 +21,60 @@ export default class ScopesScopeWorkersIndexController extends Controller {
...
+
+  /**
+   * Returns the filters object used for displaying filter tags.
+   * @type {object}
+   */
+
+  get filters() {
+    return {
+      allFilters: {
+        tags: this.groupedTags,
```
ZedLi, line 71: Rather than just sending it as is, couldn't you have transformed these into objects that match what the filters are expecting (i.e. `{ id: encodedTag, name: tag.value }`) and that way you wouldn't have to modify the filter tags component and use custom logic?

ui/admin/app/controllers/scopes/scope/workers/index.js
DhariniJeeva: I decoded it here so the query strings would show human-readable filters.  But if we are ok with having encoded values in the URL, I can move the decoding to the model hook instead

ui/admin/app/controllers/scopes/scope/workers/index.js
DhariniJeeva: ah yes, this is good point, I did a quick test and works as expected! I will push up the changes. I like that this won't modify the filter component. Thank you!

ui/admin/app/controllers/scopes/scope/workers/index.js
ZedLi: Yeah I think it's best to have it in the URL encoded since it's not great to have literal objects in the query strings

ui/admin/app/controllers/scopes/scope/workers/index.js
```diff
@@ -21,6 +21,60 @@ export default class ScopesScopeWorkersIndexController extends Controller {
...
+  queryParams = [{ tags: { type: 'array' } }];
+
+  /**
+   * Returns all unique tags of the selected worker in base64 encoded format.
+   * @type {object}
+   */
+
+  get groupedTags() {
+    const allTags = this.model
+      .flatMap((worker) => worker.allTags)
```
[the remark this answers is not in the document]
laurenolivia: Here's another way to filter on unique tags, then you could bypass using a set:
```
const allTags = this.model.flatMap((worker) => worker.canonical_tags);
    allTags.forEach((object) => {
      // output keys with: Object.keys(object)
    });
```

ui/admin/app/templates/scopes/scope/workers/index.hbs
```diff
@@ -34,64 +34,47 @@
...
+          @listPosition='bottom-left'
+          @toggleText={{t 'form.tags.label'}}
+          @itemOptions={{this.groupedTags}}
+          @checkedItems={{this.encodedTags}}
+          @applyFilter={{fn this.applyFilter 'tags'}}
+          @isGrouped={{true}}
+          as |FD selectItem itemOptions|
+        >
+          {{#each-in (group-by 'key' (decode itemOptions)) as |key groupItems|}}
+            <FD.Generic>
```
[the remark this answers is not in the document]
DhariniJeeva: Screenshot 2025-05-09 at 2 25 28 PM

makes sense, I updated the margin and the color and here's the latest change, lmk what you think

ui/admin/app/controllers/scopes/scope/workers/index.js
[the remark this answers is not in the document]
DhariniJeeva: thank you for the suggestions! I went with allTags and filtered out the duplicate tags before encoding for the filter

ui/admin/tests/integration/helpers/encode-test.js
```diff
@@ -0,0 +1,38 @@
...
+  test('it encodes undefined or null to an empty string', async function (assert) {
+    this.set('input', undefined);
+
+    await render(hbs`{{encode this.input}}`);
+    assert.strictEqual(this.element.textContent.trim(), '');
+
+    this.set('input', null);
+
+    await render(hbs`{{encode this.input}}`);
+    assert.strictEqual(this.element.textContent.trim(), '');
```
cameronperera, line 36: Same comment here as in the `decode-tests.js` file.

ui/admin/app/styles/app.scss
```diff
@@ -953,9 +953,29 @@
...
+    margin-bottom: 1rem;
+
+    .toolbar-refresher {
+      margin-left: auto;
+    }
+
+    .hds-dropdown-list-item--variant-generic {
+      padding: 0.5rem 1rem;
+      text-transform: uppercase;
+      color: var(--ui-gray)
```
ZedLi, line 967: Could you have just used the `@color` arg on the HDS components themselves?

cameronperera (commented):
Looking great. Just left a few comments.

ui/admin/app/styles/app.scss
```diff
@@ -953,9 +953,29 @@
...
+
+    .hds-dropdown-list-item--variant-checkbox {
+      padding-left: 2.25rem;
+    }
+
+    .hds-dropdown-list-item {
+      display: flex;
+      align-items: center;
+      gap: 0.2rem;
+    }
```
ZedLi, line 978: Were these HDS classes needed?  To me it seems like you could have used classes directly:

```suggestion
    .grouping {
      display: flex;
      align-items: center;
      gap: 0.2rem;
      padding: 0.5rem 1rem;
      text-transform: uppercase;
    }

    li:has(.indentation) {
      padding-left: 2.25rem;
    }
```

ui/admin/app/templates/scopes/scope/workers/index.hbs
```diff
@@ -34,64 +34,46 @@
...
+          {{/each-in}}
+        </Dropdown>
+        <ToolbarRefresher @onClick={{this.refresh}} />
+      </div>
+      <FilterTags @filters={{this.filters}} as |filter|>
+        <Hds::Tag
+          class='filter-tag'
+          @text={{get (decode filter) 'value'}}
+          @onDismiss={{fn this.removeFilter filter}}
+        />
```
ZedLi, line 75: I don't think this is needed anymore

ui/admin/app/templates/scopes/scope/workers/index.hbs
```diff
@@ -34,64 +34,46 @@
...
+          name='tags'
+          @listPosition='bottom-left'
+          @toggleText={{t 'form.tags.label'}}
+          @itemOptions={{this.groupedTags}}
+          @checkedItems={{this.tags}}
+          @applyFilter={{fn this.applyFilter 'tags'}}
+          @isGrouped={{true}}
+          as |FD selectItem itemOptions|
+        >
+          {{#each-in (group-by 'key' (decode itemOptions)) as |key groupItems|}}
```
ZedLi, line 49: Do you need to decode this? It looks like you just end up using the encoded values anyways with `item`?

ui/admin/app/templates/scopes/scope/workers/index.hbs
```diff
@@ -34,64 +34,46 @@
...
+                @weight='medium'
+              >{{key}}</Hds::Text::Display>
+            </FD.Generic>
+            {{#each groupItems as |item|}}
+              <FD.Checkbox
+                @value={{encode item}}
+                checked={{includes (encode item) this.tags}}
+                {{on 'change' selectItem}}
+              >
+                {{item.value}}
```
ZedLi, line 63: I think you could have just used the object you have from `@itemOptions` directly without having to re-encode here

ui/admin/app/styles/app.scss
DhariniJeeva: I wanted the icon and the text color to match what's currently in prod, so I moved the color to the class instead of using the `@color` arg

ui/admin/app/templates/scopes/scope/workers/index.hbs
DhariniJeeva: yes, this needs to be decoded as we pass the `Id` and the `name` to the filters 
Screenshot 2025-05-12 at 2 21 49 PM


to display the key in the dropdown we will need to decode the `id`

ui/admin/app/styles/app.scss
```diff
@@ -953,9 +953,29 @@
 
 // workers
 .workers {
-  .rose-dropdown-content {
-    max-height: 300px;
-    overflow: scroll;
+  .workers-filtering {
+    display: flex;
+    margin-bottom: 1rem;
```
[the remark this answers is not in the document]
DhariniJeeva: good catch, I removed the .workers-list-table and adjusted `.filters-applied` section to have enough margin

ui/admin/app/controllers/scopes/scope/workers/index.js
```diff
@@ -133,28 +182,20 @@ export default class ScopesScopeWorkersIndexController extends Controller {
...
   /**
-   * Refreshes worker data.
-   */
+   * Removes a filter from the selected filters and updates the query params.
+   * @param {object} tag - The tag to remove from the filters.
+   **/
   @action
-  refresh() {
-    this.router.refresh('scopes.scope.workers');
+  removeFilter(tag) {
```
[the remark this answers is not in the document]
DhariniJeeva: ah yes, removed!

ui/admin/app/styles/app.scss
ZedLi: Is the `@color='faint'` not close enough? I think it's probably better to use the HDS color scheme where we can at this point

ui/admin/app/templates/scopes/scope/workers/index.hbs
ZedLi: I'm not sure I understand, you have the `name` right there in the object to use for the dropdown, it's still not clear to me why you'd need to decode it since everything else would be using the encoded value

DhariniJeeva: I missed mentioning that the `key` we display in the dropdown is grouped in the `encodedId`

ui/admin/app/styles/app.scss
DhariniJeeva: so are we ok with having the icon in black color? if I use the @color args we either leave the icons black or handle this icon's color in the css.   
Screenshot 2025-05-12 at 3 51 59 PM

ui/admin/app/templates/scopes/scope/workers/index.hbs
ZedLi: What is `key` in this case? It's not the `name`?

ui/admin/app/styles/app.scss
ZedLi: Why not make the icon [color](https://helios.hashicorp.design/components/icon?tab=code#color) also the same?

ui/admin/app/templates/scopes/scope/workers/index.hbs
ZedLi: Ah sorry I see what you mean, I was focused on the dropdown values and not the grouping separator. I would still personally have either grouped them beforehand so you don't have to decode and re-encode in the template or just have included the original values as part of the objects

DhariniJeeva: for example `key`(which is grouped in the encodedId) is `os` and the filter name we display is actually the `value` 
Screenshot 2025-05-12 at 3 59 59 PM
Screenshot 2025-05-12 at 4 00 15 PM

the filter expects `id` and a `name` and we can't display the `key` since there could be multiple `values` for the same key

ui/admin/app/styles/app.scss
DhariniJeeva: oops I forgot we could do that for icons!

ui/admin/app/templates/scopes/scope/workers/index.hbs
DhariniJeeva: Thanks! I like the approach of not having to encode/decode in the template and pass the original and handle these in the route/controller. Updated accordingly

ui/admin/tests/integration/helpers/decode-test.js
```diff
@@ -0,0 +1,45 @@
...
+    );
+    assert.dom('[data-test]').hasText('bar');
+  });
+
+  test('it returns an empty string if the value is undefined', async function (assert) {
+    this.set('encodedString', undefined);
+
+    await render(hbs`{{decode this.encodedString}}`);
+
+    assert.strictEqual(this.element.textContent.trim(), '');
```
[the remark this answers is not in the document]
DhariniJeeva: I ended up not needing these as I'm not handling these in the templates anymore, same for the ones below

ui/admin/app/styles/app.scss
```diff
@@ -953,9 +952,29 @@
...
-    overflow: scroll;
+  .tags-list {
+    display: flex;
+    margin-bottom: 1rem;
+
+    .toolbar-refresher {
+      margin-left: auto;
+    }
+
+    .grouping {
```
ZedLi, line 963: I think the braces need to be reformatted here

ui/admin/app/styles/app.scss
```diff
@@ -953,9 +952,29 @@
...
+	      text-transform: uppercase;
+	    }
+	
+	    li:has(.indentation) {
+	      padding-left: 2.25rem;
+	    }
+  }
+
+  .filters-applied {
+    margin-bottom: 1rem
```
ZedLi, line 977: Alternatively, you could have just had 
```
  > div {
    margin-bottom: 1rem;
  }
```

Under `.workers` rather than two separate margin bottoms

ZedLi (approved):
Looks great, this is a fantastic upgrade!

ui/admin/app/styles/app.scss
DhariniJeeva: updated!

ZedLi: I think the whitespace is still wonky 😅

ZedLi: Yup looks good! Prettier would be the one that would fix this whitespace formatting instead of lint

ZedLi (approved):
Looks great, thanks for the upgrade!

priya-patel04 (approved):
Thanks for making all the changes. Looks great! 🎉

Some changes can an be addressed in the follow-up PR:
1. The Tags dropdown badge count color contrast for dark mode
Screenshot 2025-05-13 at 11 29 03 AM

2. The background for the toolbar is removed - Unsure if it's intentional

Before:

Screenshot 2025-05-13 at 11 29 07 AM

After:

Screenshot 2025-05-13 at 11 29 03 AM

DhariniJeeva:
> Thanks for making all the changes. Looks great! 🎉
> 
> Some changes can an be addressed in the follow-up PR:
> 
> 1. The Tags dropdown badge count color contrast for dark mode
> 
> Screenshot 2025-05-13 at 11 29 03 AM
> 2. The background for the toolbar is removed - Unsure if it's intentional
> 
> Before:
> 
> Screenshot 2025-05-13 at 11 29 07 AM
> After:
> 
> Screenshot 2025-05-13 at 11 29 03 AM

yeah we might have to address this separately, this is now using the generic filter component. I will add a ticket to address the dark mode for the tag count. And background removal seems intentional per this filter component's design

PR #2803, opened 2025-04-28
Update ember-notify style to use rem

priya-patel04:
# Description
Update ember-notify style to replace px with rem 

:tickets: [Jira ticket](https://hashicorp.atlassian.net/browse/ICU-15814)

## How to Test

1. Create a new resource ( ex: project ) successfully, to see success toast.
2. Try submitting a form without required field and cause an error,  to see the error toast
3. In Desktop UI cancel a session to see the toast

## Checklist


calcaide (approved):
I followed test procedure with expected results, Nicely done!! thanks for the work 🙌

cameronperera (approved):
Nice update!

lisbet-alvarez (approved):
awesome work!!

addons/rose/app/styles/rose/variables/_sizing.scss
```diff
@@ -40,8 +40,16 @@ $size-coefficients: (
...
- * Returns a value in rem units equal to the requested size.
+ * @param {string|number} $name - The size name (e.g., 'm', 'l', 'xl') or a numeric value in pixels.
+ * @return {dimension|null} - The size in rem units, or null if the input is invalid.
  */
 @function rems($name) {
-  @return size($name) * 1rem;
+  @if type-of($name) == 'string' {
+    @return size($name) * 1rem;
+  } @else if type-of($name) == 'number' {
+    @return calc($name / 16) * 1rem;
```
ZedLi, line 50: I know the default is 16px to 1rem and that's what our codebase is currently working in but in this case, shouldn't we not use a magic number for this assumption as we somewhat lose the benefit of using `rems` in the first place then? 

I like this change in convenience but if for whatever reason we change the root font size, wouldn't this result in an unexpected scaling?

priya-patel04: That is a really good point. If the base size changes, it wont scale properly. Let me find a workaround in that case.

hashicc: I did a quick test and I think the browser scaling still works:
```css
body {
  font-size: calc((32 /16) * 1rem)
}
```
https://github.com/user-attachments/assets/a0ac9d08-b9a0-46ab-8b4b-db78e5deb597

But instead of the magic number `16` could we use [`$base-font-size`](https://github.com/hashicorp/boundary-ui/blob/9b1f052a569fc1579befffda87de3a91ba5e417b/addons/rose/app/styles/rose/variables/_typography.scss#L38).


I think for our rem setting to be correct we should set the base font-size on the `html` element instead of `body` [here](https://github.com/hashicorp/boundary-ui/blob/9b1f052a569fc1579befffda87de3a91ba5e417b/addons/rose/app/styles/rose/_typography.scss#L19).

> To recap, the rem unit means "The root element's font-size" (rem stands for "root em"). The `<li>` elements inside the `<ul>` with a class of rems take their sizing from the root element (`<html>`). This means that each successive level of nesting does not keep getting larger.

ui/admin/app/styles/_notify.scss
```diff
@@ -3,10 +3,12 @@
  * SPDX-License-Identifier: BUSL-1.1
  */
 
+@use 'rose/variables/sizing';
+
 // This file is essentially a rip of the default ember notify styles with some minor tweaks for positioning.
 @-webkit-keyframes ember-notify-show {
   from {
-    right: -310px;
+    right: sizing.rems(-310);
```
ZedLi, line 11: I've been thinking about this and am wondering if some of these _should_ even be `rems`.

Here's my thought process, which is mainly there are still cases where pixels are valid and not everything should be rems. My take is that rems is fantastic for accessibility for people to be able to read things comfortably which means obvious things like typography but also _can_ include things like margin/padding/width/height as these things might want to scale to provide proper spacing for readability. 

However, it looks like these styles are for _layout_ positioning of a notification animation, which to me seems like they don't necessarily need to scale based on a user's font-size; should they animate farther from the screen with a larger font-size? To me it seems like they shouldn't but I'm open to hearing any thoughts on this

priya-patel04: Yes, I have been thinking along the same lines as well and I agree with you. We may have to keep it to `px` value for this notification animation. I was testing it and depending on the font-size it does animate farther away for larger and more inwards for smaller. 

Screenshot 2025-05-08 at 9 25 57 AM
Screenshot 2025-05-08 at 9 26 10 AM

addons/rose/app/styles/rose/variables/_sizing.scss
hashicc: I actually didn't know that there was the setting to change the font-size separately from the scaling! Setting the `html` does "fix this" although you're right it seems like for accessibility reasons we shouldn't:

> In the same manner, a user might change the base font size to fit their needs. A person with low vision would want to choose a larger size, while someone with an excellent vision can go smaller to gain real estate on their screens.

https://css-tricks.com/accessible-font-sizing-explained/#aa-avoid-setting-a-base-font-size

ZedLi: I had left an earlier comment but deleted it as I wanted to think about it a bit more. 

It looks like a lot of the properties in this PR should still remain `px`. In that case, are there any other scenarios where we would we still want to use this function to convert pixels to rem rather than using rems directly?

priya-patel04: There are few files where we do use `px` and may need to convert them to `rems`. For meanwhile I will close this PR and as we don't need these changes. I will create another task for later; once we have completed HDS migration

priya-patel04:
We don't need to convert px to rems for notify as the scaling based on user's font-size will disrupt notification animation positioning. 
However, there are few files where we do use `px` and may need to convert them to `rems`. Closing this PR and will create another task for later; once we have completed HDS migration

PR #2810, opened 2025-05-01
Icu 11937 foundation components retire rose typography styles

cameronperera:
# Description

The goal was this work is to check on the styles in `rose/_typography` stylesheet. Some things are still needed, some are still needed until we migrate more things to HDS, and some I was able to completely remove.

## How to Test

Please click all around the app. Check every page you can in both Admin and Desktop. Using a split view with a vercel build of main and a vercel build from this PR worked well for me. 

## Checklist


addons/rose/app/styles/rose/_typography.scss
```diff
@@ -12,60 +12,26 @@ html {
...
 .p {
   margin-bottom: sizing.rems(m);
 }
 
+// once we move to Hds::Link::Standalone we can remove this
 a {
   color: var(--action);
 }
-
-time {
```
cameronperera, line 69: I opted to remove this style completely since HDS does not use it. While we haven't migrated `time` to Hds::Time yet, this would get removed anyway.

addons/rose/app/styles/rose/_typography.scss
```diff
@@ -12,60 +12,26 @@ html {
...
-
-  margin-top: sizing.rems(m);
-  margin-bottom: sizing.rems(xs);
-}
-
-h2,
-.h2 {
-  @include type-mixins.type(h2, bold);
-
+.header {
```
cameronperera, line 25: I did not realize when I created the `header-spacing` (now removed) rule that we had rules for the different levels of headers. So I decided to merge the rules into one called `header`. Happy to change the naming or location if needed.

addons/rose/app/styles/rose/_typography.scss
```diff
@@ -12,60 +12,26 @@ html {
 }
 
 body {
+  // Used by Rose::Tabs and '/' in breadcrumbs
```
DhariniJeeva, line 15: thanks for adding these comments, it's helpful to understand the reasoning behind keeping them in here

addons/rose/app/styles/rose/_typography.scss
ZedLi: I notice that we removed the header (`h1, .h1`, etc) styles but we have all these individual headers now with the `header` class attached to them all, is there a reason we shouldn't have kept them?

It seems to me that's the rule rather than the exception, are there more cases with headers that shouldn't have the spacing?

cameronperera: Great question. The rules that were specific to the element, `h1` for example, were being ignored and the `margin: 0` from HDS was overriding the rule. When I added the class to a specific HDS component, the rule `.h1` was being applied. Since it was ignoring the element directly, I went the route of using a single class for any header that we want those styles applied to.

ZedLi: Ahh I see that makes sense. But do we feel like we _should_ have applied it overall to these header elements? If we do, it seems like this might be closer to an HDS override to handle the CSS specificity issue (e.g. `h2[class*=hds-typography-display-]`. I just wanted to make sure using `.header` as a utility class is conscious choice we made in how we want to style and not just a technical one we solved to get around an issue.

calcaide (approved):
Tested manually and run the e2e test (just for mental sake) and all as expected. I left a comment, great work!! 💪 🙌

addons/rose/app/styles/rose/_typography.scss
ZedLi: I would really recommend not to have to use `!important` since that can be very gnarly to deal with in the future, usually you can still get it done with the right CSS specificity

cameronperera: Sorry the delay, I have been thinking over what direction we should go with this. Please let me know your thoughts before I start making the changes.

1. I think an override is a good idea. The majority of our headers need the top and bottom margin rule.
    - We can specify the override to only apply to header tags to be safe.
3. This brings me to the second point, we do have some headers that do not need the margin rule. We could have a special class, just for those specific cases. That way we only need to apply a special class if we don't want the margin added.
4. I would also agree that `!important` is not something we want unless absolutely needed.

addons/core/tests/dummy/app/templates/smoke-test.hbs
```diff
@@ -9,7 +9,7 @@
   Ember can start up and basic features work as intended.
 </Hds::Text::Body>
 
-<Hds::Text::Display @tag='h2'>{{t
+<Hds::Text::Display @tag='h2' class='remove-header-margin'>{{t
```
cameronperera, line 12: Here is an example of using this new class to remove margin during those one-offs that we do not want margins for our header.

addons/rose/addon/styles/hds/overrides.scss
```diff
@@ -3,6 +3,17 @@
  * SPDX-License-Identifier: BUSL-1.1
  */
 
+// header margin override (HDS leaves this to implementers)
+h1[class*='hds-typography-display-'],
+h2[class*='hds-typography-display-'],
+h3[class*='hds-typography-display-'],
+h4[class*='hds-typography-display-'],
+h5[class*='hds-typography-display-'],
+h6[class*='hds-typography-display-'] {
```
cameronperera, line 12: I opted to create this for `h[1-6]`. We do not currently use h5 or h6 but I figured it would be wiser to add it here to be safe. Happy to remove if the team thinks we shouldn't have it until we actually need it.

addons/rose/app/styles/rose/_typography.scss
cameronperera: This has been updated. Please let me know your thougths.

addons/core/tests/dummy/app/templates/smoke-test.hbs
ZedLi: Does this have the right specificity? I think this would get overridden by the HDS margin overrides and wouldn't do anything

addons/rose/addon/styles/hds/overrides.scss
ZedLi: Interestingly enough, there is a relatively new selector that would work for you're asking for, the [where pseudo class](https://developer.mozilla.org/en-US/docs/Web/CSS/:where):

```
:where(h1, h2, h3, h4, h5, h6)[class*='hds-typography-display-'] {
  margin-top: 1rem;
  margin-bottom: 0.5rem;
}
```

If you're wondering why use `:where` over `:is`, it's because your `remove-header-margin` class is only a `(0, 1, 0)` in terms of specificity while your current implementation and `:is` would be `(0, 1, 1)` which would override it. With `:where`, it would leave out the element type specificity which would leave it at the same level as your remove margin class which then just means which is declared last wins.

I'm also not sure if people are aware of how CSS specificity is explicitly calculated but here's a good [resource](https://web.dev/learn/css/specificity#specificity_scoring) if you've never thought about it or only just intuitively know it.

addons/rose/app/styles/rose/_typography.scss
```diff
@@ -12,60 +12,27 @@ html {
...
+  // and link list panel
   font-family: typography.$family-primary;
 
-  // This is the only time we'll explicitly use pixels.  All other sizes are
-  // expressed in rem units.
-  font-size: typography.$base-font-size * 1px;
```
hashicc, line 19: Is there somewhere else where we set the root font size? This is related to this [suggestion](https://github.com/hashicorp/boundary-ui/pull/2803/files#r2079836275) to set the base on `html`

addons/rose/addon/styles/hds/overrides.scss
```diff
@@ -3,6 +3,17 @@
...
+  margin-top: 1rem;
```
hashicc, line 13: Would `sizing.rems` be useful here? Maybe not! I'm still trying to wrap my head around the intention of our sizing scales. One reason I like tailwind is that the scales are thought out for the context of the particular dimension.

addons/rose/addon/styles/hds/overrides.scss
cameronperera: Thanks for the detailed comment. I was not aware of `:where()` and that did resolve the specificity issue.

addons/core/tests/dummy/app/templates/smoke-test.hbs
cameronperera: Fixed.

addons/rose/app/styles/rose/_typography.scss
cameronperera: @hashicc and I did some testing and discussion and decided that we should not be setting `font-size` on the body. Now that we have moved primarily to `rem`, any `rem` sizes are ignoring the `font-size` set on body and looking at the default size set on `html`. However, we have some places that rely on `inheritance` which will recursively look at each parent until it finds a set `font-size`. By defining it on the body AND html, we had 2 sources of truth which causes mayhem (if we theoretically changed the `font-size` on body).

addons/rose/addon/styles/hds/overrides.scss
cameronperera: With recent styles related discussions, @hashicc and I talked about this. We landed on leaving this as a direct `rem` declaration since it is spacing related and not font-size.

addons/rose/app/styles/rose/_typography.scss
```diff
@@ -12,60 +12,27 @@ html {
...
-
-h2,
-.h2 {
-  @include type-mixins.type(h2, bold);
-
-  margin-top: sizing.rems(m);
-  margin-bottom: sizing.rems(xs);
+// By default, we add top and bottom margins to all headers
+// so this class can be used to remove the margin if needed
+.remove-header-margin {
```
ZedLi, line 27: In retrospect, if this is called `remove-header-margin`, shouldn't it have targeted for only header tags? Isn't this class as it is just a generic "remove margin" class?

addons/rose/app/styles/rose/_typography.scss
ZedLi: Out of curiosity, where do we depend on inheritance with font-size?

cameronperera: our `html` tag has a `font-size` that rem units are already using. The `font-size` on the body was actually blocking the inheritance as it was stopping at the body tag.

ZedLi: I mean I think that would be intentional, if you wanted `ems` to be different from `rems` and controlled in the code _but_ not remove the ability for an end user to change their default font size, that would be the way.

I was more curious with this statement and what's actually being used in code that is recurisvely looking at the font-size:
> we have some places that rely on inheritance which will recursively look at each parent until it finds a set font-size

cameronperera: Gotcha, sorry for the confusion. One example would be the Hds::Breadcrumb component. It relies on inheritance to set the font values. From my testing, it was all Hds components. My plan was to show this during techtime to make the team aware.

ZedLi: Sounds good, I'll wait until tech time since I didn't see the inheritance but I didn't look super hard, this was what I saw for breadcrumbs:

image

DhariniJeeva (approved):
🎉  awesome work! thanks for making this huge upgrade

priya-patel04 (approved):
Looks great! Thanks for making all the changes 🚀

hashicc (approved):
💯 Thanks for trying all the combinations of font-size , tricky tricky!

PR #2812, opened 2025-05-02
[Replace][Desktop] Rose::Button

laurenolivia:
# Description
This PR replaces instances of Rose::Button in ui/desktop, and fixes a bug currently in prod where hovering over our fullscreen/minimize btns do not trigger the hover settings by changing the btn background to white in the toolbar.

Our OS-specific icons (e.g. macOS traffic lights) are handled within the electron app, and do not need migrating. See [docs](https://www.electronjs.org/docs/latest/api/base-window#new-basewindowoptions).

This PR updates acceptance tests in ui/admin that were still referencing the .rose-button class.

:tickets: [Jira ticket](https://hashicorp.atlassian.net/browse/ICU-16895)

Linux:
linux

Windows:
Screenshot 2025-05-19 at 5 21 02 PM
Screenshot 2025-05-19 at 5 20 56 PM

## Checklist


ZedLi:
Can we get screenshots from windows and linux instead of just test runs?

ui/desktop/app/templates/application.hbs
```diff
@@ -21,26 +21,32 @@
...
-            @style='warning'
+          <Hds::Button
+            {{on 'click' this.close}}
             class='button-window-close'
             title={{t 'actions.close'}}
-            {{on 'click' this.close}}
+            @text={{t 'actions.close'}}
+            @icon='x'
+            @isIconOnly={{true}}
+            @style='warning'
```
priya-patel04, line 31: It seems `@style` arg is custom to [Rose::Button](https://github.com/hashicorp/boundary-ui/blob/93a57f26ee75e073427a1074fd893bf94ecf58e6/addons/rose/addon/components/rose/button/index.hbs#L10) only. Here's styling for [warning](https://github.com/hashicorp/boundary-ui/blob/9b1f052a569fc1579befffda87de3a91ba5e417b/addons/rose/app/styles/rose/components/_button.scss#L74).  We don't have it for [HDS::Button ](https://helios.hashicorp.design/components/button?tab=code). Can we remove this arg and also make sure the styling remains consistent ?

laurenolivia: Oh good catch, yep, will fix

ui/desktop/app/styles/app.scss
```diff
@@ -135,6 +135,32 @@
...
+  }
+
+  .button-window-minimize,
+  .button-window-fullscreen {
+    &:not(:disabled) {
+      &:hover {
+        background: var(--white);
+        color: var(--ui-gray-starker-4);
+
+        @media (prefers-color-scheme: dark) {
```
laurenolivia, line 153: We had a bug in the desktop client where hovering over the minimize/fullscreen btns did not trigger a color change. See video.

https://github.com/user-attachments/assets/ff737650-4786-496f-b5b5-fa8ae2a67b0b

ZedLi: Good catch but a few questions: 
* What scenario would we need the `not(:disabled)`?
* We shouldn't need another media query for dark mode, the tokens should have automatically handled it
  *  Since it's already always a black toolbar/buttons, do we need these to be changing colors when switching between light/dark?

laurenolivia: - Fair question. Those btns do not need a setting to account for being disabled.
- Correct, light/dark-mode specific settings were overcomplicating the fix. Our `var(--white)` and `var(--black)` are not overridden in the application, so I used those instead. Tested this at the system and electron app level and the fix works.

ui/desktop/app/styles/app.scss
```diff
@@ -135,6 +135,22 @@
     margin: sizing.rems(xs);
   }
 
+  .button-window-close {
+    &:not(:disabled) {
```
ZedLi, line 139: Any reason this was kept for the close button but not the minimize/fullscreen buttons?

laurenolivia: That was left in err. Removed.

ui/desktop/app/styles/app.scss
```diff
@@ -483,3 +497,8 @@
     }
   }
 }
+
+.header-spacing {
```
ZedLi, line 501: This looks like it got messed up from the merge conflict?

ZedLi (changes requested):

priya-patel04 (approved):
Great to see all rose buttons migrated! 🎉

ui/desktop/app/styles/app.scss
```diff
@@ -123,6 +123,20 @@
     margin: sizing.rems(xs);
   }
 
+  .button-window-close {
+    &:hover {
+    background: var(--failure);
```
ZedLi, line 128: It looks like this formatting didn't get fixed, were the git hooks bypassed? Can you re-run the formatter for this file

laurenolivia: It should not get bypassed! Fixing.

ZedLi (approved):
Looks good!

priya-patel04 (approved):

PR #2813, opened 2025-05-02
Refactor: Rose::Page::Layout - header, actions and breadcrumbs

priya-patel04:
# Description
We wont be able to migrate Rose::Layout::Page completely, but we can combine [<:breadcrumbs>](https://github.com/hashicorp/boundary-ui/blob/9b1f052a569fc1579befffda87de3a91ba5e417b/addons/rose/addon/components/rose/layout/page/index.hbs#L9), [<:header>](https://github.com/hashicorp/boundary-ui/blob/9b1f052a569fc1579befffda87de3a91ba5e417b/addons/rose/addon/components/rose/layout/page/index.hbs#L12) and [<:actions>](https://github.com/hashicorp/boundary-ui/blob/9b1f052a569fc1579befffda87de3a91ba5e417b/addons/rose/addon/components/rose/layout/page/index.hbs#L13) into  [ HDS::PageHeader](https://helios.hashicorp.design/components/page-header) component.

This is Part 1 of the PR focusing on scope routes for:
1. add-storage-policy
2. aliases
3. auth-methods
4. credential-stores

:tickets: [Jira ticket](https://hashicorp.atlassian.net/browse/ICU-16852)

### Before:
Screenshot 2025-05-06 at 12 03 14 PM

### After: 
Screenshot 2025-05-06 at 12 03 22 PM

## How to Test
Go to routes for below and ensure it renders the page header component properly: 
1. add-storage-policy
2. aliases
3. auth-methods
4. credential-stores

## Checklist


ui/admin/app/templates/scopes/scope/edit.hbs
```diff
@@ -13,55 +13,72 @@
...
+            doc='project'
+          )
+          (hash
+            key='isScope'
+            title='resources.scope.title'
+            description='resources.scope.description'
+            doc='scope'
+          )
+        )
+        as |scopes|
```
DhariniJeeva, line 50: could we move this to the controller? I think it would be easier to read and maintain in JS rather than building the structue directly in the template

ui/admin/app/templates/scopes/scope/credential-stores/credential-store/credentials/index.hbs
```diff
@@ -3,29 +3,32 @@
...
+        <CredentialStores::CredentialStore::Header @header={{header}} />
+      {{/let}}
+      <PH.Actions>
+        <CredentialStores::CredentialStore::Actions
+          @model={{this.credentialStore}}
+          @delete={{fn this.credentialStores.delete this.credentialStore}}
+        />
+      </PH.Actions>
+    </Hds::PageHeader>
+    <Hds::Copy::Snippet
```
hashicc, line 22: Could this be kept inside `CredentialStores::CredentialStore::Header` like it was before? That'd help reduce the duplication and maintain the existing header pattern

ui/admin/app/templates/scopes/scope/credential-stores/credential-store/credentials/index.hbs
```diff
@@ -3,29 +3,32 @@
...
-
+<Rose::Layout::Page class='new-rose-page-layout' as |page|>
   <page.header>
-    <CredentialStores::CredentialStore::Header
-      @model={{this.credentialStore}}
+    <Hds::PageHeader as |PH|>
+      <PH.Breadcrumb>
+        <Breadcrumbs::Container />
+      </PH.Breadcrumb>
+      {{#let (hash title=PH.Title description=PH.Description) as |header|}}
```
hashicc, line 12: Is it easier to just pass in `PH` as is?

```hbs
<CredentialStores::CredentialStore::Header @header={{PH}} />
```

ui/admin/app/templates/scopes/scope/credential-stores/credential-store/credentials/index.hbs
priya-patel04: Ahh yes, that is good suggestion. I will try using <PH.Generic /> for the code snippet

ui/admin/app/templates/scopes/scope/credential-stores/credential-store/credentials/index.hbs
priya-patel04: Thanks for pointing this out. Updated!

ui/admin/app/styles/app.scss
```diff
@@ -17,6 +17,16 @@
   }
 }
 
+// remove this after the new layout is in place. This is a temporary fix
+// to make the new layout work with the old layout
+.new-rose-page-layout {
```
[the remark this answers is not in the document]
priya-patel04: That is a good point updated for better understanding.

ui/admin/app/controllers/scopes/scope/edit.js
```diff
@@ -7,4 +7,33 @@ import Controller, { inject as controller } from '@ember/controller';
 
 export default class ScopesScopeEditController extends Controller {
   @controller('scopes/scope/index') scopes;
+
+  get scopeTypes() {
+    return [
+      {
+        key: 'isGlobal',
+        title: 'resources.global.title',
```
laurenolivia, line 15: Non-blocking: I noticed our translations define scope types: org, global, project [here](https://github.com/hashicorp/boundary-ui/blob/main/addons/core/translations/resources/en-us.yaml#L993-L996), as well. Wondering if that's necessary.

ui/admin/app/components/auth-methods/auth-method/header/index.hbs
```diff
@@ -3,16 +3,22 @@
...
-<Hds::Copy::Snippet @textToCopy={{@model.id}} @color='secondary' />
\ No newline at end of file
+</@header.Title>
+<@header.Description>
+  {{t 'resources.auth-method.description'}}
+</@header.Description>
+<@header.Generic>
+  <div class='copy-snippet-container'>
+    <Hds::Copy::Snippet @textToCopy={{@model.id}} @color='secondary' />
+  </div>
```
DhariniJeeva, line 23: should this be a [subtitle](https://helios.hashicorp.design/components/page-header#subtitle)? 

> Common examples of this are organization name, project name, unique IDs, resource names, etc.

ui/admin/app/components/auth-methods/auth-method/accounts/account/header/index.hbs
```diff
@@ -3,11 +3,15 @@
...
+</@header.Title>
+<@header.Description>
   {{t 'resources.account.description'}}
-</Hds::Text::Body>
-<Hds::Copy::Snippet @textToCopy={{@model.id}} @color='secondary' />
\ No newline at end of file
+</@header.Description>
+<@header.Generic>
+  <div class='copy-snippet-container'>
+    <Hds::Copy::Snippet @textToCopy={{@model.id}} @color='secondary' />
```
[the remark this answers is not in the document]
priya-patel04: Thanks for pointing out, we have plenty of space to fit the value for `textToCopy`. And also for a11y reasons, it's best to avoid truncation https://helios.hashicorp.design/components/copy/snippet#truncation

ui/admin/app/controllers/scopes/scope/edit.js
priya-patel04: We are still using those translations, these are just the `key` values to get the corresponding translations based on scope type.

ui/admin/app/components/auth-methods/auth-method/header/index.hbs
priya-patel04: Hmm good point, however if we do use Subtitle, then it will move the copy snippet button above the description. I think it makes sense to have it below description but open to what others think:
Screenshot 2025-05-14 at 4 06 13 PM

laurenolivia: Looks like we tend to place code snippets below descriptions. Perhaps we should remain consistent?

Screenshot 2025-05-14 at 5 38 27 PM
Screenshot 2025-05-14 at 5 40 12 PM

ui/admin/app/components/auth-methods/auth-method/accounts/account/header/index.hbs
```diff
@@ -3,11 +3,15 @@
...
-<Hds::Text::Body @tag='p' @size='300' class='p'>
+</@header.Title>
+<@header.Description>
   {{t 'resources.account.description'}}
-</Hds::Text::Body>
-<Hds::Copy::Snippet @textToCopy={{@model.id}} @color='secondary' />
\ No newline at end of file
+</@header.Description>
+<@header.Generic>
+  <div class='copy-snippet-container'>
```
ZedLi, line 14: Did we need this extra wrapping div for the class rather than just attaching the class directly to `<Hds::Copy::Snippet>`?

addons/rose/app/styles/rose/components/layout/_page.scss
```diff
@@ -23,22 +23,25 @@
...
       gap: sizing.rems(xxs);
 
       .doc-link {
         display: flex;
         align-items: center;
       }
     }
-  }
+    /* stylelint-disable selector-class-pattern */
+    .hds-page-header__body {
```
ZedLi, line 38: Any reason you added margin directly with the HDS class versus adding a custom class but used a custom class for `copy-snippet-container` rather than just directly selecting for the HDS copy snippet? It seems like it could've been one way or the other

This feels like a scenario where we should've just used a utility class rather than these specific styles but I know we haven't standardized on that yet.

ZedLi (commented):
This is looking really good!! Just left a few minor comments

ui/admin/app/styles/app.scss
```diff
@@ -17,6 +17,16 @@
   }
 }
 
+// remove this after the new layout is in place. This is a temporary fix
+// to make the new layout work with the old layout
+.rose-layout-page.hds-migration-overrides {
```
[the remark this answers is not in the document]
priya-patel04: Yes right, we would need the override for both admin and DC. Updated!

ui/admin/app/styles/app.scss
```diff
@@ -1037,3 +1047,9 @@
     margin-left: 1.5rem;
   }
 }
+
+// auth method header
+.primary-badge {
+  margin-right: sizing.rems(xxs);
+  margin-left: sizing.rems(xxs);
```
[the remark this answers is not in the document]
priya-patel04: Ahh yes, I didnt want to update gap for header as it will increase a smidge gap for the `doc-link`. But I think it looks okay with increased gap:

Before: 

Screenshot 2025-05-15 at 10 07 23 AM

After: 

Screenshot 2025-05-15 at 10 07 09 AM

addons/rose/app/styles/rose/components/layout/_page.scss
priya-patel04: Thanks for catching this. I used HDS classes to avoid adding more custom class for just margin. But turns out we can just leverage existing `rose-page-layout-header` class instead.  
And for HDS copy snippet - Since the parent container has flex property, I added a wrapper `div` with custom class to make sure it does not take full-width of it's parent. I found a workaround and used HDS classes to avoid extra custom class for this as well.

But yes, you are right - ideally a utility class would be more consistent and scalable, once we have standardization that’s in place, these margin adjustments could be handled more uniformly.

ui/admin/app/components/auth-methods/auth-method/accounts/account/header/index.hbs
priya-patel04: Thanks for pointing this out. Since the parent container has flex property, I added a wrapper `div` with custom class to make sure it does not take full-width of it's parent. I found a workaround and used HDS classes to avoid extra custom class for this as well.

ZedLi (approved):
Looks great, thanks for the update!

ui/admin/app/styles/app.scss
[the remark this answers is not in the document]
ZedLi: Yeah I was thinking the same thing, to me it seems like maybe the space was too small to start with for the `doc-link` if adding anything else makes it look like there's not enough whitespace between the elements

cameronperera (approved):

ui/admin/app/components/auth-methods/auth-method/header/index.hbs
DhariniJeeva: makes sense, looks like we also don't need an extra div anymore so I'm good with leaving it as is

PR #2817, opened 2025-05-05
Add check to only show windows buttons on windows

calcaide:
# Description

When using mirage on desktop client ONLY show windows action buttons when using windows.

## Screenshots:

Before using mac showing both mac and windows:
before_mac

After mac ONLY showing mac action buttons:
after_mac

Windows showing ONLY windows action buttons:
after_windows

Linux (Ubuntu) showing windows action buttons:
Screenshot 2025-05-07 at 8 26 41 AM

## How to Test
- Run desktop client using mirage in Mac and see only Mac action buttons.
- Run desktop client using mirage in Windows and see only Windows action buttons.


addons/api/mirage/scenarios/ipc.js
```diff
@@ -199,7 +199,7 @@ export default function initializeMockIPC(server, config) {
      * Check for OS chrome state
      */
     showWindowActions() {
-      return true;
+      return window.navigator?.userAgentData?.platform === 'Windows';
```
lisbet-alvarez, line 202: should these window actions also be shown on linux?

calcaide: Good call! building a VM on linux to validate 😉

ZedLi: Which scenarios is `userAgentData` available? I don't see it firefox or safari (is this chromium only?) but I do see it in `navigator.userAgent` though it would require you to pull it out. If we go down this path, should we use that one instead or as a backup field? And default to true if we don't find the properties?

calcaide: `userAgenData` is indeed not present on firefox, and for the moment available in chromium. [More info about it here](https://developer.mozilla.org/en-US/docs/Web/API/Navigator/userAgentData). I assume, since this is DC that runs on Electron which uses chrome it was ok to use, also because the `userAgent.platform` [its been deprecated as per this information](https://developer.mozilla.org/en-US/docs/Web/API/Navigator/platform).

I will look for a more compatible solution, open to suggestions 😉

DhariniJeeva (approved):
thanks for this fix!

priya-patel04 (approved):
Thanks for the fix!

hashicc (approved):
🕺 appreciate all the fixes

lisbet-alvarez (approved):
Thank you for fixing this!

PR #2820, opened 2025-05-07
Use `@use` for scss explicit imports

hashicc:
# Description
This pr migrates [deprecated `@import`](https://sass-lang.com/blog/import-is-deprecated/) syntax to `@use` within `boundary-ui`, which removes `Deprecation Warning [import]` warnings seen in the build process.

The `sass-migrator` cli tool was used and then reverted any variables that were being re-used from global scope. For example the tool took:
```scss
$m: sizing.rems(m); // 16
$s: sizing.rems(s); // 12
$xs: sizing.rems(xs); // 8
$xxs: sizing.rems(xxs); // 4
$xxxs: sizing.rems(xxxs); // 2
$xxxxs: sizing.rems(xxxxs); // 1
```

and converted it to:

```scss
@use "../card/link";

$m: sizing.rems(m); // 16
$s: sizing.rems(s); // 12
$xs: sizing.rems(xs); // 8
$xxs: sizing.rems(xxs); // 4
link.$xxxs: sizing.rems(xxxs); // 2
$xxxxs: sizing.rems(xxxxs); // 1
```
While previously the `$xxxs` variable was being re-declared from global scope, the `sass-migrator` tried to preserve that behavior by importing `link` and re-assigning the `$xxxs` variable. Instead these should variables should be preserved "as is" within the local module scope. There was [one case](https://github.com/hashicorp/boundary-ui/pull/2820/files#r2078305425) where a local `scss` file depended on a global variable that was not (re-)declared in the file, the rest were just re-using variables that already existed globally.

The `sass-migrator` tool also identified cases where internal sass functions were being used. In these cases it added the appropriate `@import` and updated the reference to use the imported module.

⚠️ Note: There's still one `@import` deprecation in the [`@hashicorp/ember-asciinema-player` dependency](https://github.com/hashicorp/ember-asciinema-player/blob/9dd2072b3f8ae0ff1015965c99bcbf493420ec88/app/styles/%40hashicorp/ember-asciinema-player.scss#L6)

## How to Test
* Ember builds should work without failure or mention of the `@import` deprecation (except for `@hashicorp/ember-asciinema-player` dependency)
* Styles should still work and look the same (clicking around [deployed environments](https://github.com/hashicorp/boundary-ui/pull/2820#issuecomment-2858638546))

## Checklist

### DIFF
--- a/addons/rose/app/styles/rose/components/form/_select.scss
+++ b/addons/rose/app/styles/rose/components/form/_select.scss
@@ -1,5 +1,3 @@
-@use '../button';
-
 /**
  * Copyright (c) HashiCorp, Inc.
  * SPDX-License-Identifier: BUSL-1.1
@@ -7,6 +5,8 @@
 
 @use '../../variables/sizing';
 @use '../../utilities/type';
+
+$size-xxs: sizing.rems(xxs); // 4
 
 .rose-form-select {
   margin-bottom: sizing.rems(m);
@@ -24,7 +24,7 @@
   padding: sizing.rems(xs) sizing.rems(xxl) sizing.rems(xs) sizing.rems(xs);
   border: sizing.rems(xxxxs) solid var(--ui-gray-subtler-2);
   border-radius: sizing.rems(xxxs);
-  box-shadow: 0 button.$size-xxs button.$size-xxs
+  box-shadow: 0 $size-xxs $size-xxs
     rgba(var(--stark-components), var(--opacity-2));
   background: var(--ui-gray-subtler-5)
     url('data:image/svg+xml;charset=utf8,%3Csvg%20viewBox%3D%220%200%2024%2024%22%20fill%3D%22none%22%20xmlns%3D%22http%3A%2F%2Fwww.w3.org%2F2000%2Fsvg%22%3E%3Cpath%20fill-rule%3D%22evenodd%22%20clip-rule%3D%22evenodd%22%20d%3D%22M12%205.83L15.17%209l1.41-1.41L12%203%207.41%207.59%208.83%209%2012%205.83zm0%2012.34L8.83%2015l-1.41%201.41L12%2021l4.59-4.59L15.17%2015%2012%2018.17z%22%20fill%3D%22%23525761%22%2F%3E%3C%2Fsvg%3E')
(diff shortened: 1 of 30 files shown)


addons/rose/app/styles/rose/components/form/_select.scss
```diff
@@ -6,6 +6,8 @@
 @use '../../variables/sizing';
 @use '../../utilities/type';
 
+$size-xxs: sizing.rems(xxs); // 4
```
hashicc, line 9: This variable was previously expected to be available globally. It's used on [line 27](https://github.com/hashicorp/boundary-ui/pull/2820/files#diff-7a0a66e0bf65a1140f82f3672e78d0517a4bf609dced637f2eb48cf5531a674aR27).

ZedLi: I think this is fine but I'm also not sure why it was originally just this one that was using a variable when there are other `sizing.rem` usages that are repeated and used directly

ZedLi (approved):
Thanks for the upgrade!

addons/rose/app/styles/rose/components/form/_select.scss
cameronperera: This is odd. We use `sizing.rems()` directly just a few lines above. Should we remove the usage of `$size-xxs` and use `sizing.rems(xxs)` directly?

addons/rose/app/styles/rose/components/layout/_sidebar.scss
```diff
@@ -1,3 +1,5 @@
+@use '../../variables/color/sass';
```
cameronperera, line 1: nit: Should this `@use` be below the copyright header?

addons/rose/app/styles/rose/components/form/_select.scss
hashicc: I was mostly trying to preserve the existing variable usage which was using a global variable. But agreed, in context of the file it's not necessary, I'll update it to use the function directly

addons/rose/app/styles/rose/components/layout/_sidebar.scss
hashicc: You bet, thanks. I thought I got call these

ZedLi (approved):

cameronperera (approved):

priya-patel04 (approved):
Nice upgrade!

DhariniJeeva (approved):
thanks for the upgrading this!

PR #2815, opened 2025-05-05
[Retire] Rose::Form::Label

laurenolivia:
# Description
This PR retires the Rose::Form::Label component from our codebase.

[PR #2821](https://github.com/hashicorp/boundary-ui/pull/2821) has to be merged, first.

:tickets: [Jira ticket](https://hashicorp.atlassian.net/browse/ICU-9426)

## How to Test
Tests should pass ✅

## Checklist

- [ ] I have added before and after screenshots for UI changes
- [ ] I have added JSON response output for API changes
- [ ] I have added steps to reproduce and test for bug fixes in the description
- [ ] I have commented on my code, particularly in hard-to-understand areas
- [ ] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works

PR #2821, opened 2025-05-07
[Retire] Rose::Form::Input

laurenolivia:
# Description
This PR retires the Rose::Form::Input component from our codebase.

Once merged, we can then rebase and merge [PR #2815](https://github.com/hashicorp/boundary-ui/pull/2815)

:tickets: [Jira ticket](https://hashicorp.atlassian.net/browse/ICU-15716)

## How to Test
Tests should pass ✅ 

## Checklist

### DIFF
--- a/addons/rose/app/styles/rose/components/form/_index.scss
+++ b/addons/rose/app/styles/rose/components/form/_index.scss
@@ -6,7 +6,6 @@
 @use 'label';
 @use 'errors';
 @use 'helper-text';
-@use 'input';
 @use 'checkbox';
 @use 'select';
 @use 'fieldset';


ZedLi (approved):
Looks good, thanks for the cleanup!

addons/rose/tests/integration/components/rose/form-test.js
```diff
@@ -44,7 +44,7 @@ module('Integration | Component | rose/form', function (hooks) {
         @disabled={{false}}
         as |form|
       >
-        <form.input @label="Input" @name="input-field" />
+        <Hds::Form::TextInput::Field name="input-field" />
```
hashicc, line 47: It looks like this was previously testing the yield from `form`, providing `form.input`. I can see that there's a pattern here of other `Hds` components here but since they aren't related to the contextually yielded components I think any `Hds::` can be deleted and the assertions updated

addons/rose/tests/integration/components/rose/form-test.js
```diff
@@ -96,7 +96,7 @@ module('Integration | Component | rose/form', function (hooks) {
         @showEditToggle={{true}}
         as |form|
       >
-        <form.input @label="Label" @value="value" />
+        <Hds::Form::TextInput::Field @value="value" disabled={{if form.isEditable false true}} />
```
hashicc, line 99: Similar thing here and for the next test, these were previously testing the `disabled` state that was contextually wired on `<form.input />` automatically. Since this wiring isn't supported any more can we also remove the assertion checking for it to be disabled? I think at this point we're testing the `Hds::Form::TextInput::Field` instead of `Rose::Form`

hashicc (changes requested):
Thanks for the final removal of `Rose::Form::Input`!

Two minor bits that would be great to clean up:
1. I think the `Rose::Form` tests could be updated to remove assertions that were previously testing the contextually yielded `input`.
2. I found two other references to `Rose::Form::Input` that could be removed [in this file](https://github.com/hashicorp/boundary-ui/blob/9b1f052a569fc1579befffda87de3a91ba5e417b/addons/rose/tests/dummy/app/templates/playground.hbs#L65).

laurenolivia:
> Thanks for the final removal of `Rose::Form::Input`!
> 
> Two minor bits that would be great to clean up:
> 
> 1. I think the `Rose::Form` tests could be updated to remove assertions that were previously testing the contextually yielded `input`.
> 2. I found two other references to `Rose::Form::Input` that could be removed [in this file](https://github.com/hashicorp/boundary-ui/blob/9b1f052a569fc1579befffda87de3a91ba5e417b/addons/rose/tests/dummy/app/templates/playground.hbs#L65).

@hashicc RE: I found two other references to Rose::Form::Input that could be removed [in this file](https://github.com/hashicorp/boundary-ui/blob/9b1f052a569fc1579befffda87de3a91ba5e417b/addons/rose/tests/dummy/app/templates/playground.hbs#L65).

Is the overall goal to hollow out this file, until it can eventually be deleted? The rose input components have nested rose btns, so just want to clarify your feedback since removing that component also removes the associated btns.

hashicc:
> Is the overall goal to hollow out this file, until it can eventually be deleted? The rose input components have nested rose btns, so just want to clarify your feedback since removing that component also removes the associated btns.

The name of the file is `playground.hbs` and it's in the dummy app, so I'm assuming it's just a place that example snippets of code can be tested and shared. I'm not sure we need the file at all, might be a good question for tech time. But for your PR, I think at least deleting the `Rose::Form::Input` including the the associated buttons within the input contextual components is fine.

hashicc (approved):
🚀 Thanks for cleaning up `Rose::Form::Input`, another rose component gone!

PR #2833, opened 2025-05-15
chore: desktop header nav dropdown changes

DhariniJeeva:
# Description

This PR addresses the background color of the dropdown button in desktop and updates the rose utility dropdown for sign out to HDS

before:
Screenshot 2025-05-13 at 11 32 51 AM

after:

Screenshot 2025-05-15 at 11 06 29 AM

## Checklist


addons/rose/app/styles/rose/components/header/_dropdown.scss
```diff
@@ -10,28 +10,9 @@
...
-          padding-left: sizing.rems(xs);
-        }
-
-        &:hover,
-        &:focus {
-          background-color: transparent;
-        }
-      }
-    }
+   .hds-dropdown > button {
```
DhariniJeeva, line 13: this is a temp fix, we will be retiring this header component when we integrate [hds side nav](https://helios.hashicorp.design/components/app-side-nav?tab=code) in desktop

addons/rose/app/styles/rose/components/header/_dropdown.scss
```diff
@@ -3,35 +3,10 @@
...
+    background: var(--black);
+    border: none;
+    color: var(--ui-gray-subtler-3);
```
ZedLi, line 10: Is the contrast a little low in dark mode with this? Should it just be `--white`?

image

ZedLi (approved):
Looks good, left 2 minor comments!

addons/rose/app/styles/rose/components/header/_dropdown.scss
```diff
@@ -3,35 +3,11 @@
...
-        }
-
-        &:hover,
-        &:focus {
-          background-color: transparent;
-        }
-      }
-    }
+  // Temporary fix: This header component will be retired upon integrating HDS side nav for desktop.
+   .hds-dropdown > button {
```
hashicc, line 8: I know this is supposed to be temporary but it looks like the [toggle button](https://github.com/hashicorp/boundary-ui/pull/2833/files?w=1#diff-de7e5aba3ecc486ff8ba63bcd57bf8e6660eb3d3dd45530d712a2417548bf65cR50) supports "splattributes" so I'm wondering if instead of targeting an `.hds-*` class we could use our own:

```hbs
<dd.ToggleButton class="header-dropdown-button-override" />
```

I think a `data-test-*` on the toggle and the `<dd.Interactive />`  would be supported for the tests as well

e2e-tests/desktop/tests/authentication.spec.js
```diff
@@ -21,15 +21,14 @@ test.describe('user/password authentication tests', () => {
       /.*\/scopes\/global\/projects\/targets$/,
     );
 
+    // click dropdown
     await electronPage
-      .locator('details')
-      .filter({ hasText: 'Sign Out' })
+      .getByRole('button', { name: 'admin' })
```
hashicc, line 26: 💯 good catch!

> If you absolutely must use CSS or XPath locators, you can use [`page.locator`()](https://playwright.dev/docs/api/class-page#page-locator) to create a locator that takes a selector describing how to find an element in the page.

`getByRole` is much preferred

hashicc (changes requested):
One minor suggestion around trying to using class and data-test attributes directly so we can limit the amount of `.hds-*` classes being targeted. I didn't test it myself but looking at the hds source it should work

addons/rose/app/styles/rose/components/header/_dropdown.scss
DhariniJeeva: I like the idea of our own class with a clear purpose but I want to bring up a point here, if we go with this custom class route, we will have to add the classname to this [toggleButton](https://github.com/hashicorp/boundary-ui/pull/2833/files#diff-de7e5aba3ecc486ff8ba63bcd57bf8e6660eb3d3dd45530d712a2417548bf65cR50) and the [toggleIcon](https://github.com/hashicorp/boundary-ui/pull/2833/files#diff-de7e5aba3ecc486ff8ba63bcd57bf8e6660eb3d3dd45530d712a2417548bf65cR55) in the dropdown and as well as for this [dropdown](https://github.com/hashicorp/boundary-ui/blob/80ffa856bd2a0393450709e4a704da7393be38e2/ui/desktop/app/templates/scopes/scope.hbs#L11) in the header nav on the left side.

DhariniJeeva: we could add the override class to the dropdowns and target button instead of using the HDS class name directly here. Let me know what you think.

hashicc: That makes sense. There were two things I think would be good to balance: trying to reduce the number of selectors, nesting or selector complexity to overcome specificity, and avoiding `hds-*` class names if possible. Whatever you think has the best clarity and maintainability

DhariniJeeva: sounds good! I added the custom class and use that to target buttons within the header

hashicc (approved):
🚀 Thanks for the updates and noticing the color fix

ZedLi (approved):
Thanks for the fix!

priya-patel04 (approved):

PR #2838, opened 2025-05-19
[Retire] Rose::Button

laurenolivia:
# Description
This PR retires the Rose::Button component from our codebase.

PR [#2812](https://github.com/hashicorp/boundary-ui/pull/2812) must be merged, first.

The test in this PR that is not currently passing will be resolved once the noted PR above is merged.

:tickets: [Jira ticket](https://hashicorp.atlassian.net/browse/ICU-16830)

## Checklist


ZedLi:
It looks like there's still references here that need to get updated
https://github.com/hashicorp/boundary-ui/blob/f12ae90aba77d7d521d262c8cea238654386fd1f/ui/admin/tests/acceptance/host-catalogs/host-sets/update-test.js#L300

and in our dummy app file which likely can just be deleted

laurenolivia:
> It looks like there's still references here that need to get updated
> 
> https://github.com/hashicorp/boundary-ui/blob/f12ae90aba77d7d521d262c8cea238654386fd1f/ui/admin/tests/acceptance/host-catalogs/host-sets/update-test.js#L300
> 
> and in our dummy app file which likely can just be deleted

Thanks. Thought I caught all of those. Fixed.

ZedLi (approved):
Thanks for the cleanup!

addons/rose/tests/dummy/app/templates/index.hbs
```diff
@@ -4,33 +4,16 @@
 }}
 
 <Rose::Toolbar as |toolbar|>
-  <Rose::Button @style='inline-link-action' @iconLeft='flight-icons/svg/x-16'>
-    Clear something
-  </Rose::Button>
+  <Hds::Button @icon='x' @isInline={{true}} @text='Clear something' />
```
DhariniJeeva, line 7: I thought we voted to get rid of dummy app during our tech time. should we just get rid of this file instead of updating it with HDS components?

DhariniJeeva (approved):
looks good. 

left a comment about updating components within dummy app

addons/rose/tests/dummy/app/templates/index.hbs
hashicc: We talked about getting rid of the `playground.hbs` as part of the dummy app's templates. But for the same reason I don't know if we need this `index.hbs` either since a bunch of random UI components aren't helpful. The rose addon doesn't have any acceptance tests so I doubt this markup is being used or referenced.

That's all to say I agree that we should remove this file as well

ZedLi: We should keep the dummy app and I think we should keep at least it being able to run the rose addon separately. I'm of the opinion we should keep some placeholder template at the minimum, it doesn't need to have any random components but we should at least have it be able to load something.

laurenolivia: Maybe we should chat about this again, as a team.

ZedLi: I think we're basically in agreement, which is fundamentally we don't need random components in our templates or any extra routes and those can be deleted. My only comment was that we should keep whatever the default template file is (without any extra components) just so it at least has a placeholder template to load if you did try and run rose.

Let me know if this is unclear.

PR #2839, opened 2025-05-20
fix: 🐛 refocus on route changes when attempting to search

lisbet-alvarez:
# Description

On any type of route change refocus changed to be the side nav skip link in order to facilitate keyboard navigation (using "tab" key). Therefore, when trying to search since queryParams will be updated, this would be considered a route change causing focus to shift away from search box. This PR adds a check specifically for when route name isn't changing and search queryParam is being used in order to prevent focus being removed from search box. 
For more information on custom route change validation for ember-ally-refocus [go here](https://helios.hashicorp.design/components/side-nav?tab=code#:~:text=a11yRefocusRouteChangeValidator).
Before:

https://github.com/user-attachments/assets/cfa3f151-f0a9-402c-b515-74ce4379b915

After:

https://github.com/user-attachments/assets/7c143f2b-1a48-4cfd-aca6-806a90cd5cf2

## How to Test

1. Go to any resource that has "search" (example: targets).
2. Type in the search box a target name and validate focus is maintained.

## Checklist


ui/admin/app/controllers/application.js
```diff
@@ -7,6 +7,7 @@ import Controller from '@ember/controller';
 import { service } from '@ember/service';
 import { getOwner } from '@ember/application';
 import { action } from '@ember/object';
+import { defaultValidator } from 'ember-a11y-refocus';
```
ZedLi, line 10: Should we actually import this rather that depending it being available from a dependency?

ui/admin/app/controllers/application.js
```diff
@@ -105,4 +106,19 @@ export default class ApplicationController extends Controller {
...
+  /**
+   * Add custom route change validation to prevent refocus when
+   * user is trying to type in search box.
+   * @param {object} transition
+   * @returns
+   */
+  customRouteChangeValidator(transition) {
+    if (
+      transition.to?.name === transition.from?.name &&
+      typeof transition.to.queryParams.search === 'string'
```
ZedLi, line 118: Hmm is the intention to only have it refocus on a page change or anything that's not search? I can imagine it being still jarring if you're a keyboard user and having your focus being reset just for using a filter and having to go through the page again to goto the next filter

ZedLi (commented):
Great fix! Left a few questions

ui/admin/app/controllers/application.js
lisbet-alvarez: good point! Ill add `ember-ally-refocus` as a dep. in admin ui

ui/admin/app/controllers/application.js
lisbet-alvarez: Yup that is correct.
Those are good points I'll modify this condition to be less strict

ui/admin/app/controllers/application.js
```diff
@@ -105,4 +111,21 @@ export default class ApplicationController extends Controller {
...
+   * Add custom route change validation to prevent refocus when
+   * user is attempting to search, filter, or sort.
+   * @param {object} transition
+   * @returns {boolean}
+   */
+  customRouteChangeValidator(transition) {
+    const urlHasQueryParams = /^[\S]+[?][\S]+$/;
+    if (
+      transition.to?.name === transition.from?.name &&
+      urlHasQueryParams.test(this.router.currentURL)
```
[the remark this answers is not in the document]
lisbet-alvarez: I checked all the places we are using query params and u are correct they were all for search, filtering, sort, or type change (on a new form). I did attempt to use `allyRefocusExcludeAllQueryParams` but it just checks queryParams are defined in that route and there was one inconsistency. For example when transitioning from `scopes.scope.scopes.index` (actual url being `scopes/global/scopes`) to another `scopes.scope.scopes.index` route (where actual url is `scopes/org_id/scopes` focus would not be reset. Same thing happens if I only check route name. 
Here is an example using `a11yRefocusExcludeAllQueryParams`:

https://github.com/user-attachments/assets/da698a56-4b67-412a-aa8b-211dde239385

Perhaps this is minor and for simplicity easier to use `allyRefocusExcludeAllQueryParams`?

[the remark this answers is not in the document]
ZedLi: Ah interesting, is this only with the scope picker or also just navigating between scopes?

If that's the case and we want to keep this check, are you able to just use `transition.to.queryParams` instead of using regex to parse the URL or will that not work?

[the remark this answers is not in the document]
hashicc: [Is this current logic roughly the same as `excludeAllQueryParams`?](https://github.com/ember-a11y/ember-a11y-refocus/blob/40df81d974420c3ef824553e5114d195ddc644d4/addon/src/components/navigation-narrator.gts#L135-L141). Refactoring to what @ZedLi is suggesting I think would bring it even closer to that implementation.

---

> Same thing happens if I only check route name.

This might work as a tiebreaker in the case the route names are the same:

The `to` and `from` are [`RouteInfo`](https://api.emberjs.com/ember/6.3/classes/RouteInfo)

For the `params` property:
> The values of the route's parameters. These are the same params that are received as arguments to the route's model hook. Contains only the parameters valid for this route, if any (params for parent or child routes are not merged).

I'm wondering if in this case the `params` would contain and then you could compare a different even if the route matches:
* `scopes/global/scopes` (`["global"]` in `transition.from.params`)
* `scopes/org_id/scopes` (`["org_id"]` in `transition.to.params`)

That might not have all the information so going to all the `transition.from.parent` which is also a `RouteInfo` and building the total `params `for all `from`... and then doing the same for all `to`, and then comparing: `allFrom.join("") === allTo.join("")`?

Maybe? 🤷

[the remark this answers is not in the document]
lisbet-alvarez: @ZedLi 
Same thing happens when navigating in between scopes using table links.
So I tried using queryParams but it seems to be a bit tough to work with since sometimes it includes queryParams default values.

@hashicc 
Actually, did think about doing something like that since there actually already is a [util for finding all dynamic segments](https://github.com/hashicorp/boundary-ui/blob/main/ui/admin/app/utils/param-value-finder.js) (`params`) tho ultimately decided on using regex because it seemed like a simpler solution. But if we like the approach of checking the `params` more im open to trying that instead.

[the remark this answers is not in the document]
hashicc: There's a few cases here, maybe it would be good to write some tests with this PR.

Ah good find, that `paramValueFinder` is exactly it. The regex looks like it handles the query params case but does it handle the same to/from route with different dynamic segment case?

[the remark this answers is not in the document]
lisbet-alvarez: yes u are correct! Tests were also helpful in keeping track of all the cases, thank you!

ui/admin/tests/unit/controllers/application-test.js
```diff
@@ -58,4 +60,55 @@ module('Unit | Controller | application', function (hooks) {
...
+    };
+
+    assert.true(controller.customRouteChangeValidator(transition));
+
+    transition.from = undefined;
+
+    assert.true(controller.customRouteChangeValidator(transition));
+  });
+
+  test('customRouteChangeValidator returns true when transitioning between different routes with the same name', function (assert) {
```
ZedLi, line 77: Would it be more accurate to say this is transitioning between same routes but different path parameters?

ZedLi (approved):
Looks good! Thanks for this investigation and tests!

ui/admin/tests/unit/controllers/application-test.js
lisbet-alvarez: yeah and thats clearer too. thank u, ill update the wording

cameronperera (approved):
Nice work!

ZedLi (dismissed):

ui/admin/tests/unit/controllers/application-test.js
```diff
@@ -58,4 +60,60 @@ module('Unit | Controller | application', function (hooks) {
 
     assert.true(featuresService.isEnabled('ssh-target'));
   });
+
+  test('customRouteChangeValidator returns true when transitioning between different routes', function (assert) {
```
ZedLi, line 64: Not blocking but I also wonder if these tests are similar enough to get parametrized?

lisbet-alvarez: great suggestion!! 
updated the tests!

ZedLi (approved):
Looks amazing, love the new tests!

PR #2841, opened 2025-05-21
Icu 17092 replace hosts connect modal with table in target detail page

cameronperera:
# Description

This PR removes the host modal when trying to connect to a target with more than one host source. Now a target with more than one host source will display a table that allows the user to select a specific host to connect to or use the "Quick Connect" button in the top right corner of the target detail page.

**Note: This PR does not add pagination or search support.**

:tickets: [Jira ticket](https://hashicorp.atlassian.net/browse/ICU-17092)

https://github.com/user-attachments/assets/17633a5c-d88b-45f8-bba7-068a15d78c7a

## How to Test

Start a `boundary dev` instance and start the desktop using this branch. Connect to the different types of Targets and the connect experience should be the same outside of no modal. Should see a table with host sources now.
Target with direct address
Target with one host source (will not display host source table and should connect right away)
Target with more than one host source (should display table and button in top right should say "Quick Connect")

## Checklist


e2e-tests/desktop/tests/targets.spec.js
```diff
@@ -166,13 +166,18 @@ test.describe('Targets tests', () => {
...
-
-      await authedPage.getByRole('button', { name: host }).click();
+      // eslint-disable-next-line playwright/no-conditional-in-test
+      if (host === 'Quick Connect') {
+        await authedPage.getByRole('button', { name: host }).click();
+      } else {
+        await authedPage
+          .getByRole('table')
+          .getByRole('cell', { name: host })
+          .locator('..')
```
ZedLi, line 177: We should avoid using `.locator('..')` when possible, there's usually other ways of getting the right locator as using xpath is a last resort.

You might have been able to do something like this (I didn't confirm this):

```
authedPage
  .getByRole('row')
  .filter({ hasText: host})
  .getByRole('button', { name: 'Connect' })
  .click();
```

ZedLi (approved):
Looks great, love all the code and logic you were able to remove with this change!

e2e-tests/desktop/tests/targets.spec.js
moduli: Nice

ZedLi (approved):

DhariniJeeva (approved):
this looks great! thanks for the work

PR #2850, opened 2025-05-22
fix session route model caching

hashicc:
# Description

## Checklist


ui/admin/app/routes/scopes/scope/sessions/index.js
```diff
@@ -41,9 +42,8 @@ export default class ScopesScopeSessionsIndexRoute extends Route {
     },
   };
 
-  allSessions;
-  associatedUsers = [];
-  associatedTargets = [];
+  @tracked
```
hashicc, line 45: I don't think `@tracked` is actually needed but it's probably a good thing to have

ui/admin/app/routes/scopes/scope/sessions/index.js
```diff
@@ -94,104 +92,111 @@ export default class ScopesScopeSessionsIndexRoute extends Route {
...
+        };
+
+        targets = await this.store.query('target', {
+          scope_id: scope.id,
+          query: { filters },
+        });
+      }
     }
+
+    return Object.freeze({
```
hashicc, line 194: No longer directly sets anything on `this` just returns the context object

ui/admin/app/routes/scopes/scope/sessions/index.js
```diff
@@ -201,12 +206,7 @@ export default class ScopesScopeSessionsIndexRoute extends Route {
...
   @action
   async refreshAll() {
-    const scope = this.modelFor('scopes.scope');
-
-    await this.getAllSessions(scope.id);
-    await this.getAssociatedUsers(scope);
-    await this.getAssociatedTargets(scope);
-
+    this.scopeContext = undefined;
     return super.refresh(...arguments);
```
hashicc, line 210: `refresh` will call all the model hooks which will refresh the `scopeContext`

ui/admin/app/routes/scopes/scope/sessions/index.js
```diff
@@ -117,104 +115,111 @@ export default class ScopesScopeSessionsIndexRoute extends Route {
...
-        scope_id: scope.id,
-        query: { filters },
-      };
-      this.associatedTargets = await this.store.query(
-        'target',
-        associatedTargetsQuery,
-      );
+    //
+    // `targets`
+    //
```
cameronperera, line 190: for my learning: Is there a special purpose for this style of comment?

ui/admin/app/routes/scopes/scope/sessions/index.js
```diff
@@ -117,104 +115,111 @@ export default class ScopesScopeSessionsIndexRoute extends Route {
...
-      if (!this.allSessions) {
-        await this.getAllSessions(scope_id);
-      }
-      if (this.associatedUsers.length === 0) {
-        await this.getAssociatedUsers(scope);
-      }
-      if (this.associatedTargets.length === 0) {
-        await this.getAssociatedTargets(scope);
+      // refresh the saved current scope context when the scope changes or if it has not been previously set
+      if (this.scopeContext?.scopeId !== scope.id) {
```
lisbet-alvarez, line 119: Im wondering perhaps we should also refresh the saved content when `allSessions` is an empty array because if a user has no sessions on first load they will have to manually refresh the browser to see newly created sessions. Any other time they can use our `Refresh` button to see new changes.

hashicc: Is the issue when `sessions` on [115](https://github.com/hashicorp/boundary-ui/pull/2850/files#diff-9d08078089e21a92f6a89fe0ee7984667857fd394d1ccb5ca8533ca531a63bd0R115) has results but `allSessions` doesn't get updated because the `scope.id` is the same so they wouldn't see the new results although we know there are some?

> will have to manually refresh the browser

Wouldn't the `Refresh` button work in this case also since it would clear the `this.scopeContext`?

ui/admin/app/routes/scopes/scope/sessions/index.js
hashicc: I toyed with roughly what the class-based approach for the scope context could look like, it seems a bit verbose but also encapsulates the logic separately from the route:

```js
class ScopeContext {
  #scope;
  #data;

  constructor(scope) {
    this.#scope = scope;
  }

  get store() {
    return getOwner(this).lookup('service:store');
  }

  get can() {
    return getOwner(this).lookup('service:can');
  }

  get data() {
    if (!this.#data) {
      throw new Error('ScopeContext data has not been fetched yet.');
    }

    return this.#data;
  }

  async fetch() {
    const scope = this.#scope;
    const orgScope = await this.store.findRecord('scope', this.scope.scope.id);
    const globalScope = await this.store.findRecord('scope', 'global');

    const allSessions = await this.#getAllSessions();
    const users = this.#getUsers(allSessions, globalScope, orgScope);
    const targets = this.#getTargets(allSessions);

    this.#data = { scopeId: scope.id, allSessions, users, targets };
  }

  async #getAllSessions() {
    const scope = this.#scope;
    const allSessions = await this.store.query(
      'session',
      {
        scope_id: scope.id,
        include_terminated: true,
        query: { filters: { scope_id: [{ equals: scope.id }] } },
      },
      {
        pushToStore: false,
      },
    );

    return allSessions;
  }

  async #getUsers(allSessions, globalScope, orgScope) {
    let users = [];

    if (
      this.can.can('list model', globalScope, { collection: 'users' }) &&
      this.can.can('list model', orgScope, { collection: 'users' })
    ) {
      const uniqueSessionUserIds = [
        ...new Set(
          allSessions
            .filter((session) => session.user_id)
            .map((session) => session.user_id),
        ),
      ];

      const filters = {
        id: {
          values: uniqueSessionUserIds.map((userId) => ({
            equals: userId,
          })),
        },
      };

      users = await this.store.query('user', {
        scope_id: 'global',
        recursive: true,
        query: { filters },
      });
    }

    return users;
  }

  async #getTargets(allSessions) {
    let targets = [];
    const scope = this.#scope;

    if (this.can.can('list model', scope, { collection: 'targets' })) {
      const uniqueSessionTargetIds = [
        ...new Set(
          allSessions
            .filter((session) => session.target_id)
            .map((session) => session.target_id),
        ),
      ];

      const filters = {
        id: {
          values: uniqueSessionTargetIds.map((targetId) => ({
            equals: targetId,
          })),
        },
      };

      targets = await this.store.query('target', {
        scope_id: this.scope.id,
        query: { filters },
      });
    }

    return targets;
  }
}
```

ui/admin/app/routes/scopes/scope/sessions/index.js
lisbet-alvarez: Thank for looking into this, and yup the modification u made does fix the case I was referring to.

ui/admin/app/routes/scopes/scope/sessions/index.js
```diff
@@ -117,104 +115,124 @@ export default class ScopesScopeSessionsIndexRoute extends Route {
...
+      //   then we know the previous `allSessions` is stale
+      const scopeContextSessionIds =
+        this.scopeContext?.allSessions?.map(({ id }) => id) ?? [];
+      const scopeContextHasAllQueriedSessions = sessions.every((session) =>
+        scopeContextSessionIds.includes(session.id),
+      );
+      const shouldRefreshScopeContext =
+        !this.scopeContext ||
+        this.scopeContext.scopeId !== scope.id ||
+        !scopeContextHasAllQueriedSessions;
```
ZedLi, line 131: I'm making sure I understand what the root of the issue is and it looks like it's because we're not properly refreshing `allSessions` bc when changing between scopes, we still end up using the same sessions route with the previously retrieved `allSessions` and `allSessions` doesn't refresh to get all the sessions for a possible new scope? 

If that's accurate, do we need to do a check with `scopeContextHasAllQueriedSessions`? Should we not just refresh any time the scope changes in the route?

hashicc: There's a few cases here, one is when the scope changes:
`this.scopeContext.scopeId !== scope.id`

This other case is when the scope id stays the same, but we know the cache is invalid. We can know it's invalid because we always get query `sessions` based on the current filters. If the `allSessions` doesn't have the same `sessions` that we've just queried it can be refreshed. That's also important because if `allSessions` was cached for the current scope with an empty array, and a new `sessions` is queried and results in a non-empty array, then [nothing will be shown](https://github.com/hashicorp/boundary-ui/pull/2850/files#diff-6b303cc385ad7795e93450c2e93d4f7ccfc45669323715409e4c6e91c2a66b44R23). Another related case to this check is where a session was added in `sessions` that `allSessions` doesn't have and that new session has new target or users filter options, then the `scopeContext` also needs to be refreshed.

ZedLi: I think with the same scope ID, that was somewhat the tradeoff we made since we knew the "cache" could be outdated and that was the acknowledged risk and we didn't add extra logic to figure out when to "invalidate the cache", but I didn't expect changing scopes wouldn't reset the route variables. This is why I was wondering if we should just explicitly clear it anytime we change scopes, but I was mainly asking if the fundamental issue was `allSessions` not being up to date for all of these scenarios. 

But to give more context, this was originally an issue that I was worried about when querying for `allTargets` originally in desktop client, which could be slow because if they had a lot of targets, we would process a lot of data for targets on any filter/search change (since we query for targets _twice_ on every input change). However, quite a few things have changed since then, namely the cache daemon has a hard limit on the amount of data that it returns now _and_ we realized we didn't actually need an `allTargets` and could figure out the right state without it so didn't need to query for it in the first place.

But that brings us back to `allSessions` which was still following the same pattern as `allTargets`. I don't think there was ever a worry about having too many sessions and how long that would take (since a user has to start up all these sessions, it'd be unlikely to be a huge number), especially on admin UI where you can't grab _all_ sessions from the global scope (a user has to drill down to an actual project). 

With all that said, I'm questioning if it makes sense to do all these extra checks and logic for `allSessions` still. I believe this would also be an issue in DC and likely we should just query every time there. We could also query every time in admin UI but the difference is there's a tradeoff as we have multiple dependent queries based off of `allSessions`. It looks like `associatedUsers` and `associatedTargets` actually only depend on `allSessions` so we could likely fire both off and not have them be sequential but it's still technically "two" extra requests we make every time.

Do you know how much longer loads would take if we did this rather than complicating things by trying to figure out when `allSessions` is out of date and needs to be refetched?

ui/admin/app/routes/scopes/scope/sessions/index.js
calcaide: I really like the class approach. Not saying we have to implement it, but it facilitates (a lot in my case) the full understanding of all the bit and pieces to coudl fetch the proper sessions and context 🙌

hashicc:
Based on this discussion: https://github.com/hashicorp/boundary-ui/pull/2850#discussion_r2116188870 and chatting with @ZedLi, along with some testing, around the realistic number of sessions that would be used it was decided to remove the caching altogether. https://github.com/hashicorp/boundary-ui/pull/2862 will replace this PR to fix this bug

PR #2854, opened 2025-05-23
feat: 🎸 add pagination to host source table

cameronperera:
✅ Closes: https://hashicorp.atlassian.net/browse/ICU-17093

# Description

This PR adds the ability to display a portion of the host sources (10 by default) and allow the user to click to the next page(s) if needed. Along with selecting different options for items per page.

**Note: This feature is not using `<Rose::Pagination />` because that component is tied to the API and utilized queryParams from the route. So I use the `<Hds::Pagination::Numbered />` component directly and the page and pageSize handling lives in the controller. This work will be removed and replaced by `<Rose::Pagination />` once the backend supports host source search and filtering.** 

https://github.com/user-attachments/assets/0540c44b-11fe-436e-b981-47f58e62b1c2

## How to Test

Spin up the desktop client and visit a target that has more than 2 host sources. To better test, find a target with more than 10 host sources (typically the first target in a vercel build) to test the page change and "items per page" change.

## Checklist

### DIFF
--- a/e2e-tests/desktop/tests/pagination.spec.js
+++ b/e2e-tests/desktop/tests/pagination.spec.js
@@ -1,0 +1,151 @@
+/**
+ * Copyright (c) HashiCorp, Inc.
+ * SPDX-License-Identifier: BUSL-1.1
+ */
+
+import { expect, test } from '../fixtures/baseTest.js';
+import * as boundaryHttp from '../../helpers/boundary-http.js';
+
+let hosts = [];
+let org;
+let targetWithManyHosts;
+
+test.beforeEach(async ({ request, targetAddress, targetPort }) => {
+  org = await boundaryHttp.createOrg(request);
+  const project = await boundaryHttp.createProject(request, org.id);
+
+  // Create host set
+  const hostCatalog = await boundaryHttp.createStaticHostCatalog(
+    request,
+    project.id,
+  );
+
+  // Create hosts
+  const hostCount = 15;
+  for (let i = 0; i < hostCount; i++) {
+    const host = await boundaryHttp.createHost(request, {
+      hostCatalogId: hostCatalog.id,
+      address: targetAddress,
+      name: `Host ${i + 1}`,
+    });
+    hosts.push(host);
+  }
+
+  const hostSet = await boundaryHttp.createHostSet(request, hostCatalog.id);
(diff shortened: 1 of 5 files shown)


ui/desktop/app/controllers/scopes/scope/projects/targets/target.js
```diff
@@ -20,11 +20,23 @@ export default class ScopesScopeProjectsTargetsTargetController extends Controll
...
+  @tracked pageSize = 10;
+  @tracked totalItems = this.model.hosts.length;
+  @tracked paginatedHosts = this.paginateHosts();
+
   @tracked isConnecting = false;
   isConnectionError = false;
 
   // =methods
 
+  paginateHosts() {
```
ZedLi, line 33: Could this have just been a getter?

ZedLi (commented):
Are we able to add an e2e test case for this as well? The test can then be used for searching once that's possible

ui/desktop/app/controllers/scopes/scope/projects/targets/target.js
cameronperera: It sure can. I'll update.

calcaide: asking for learning purposes: when we mean this being a getter, do we mean moving it to the target model? or something else?

e2e-tests/desktop/tests/pagination.spec.js
```diff
@@ -0,0 +1,151 @@
...
+      .selectOption('30');
+    await expect(
+      authedPage.getByRole('cell', { name: hosts[0].name, exact: true }),
+    ).toBeVisible();
+    await expect(
+      authedPage.getByRole('cell', {
+        name: hosts[hosts.length - 1].name,
+        exact: true,
+      }),
+    ).toBeVisible();
```
ZedLi, line 133: Couldn't we assert that there's 15 items now instead? And for the 10 items option that there's 10 rows? Rather than just looking at the first and last item

ZedLi (approved):
Looks great!

e2e-tests/desktop/tests/pagination.spec.js
cameronperera: Yeah that would be a better approach. I will update that now.

PR #2859, opened 2025-05-27
Update ember-electron to the latest version

calcaide:
# Description

Update ember-electron from `6.x` to `7.0.0` (latest)

Previously, during ember upgrade work, we updated `ember-electron` to `6.x` and not `7.0.0` due to the latest version not supporting ember < 4.8, now that we are on ember 5.12 we are compliant and the issue does not reproduce anymore.

More info on the context [within the initial `ember-electron` PR](https://github.com/hashicorp/boundary-ui/pull/2740)

## How to Test
- Build in local environment successfully the desktop client.
- Use boundary-desktop-releases to generate artifacts with this new version and test them.


ZedLi (approved):
Thanks for the update!

hashicc (approved):
🚀 ember and node version bumps were the only [breaking changes](https://github.com/adopted-ember-addons/ember-electron/releases/tag/v7.0.0) as you mentioned!
thanks for following up on this dependency!

DhariniJeeva (approved):
tested locally and everything looks good! thanks for the upgrade

hashicc (approved):
Sorry for breaking your PR with pnpm!

cameronperera (approved):

PR #2860, opened 2025-05-29
refactor: 💡 add custom sort to resources

lisbet-alvarez:
# Description

* Add custom sort to `type` for targets.
* Add custom sort to `state` for session recordings.
* Add custom sort to grants applied column for roles.
* Update sort-results util to sort by `id` when `name` is missing.
(spike for this can be found [here](https://github.com/hashicorp/boundary-ui/pull/2847))

 :tickets: [Jira ticket](https://hashicorp.atlassian.net/browse/ICU-17198)

Roles (grants applied column):
image

Targets (type column):
image

Session Recordings (status column):
image

## How to Test

Go to listed resources and validate sorting is functioning as intended.

## Checklist


addons/api/addon/utils/sort-results.js
```diff
@@ -26,6 +26,12 @@ function getSortableValue(schema, record, attribute) {
     return record.attributes.created_time;
   }
 
+  // When name is an optional attribute, use id to sort instead.
+  if (schema.attributes.has(attribute) && attribute === 'name') {
+    const name = record.attributes[attribute];
+    return name ? name : record.id;
+  }
+
```
hashicc, line 34: I am thinking about someone querying by `name` and then getting a result set that has additional sorting by `id` and being surprised. This does come up enough so it does seem like a path to be paved.

What if a sort function was provided by the addon that offered the same functionality.

```js
// maybe export from a different path that is more accessible than `utils/sort-results.js`
export const sortNameWithIdFallback = (a, b) {
  // 
}
```

```js
// some route
import { sortNameWithIdFallback } from "api/utils/sort-results";

//...
const sort = this.sortAttribute === 'name' ? { customSort: sortNameWithIdFallback, direction: sortDirection } : { attribute: sortAttribute, direction: sortDirection }
this.store.query('target', { query: { sort }});
```

I think the goal of pushing special cases into sort functions and generalizing reusable sort functions that can be shared might be a pattern that scales well

lisbet-alvarez: Thank u those are good points!
A reusable sort function seems like a better pattern. I appreciate the well outlined examples as well!

hashicc (changes requested):
💯 Incredible work, way to get this fully implemented!
It would be great to have a test in `indexed-db-handler-test.js` since it tests things fully and at one layer higher than `sort-results-test.js` and it'll catch that things wire up  closer to the public interface.

My first pass was focused on just the "base layer" for custom sorting. I'm going to look over the app code in `ui/admin` now. Again, really well done! Thanks for diving in

addons/api/tests/unit/utils/sort-results-test.js
```diff
@@ -421,4 +421,96 @@ module('Unit | Utility | sortResults', function (hooks) {
...
+      const schema = { attributes: new Map() };
+      schema.attributes.set('name', { type: 'string' });
+      const sortedResults = sortResults(shuffledResults, { querySort, schema });
+      assert.deepEqual(
+        sortedResults.map(({ attributes: { type } }) => type),
+        ['tcp', 'tcp', 'ssh', 'ssh'],
+      );
+    });
+
+    test('it sorts by ascending then descending `name` using string sorting', async function (assert) {
```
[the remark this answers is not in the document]
lisbet-alvarez: Good point, since setup is very similar.
Also another great point, Ill use a larger variety of mock types instead to make the test clearer. Thank u for the feedback!

ui/admin/app/routes/scopes/scope/targets/index.js
```diff
@@ -62,6 +64,16 @@ export default class ScopesScopeTargetsIndexRoute extends Route {
     return this.retrieveData.perform({ ...params, useDebounce });
   }
 
+  sortType = (recordA, recordB) => {
```
calcaide, line 67: **Asking for learning purposes**: In other domeains such as host-catalogs, we import the custom sorting function from a utils file, in example sortNameWithIdFallback.
But I see cases were rather than import the custom sorting function from a util file we define it within the route file itself, in example sortState or sortGrantsApplied. Why we have these 2 patterns?

I actually like the pattern of defining the custom sort function within a util file, so it keeps separation of concerns and facilitates unit testing, like sortNameWithIdFallback does.

lisbet-alvarez: Great question! We import `sortNameWithIdFallback` to avoid duplication since its a custom sorting function that can be reused for various resources (that have sort on `name` column & `name` is an optional field). For the other custom sort functions they are only used for that particular resource so it is defined within the route that handles sorting (since its only used in one place to avoid overhead of an additional import).

calcaide: Ahh that makes sense, I assumed the one in utils was to facilitate reuse, thank you for clarifying. 🥳

calcaide (approved):
As previously mentioned, great work Lis!! This is ready to ship!!

Thank you for adding host-catalog and credential-stores domains custom sorting 😉 . As discussed on the last sorting poddle, happy to add auth-methods optional name custom sorting after this PR gets merge! 🙌

hashicc (dismissed):
💯🏅🚀 
Left one small refactor non-blocking suggestion. Everything works great. It's really awesome to see how well the custom sort addition turned out.

In the future maybe we can have a helper that helps with the set up for the `sort` query object... and now I'll take off my over-optimization hat 

```js
const sort = createQuerySort(attribute, direction, {
  name: sortNameWithIdFallback, // imported sort function
  someAttribute: true // returns default { attribute, direction }
  state: (a, b) =>  /* inline custom sort function */
});

this.store.query('resource', { query: { sort } });
```

PR #2861, opened 2025-05-30
feat: 🎸 add search capability for host sources

cameronperera:
✅ Closes: https://hashicorp.atlassian.net/browse/ICU-17094

# Description

This PR adds the ability to search for a specific host source(s) to narrow down the results. 

https://github.com/user-attachments/assets/07d7ad16-8f9d-466f-959d-712e1ed2649c

## How to Test

Using a vercel or local instance of boundary, visit a target that has more than 1 host source and you should see the search bar and be able to search for a specific host source(s). The search will check for name, description, and address.

## Checklist


ui/desktop/app/styles/app.scss
```diff
@@ -244,6 +244,12 @@
   }
 }
 
+.target-details-body {
+  .hds-segmented-group {
```
ZedLi, line 248: Isn't this margin normally handled by the `.search-filtering` css class?

ZedLi (commented):
I also notice flickering when changing the page size, do you see that too?

ui/desktop/app/controllers/scopes/scope/projects/targets/target/index.js
```diff
@@ -0,0 +1,71 @@
...
+      (this.page - 1) * this.pageSize,
+      this.page * this.pageSize,
+    );
+  }
+
+  get hosts() {
+    let filteredHosts = this.model.hosts;
+    if (this.search) {
+      filteredHosts = this.model.hosts.filter((host) => {
+        const searchTerm = this.search?.toLowerCase() || '';
```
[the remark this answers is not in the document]
cameronperera: Fixed.

ui/desktop/app/styles/app.scss
ZedLi: Is there some template hierarchy thing happening here? I notice two `search-filtering` classes

image

cameronperera: Weird. Let me take a look.

ZedLi: Why are two being applied though? It seems like we have some weird parent/child templates as I notice two rose layouts and two rose headers. Maybe this is something we should investigate later but I'm not sure I would expect that

ZedLi (approved):
Looks good!

PR #2862, opened 2025-05-30
Remove cached resources on sessions route

hashicc:
# Description
This fixes the bugs in the cache mentioned in https://github.com/hashicorp/boundary-ui/pull/2850 by removing the cache. The cache was determined not to be worth the added complexity. The "cache" here is referred to the stored references to `allSessions`, `associatedUsers`, and `associatedTargets` that were saved on the route for the current scope.

## How to Test
1. When visiting a project that has no sessions, visiting a different project with sessions should show those sessions
2. When visiting a project that has users associated with its sessions, visiting a different project with different users associated with its session should show the updated users in the users filter
2. When visiting a project that has targets associated with its sessions, visiting a different project with different targets associated with its session should show the updated targets in the targets filter

## Checklist

### DIFF
--- a/ui/admin/app/routes/scopes/scope/sessions/index.js
+++ b/ui/admin/app/routes/scopes/scope/sessions/index.js
@@ -49,16 +49,12 @@ export default class ScopesScopeSessionsIndexRoute extends Route {
     },
   };
 
-  allSessions;
-  associatedUsers = [];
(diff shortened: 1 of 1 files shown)


lisbet-alvarez (approved):
Looks great! Tested locally and it fixes the issues!

PR #2872, opened 2025-06-05
Delete rose cards and rose card old styles

calcaide:
# Description

Delete rose cards and rose card old styles.

I found out these styles still within rose but I couldn't find them being use, so decide to delete them, please let me know if I miss something and this styles still need.

## Process to try find a usage of these styles:
- I look for usage of `Rose::Card` or `Rose::Cards` in all boundary-ui with 0 results.
- I look for usage or menions of `@use card` or `@use card`, in case other styles were reusing these, 0 results.
- I look for usage of `.rose-card` or other classes from this component with 0 results.
- I run Admin and Desktop UI's and perform a manual smoke test without seeing anything broken.


laurenolivia (approved):
💅 Thank you

priya-patel04 (approved):
Thanks for catching this and the clean up! 🙌

DhariniJeeva (approved):
thanks for the cleanup!

PR #2876, opened 2025-06-09
Check format in ci

hashicc:
# Description

## Checklist


hashicc:
Closed this because we actually have commit hooks that I forgot about. These commit hooks almost transparently take staged files and reformat them before committing. This is better than having it in CI since the feedback loop is much shorter. Even though files could skip the format commit hook and still make it into `main` this happens relatively rare to warrant this extra work in CI

PR #2882, opened 2025-06-11
refactor: 💡 add title to primary column in auth-methods table

lisbet-alvarez:
# Description

Small update to column header for 'Primary' in auth-methods table.

Screenshot 2025-06-10 at 5 22 33 PM

## Checklist

### DIFF
--- a/addons/core/translations/resources/en-us.yaml
+++ b/addons/core/translations/resources/en-us.yaml
@@ -63,6 +63,7 @@ auth-method:
     finding: Finding
     always: Always
     primary: Primary
+    primary-method: Primary method?
   types:
     password: Password
     oidc: OIDC
--- a/ui/admin/app/templates/scopes/scope/auth-methods/index.hbs
+++ b/ui/admin/app/templates/scopes/scope/auth-methods/index.hbs
@@ -113,9 +113,7 @@
                 @onClickSort={{fn H.setSortBy 'id'}}
                 @sortOrder={{if (eq 'id' H.sortBy) H.sortOrder}}
               >{{t 'form.id.label'}}</H.ThSort>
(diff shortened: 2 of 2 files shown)


addons/core/translations/resources/en-us.yaml
```diff
@@ -63,6 +63,7 @@ auth-method:
     finding: Finding
     always: Always
     primary: Primary
+    primary-method: Primary method?
```
calcaide, line 66: None blocking comment: Why change from Primary to Primary method? I do agree facilitates understanding for users, but I think it will have the side effect of doing the table cell wider? Not sure since I didn't try it but something to have in mind.

lisbet-alvarez: Oh this particular change was design feedback to add a visible table header. Before the header was visually hidden.

calcaide: @lisbet-alvarez 

On my end it's fine, I just mentioned to have it in mind. If design is ok with it, let's do it!
I quickly check HDS documentation on table to "limit" a bit the width of the cell, but tbh I think we should leave the table to use its defaults.

laurenolivia (commented):
Nice work. A couple non-blocking questions.

cameronperera (approved):

priya-patel04 (approved):
Nice work!

addons/core/translations/resources/en-us.yaml
lisbet-alvarez: I think for this case the columns all adjusted to be wider to fill up the page.
In regards to the wording, of the columns it is based off of provided figma designs.

PR #2885, opened 2025-06-12
chore: 🤖 upgrade got package to solve vulnerability

lisbet-alvarez:
# Description

Dependency tree:
`got@9.6.0` -> `package-json@6.5.0` -> `ember-try-config@4.0.0(encoding@0.1.13)` -> `ember-try@3.0.0(encoding@0.1.13)`

Attempting to upgrade ember-try did not cause `got` to update to a non-vulnerable version. `ember-try-config` is also already at newest released version and `package-json` only goes up to 6.5 for v6. Therefore, as a last resort we force `got` to resolve to newest, non-vulnerable version.

Resolves [alert #184](https://github.com/hashicorp/boundary-ui/security/dependabot/184)

## How to Test

[Desktop release test run](https://github.com/hashicorp/boundary-desktop-releases/actions/runs/15935266470)

Run `pnpm test:ember-compatibility` in all addons (api, auth, rose, core). 
`ember try` should run tests and print output detailing which scenarios passed or failed. 

## Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -47,7 +47,8 @@
       "css-select>nth-check": "^2.0.1",
       "node-gyp": "^10.0.0",
       "ansi-html": "^0.0.9",
-      "fireworm>async": "^2.6.4"
+      "fireworm>async": "^2.6.4",
+      "ember-cli-babel>@babel/runtime": "7.27.6"
     },
     "ignoredBuiltDependencies": [
       "@parcel/watcher",
--- a/package.json
+++ b/package.json
@@ -49,7 +49,7 @@
       "ansi-html": "^0.0.9",
       "fireworm>async": "^2.6.4",
       "ember-cli-babel>@babel/runtime": "7.27.6",
-      "package-json>got": "^14.4.7" 
+      "ember-try-config>package-json": "^10.0.1" 
     },
     "ignoredBuiltDependencies": [
       "@parcel/watcher",
--- a/package.json
+++ b/package.json
@@ -48,7 +48,8 @@
       "node-gyp": "^10.0.0",
       "ansi-html": "^0.0.9",
       "fireworm>async": "^2.6.4",
-      "ember-cli-babel>@babel/runtime": "7.27.6"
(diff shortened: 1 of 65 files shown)


calcaide (approved):
I did test build and run (smoke test) Admin and Desktop UI, and they work as expected.
Awesome work Lis!! Thank you for taking care of this vulnerability. 🙌 💪

package.json
```diff
@@ -48,7 +48,8 @@
       "node-gyp": "^10.0.0",
       "ansi-html": "^0.0.9",
       "fireworm>async": "^2.6.4",
-      "ember-cli-babel>@babel/runtime": "7.27.6"
+      "ember-cli-babel>@babel/runtime": "7.27.6",
+      "package-json>got": "^14.4.7"
```
ZedLi, line 52: Should we have just overriden [package-json](https://github.com/sindresorhus/package-json/releases/tag/v10.0.0) to 10+ instead since it looks like they stop using `got`?

ZedLi (commented):
How should I be confirming `ember-try` still works properly?

package.json
lisbet-alvarez: good point, upgrading package-json is cleaner solution.
also, i am attempting to run `ember try` in the addons that use it and i am getting an error (same error on `main`) im going to look into that.

lisbet-alvarez: description has been updated w/ how to test `ember try`

ZedLi: What's the expectation for `ember try`? I'll admit I never ran it before so it's likely it was already failing but I get failures for `EMBROIDER_TEST_SETUP_OPTIONS` with `safe` and `optimized`

ZedLi (approved):
Thanks for the upgrade! I had 2 failing scenarios for ember-try but I'm not sure if it's always been like that or just me

package.json
lisbet-alvarez: Yeah im not sure either. This was also the first time I ran ember try. I am planning to look more into the cases that failed (i should create a ticket for it).

PR #2891, opened 2025-06-18
[Desktop][Bug fix]: Stop target sessions before de-authenticating user session

laurenolivia:
# Description
This PR adds a prompt to the desktop client using [Hds::Modal](https://helios.hashicorp.design/components/modal) when a user attempts to sign out. The prompt requires a user to confirm their sign out attempt (or cancel), and signing out will stop all active and pending target sessions before de-authenticating a user session.

This PR also migrates our current usage of the [dialog.showMessageBoxSync](https://www.electronjs.org/docs/latest/api/dialog#dialogshowmessageboxsyncwindow-options) in the electron app to Hds::Modal.

Please see ticket details and comment section for more information about the bug that was reported.
:tickets: [Jira ticket](https://hashicorp.atlassian.net/browse/ICU-16773)

## Checklist


ui/desktop/electron-app/src/preload.js
```diff
@@ -61,3 +61,11 @@ process.once('loaded', () => {
     }
   });
 });
+
+contextBridge.exposeInMainWorld('electron', {
+  ipcRenderer: {
```
ZedLi, line 66: We shouldn't expose the bare ipcRenderer as that can be a [security concern](https://www.electronjs.org/docs/latest/tutorial/context-isolation#security-considerations).

ui/desktop/electron-app/src/ipc/handlers.js
```diff
@@ -116,6 +121,21 @@ handle('toggleFullscreenWindow', () => {
  */
 handle('closeWindow', () => app.quit());
 
+/**
+ * Toggle when signout is in progress
+ * to prevent app from quitting when signout modal is visible
+ */
+handle('setSignoutInProgress', (value) => (app.signoutInProgress = value));
```
ZedLi, line 128: I don't think you need this handler anymore now that you send an event from electron directly

ui/desktop/electron-app/src/index.js
```diff
@@ -284,18 +285,23 @@ app.on('ready', async () => {
...
-
-    const buttonId = dialog.showMessageBoxSync(null, dialogOpts);
-    buttonId === 0 ? sessionManager.stopAll() : event.preventDefault();
+    try {
+      await mainWindow.webContents.send('inform-ember-of-app-closure');
+      event.preventDefault();
+      return;
+    } catch (e) {
+      console.log(`Exception: ${e}`);
+    }
```
ZedLi, line 304: I don't think you need the `signoutInProgress` here anymore. All this function needs to do is to check if it has running sessions and if it does, send the signal to let the UI handle it and prevent the app from quitting since the UI will send the signal to quit now.

ui/desktop/app/controllers/application.js
```diff
@@ -75,4 +102,34 @@ export default class ApplicationController extends Controller {
         rootEl.classList.remove('rose-theme-light');
     }
   }
+
+  @action
+  signoutAttempt() {
+    this.showSignoutModal = true;
```
ZedLi, line 108: I don't see any logic in when this should be activated, it looks like this popup will always appear when trying to logout even if there are no active sessions. If we remove `setSignoutInProgress`, we shouldn't need this `signoutAttempt` action anymore and just do everything in `invalidateSession`

ui/desktop/app/controllers/application.js
```diff
@@ -15,14 +17,34 @@ export default class ApplicationController extends Controller {
...
+
+  @tracked showSignoutModal = false;
+  @tracked showCloseSessionsModal = false;
+
+  constructor() {
+    super(...arguments);
+
+    // Check if Electron and ipcRenderer are available
+    if (window.electron && window.electron.ipcRenderer) {
+      window.electron.ipcRenderer.on('inform-ember-of-app-closure', () => {
```
ZedLi, line 30: Can we call the handler something like `onAppQuit`?

ui/desktop/app/templates/application.hbs
```diff
@@ -155,4 +156,60 @@
       </M.Footer>
     </Hds::Modal>
   {{/if}}
-</PendingConfirmations>
\ No newline at end of file
+</PendingConfirmations>
+
+{{#if this.showSignoutModal}}
```
ZedLi, line 161: I don't think you actually need two separate modals, you can just use one modal and have the handlers handle the logic for what to do when confirming or canceling and use your two variables to handle what text to display and when the modal should be open.

ui/desktop/app/controllers/application.js
```diff
@@ -15,14 +17,34 @@ export default class ApplicationController extends Controller {
   @service clusterUrl;
   @service flashMessages;
 
+  // =tracked properties
+
+  @tracked showSignoutModal = false;
+  @tracked showCloseSessionsModal = false;
```
ZedLi, line 23: It might make more sense to call these something like `isLoggingOut` and `isAppQuitting` if we can use one modal

ui/desktop/app/controllers/application.js
```diff
@@ -75,4 +102,34 @@ export default class ApplicationController extends Controller {
...
+
+  @action
+  signoutAttempt() {
+    this.showSignoutModal = true;
+    // Prevents a user from quitting app while signout modal is present
+    this.ipc.invoke('setSignoutInProgress', true);
+  }
+
+  @action
+  cancelSignout() {
```
ZedLi, line 114: You could consolidate these cancel functions to just set both variables to false to use one modal

ui/desktop/app/controllers/application.js
```diff
@@ -75,4 +102,34 @@ export default class ApplicationController extends Controller {
...
+    this.ipc.invoke('setSignoutInProgress', false);
+  }
+
+  @action
+  closeAppAttempt() {
+    this.showCloseSessionsModal = true;
+  }
+
+  @action
+  confirmCloseSessions() {
```
ZedLi, line 125: You could just use a shared function here, you just need to check if the app is quitting from the variable your electron handler set to determine whether we should call `this.session.invalidate()` or to close the app.

ui/desktop/electron-app/src/ipc/handlers.js
```diff
@@ -116,6 +121,21 @@ handle('toggleFullscreenWindow', () => {
...
+
+/**
+ * Toggle when attempt to close app is in progress
+ * and prevent closing app until user has engaged with modal
+ */
+handle('closeSessionsAndQuit', async () => {
```
ZedLi, line 134: This should only need to just close the app since you already have a handler to stopAll

ui/desktop/app/controllers/scopes/scope/projects/settings/index.js
```diff
@@ -18,4 +18,12 @@ export default class ScopesScopeProjectsSettingsIndexController extends Controll
   toggleTheme({ target: { value: theme } }) {
     return this.application.toggleTheme(theme);
   }
+
+  /**
+   * Call Application controller's signoutAttempt method
+   */
+  @action
+  signoutAttempt() {
```
ZedLi, line 26: Mentioned elsewhere but I don't think this is necessary as you can keep the original `invalidateSession` function

ui/desktop/app/controllers/application.js
laurenolivia: Yep! sure can

ui/desktop/electron-app/src/preload.js
```diff
@@ -61,3 +61,11 @@ process.once('loaded', () => {
...
+    on(channel, callback) {
+      ipcRenderer.on(channel, callback);
```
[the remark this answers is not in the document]
laurenolivia: So not on `ipcRenderer.removeListener()` ?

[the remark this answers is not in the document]
ZedLi: You return a function that executes `ipcRenderer.removeListener()` which then gets called by the UI

ui/desktop/tests/acceptance/authentication-test.js
```diff
@@ -235,4 +250,105 @@ module('Acceptance | authentication', function (hooks) {
...
+
+    assert.dom(CLOSE_SESSIONS_MODAL_SELECTOR).isNotVisible();
+    assert.ok(stopAllSessions.calledOnce);
+    assert.notOk(currentSession().isAuthenticated);
+  });
+
+  test('attempting to quit app when signout modal is present triggers the close sessions modal', async function (assert) {
+    const quitApp = this.ipcStub.withArgs('closeWindow');
+    // onAppQuit is the channel that handles when a user triggers the before-quit event in the electron app
+    window.electron = {
```
ZedLi, line 314: We should ideally mock out things on `window` since I don't think this would get cleaned up between test runs.

ui/desktop/electron-app/src/ipc/handlers.js
```diff
@@ -116,6 +121,14 @@ handle('toggleFullscreenWindow', () => {
  */
 handle('closeWindow', () => app.quit());
 
+/**
+ * Check if session manager has running sessions
+ * Return boolean
+ */
+handle('hasRunningSessions', async () => {
+  return sessionManager.hasRunningSessions;
+});
```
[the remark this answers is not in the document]
laurenolivia: Oh ok, gotcha. Will remove.

ui/desktop/app/templates/application.hbs
```diff
@@ -155,4 +155,39 @@
       </M.Footer>
     </Hds::Modal>
   {{/if}}
-</PendingConfirmations>
\ No newline at end of file
+</PendingConfirmations>
+
+{{#if (or this.isLoggingOut this.isAppQuitting)}}
+  <Hds::Modal @onClose={{this.cancel}} data-test-close-sessions-modal as |M|>
```
hashicc, line 161: It seems like there's two separate cases that are being merged:
* `isAppQuitting`
* `isLoggingOut`

The `if this.isAppQuitting` / `else` checks in the template and `confirmCloseSessions` attempt to share the markup and the action. It might be easier to have these as separate modals with two separate actions and then the `or` on the top becomes two `if`s. This also improves the readability because the `else` in the `if this.isAppQuitting` / else` is the `isLoggingOut` case which becomes clearer when it's a separate `{{#if isLoggingOut}}` case

hashicc (changes requested):
This PR really encompasses a lot! Nice work wiring it all together from the electron app through to the desktop. I learned a few new electron apis in reviewing this too. I wanted to share the feedback I have so far, I still need to go through the tests

ui/desktop/tests/acceptance/authentication-test.js
```diff
@@ -235,4 +249,108 @@ module('Acceptance | authentication', function (hooks) {
...
   });
+
+  test('signing out with running sessions renders signout modal', async function (assert) {
+    this.ipcStub.withArgs('hasRunningSessions').returns(true);
+
+    await visit(urls.authenticate.methods.global);
+
+    await fillIn('[name="identification"]', 'test');
+    await fillIn('[name="password"]', 'test');
+    await click('[type="submit"]');
```
lisbet-alvarez, line 260: Instead of signing in manually can we use `authenticateSession` function instead or is there a particular reason for this?

ui/desktop/tests/acceptance/authentication-test.js
```diff
@@ -15,10 +22,17 @@ import {
...
 
 module('Acceptance | authentication', function (hooks) {
   setupApplicationTest(hooks);
   setupMirage(hooks);
+  setupStubs(hooks);
+
+  const SIGNOUT_BTN_SELECTOR = '[data-test-nav-signout-btn]';
+  const CLOSE_SESSIONS_MODAL_SELECTOR = '[data-test-close-sessions-modal]';
+  const CONFIRM_BUTTON = '.hds-modal__footer .hds-button--color-primary';
+  const CANCEL_BUTTON = '.hds-modal__footer .hds-button--color-secondary';
```
lisbet-alvarez, line 35: For naming conventions here can we stick to `BTN` instead of `BUTTON` like in line 32. And I think the new convention is to leave out the word `SELECTOR` similar to line 35.

ui/desktop/app/controllers/scopes/scope/projects/settings/index.js
```diff
@@ -18,4 +18,12 @@ export default class ScopesScopeProjectsSettingsIndexController extends Controll
...
     return this.application.toggleTheme(theme);
   }
+
+  /**
+   * Call Application controller's checkForSessionsRunning method
+   */
+  @action
+  checkForSessionsRunning() {
+    return this.application.checkForSessionsRunning();
+  }
```
lisbet-alvarez, line 28: It's not necessary to create a wrapper function to call the application function. `this.application.checkForSessionRunning` can be passed in the template directly for simplicity.

lisbet-alvarez (commented):
The modal is a very nice improvement! It's looking good, just left a couple of comments/suggestions.

ui/desktop/tests/acceptance/authentication-test.js
hashicc: I unresolved this because I think @ZedLi's comment is still relevant. Instead of mocking `window` directly, if the `window` service is used (as mentioned above) from this [addon](https://github.com/CrowdStrike/ember-browser-services), then it looks like it should work using the test helper `setupBrowserFakes`. The example [here](https://github.com/CrowdStrike/ember-browser-services?tab=readme-ov-file#window) (under _Alternatively, specific APIs of the window can be stubbed with an object_) should be a good one to follow

ui/desktop/electron-app/src/preload.js
```diff
@@ -61,3 +61,16 @@ process.once('loaded', () => {
     }
   });
 });
+
+/**
+ * Listener on electron app when user triggers before-quit event
+ */
+contextBridge.exposeInMainWorld('electron', {
+  onAppQuit: (callback) => {
+    ipcRenderer.on('onAppQuit', callback);
```
hashicc, line 70: This way of passing in the callback still gives the `event` argument to the `callback` function which the [electron security docs](https://www.electronjs.org/docs/latest/tutorial/security#how-17) mention is a security risk. This is similar to the "_// Also bad_" example in the code block in the docs, but can be refactored to the "// Good" example

The `terminal.receive` should also be refactored to not use this pattern:
```
ipcRenderer.on(incomingDataChannel, callback);
```

ui/desktop/tests/acceptance/authentication-test.js
```diff
@@ -235,4 +249,108 @@ module('Acceptance | authentication', function (hooks) {
...
+    await click(CONFIRM_BUTTON);
+
+    assert.dom(CLOSE_SESSIONS_MODAL_SELECTOR).isNotVisible();
+    assert.ok(stopAllSessions.calledOnce);
+    assert.notOk(currentSession().isAuthenticated);
+  });
+
+  test('attempting to quit app when signout modal is present triggers the close sessions modal', async function (assert) {
+    window.electron = {
+      onAppQuit: (e) => {
```
hashicc, line 313:
```suggestion
      onAppQuit: (callback) => {
```
I think the `e` usually signifies an `event` or `error`, I think renaming this argument to `callback` would be clearer

ui/desktop/tests/acceptance/authentication-test.js
```diff
@@ -235,4 +249,108 @@ module('Acceptance | authentication', function (hooks) {
     assert.dom('[data-test-no-auth-methods]').includesText('No Auth Methods');
     assert.dom('.change-origin').exists();
   });
+
+  test('signing out with running sessions renders signout modal', async function (assert) {
```
hashicc, line 253: These three acceptance tests are really easy to ready and cover the features well!

ui/desktop/tests/unit/controllers/application-test.js
```diff
@@ -34,12 +34,14 @@ module('Unit | Controller | application', function (hooks) {
     assert.ok(controller.minimize);
     assert.ok(controller.toggleFullScreen);
     assert.ok(controller.close);
+    assert.ok(controller.confirmCloseSessions);
+    assert.ok(controller.checkForSessionsRunning);
   });
 
   test('invalidateSession action de-authenticates a user', async function (assert) {
```
hashicc, line 41: I think this test can be deleted now. It was previously testing the controller as a controller unit test but now it's testing the `session` service which is an addon that has its own tests covering this

hashicc (changes requested):
The tests look pretty good!

The main bits to review are:
* Using the browser/window` service to make it easier to mock `setupWindowMock` on electron
* Changing `onAppQuit` callback so that it doesn't receive the event

ui/desktop/electron-app/src/preload.js
ZedLi: Good call out on that 💯, I wasn't even aware it passed along the event as part of the callback

ui/desktop/app/templates/application.hbs
ZedLi: I actually originally suggested combining the modal since my main worry was having both modals active at the same time (since a user can quit while logging out). I wanted to just make sure that only one is ever active but I can see how it's harder to read. I'd be fine either way but we just need to make sure the `if` statements only allow one at a time and for quitting to take precedence

laurenolivia: Thanks for the input. The original implementation included two modals. Just to be clear, are we all agreeing to revert this work back to two modals?

ui/desktop/tests/acceptance/authentication-test.js
laurenolivia: Totally. No, no particular reason. I noticed this is what we are doing in other tests, so I implemented the same convention. Will change it.

ui/desktop/app/templates/application.hbs
hashicc: Got it, I think we can leave it as it is. I think what would have really helped me was realizing the `else` case wasn't a generic fallback but actually the `isLoggingOut` case, so maybe we can just have a comment on the `else` block in the template and the action.

ui/desktop/app/services/session.js
```diff
@@ -13,6 +13,7 @@ export default class SessionService extends BaseSessionService {
   /**
    * Extend ember simple auth's handleAuthentication method
    * so we can hook in and add the user's token to the cache daemon
+   * Note: this.session.invalidate() comes from Ember Simple Auth BaseSessionService
```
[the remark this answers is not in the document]
laurenolivia: Yea great question. The `invalidate()`method comes from Ember Simple Auth BaseSessionService ([docs](https://github.com/mainmatter/ember-simple-auth/blob/master/README.md)). We get to access that since we extend the base service.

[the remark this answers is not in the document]
laurenolivia: @cameronperera Realizing my last reply was an incomplete thought:

I added that comment because I thought it would be more helpful to know where the method comes from since we explicitly use it, without really making it clear in the code that it's inherent to BaseSessionService. Like if someone were following the breadcrumbs back to this service, there's no mention.

Does that make more sense? Lmk if you don't think the comment belongs here.

e2e-tests/desktop/tests/scope.spec.js
```diff
@@ -54,7 +54,7 @@ test.describe('Scope tests', () => {
     await expect(headerNavLocator).toBeVisible();
     await expect(
       headerNavLocator.locator('.hds-dropdown-toggle-button'),
-    ).toHaveText('Global');
+    ).toHaveText('Global', { ignoreCase: true });
```
DhariniJeeva, line 57: just curious- was this addition meant to address a failing test? I noticed we didn't change any labels, so I was wondering why we had add the ignoreCase for this text.

DhariniJeeva (approved):
looks good!

PR #2901, opened 2025-06-23
Add api client for playwright e2e tests

hashicc:
# Description
This pull request adds a generated api-client for e2e tests. The `README.md` updates capture most of the context and intention around using the generated api-client. Everything in the `api-client` folder is generated and can mostly be skipped for the purposes of the review

## How to Test
* The two tests that have been refactored to use the `apiClient` fixture should continue to pass
* The added `api-client.spec.js` that tests specific functionality of the fixture should also pass

## Checklist


moduli (commented):
leaving a couple of comments for now.

e2e-tests/package.json
```diff
@@ -14,6 +14,8 @@
     "lint:js:fix": "eslint . --fix",
     "format": "concurrently \"npm:format:*\" --names \"format:\"",
     "format:js": "prettier --write './**/*.{js,mjs}'",
+    "generate:api-client": "pnpm clean:api-client && openapi-generator-cli generate -g typescript-fetch -o api-client -i https://raw.githubusercontent.com/hashicorp/boundary/refs/heads/main/internal/gen/controller.swagger.json",
+    "clean:api-client": "rm -rf api-client || true",
```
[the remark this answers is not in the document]
moduli: mmm maybe I'm mistaken, which happens often 😅

e2e-tests/package.json
```diff
@@ -14,6 +14,8 @@
     "lint:js:fix": "eslint . --fix",
     "format": "concurrently \"npm:format:*\" --names \"format:\"",
     "format:js": "prettier --write './**/*.{js,mjs}'",
+    "generate:api-client": "pnpm clean:api-client && openapi-generator-cli generate -g typescript-fetch -o api-client -i https://raw.githubusercontent.com/hashicorp/boundary/refs/heads/main/internal/gen/controller.swagger.json",
```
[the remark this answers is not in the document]
hashicc: Since it can be generated at any time time and against different branches or versions of `boundary` I'm not sure what would be considered outdated without trying to over complicate the checks. The openapi client files are checked into `boundary-ui` so what is committed should work with alongside the `e2e-tests` for the same commit.  But maybe there's something easier that I'm missing, did you have something in mind?

When a new api is needed then it'd have to be re-generated (which somewhat relates to your above comment, generating it for something other than `main` or wherever the new api lives). As for how people would know, I would add some documentation to the `README` so folks would hopefully learn that way. Also if anyone tries to edit one of the files to add the missing api hopefully they would see the notice at the top of the file:
```
 * NOTE: This class is auto generated by OpenAPI Generator (https://openapi-generator.tech).
 * https://openapi-generator.tech
 * Do not edit the class manually.
 ```

In terms of deprecations or breaking changes, all of these files would roll up into a call that would likely result in a `Bad Request` response and hopefully with the extra logging it'd be easier to determine what requests are failing. I'm not sure if this is an issue we've run into that often though and it's probably one that would occur with a hand rolled client or even on the ember side as there'd probably be changes needed there as well for any breaking changes

File: e2e-tests/helpers/boundary-api-client.ts [hunk 3]
```diff
@@ -0,0 +1,205 @@
...
+}>({
+  apiClient: async ({ request, controllerAddr }, use) => {
+    if (!controllerAddr) {
+      throw new Error('`controllerAddr` must be set in the base fixture');
+    }
+
+    const boundaryApi = new BoundaryApi(controllerAddr);
+    await use(boundaryApi);
+
+    for (const [resourceType, resources] of Object.entries(
```
ZedLi, line 161: Very slick, I love how this is handled for resource cleanup by directly integrating with playwright

e2e-tests/helpers/boundary-api-client.ts
```diff
@@ -0,0 +1,205 @@
...
+        }
+
+        return context.response;
+      },
+    };
+
+    const openapiConfiguration = new BoundaryApiClient.Configuration({
+      basePath: controllerAddr,
+      headers: { Authorization: `Bearer ${process.env.E2E_TOKEN}` },
+      middleware: [
```
ZedLi, line 65: Super cool, I didn't think/realize you could do middleware with these clients

e2e-tests/helpers/boundary-api-client.ts
```diff
@@ -0,0 +1,205 @@
...
+  async post(context: BoundaryApiClient.ResponseContext) {
+    if (context.response.status === 204 && context.response.body === null) {
+      context.response.json = async () => undefined;
+    }
+
+    return context.response;
+  },
+};
+
+const betterErrorHandlingMiddleware = {
```
ZedLi, line 119: I like how logging is handled in the context of the client middleware 🚀

ZedLi (commented):
Really like this so far, I think it's a huge step up in convenience and dev ex!!

e2e-tests/helpers/boundary-api-client.ts
```diff
@@ -0,0 +1,205 @@
...
+  createdResources: Record<string, { id: string }[]> = {};
+  skipCleanupResources: { id: string }[] = [];
+
+  constructor(
+    private readonly controllerAddr: string
+  ) {}
+
+  get clients() {
+    const { openapiConfiguration } = this;
+    const clients = {
```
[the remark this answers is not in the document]
hashicc: That's right. As I worked through this iteration it seemed like creating a wrapper wouldn't be as useful. The method names are a bit verbose but they do match the swagger openapi operation id, and everything else maps really well between the swagger openapi spec and the clients which I think creates a good mental model to know what's supported and should be expected. Also @johanbrandhorst's comments on the `attributes` come in on the types when hovered, the work to hoist the references could probably be done but is just another layer of indirection.

To still provide some of the benefits of having a wrapper I think helpers could be created (either locally to the test file, imported, or attached to the `BoundaryApiClient` class) that can be spread into the arguments of the clients. 

Roughly like the following example, the input could be typed directly to hint at which resource it's intended for, and this allows what's passed to the client to be clear (type safety on the client's arguments, `console.log()` to verify that the helper output looks correct, etc).

```js
function targetHelper(scopeId?: string, type?: 'ssh' | 'tcp'): ScopeServiceCreateScopeRequest {
  return {
    item: {
      scopeId: scopeId || 'global',
      type: type || 'tcp',
    }
  };
}

type NamedResourceInput = { item: { name: string } };
function randomResourceName(prefix: string): NamedResourceInput {
  return {
    item: {
      name: `${prefix}-${nanoid()}`
    }
  }
}

const target = await apiClient.clients.targets.targetServiceCreateTarget(
  deepMerge(targetHelper(), randomResourceName('target')),
);
```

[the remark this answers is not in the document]
ZedLi: Yeah that's fair, I was just wondering what your thoughts were on that but either way it sounds like even if we wanted an abstraction, it would be at a level above which I agree with, thanks for the explanation!

e2e-tests/package.json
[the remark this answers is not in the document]
hashicc: @ZedLi any thoughts on ensuring consistency of the generated files?

e2e-tests/package.json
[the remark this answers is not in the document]
hashicc: I pushed a commit that makes this easier to generate against different commits, using `OPENAPI_SWAGGER_URL_OR_FILE` this way it can be relative to a local path or a public url like `https://raw.githubusercontent.com/hashicorp/boundary/release/0.19.x/internal/gen/controller.swagger.json`. Otherwise it defaults to pulling the `controller.swagger.json` from `main` if the env var isn't specified

e2e-tests/admin/tests/pagination.spec.js
```diff
@@ -5,159 +5,166 @@
...
 
 test(
   'Search and Pagination (Targets)',
   { tag: ['@ce', '@ent', '@aws', '@docker'] },
-  async ({ page, request }) => {
-    let org;
-    try {
-      org = await boundaryHttp.createOrg(request);
-      const project = await boundaryHttp.createProject(request, org.id);
+  async ({ page, apiClient }) => {
```
hashicc, line 13: Example of a test that is refactored to using the `apiClient`

e2e-tests/admin/tests/pagination.spec.js
```diff
@@ -5,159 +5,166 @@
...
-      ).toBeHidden();
-      await expect(
-        page.getByRole('link', { name: targets[targets.length - 2].name }),
-      ).toBeVisible();
-      await expect(
-        page.getByRole('link', { name: targets[targets.length - 1].name }),
-      ).toBeVisible();
-    } finally {
-      if (org.id) {
-        org = await request.delete(`/v1/scopes/${org.id}`);
```
hashicc, line 159: Clean up on error isn't needed any more. The `apiClient` fixture should handle any clean up of created resources regardless of test outcome

e2e-tests/package.json
```diff
@@ -14,6 +14,8 @@
     "lint:js:fix": "eslint . --fix",
     "format": "concurrently \"npm:format:*\" --names \"format:\"",
     "format:js": "prettier --write './**/*.{js,mjs}'",
+    "generate:api-client": "pnpm clean:api-client && openapi-generator-cli generate -g typescript-fetch -o api-client --auth $(gh auth token) -i ${OPENAPI_SWAGGER_URL_OR_FILE-https://raw.githubusercontent.com/hashicorp/boundary/main/internal/gen/controller.swagger.json}",
```
hashicc, line 17: This can be ran to (re)generate the api-client

e2e-tests/helpers/boundary-api-client.ts
```diff
@@ -0,0 +1,358 @@
...
+
+// These constants represent assumptions around the boundary api
+const BoundaryApiStatusCodes = Object.freeze({
+  created: 200,
+  deleted: 204,
+  notFound: 404,
+});
+const BOUNDARY_API_CREATED_RESOURCE_HTTP_CONTENT_TYPE = 'application/json';
+
+// the openapi spec generated incorrectly specifies that DELETE responses will be a 200 with a json body,
```
ZedLi, line 19: Is this something that needs to get fixed upstream so we don't have to patch it?

e2e-tests/helpers/boundary-api-client.ts
```diff
@@ -0,0 +1,358 @@
...
+        );
+      }
+    }
+
+    return context.response;
+  },
+};
+
+function extractPluralizedResourceTypeFromUrl(url: string): string | undefined {
+  const match = new URL(url).pathname.match(/\/v1\/([a-z]+)(|\/.*)/);
```
ZedLi, line 68: What's the purpose of the or statement (`|`) in this regex? Or even the second capturing group at all

e2e-tests/helpers/boundary-api-client.ts
```diff
@@ -0,0 +1,358 @@
...
+    const { controllerAddr, createdResources: resources } = this;
+
+    const captureCreatedResourcesMiddleware = {
+      async post(context) {
+        const resourceWasCreated =
+          context.response.status === BoundaryApiStatusCodes.created &&
+          context.response.headers.get('content-type') ===
+            BOUNDARY_API_CREATED_RESOURCE_HTTP_CONTENT_TYPE;
+
+        // if the respronse doesn't represent a created resource, return early
```
ZedLi, line 194:
```suggestion
        // if the response doesn't represent a created resource, return early
```

e2e-tests/helpers/boundary-api-client.ts
```diff
@@ -0,0 +1,358 @@
...
+        const pluralizedResourceType = extractPluralizedResourceTypeFromUrl(
+          context.response.url,
+        );
+        // the response `json` can only be read once, and needs to be read later by the generated api client,
+        // instead we can clone the response and this allows us to read the `json` on the cloned response only
+        const json = await context.response.clone().json();
+
+        // it's assume that json response has an `id` in order for it to be a created resource
+        if (pluralizedResourceType && typeof json?.id === 'string') {
+          resources[pluralizedResourceType] ??= [];
```
ZedLi, line 209: TIL about the [nullish coalescing assignment](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing_assignment), neat

ZedLi (commented):
This is looking really good, just had a few smaller questions (also does this PR lag in github for anyone else?)

e2e-tests/helpers/boundary-api-client.ts
hashicc: I think this is a [change in the boundary api](https://developer.hashicorp.com/boundary/docs/api-clients/api#status-codes) that is not reflected/updated in the swagger open api docs:

> (Starting with Boundary 0.2.1, delete actions return 204 on success.)

@johanbrandhorst does that seem accurate? Any plans to update this?

e2e-tests/helpers/boundary-api-client.ts
hashicc: The `|` was allowing for "nothing" on the left _or_ "/" and anything else on the right, but removing the second capture group completely still matches the part I need. I'll remove it!

e2e-tests/helpers/boundary-api-client.ts
johanbrandhorst: Interesting, yeah, sounds like we need to update the spec for that! It should be possible to fix with some annotations in the right places. I'll create a ticket for it.

johanbrandhorst: https://hashicorp.atlassian.net/browse/ICU-17477

ZedLi (dismissed):
This looks great, I loved the middleware pattern you used for the API client, this is going to make things a lot easier! 🎉

ZedLi (approved):

calcaide (approved):
Run the test and passing as described in the how to test section.

Thank you for this awesome work!! this improves DX experience A LOT and facilitates e2e testing 🙌  (insert-chef-kiss-emoji)

DhariniJeeva (dismissed):
awesome work!! tested and works as expected 🎉

priya-patel04 (approved):
Thank you for the effort - API client is a great improvement to our e2e tests! 🙌

e2e-tests/helpers/boundary-api-client.ts
```diff
@@ -0,0 +1,361 @@
...
+      ),
+
+      /** {@link https://developer.hashicorp.com/boundary/docs/concepts/domain-model/groups Groups Documentation} */
+      Group: new BoundaryApiClient.GroupServiceApi(openapiConfiguration),
+
+      /** {@link https://developer.hashicorp.com/boundary/docs/concepts/domain-model/host-catalogs Host Catalogs Documentation} */
+      HostCatalog: new BoundaryApiClient.HostCatalogServiceApi(
+        openapiConfiguration,
+      ),
+
```
moduli, line 130: Is there a `Host` service missing here?

moduli (dismissed):

e2e-tests/admin/tests/api-client.spec.js
```diff
@@ -16,10 +19,24 @@ const scopeId = 'global';
...
+      let apiClientTestAfter;
+      const setapiClientTestAfter = (fn) => {
+        apiClientTestAfter = fn;
+      };
+      await use(setapiClientTestAfter);
+      await apiClientTestAfter?.();
+    },
+  },
+  { scope: 'test', auto: true },
+);
```
hashicc, line 34: This turned out to be a better way to do assertions post-test rather than having a test that depended on a previous test being ran

e2e-tests/helpers/boundary-api-client.ts
```diff
@@ -64,8 +63,10 @@ const betterErrorHandlingMiddleware = {
   },
 };
 
-function extractPluralizedResourceTypeFromUrl(url: string): string | undefined {
-  const match = new URL(url).pathname.match(/\/v1\/([a-z-]+)/);
+function extractPluralizedResourceTypeFromResourceCreationUrl(
+  url: string,
+): string | undefined {
+  const match = new URL(url).pathname.match(/^\/v1\/([a-z-]+)$/);
```
hashicc, line 69: One of the issues I ran into was getting a false positive with `sessions` from the sessions cancel url, because it checked the other boxes I was using to assume if a resource was created:
* response was a `200`
* response was json
* response contained an id

This regex is stricter expecting a `/v1/resource` string and nothing trailing

e2e-tests/helpers/boundary-api-client.ts
```diff
@@ -261,47 +272,40 @@ export const boundaryApiClientTest = base.extend<{
         clients.Account,
       ),
       aliases: clients.Alias.aliasServiceDeleteAlias.bind(clients.Alias),
-      auth_tokens: clients.AuthToken.authTokenServiceDeleteAuthToken.bind(
+      ['auth-tokens']: clients.AuthToken.authTokenServiceDeleteAuthToken.bind(
```
hashicc, line 275: With `storage-buckets` I actually caught my mistake that these need to be keyed off the same way they are in urls which is with a `-`, not a `_`

e2e-tests/helpers/boundary-api-client.ts
```diff
@@ -261,47 +272,40 @@ export const boundaryApiClientTest = base.extend<{
...
       roles: clients.Role.roleServiceDeleteRole.bind(clients.Role),
       scopes: clients.Scope.scopeServiceDeleteScope.bind(clients.Scope),
-      session_recordings:
-        clients.SessionRecording.sessionRecordingServiceDeleteSessionRecording.bind(
-          clients.SessionRecording,
-        ),
-      // this is not a delete method on the resource, but cancelling a session is the equivalent cleanup action
-      sessions: clients.Session.sessionServiceCancelSession.bind(
-        clients.Session,
-      ),
```
hashicc: In writing the more exhaustive test to check that storage buckets I caught that these resources can't be created by the api so they can never be automatically cleaned up, so they shouldn't have clean up method entries

e2e-tests/helpers/boundary-api-client.ts
```diff
@@ -333,33 +338,45 @@ export const boundaryApiClientTest = base.extend<{
...
+              let responseDetails;
+              if (error.response) {
+                const clonedResponse = error.response.clone();
+                const { status, statusText, url } = clonedResponse;
+                responseDetails = {
+                  status,
+                  statusText,
+                  url,
+                  body: await clonedResponse.text(),
+                };
```
hashicc, line 368: This was helpful in troubleshooting in the case of a storage bucket not getting cleaned up, the error was related to a foreign key constraint and that wasn't being logged until this was added (mostly because the response body needs to be extracted for that information

hashicc (commented):
@moduli thanks for the tip on the storage bucket. I wrote a test that covers that case and it actually forced me to revisit a few assumptions, catch some mistakes I made in the implementation, and just refactor a few things that make more sense. Whenever you get a chance I'd love another review!

ZedLi (dismissed):
Looks great, thanks for the updated changes and tests! Just had a few minor comments but otherwise this has been great to see become a reality from just an idea 🚀

ZedLi (approved):
Thanks for all the due diligence in testing and getting the clients working!

PR #2905, opened 2025-06-24
PoC no unused services within addons

calcaide:
# Description

This PR is a quick PoC of an old ticket that we discussed within tech time ( March 21 2024). As a result I opened [this ticket](https://hashicorp.atlassian.net/browse/ICU-14874). 

The coverage of this initial PR is addons only, if this is welcome by the team 2 more PR's will follow up, one for Admin other for Desktop.

**For the record, where this will be more useful, simply because there are way more rule matches, is within Admin and Desktop.**

I would like to disscuss: Adding ember/no-unused-services as a rule:
This rule isn't included on the recommended config from eslint because it has limitations, [more info here.](https://github.com/ember-cli/eslint-plugin-ember/blob/master/docs/rules/no-unused-services.md) 
In a nutshell, can give false positives, i.e. if a service is included within a route file and not use it within the route file but within the template that renders such route, the linter will mark it as a warning/error but it's not. Because for the template to access, the service should be included.
In that case, the rule becomes "noise" rather than help.

There is a path to avoid the noise, which is disable the eslint rule within the code with a comment `// eslint-disable-next-line ember/no-unused-services`. Not ideal, but a way of not creating noise. [Example here](https://github.com/hashicorp/boundary-ui/compare/icu-14874-no-unused-services-addons?expand=1#diff-c65b190678bd171685ef9d33c57d394b39c1bd65b77547a67b6944e87f636f5eR17).

API with rule as warning BEFORE fixing it:
api-before-fix

API with rule as warning AFTER the fix:
api-after-fix

Core with rule as warning BEFORE fixing it:
core-before-fix

Core with rule as warning After fixing it:
core-after-fix

### DIFF
--- a/addons/core/addon/components/pending-confirmations/index.js
+++ b/addons/core/addon/components/pending-confirmations/index.js
@@ -14,7 +14,8 @@ import { action } from '@ember/object';
  */
 export default class PendingConfirmationsComponent extends Component {
   // =services
-
+  // confirm service use within template
+  // eslint-disable-next-line ember/no-unused-services
   @service confirm;
 
   // =actions
--- a/addons/api/.eslintrc.js
+++ b/addons/api/.eslintrc.js
@@ -31,6 +31,7 @@ module.exports = {
     'ember/no-computed-properties-in-native-classes': 'off',
     'ember/no-assignment-of-untracked-properties-used-in-tracking-contexts':
       'off',
+    'ember/no-unused-services': 1,
   },
   overrides: [
     // node files
--- a/addons/api/addon/models/session-recording.js
+++ b/addons/api/addon/models/session-recording.js
@@ -5,7 +5,6 @@
 
 import GeneratedSessionRecordingModel from '../generated/models/session-recording';
 import { hasMany } from '@ember-data/model';
-import { service } from '@ember/service';
 
 export const TYPE_SESSION_RECORDING_SSH = 'ssh';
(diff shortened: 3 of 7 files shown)


addons/core/addon/components/pending-confirmations/index.js
```diff
@@ -14,7 +14,8 @@ import { action } from '@ember/object';
  */
 export default class PendingConfirmationsComponent extends Component {
   // =services
-
+  // confirm service use within template
+  // eslint-disable-next-line ember/no-unused-services
```
ZedLi, line 18: Judging by the rule's [limitations](https://github.com/ember-cli/eslint-plugin-ember/blob/master/docs/rules/no-unused-services.md), how frequent do we have unused services vs using the service only in the template or other scenarios? I'm wondering if we'll have more exceptions than valid catches in the end

laurenolivia (commented):
Thanks for sharing the doc references for context. Questions I'd probably consider are:

1. Does the benefit of reducing noise in the terminal outweigh the risk of potential false positives (plus the other limitations), if we implement this rule?
2. Is there another way we can achieve reducing terminal noise, without utilizing this linter?

calcaide:
@ZedLi 

Most of the warning this linting tool provides are false positives due to its limitations. In example, lots of reports we are importing services that we don't use on the controllers or routes but we need them on the templates.
So it generates more noise than solutions, Closing this PR and the ticket.

PR #2918, opened 2025-07-02
fix: 🐛 transition aborted error when deleting or saving

lisbet-alvarez:
# Description

When awaiting a transition to a resource list view that implements search/filtering the original transition gets aborted since there is some sort of reload going on.
For example when saving a resource that does not have `read` permissions a user is redirected to resource list view by default and in some resources the transition was erroring out via alert toast.
Additionally when deleting a user is also redirected to list view and in some resources this transition was also erroring out.

Example of buggy behavior w/ User resource.
Before:

https://github.com/user-attachments/assets/944fbe04-eb64-41c9-a2cf-419c04c15837

After:

https://github.com/user-attachments/assets/03570fed-6910-45aa-bb92-9231d7763ead

## How to Test

To test delete action for aliases, and workers:
1. Navigate to existing resource.
2. Click on manage dropdown and click "Delete"
3. Delete action should result in success alert msg.

To test save action for workers, users, targets, roles, host-catalogs, groups, cred-stores, auth-methods, & aliases
1. Login to admin ui using an account that only allows create or list actions on the specified resource (no read action).
2. Navigate to list view of the resource & click on "New".
3. Fill in information and click "Save"
4. User should be redirected to list view & should see green success alert msg.

## Checklist

## PCI review checklist

### DIFF
--- a/ui/admin/app/controllers/scopes/scope/aliases/index.js
+++ b/ui/admin/app/controllers/scopes/scope/aliases/index.js
@@ -87,7 +87,7 @@ export default class ScopesScopeAliasesIndexController extends Controller {
     if (this.can.can('read model', alias)) {
       await this.router.transitionTo('scopes.scope.aliases.alias', alias);
     } else {
-      await this.router.transitionTo('scopes.scope.aliases');
+      this.router.transitionTo('scopes.scope.aliases');
     }
     await this.router.refresh();
   }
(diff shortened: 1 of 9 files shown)


ZedLi (approved):
Thanks for the fix! It definitely resolves the issue but I've always wondered if we're covering some other issue up, at some point we should look into a best practice on when/if we should await these since they are technically [thennable](https://api.emberjs.com/ember/release/classes/transition/) and it's easy to make this mistake.

PR #2919, opened 2025-07-03
chore: 🤖 Allow DC E2E tests to run in ent

ZedLi:
# Description
Change the run condition for DC E2E tests to allow them to run in ent


DhariniJeeva (approved):
thanks for enabling the test!

cameronperera (approved):

PR #2930, opened 2025-07-15
fix: 🐛 OIDC authentications not properly showing username

lisbet-alvarez:
✅ Closes: https://hashicorp.atlassian.net/browse/ICU-17231

# Description

## Checklist

## PCI review checklist

### DIFF
--- a/addons/auth/addon/authenticators/base.js
+++ b/addons/auth/addon/authenticators/base.js
@@ -6,6 +6,7 @@
 import SimpleAuthBaseAuthenticator from 'ember-simple-auth/authenticators/base';
 import { resolve, reject } from 'rsvp';
 import { waitForPromise } from '@ember/test-waiters';
+import { service } from '@ember/service';
 
 /**
  * Encapsulates common authenticator functionality.
@@ -19,6 +20,10 @@ import { waitForPromise } from '@ember/test-waiters';
  * All other responses should resolve the session restoration successfully.
  */
 export default class BaseAuthenticator extends SimpleAuthBaseAuthenticator {
+  // =services
+
+  @service store;
+
   // =unimplemented methods
 
   /**
@@ -101,7 +106,7 @@ export default class BaseAuthenticator extends SimpleAuthBaseAuthenticator {
    * @param {string} username
    * @return {object}
    */
-  normalizeData(data, username) {
+  async normalizeData(data) {
     // Pull fields up from `data.attributes` for easier access in JavaScript.
     // The `attributes` field exists on the Go side for its convenience but is
     // unnecessary here.
@@ -109,7 +114,33 @@ export default class BaseAuthenticator extends SimpleAuthBaseAuthenticator {
     // Add booleans indicated the scope type
     data.isGlobal = data?.scope?.type === 'global';
     data.isOrg = data?.scope?.type === 'org';
-    if (username) data.username = username;
+
+    try {
+      const adapter = this.store.adapterFor('application');
+      const findAccountRecordURL = adapter.buildURL(
+        'account',
+        data?.account_id,
+        {},
+        'findRecord',
+      );
+      const response = await waitForPromise(
+        fetch(findAccountRecordURL, {
+          method: 'get',
+          headers: { Authorization: `Bearer ${data.attributes.token}` },
+        }),
+      );
+
+      const authenticatedAccount = await response.json();
+
+      if (response.status === 200) {
+        const { email, full_name, login_name, subject } =
+          authenticatedAccount.attributes;
+        data.username = email || full_name || login_name || subject;
+      }
+    } catch (_) {
+      /* no op */
+    }
+
     return data;
   }
 
--- a/ui/admin/app/templates/application.hbs
+++ b/ui/admin/app/templates/application.hbs
@@ -63,13 +63,13 @@
               data-test-side-nav-user-dropdown
               as |dd|
             >
+              <dd.ToggleIcon @icon='user' @text={{t 'titles.user-menu'}} />
               {{#if this.session.data.authenticated.username}}
-                <dd.ToggleButton
-                  @icon='user'
+                <dd.Title @text={{t 'titles.authenticated'}} />
+                <dd.Description
                   @text={{this.session.data.authenticated.username}}
                 />
-              {{else}}
-                <dd.ToggleIcon @icon='user' @text={{t 'titles.user-menu'}} />
+                <dd.Separator />
               {{/if}}
 
               <dd.Interactive @route='account.change-password'>{{t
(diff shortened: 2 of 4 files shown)


addons/auth/addon/authenticators/base.js
```diff
@@ -101,15 +106,41 @@ export default class BaseAuthenticator extends SimpleAuthBaseAuthenticator {
...
+    try {
+      const adapter = this.store.adapterFor('application');
+      const findAccountRecordURL = adapter.buildURL(
+        'account',
+        data?.account_id,
+        {},
+        'findRecord',
+      );
+      const response = await waitForPromise(
+        fetch(findAccountRecordURL, {
```
ZedLi, line 127: Does it make sense to make the call here in the `authenticator`? Should we instead do this in the actual application since it seems like the authenticator wouldn't care about fetching the account

ui/admin/app/templates/application.hbs
```diff
@@ -70,42 +70,76 @@
...
+                  <dd.Radio
+                    @value={{theme.value}}
+                    checked={{eq this.session.data.theme theme.value}}
+                    {{on 'change' (fn this.toggleTheme theme.value)}}
+                  >
+                    {{t (concat 'themes.' theme.label)}}
+                  </dd.Radio>
+                {{/each}}
+              </Hds::Dropdown>
+            {{else}}
```
lisbet-alvarez, line 113: This second drop down is rendered with a `ToggleIcon` instead of a `ToggleButton`. We can't do the `this.session.data.authenticated.username` check inside the dropdown since on a re-render the event listeners seem to not get properly re-attached.

ui/admin/app/routes/application.js
```diff
@@ -75,6 +76,23 @@ export default class ApplicationRoute extends Route {
...
         await this.sqlite.setup(formatDbName(userId, hostUrl));
       }
+
+      if (this.session.data?.authenticated?.account_id) {
+        try {
+          const account = await this.store.findRecord(
+            'account',
+            this.session.data.authenticated.account_id,
+          );
+          const username =
```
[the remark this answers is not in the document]
ZedLi: It looks like the account model returns the same values just in a different order, though shouldn't we use the same order either way?
https://github.com/hashicorp/boundary-ui/blob/2db63ae2f84423aa08eb468204081ac2ec1db9e2/addons/api/addon/models/account.js#L51

Are you referring to the one on user?
https://github.com/hashicorp/boundary-ui/blob/9b1f052a569fc1579befffda87de3a91ba5e417b/addons/api/addon/models/user.js#L51

ui/admin/app/templates/application.hbs
```diff
@@ -70,42 +70,76 @@
...
+                data-test-side-nav-user-dropdown
+                as |dd|
+              >
                 <dd.ToggleButton
                   @icon='user'
                   @text={{this.session.data.authenticated.username}}
+                  {{hds-tooltip this.session.data.authenticated.username}}
+                  class='user-menu-dropdown__toggle-button'
+                />
+                <dd.Title @text={{t 'titles.authenticated'}} />
```
ZedLi, line 86: Is it redundant to have text that just says `Signed in`? Is that not obvious?

ui/admin/app/services/session.js
```diff
@@ -22,6 +24,23 @@ export default class SessionService extends BaseSessionService {
       await this.sqlite.setup(formatDbName(userId, hostUrl));
     }
 
+    if (this.data?.authenticated?.account_id) {
```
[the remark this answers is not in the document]
ZedLi: FWIW it seems ember simple auth suggests a [current user service](https://github.com/mainmatter/ember-simple-auth/blob/master/guides/managing-current-user.md) but I do agree I don't think we need to set another field back in the authenticated data.

ui/admin/app/templates/application.hbs
```diff
@@ -70,42 +70,76 @@
...
               </Hds::Dropdown>
             {{/if}}
-
-            <Hds::Dropdown
-              @enableCollisionDetection={{true}}
-              data-test-side-nav-user-dropdown
-              as |dd|
-            >
-              {{#if this.session.data.authenticated.username}}
+            {{#if this.session.data.authenticated.username}}
```
ZedLi, line 73: In what scenarios would this not be true?

ui/admin/app/routes/application.js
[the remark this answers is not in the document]
lisbet-alvarez: Yup u are correct. I was referring to the one in the account model. Yeah ill update to use that getter instead.

ui/admin/app/templates/application.hbs
lisbet-alvarez: after login when the application route renders the side nav isAuthenticated is true but the value of username has not been set yet.

ui/admin/app/templates/application.hbs
lisbet-alvarez: good point!

ui/admin/tests/helpers/index.js
```diff
@@ -18,19 +19,19 @@ function setupApplicationTest(hooks, options) {
...
-  // application test, you could do:
-  //
+
+  setupMirage(hooks);
+
   hooks.beforeEach(async function () {
-    await authenticateSession();
+    this.server.create('scope', { id: 'global' }, 'withGlobalAuth');
+    const account_id = this.server.schema.accounts.first().id;
+    await authenticateSession({ account_id });
```
lisbet-alvarez, line 28: Thoughts on refactoring our test cases to use this helper in order to provide an account_id  to the `authenticateSession`? I only pushed up changes for `tests/acceptance/accounts` as a POC.

cameronperera (dismissed):
Nice work Lis!

ZedLi (dismissed):
Looks good, I assume the test changes you've already made are now optional (which is fine) because having a username can fail?

ui/admin/app/templates/application.hbs
```diff
@@ -71,41 +71,73 @@
               </Hds::Dropdown>
             {{/if}}
 
-            <Hds::Dropdown
-              @enableCollisionDetection={{true}}
-              data-test-side-nav-user-dropdown
-              as |dd|
-            >
-              {{#if this.session.data.authenticated.username}}
+            {{#if this.session.username}}
```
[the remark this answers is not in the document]
lisbet-alvarez: Great question! This second drop down is rendered with a ToggleIcon instead of a ToggleButton. We can't do the `this.session.username` check inside the dropdown since on a re-render the event listeners seem to not get properly re-attached causing the dropdown to not expand (unless page is fully refreshed).

DhariniJeeva (dismissed):
thanks for the fix!

priya-patel04 (dismissed):
Great work! Thanks for the fix

PR #2932, opened 2025-07-15
feat: 🎸 Add a shared service between web workers

ZedLi:
✅ Closes: https://hashicorp.atlassian.net/browse/ICU-17409

# Description
This PR is the foundation to allowing multiple tab access to SQLite. We add a `shared-service` (not to be confused with an ember service) that is shared among all web workers. This allows multiple tabs to interact with SQLite by only allowing one tab to be active as the "provider" at a time. All other tabs funnel requests to the provider tab and do so by using  the shared proxy where the actual SQLite methods will live. 

It might help to test using PR #2933 to be able to interact with targets. 

A simplified example sequence diagram for a user trying to fetch a resource from a tab that doesn't have direct access to SQLite, note that the adding of event listeners is not indicated here. 

Web Worker Shared Service

## How to Test
Confirm that opening another tab doesn't crash/cause an error.

## Checklist

## PCI review checklist

### DIFF
--- a/addons/api/addon/workers/shared-service.js
+++ b/addons/api/addon/workers/shared-service.js
@@ -1,0 +1,242 @@
+/**
+ * Copyright (c) HashiCorp, Inc.
+ * SPDX-License-Identifier: BUSL-1.1
+ */
+
+import { v4 as uuidv4 } from 'uuid';
+
+let clientChannel = null;
+const requestsInFlight = new Map();
+
+export default (serviceName, target, onProviderChange) => {
+  const commonChannel = new BroadcastChannel(
+    `shared-service-common-channel-${serviceName}`,
+  );
+
+  let readyResolve = null;
+  const status = {
+    // Set a promise that doesn't resolve until our setup is finished
+    ready: new Promise((resolve) => {
+      readyResolve = resolve;
+    }),
+    // Check if there's an existing provider
+    isServiceProvider: (async () => {
+      const locks = await navigator.locks.query();
+      const isProvider = !locks.held?.find((lock) => lock.name === serviceName);
+      if (!isProvider) {
+        await onNotProvider();
+      }
+      return isProvider;
+    })(),
+  };
+
+  // This lock is used to ensure that only one tab can become the provider with an indefinite lock.
+  navigator.locks.request(serviceName, { mode: 'exclusive' }, async () => {
+    await onBecomeProvider();
+    await new Promise(() => {});
+  });
+
+  const onNotProvider = async () => {
+    const clientId = getClientId();
+    clientChannel = new BroadcastChannel(
+      getClientBroadcastChannelName(clientId, serviceName),
+    );
+
+    const register = async () =>
+      new Promise((resolve) => {
+        const onRegisteredListener = ({ data }) => {
+          if (data.clientId === clientId && data.type === 'registered') {
+            commonChannel.removeEventListener('message', onRegisteredListener);
+            resolve();
+          }
+        };
+
+        // Add an event listener to check for the registration response
+        commonChannel.addEventListener('message', onRegisteredListener);
+        commonChannel.postMessage({ type: 'register', clientId });
+      });
+
+    commonChannel.addEventListener('message', async ({ data }) => {
+      // Check if this was a provider change. If so and we're not the new provider, re-register.
+      if (data.type === 'providerChange' && !(await status.isServiceProvider)) {
+        await onProviderChange?.(await status.isServiceProvider);
+        await register();
+
+        // If there were requests in flight when the provider was changed,
+        // we need to resend them to the new provider.
+        if (requestsInFlight.size > 0) {
+          requestsInFlight.forEach(
+            ({ method, args, resolve, reject }, uuid) => {
+              requestFromProvider(uuid, resolve, reject, method, args);
+            },
+          );
+        }
+      }
+    });
+
+    await register();
+
+    readyResolve?.();
+    readyResolve = null;
+  };
+
+  const onBecomeProvider = async () => {
+    status.isServiceProvider = Promise.resolve(true);
+
+    if (readyResolve === null) {
+      status.ready = new Promise((resolve) => {
+        readyResolve = resolve;
+      });
+    }
+
+    // Set an event listener for other tabs to use to register with the provider tab
+    commonChannel.addEventListener('message', async (event) => {
+      const { clientId, type } = event.data;
+      if (type !== 'register') {
+        return;
+      }
+
+      const clientChannel = new BroadcastChannel(
+        getClientBroadcastChannelName(clientId, serviceName),
+      );
+
+      // Acquire another lock that immediately is waiting for the
+      // initial lock on the client tab to be resolved to handle cleanup.
+      navigator.locks.request(clientId, { mode: 'exclusive' }, async () => {
+        // The client tab was closed, clean up.
+        clientChannel.close();
+      });
+
+      // Setup an event listener for communication between the provider and subscriber to respond to requests
+      clientChannel.addEventListener('message', async ({ data }) => {
+        if (data.type === 'response') {
+          return;
+        }
+        const { method, args, id } = data;
+
+        let result, error;
+        try {
+          result = await target[method](...args);
+        } catch (e) {
+          // Error objects won't be properly serialized, we need to reconstruct an error like object
+          error =
+            e instanceof Error
+              ? Object.fromEntries(
+                  Object.getOwnPropertyNames(e).map((k) => [k, e[k]]),
+                )
+              : e;
+        }
+
+        clientChannel.postMessage({
+          id,
+          type: 'response',
+          result,
+          error,
+          method,
+        });
+      });
+
+      // Send back a message to the client that we finished registering
+      commonChannel.postMessage({ type: 'registered', clientId, serviceName });
+    });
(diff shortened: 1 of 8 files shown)


addons/api/addon/workers/shared-service.js
```diff
@@ -0,0 +1,241 @@
...
+  const onBecomeProvider = async () => {
+    status.isServiceProvider = Promise.resolve(true);
+
+    if (readyResolve === null) {
+      status.ready = new Promise((resolve) => {
+        readyResolve = resolve;
+      });
+    }
+
+    // Set an event listener for other tabs to register against with the provider tab
```
laurenolivia, line 91:
```suggestion
    // Set an event listener for other tabs to register with the provider tab
```

addons/api/addon/workers/shared-service.js
```diff
@@ -0,0 +1,241 @@
...
+      }
+
+      const clientChannel = new BroadcastChannel(
+        getClientBroadcastChannelName(clientId, serviceName),
+      );
+
+      // Acquire another lock that immediately is waiting for the
+      // initial lock on the client tab to be resolved to handle cleanup.
+      navigator.locks.request(clientId, { mode: 'exclusive' }, async () => {
+        // The client tab was closed, clean up.
```
laurenolivia, line 105: We're defining a client tab with an active sqlite connection as "resolved" to mean the tab was closed, but not when the user switches to another tab?

ZedLi:
> Since the OPFS API is only available in certain browsers, how are we communicating this to our users? For example, it's not supported with Brave browser ([ref](https://developer.mozilla.org/en-US/docs/Web/API/File_System_API/Origin_private_file_system#browser_compatibility)).

Are you encountering any issues? I don't see that it's not explicitly supported in that link, and it looks like brave is another chromium based browser so I would expect it would be OK.

We only support the major browsers (Chrome, Firefox, Safari) either way, everything else is best effort

hashicc (commented):
This has been fun to review. A few thoughts so far on the first half, nothing blocking. I'm going to look over the proxy part next

addons/api/addon/workers/shared-service.js
```diff
@@ -0,0 +1,242 @@
...
+
+    commonChannel.addEventListener('message', async ({ data }) => {
+      // Check if this was a provider change. If so and we're not the new provider, re-register.
+      if (data.type === 'providerChange' && !(await status.isServiceProvider)) {
+        await onProviderChange?.(await status.isServiceProvider);
+        await register();
+
+        // If there were requests in flight when the provider was changed,
+        // we need to resend them to the new provider.
+        if (requestsInFlight.size > 0) {
```
DhariniJeeva, line 67: nice resend logic here! 

do we need to think about debouncing or adding any guard here in case of multiple provider changes? Might not be an issue right now, but I was just wondering if this is something to consider.

ZedLi: I'm not sure if we need to worry about that, it's already likely uncommon for requests to be in flight that will need to be re-tried, and even if there are multiple provider changes happening around the same time, what ends up happening is just some wasted requests as they don't finish.

addons/api/addon/workers/shared-service.js
```diff
@@ -0,0 +1,242 @@
...
+      if (!isProvider) {
+        await onNotProvider();
+      }
+      return isProvider;
+    })(),
+  };
+
+  // This lock is used to ensure that only one tab can become the provider with an indefinite lock.
+  navigator.locks.request(serviceName, { mode: 'exclusive' }, async () => {
+    await onBecomeProvider();
```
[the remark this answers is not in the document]
ZedLi: This is really great feedback. I took some of it to try and refactor this file to be a bit easier to read.

I did end up moving the IIFE that checks for the service provider down which had the nice side effect of not having the variable be a promise that needs to be awaited. However, I didn't quite extract it out to the `setup` function you suggested. Calling the setup in both the lock and immediately means initial providers will end up calling `onBecomingProvider` twice. I think it makes more sense to keep those functions still separate. 

I liked your premise of setting `ready` to the `setup` promise and just instead set ready to the individual `onBecomeProvider` or `onNotProvider` promises. I also moved/extracted those methods out of the main function to try and reduce the clutter inside the method which hopefully helps makes it easier to read.

Let me know if these changes make sense or if I missed any suggestions/ideas.

addons/api/addon/workers/utils/shared-service.js
```diff
@@ -0,0 +1,269 @@
...
+    readyResolve();
+    readyResolve = null;
+    await new Promise(() => {});
+  });
+
+  (async () => {
+    let providerExists = false;
+
+    // Make sure a lock was acquired which means we have a provider
+    while (!providerExists) {
```
lisbet-alvarez, line 51: for learning purposes: in what scenarios would there be no provider at all? I was thinking that there would be a provider within the locks held?

addons/api/addon/workers/utils/shared-service.js
```diff
@@ -0,0 +1,269 @@
...
+  });
+
+  await register();
+};
+
+// When we become the provider, we do two things:
+// 1. Set an event listener to check for clients trying to register with us. As part of registration,
+//    we also setup a listener with the individual client to listen for requests.
+// 2. Send a message to all tabs that we've now become the provider
+const onBecomeProvider = async ({
```
calcaide, line 166: None blocking comment:  If the provider tab is close, how do we re-assign the other tab to be the provider? Correct me if I am wrong, but for a tab to know it's the provider, we are using the `status.isServiceProvider` right?

lisbet-alvarez, line 166: So is this function only called at the very beginning when the shared service is initialized and when the provider changes onProviderChange is called?

addons/api/addon/workers/utils/shared-service.js
```diff
@@ -0,0 +1,269 @@
...
+    const clientChannel = new BroadcastChannel(
+      getClientBroadcastChannelName(clientId, serviceName),
+    );
+
+    // Acquire another lock that immediately is waiting for the
+    // initial lock on the client tab to be resolved to handle cleanup.
+    navigator.locks.request(clientId, { mode: 'exclusive' }, async () => {
+      // The client tab was closed, clean up.
+      clientChannel.close();
+    });
```
lisbet-alvarez, line 188: On line 185-188 so basically as soon as the original lock on the `clientId` releases the lock we set this up to close the `clientChannel`, but wouldn't the client channels need to stay open to listen in on messages sent by the common channel? I am most likely misunderstanding something.

lisbet-alvarez (commented):
Left a couple of questions to better understand the shared service

addons/api/addon/workers/utils/shared-service.js
```diff
@@ -0,0 +1,269 @@
...
+
+  let { promise: ready, resolve: readyResolve } = Promise.withResolvers();
+  const status = {
+    ready,
+    isServiceProvider: false,
+  };
+
+  const proxy = getProxy(target, status);
+
+  // This lock is used to ensure that only one tab can become the provider with an indefinite lock.
```
calcaide, line 25: none blocking comment: nice and clever using `locks` I didn't know about it, after reading the web locks api in MDN I think this is very nice approach 🙌

addons/api/addon/workers/utils/shared-service.js
calcaide: Definetly will wait for Zed answer, but after reading your question I wanted to double check my understanding of this part, so my thoguhts here are thinking loud to check if we understood this properly. 

My understanding here is the clientChannel responds to a client tab, so once that is close we want to also close the channel because the messages that might receive back aren't need/valid anymore.

calcaide (approved):
Thanks for the terrific work!! Learned a bit with this PR about locks and good interaction between channels!! can't wait to start using it at implementation stage

DhariniJeeva (approved):
awesome work on this! 🎉

addons/api/addon/workers/utils/shared-service.js
ZedLi: The re-assignment happens automatically. What happens is the lock here will immediately call `onBecomeProvider` once the old one releases its lock (due to the tab closing). So what is happening is _all_ tabs are waiting to become the provider but are prevented from doing so until the previous one is released. 

 https://github.com/hashicorp/boundary-ui/blob/3f38cf2d2d76d1bc593a8900d541ad560855dad1/addons/api/addon/workers/utils/shared-service.js#L26-L45

ZedLi: Yup that's correct

addons/api/addon/workers/utils/shared-service.js
ZedLi: Carlos is right, in this case what happens is the client tab was closed (which means there shouldn't be communication anymore) which means we can clean up the broadcast channel. If the client tab was closed, that means the provider still has a handle to that broadcast channel (but it's the only one subscribed now and would just send messages to itself). That's why we need to clean it up and this lock basically ensures the moment the client tab is closed, it will close it.

The client channels don't listen in on the common channel messages, those are two different broadcast channels.

Let me know if it's still unclear and I can walkthrough if you'd like

priya-patel04 (approved):
Amazing provider-client pattern implementation!! 🙌

PR #2933, opened 2025-07-17
feat: 🎸 Add sqlite handler and parser

ZedLi:
✅ Closes: https://hashicorp.atlassian.net/browse/ICU-17410
✅ Closes: https://hashicorp.atlassian.net/browse/ICU-17438
# Description
Add a handler for interacting with the sqlite DB service as well as a parser to generate the queries used by sqlite. This PR also includes targets as an example. 

image

## How to Test
Use this [cluster](http://boundary-alb-lswjurcf-1368391987.us-east-1.elb.amazonaws.com:9200) with 400k targets as a playground. DM me for credentials. 

Try using both normal `LIKE` queries and `FTS5` by removing or keeping `target` in this [set](https://github.com/hashicorp/boundary-ui/blob/a57aff356d5b96abb21d61955a8e8afef7402a12/addons/api/addon/services/sqlite-db.js#L25).

Try making requests in multiple tabs and see what happens, try console logging inside `shared-service.js` to see how different tabs coordinate. 

Download the DB file using the dev dropdowns and try playing around with the `sqlite3` cli. 

## Checklist

## PCI review checklist


ui/admin/tests/acceptance/targets/list-test.js
```diff
@@ -269,30 +264,6 @@ module('Acceptance | targets | list', function (hooks) {
     assert.dom(selectors.TABLE_SESSIONS_ID(instances.session.id)).isVisible();
   });
 
-  test('targets table is sorted by `created_time` descending by default', async function (assert) {
```
ZedLi, line 272: I removed this test since right now by default I don't add a sort on `created_time` to every request since it _should_ already be sorted in that order by backend.

My main thinking on this is if we do add the default sort to every request, that means we likely need to add an index to `created_time` for every resource as well which I would like to avoid since there isn't a big benefit besides ensuring the right order on the off chance that it got out of order at some point while the main con is more disk space usage for all the indexes (and maybe slower writes).

Let me know your thoughts on this.

addons/api/addon/services/sqlite.js
```diff
@@ -0,0 +1,120 @@
...
+    name: 'name',
+    description: 'description',
+    address: 'address',
+    scope_id: 'scope.scope_id',
+    created_time: 'created_time',
+  },
+};
+
+// A list of tables that we support searching using FTS5 in SQLite.
+export const searchTables = new Set(['target']);
```
ZedLi, line 25: If we end up always using or not using FTS5 at all for searching, we can remove this

addons/api/addon/workers/sqlite-worker.js
```diff
@@ -6,21 +6,155 @@
...
+        const [row] = db.exec('PRAGMA user_version;', { rowMode: 'object' });
+        // Check if we're on a newer schema version and clear the database if so
+        if (SCHEMA_VERSION > row.user_version) {
+          db.exec(CLEAR_DB);
+        }
+
+        db.exec(CREATE_TABLES(SCHEMA_VERSION));
+        done = true;
+      } catch (err) {
+        // If we get a SQLite3Error and the pool is at capacity, we'll assume the error is
```
ZedLi, line 55: The reason we have this whole retry system in place is because in the event of multiple users, we might run out of file handles. The initial capacity is 6, (see this [documentation](https://sqlite.org/wasm/doc/trunk/persistence.md#vfs-opfs-sahpool)) and this is because this particular implementation pre-opens OPFS access handles ahead of time for usage. If we run out, we can just add more and retry the operation. Take a look at this [comment](https://github.com/rhashimoto/wa-sqlite/discussions/67#discussioncomment-4249212) for more info (Mainly the second paragraph). 

This is what it looks like in practice:

Image

This also makes it easy to clear out and reset everything (for all users) by removing the VFS but that could also be considered a con since there's less isolation between personas.

The alternative to this could be to just register a VFS for _each_ persona (see the above documentation again for setting a `name` of the VFS, we use normally use the default `opfs-sahpool`). This feels a bit less conventional and conceptually we're treating it as if each persona has it's own file system instead of just a DB file in the file system. However, this could likely remove the need to check for capacity since it's unlikely we would hit the cap as each persona gets it's own VFS. 

This is what it looks like:
Image

Let me know if you have any thoughts on this.

addons/api/addon/handlers/sqlite-handler.js
```diff
@@ -4,16 +4,193 @@
...
+import { modelMapping } from 'api/services/sqlite';
+import { get } from '@ember/object';
+import { typeOf } from '@ember/utils';
+import { hashCode } from '../utils/hash-code';
+import { generateSQLExpressions } from '../utils/sqlite-query';
 
 export default class SqliteHandler {
+  @service sqlite;
+
+  batchLimit = 5_000;
```
calcaide, line 17: Asking for learning purposes: I do understand batchLimit being a property only used within this class, seems to me is declared as type number, but never seen `5_000` to declare `5000` what is the difference?

addons/api/addon/utils/sqlite-query.js
```diff
@@ -0,0 +1,160 @@
...
+  if (paginationClause) {
+    parameters.push(pageSize, (page - 1) * pageSize);
+  }
+
+  const selectClause = `SELECT ${select ? select.join(', ') : '*'} FROM ${resource}`;
+
+  return {
+    // Replace any empty newlines or leading whitespace on each line to be consistent with formatting
+    // This is mainly to help us read and test the generated SQL as it has no effect on the actual SQL execution.
+    sql: `
```
calcaide, line 59: None blocking comment: Clever!! I like the separation of the query elements/clauses

addons/api/addon-test-support/helpers/sqlite.js
```diff
@@ -0,0 +1,37 @@
...
+export function setupSqlite(hooks) {
+  hooks.beforeEach(async function () {
+    const sqliteDb = this.owner.lookup('service:sqlite');
+    await sqliteDb.setup(`test-sqlite-${uuidv4()}`);
+  });
+
+  hooks.afterEach(async function () {
+    const sqliteDb = this.owner.lookup('service:sqlite');
+
+    await sqliteDb.deleteDatabase();
```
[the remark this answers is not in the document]
ZedLi: Yeah, in this case if you look at the code for this function, it can figure out what what the current name of the DB is automatically so we don't need to pass it in. You can technically have more than 1 DB running at a time but they'd just have their own file they're interacting with.

addons/api/addon/handlers/sqlite-handler.js
ZedLi: This was originally used in the `indexed-db-handler` for 10,000 so maybe it's less necessary now but this is just a [numeric separator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Lexical_grammar#numeric_separators) which helps with readability (also see the original [proposal](https://github.com/tc39/proposal-numeric-separator), this was added in ECMAScript 2021). They're functionally the same thing.

addons/api/addon/workers/sqlite-worker.js
```diff
@@ -6,21 +6,155 @@
...
+          poolUtil.getCapacity() === poolUtil.getFileCount()
+        ) {
+          await poolUtil.addCapacity(1);
+        } else {
+          console.error('SQLite initialization error:', err);
+          throw err;
+        }
+      }
+
+      // If we have more than 20 files, that might indicate a problem so let's just
```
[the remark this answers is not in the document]
ZedLi: It was an arbitrary number and I picked it bc it seemed high. Hopefully it's unlikely that it gets hit but this was more of a just in case "something might be wrong scenario, let's reset". 

I tested creating a lot of different users which is likely the only way this would ever happen but it's unlikely a user would have more than a few so ideally this would never get hit.

addons/api/addon-test-support/helpers/sqlite.js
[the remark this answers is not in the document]
calcaide: ahh that is right, using the `db` that is assigned and initialize, thanks for the pointer!

addons/api/addon/workers/sqlite-worker.js
ZedLi: So technically it's variable but initializing takes 2 entries but removes the second one after. It likely can support around 5 users in normal use cases when I was testing (n-1).

addons/api/addon/workers/utils/schema.js
```diff
@@ -0,0 +1,75 @@
...
+-- Note that this only creates the FTS index and cannot reference the target content
+CREATE VIRTUAL TABLE IF NOT EXISTS target_fts USING fts5(
+    id,
+    type,
+    name,
+    description,
+    address,
+    scope_id,
+    created_time,
+    content='',
```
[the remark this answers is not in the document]
ZedLi: I actually did but originally discounted it because of this drawback:

> Substrings consisting of fewer than 3 unicode characters do not match any rows when used with a full-text query.

As I wanted all searches to use the full text searching and anything less than 3 returned nothing but in retrospect I could have just checked to see the length of the search and if it's less than 3, just fall back to the normal `LIKE` query.

The tradeoff then is that searches now are basically the same unless it's 2 characters or less which then still does a full table scan and is slower _and_ that this index uses considerably more storage.

I just tested the storage usage and the FTS5 index was 3.5x bigger than the standard tokenized index, which increased storage by ~39% vs 11% in one cluster, though I don't have my other cluster to compare to anymore. I'm not sure how much extra storage it'd use in a cluster where all the fields are populated.

Do we think this tradeoff is better?

addons/api/addon/workers/sqlite-worker.js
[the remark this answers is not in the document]
calcaide: Gotcha, yeah it's a good error handling mechanism and 20 looks a good threshold.

addons/api/addon/workers/sqlite-worker.js
```diff
@@ -6,21 +6,155 @@
...
+      bind: items.flat(),
+      rowMode: 'object',
+    });
+  },
+  deleteResource: ({ resource, ids }) => {
+    if (ids.length === 0) {
+      return [];
+    }
+
+    // TODO: Test with a large amount of deletions, should we chunk here as well?
```
calcaide, line 151: None blocking comment: at first read of this comment I couldn't think of a case with such amount of deletions that will require us to chunk them, but giving a second though I might be missing context.

What cases do you think might require us to chunk deletions?

Perhaps a user of a company with big infra that has a quite outdated DB in memory? I guess that will require quite a bit of deletions of the old data and insertions of the new one.

In any case, IMHO no need to change this within this PR, we can take care of this later during the implementation phase.

ZedLi: Yeah exactly, that's why I didn't do anything yet since it's also a pain to test to have that many deleted resources, but it would basically be if for whatever reason they deleted more than 32,766 resources in between refreshes.

addons/api/addon/workers/sqlite-worker.js
```diff
@@ -6,20 +6,149 @@
...
+        }
+
+        db.exec(CREATE_TABLES(SCHEMA_VERSION));
+        done = true;
+      } catch (err) {
+        // If we get a SQLite3Error and the pool is at capacity, we'll assume the error is
+        // because we hit the cap so we add more capacity and see if that solves it.
+        if (
+          err.name === 'SQLite3Error' &&
+          poolUtil.getCapacity() === poolUtil.getFileCount()
```
lisbet-alvarez, line 59: question for understanding: why not check for `getCapacity` >= to `getFileCount` and also move the `getCapacity` > 20 check above this one?

addons/api/addon/handlers/sqlite-handler.js
```diff
@@ -4,16 +4,193 @@
...
+      schema,
+      payload,
+      null,
+      'query',
+    );
+
+    const { data: payloadData } = normalizedPayload;
+
+    // Store the new data we just got back from the API refresh
+    const items = payloadData.map((datum) => {
```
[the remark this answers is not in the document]
ZedLi: Great points, I took your feedback and changed the `INSERT_STATEMENTS` to be able to generate the columns by taking in the `modelMapping` which makes the statement generic and works for all types.  I think it'd be fine to just pass in the `modelMapping` and have the worker apply the mapping so it's still really controlled by the consumer. However, I didn't change the creation of `items` for two reasons:

1. The ember object itself can sometimes have properties that can't be properly cloned with [structured clone](https://developer.mozilla.org/en-US/docs/Web/API/Web_Workers_API/Structured_clone_algorithm) so it can error out when trying to serialize it and send it to the worker. I'd have to transform it to standard POJO from the ember object regardless to guarantee no issues. 
2. I still like the idea that the sqlite worker just inserts the data as is and is completely unaware of what the models need to look like and that's falls under the responsibility of the consumer. Right now it ends up being a little leaky by still passing in the `modelMapping` but I think that's a fair tradeoff for the convenience you suggested.

 > An alternative might be to have a test that introspects the table and compares it against modelMapping for all expected columns. I think a similar test would work to ensure that the fts table is kept in sync too

Can you expand on what kind of test and how it would look like? Is the suggestion I query the DB for metadata regarding the tables themselves?

addons/api/addon/workers/sqlite-worker.js
ZedLi: You're right, it likely can be moved inside the catch which I did. I still kept it after this if check though because it only needs to be checked _after_ we add capacity and should be the last thing we do since you can't interact anymore after removing the VFS, otherwise you'd need to `continue` the loop early.

lisbet-alvarez: ah ok, thank for explanation!

addons/api/addon/workers/sqlite-worker.js
calcaide: Thanks for the explanation

addons/api/addon/handlers/sqlite-handler.js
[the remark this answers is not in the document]
hashicc: > Can you expand on what kind of test and how it would look like? Is the suggestion I query the DB for metadata regarding the tables themselves?

Yup! Using [`table_info`](https://sqlite.org/pragma.html#pragma_table_info) and then just comparing that the `modelMapping` has all the keys

addons/api/addon/utils/sqlite-query.js
```diff
@@ -0,0 +1,160 @@
...
+      ${selectClause}
+      ${whereClause}
+      ${orderByClause}
+      ${paginationClause}`
+      .replace(/^\s+/gm, '')
```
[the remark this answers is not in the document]
hashicc: hah, that's exactly right. I think that's perfect

addons/api/addon/workers/utils/schema.js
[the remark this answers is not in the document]
hashicc: Thanks for the comparisons, that's pretty interesting. I agree, being conservative on storage is better than the performance gain until there's a case that dictates otherwise

calcaide (approved):
Terrific work!!

I have no blocking feedback, all my comments have been for learning purposes and to understand better the context. Definitely, I will keep this PR bookmark because it will be super useful while implementing the migration within the UI domains.

Thank you for such a heavy lift and lots of learnings taken from this work. 🙌 🚀

ui/admin/tests/acceptance/targets/list-test.js
ZedLi: I also wanted to point out that I changed the index to a composite index here: 
https://github.com/hashicorp/boundary-ui/blob/b64228d96680458dc8dbe7a5c8abec49626e9324/addons/api/addon/workers/utils/schema.js#L33

With two individual indexes (one on `scope_id` and one on `created_time`), only the `scope_id` was getting used to filter and the ORDER BY won't end up using the `created_time` index. The composite index has the benefit of still being used for just a `scope_id` WHERE clause too so it won't be any worse.

addons/api/addon/handlers/sqlite-handler.js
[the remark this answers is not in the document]
ZedLi: Good call, I didn't think I could easily unit test this but it was simpler than I thought and works well

addons/api/addon/workers/sqlite-worker.js
ZedLi: I added a test for this and it did fail so I fixed the issue

addons/api/addon/handlers/sqlite-handler.js
```diff
@@ -4,16 +4,193 @@
...
+          page,
+          pageSize,
+          select: ['data'],
+        });
+
+        const rows = await this.sqlite.fetchResource({
+          sql,
+          parameters,
+        });
+        console.timeLog(`SQLite fetch ${type}`, 'rows');
```
DhariniJeeva, line 107: do we plan to keep the console.time() logs for performance tracking long term or is it just temporary for debugging/devlopment purposes?

addons/api/addon/services/sqlite.js
```diff
@@ -0,0 +1,119 @@
...
+// The JSON data column is not included here but is assumed to always be last.
+export const modelMapping = {
+  target: {
+    id: 'id',
+    type: 'type',
+    name: 'name',
+    description: 'description',
+    address: 'address',
+    scope_id: 'scope.scope_id',
+    created_time: 'created_time',
```
DhariniJeeva, line 19: will we be adding mappings for the rest of the resources and their fields here as well?

addons/api/addon/services/sqlite.js
```diff
@@ -0,0 +1,119 @@
...
+  }
+
+  deleteResource(resource, ids) {
+    return this.worker.postMessage({
+      method: 'deleteResource',
+      payload: { resource, ids },
+    });
+  }
+
+  async downloadDatabase() {
```
DhariniJeeva, line 85: just curious- is the downloadDatabase feature mainly for dev/debug workflow?

addons/api/addon/utils/sqlite-query.js
```diff
@@ -0,0 +1,166 @@
...
+        parameters.push(`%${search}%`);
+        return `${field}${OPERATORS['contains']}`;
+      }),
+    ),
+  );
+  conditions.push(searchConditions);
+}
+
+// Comparison Operators
+const OPERATORS = {
```
DhariniJeeva, line 153: minor nitpick: would it make sense to move the constants to the top of the file for visibility?

addons/api/addon/services/sqlite.js
ZedLi: Yup, that will be handled in separate PRs for each resource

addons/api/addon/utils/sqlite-query.js
ZedLi: I can, but I treated these more as functions which I prefer being declared after where they're used so when you're reading the file, you first see the context where it's being used rather than a const/function that you skip over initially to see where it's being used.

I definitely would declare them at the top if they were for example, just static strings

addons/api/addon/services/sqlite.js
ZedLi: Yup it is, it's in the dropdown that's dev only

hashicc (dismissed):
Image

Image

🆒 💯 🚀 

It's so cool being able to download the database and actually interact with the data with any tool. The browser tools for indexeddb always felt a bit crufty and limiting which made troubleshooting trickier. From the querying language, to the performance, to managing the database, having sqlite feels like such a win.

Thanks for all the effort you've put into getting this working and refining it.

addons/api/addon/handlers/sqlite-handler.js
DhariniJeeva: oops missed that 😅

addons/api/addon/utils/sqlite-query.js
DhariniJeeva: thanks for the clarification, that makes sense.

DhariniJeeva (approved):
awesome job! 🎉  

Appreciate the clear explanations and the effort you put into this!

addons/api/addon/handlers/sqlite-handler.js
```diff
@@ -4,16 +4,203 @@
...
+        } = data;
+        const supportedModels = Object.keys(modelMapping);
+
+        // Go through normal flow if we don't yet support the model
+        // or if we don't have a sqlite db instance
+        if (!supportedModels.includes(type) || !this.sqlite) {
+          return next(context.request);
+        }
+
+        const adapter = store.adapterFor(type);
```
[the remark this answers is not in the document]
ZedLi: If you're talking specifically about a random sqlite operation failing, I don't think fallbacks are necessary as that would be more on re-tries rather than fallbacks. If SQLite is not available _at all_, that's a different problem. 

There's a wider question for product that's TBD if we want to keep indexedDB to fallback on since this has stricter requirements (HTTPS) and what to do if for some reason we can't use sqlite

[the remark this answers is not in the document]
priya-patel04: Ahh okay got it. Yeah I was asking more about random sqlite op failing, but make sense. Thanks for clarifying.

PR #2958, opened 2025-08-06
Icu 17516 use in memory sqlite in non secure contexts

ZedLi:
✅ Closes: https://hashicorp.atlassian.net/browse/ICU-17516

# Description
Add checks to see if we're in a [secure context](https://developer.mozilla.org/en-US/docs/Web/Security/Secure_Contexts) as the OPFS and weblock APIs require them.

If we're not in a secure context, just start up SQLite in memory so a user still can still get full functionality just without persistence. This would then allow us to remove indexed DB in the future as we wouldn't need to ever fallback.
 
## How to Test
Setup the branch in an http environment that's not localhost. 
You could do this using enos or using a tool like ngrok. If using ngrok, you'd need to force the scheme to be http with something like `ngrok http --scheme=http 4200`. Your browser might also try and force you onto HTTPS with ngrok's urls. For Firefox, I had to set `network.stricttransportsecurity.preloadlist` to false in `about:config`.

Try visiting targets and you should be able to use it like normal, refresh and confirm nothing is persisted. 

## Checklist

## PCI review checklist

### DIFF
--- a/addons/api/addon/workers/sqlite-worker.js
+++ b/addons/api/addon/workers/sqlite-worker.js
@@ -25,6 +25,7 @@ let sqlite3;
 // https://www.sqlite.org/limits.html
 const MAX_HOST_PARAMETERS = 32766;
 const SCHEMA_VERSION = 1;
+const isSecure = self.isSecureContext;
 
 const methods = {
   initializeSQLite: async () => {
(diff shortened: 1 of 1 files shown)


addons/api/addon/workers/sqlite-worker.js
```diff
@@ -37,21 +38,32 @@ const methods = {
...
+      }
+
+      db.exec(CREATE_TABLES(SCHEMA_VERSION));
+      // Force an analyze on initial connection
+      db.exec('PRAGMA optimize=0x10002;');
+    }
+
+    // Instantiate an in memory DB if we're not in a secure context as OPFS is unavailable otherwise
+    if (!isSecure) {
+      db = new sqlite3.oo1.DB(':memory:');
```
calcaide, line 55: None blocking comment and only asking for leraning purposes: Did we test/try this with a big amount of data? I am wondering the memory consumption with a big load, I am sure shouldn't be a problem, specially after the changes you made to batch the digestion of big amounts.

calcaide (approved):
Very clever solution!! this simplifies a lot from the planned fallback and keeps consistent with the sqlite choice nice work!! 🙌 

Tested and I can confirm the usage of SQLite within targets and the none persistence of it, also indexedDB shows 0 entries on targets as expected

lisbet-alvarez (approved):
Works great & verified nothing persisted!

hashicc (approved):
Works great! An alternative way I found to test this is with this `/etc/hosts` entry:
```
127.0.0.1 localhost boundary
```

And then booting up the app with:
```
ENABLE_MIRAGE=false API_HOST=http://boundary:9200 pnpm start
```

Then both `http://boundary:4200` and `http://localhost:4200` work. The `API_HOST` has to be `http://boundary:9200` for it to work for both.

Image

Image

PR #2964, opened 2025-08-13
fix: 🐛 fix e2e-tests api client fixture dependency on `apiClientTestAfter`

hashicc:
# Description
For tests that don't use `apiClientTestAfter` fixture it still has to be provided for any test that uses `apiClient` to fulfill the dependency. To work around this a noop function can be provided instead and then overridden in the `api-client.spec.js` test itself.

This was introduced in this [commit](https://github.com/hashicorp/boundary-ui/pull/2901/commits/c15562a7a30da21c61e15d0fdab6dfad55ce1d6d) when the `apiClientTestAfter` fixture was introduced. The `api-client.spec.js` tests were passing locally but I didn't test the effect of the `apiClientTestAfter` fixture on other e2e tests. The [PR](https://github.com/hashicorp/boundary-ui/pull/2901) that introduced this issue passed CI because desktop e2e tests weren't ran (even though those tests are currently not blocking, the check would have failed).

## How to Test
Relying on the desktop client test to pass on this branch. It was [previously failing](https://github.com/hashicorp/boundary-ui/actions/runs/16941334026/job/48010986680#step:13:174).

## Checklist

## PCI review checklist

### DIFF
--- a/e2e-tests/helpers/boundary-api-client.ts
+++ b/e2e-tests/helpers/boundary-api-client.ts
@@ -246,6 +246,9 @@ export const boundaryApiClientTest = base.extend<{
   controllerAddr?: string;
   apiClientTestAfter?: () => {};
 }>({
+  // this is a noop so that the `apiClient` fixture doesn't fail for other tests,
+  // the `apiClientTestAfter` fixture is used in api-client tests to provide a hook for assertions
+  apiClientTestAfter: () => {},
   apiClient: async ({ controllerAddr, apiClientTestAfter }, use) => {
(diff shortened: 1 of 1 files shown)


ZedLi (approved):
Thanks for the fix!

DhariniJeeva (approved):
thanks for the fix!

hashicc:
@ZedLi Thanks for investigating. I updated the PR body with a bit more context of how this got through. Tested in CI against desktop and locally tests are running successfully against `admin:ent:docker`.

PR #2969, opened 2025-08-22
feat: 🎸 Add sqlite support for auth methods

ZedLi:
# Description
Added sqlite support for auth methods. 

## How to Test
Use this [cluster](http://boundary-alb-dvtowfcd-362873315.us-east-1.elb.amazonaws.com:9200/) and goto the org that's indicated that has many auth methods.

Keep in mind that we hit the API for every request so any search or filtering will take a bit. Try the normal use case of a few auth methods in the global scope as well.

## Checklist

## PCI review checklist

### DIFF
--- a/addons/api/addon/handlers/sqlite-handler.js
+++ b/addons/api/addon/handlers/sqlite-handler.js
@@ -81,8 +81,15 @@ export default class SqliteHandler {
                 await writeToDbPromise;
               }
             } catch (err) {
-              console.error('Error retrieving data:', err);
-              throw err;
+              payload = await this.retryQueryFailure({
+                err,
+                type,
+                adapter,
+                store,
+                schema,
+                remainingQuery,
+              });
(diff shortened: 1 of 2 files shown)


addons/api/addon/handlers/indexed-db-handler.js
```diff
@@ -64,6 +64,11 @@ export default class IndexedDbHandler {
           if (storeToken) {
             const tokenObj = await indexedDb.token.get(tokenKey);
             listToken = tokenObj?.token;
+          } else {
```
ZedLi, line 67: This was a bug that got missed when I added batching to requests. It was mistakenly deleting on every batch insert instead of just once the first time so you only ever got the last page of results.

lisbet-alvarez (dismissed):
Great work!!

lisbet-alvarez (approved):
Great work!! Thank u!

PR #2977, opened 2025-08-28
[FE-112] Boundary Enterprise a11y audit- User details form

venusang:
# Description
This updates the input fields on the User details page so that they are `readonly` instead of `disabled`.  Additionally it changes the edit button text from `Edit form` to `Edit User details` so that it is more descriptive.  Lastly, `aria-describedby` was also added to the button so that it is a11y compliant.  

:tickets: [FE-112](https://hashicorp.atlassian.net/browse/FE-112)

fe-112-before-after

## How to Test
Navigate to a User in the Users list.  Verify that the input fields are readonly (you should be able to highlight the text with your mouse) and that the submit button reflects the text changes shown in the after screenshot.

## Checklist

## PCI review checklist

[FE-112]: https://hashicorp.atlassian.net/browse/FE-112?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/addons/core/translations/resources/en-us.yaml
+++ b/addons/core/translations/resources/en-us.yaml
@@ -573,7 +573,7 @@ user:
     delete: Delete User
     edit-details:
       button-text: 'Edit User details'
-      aria-describedby: 'user-details'
+      aria-describedby: 'edit-user-details'
 role:
   title: Role
   title_plural: Roles
--- a/addons/core/translations/resources/en-us.yaml
+++ b/addons/core/translations/resources/en-us.yaml
@@ -571,10 +571,7 @@ user:
   actions:
     add-accounts: Add Accounts
     delete: Delete User
-    edit-details:
-      button:
-        text: Edit User details
-        aria-label: 'edit form: users'
+    edit-details: 'Edit User details'
 role:
   title: Role
   title_plural: Roles
--- a/addons/core/translations/resources/en-us.yaml
+++ b/addons/core/translations/resources/en-us.yaml
@@ -571,7 +571,9 @@ user:
   actions:
     add-accounts: Add Accounts
     delete: Delete User
-    edit-details: 'Edit User details'
+    edit-details:
+      button-text: 'Edit User details'
+      aria-describedby: 'user-details'
 role:
   title: Role
   title_plural: Roles
--- a/addons/core/translations/resources/en-us.yaml
+++ b/addons/core/translations/resources/en-us.yaml
@@ -571,6 +571,10 @@ user:
   actions:
     add-accounts: Add Accounts
     delete: Delete User
+    edit-details:
+      button:
+        text: Edit User details
+        aria-label: 'edit form: users'
 role:
   title: Role
   title_plural: Roles
--- a/ui/admin/app/components/form/user/index.hbs
+++ b/ui/admin/app/components/form/user/index.hbs
@@ -4,7 +4,6 @@
 }}
 
 <Rose::Form
-  id={{t 'resources.user.actions.edit-details.aria-describedby'}}
   @onSubmit={{@submit}}
   @cancel={{@cancel}}
   @disabled={{@model.isSaving}}
@@ -53,9 +52,6 @@
 
   {{#if (can 'save model' @model)}}
     <form.actions
-      @ariaDescribedBy={{t
-        'resources.user.actions.edit-details.aria-describedby'
-      }}
       @enableEditText={{t 'resources.user.actions.edit-details.button-text'}}
       @submitText={{t 'actions.save'}}
       @cancelText={{t 'actions.cancel'}}
--- a/ui/admin/app/components/form/user/index.hbs
+++ b/ui/admin/app/components/form/user/index.hbs
@@ -16,7 +16,7 @@
     @value={{@model.name}}
     @isInvalid={{@model.errors.name}}
     @isOptional={{true}}
-    disabled={{form.disabled}}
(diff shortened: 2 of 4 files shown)


cameronperera (dismissed):
LGTM

ui/admin/app/components/form/user/index.hbs
```diff
@@ -52,7 +52,8 @@
 
   {{#if (can 'save model' @model)}}
     <form.actions
-      @enableEditText={{t 'actions.edit-form'}}
+      @enableEditText={{t 'resources.user.actions.edit-details.button.text'}}
+      aria-label={{t 'resources.user.actions.edit-details.button.aria-label'}}
```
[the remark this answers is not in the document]
venusang: I implemented this based on the recommendations outlined in the ticket [here](https://hashicorp.atlassian.net/browse/FE-112).

> 8.2 “Edit Form” button text
> Issue: Due to the structure of the page content, it was really difficult to tell what form I would be editing while using only a screen reader to navigate the page. I would consider adding the form name to the button so that it would be more informative for the user. Suggest adding an aria-label attribute to the button, where the value is “edit form: users”. Note that the visible button label matches the first part of the aria-label attribute value, conforming to [WCAG Success Criterion 2.5.3 Label In Name](https://www.w3.org/WAI/WCAG22/Understanding/label-in-name.html).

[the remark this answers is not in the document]
ZedLi: I took the ticket to mean to add an `aria-label` to make the existing `Edit form` button more descriptive for screen readers such that `Edit form: users` now properly describes the existing button. I also assume just changing the text of the button without an aria label would have accomplished that but it sounds like the recommendation is to add the label. 

But either way reading the linked WCAG success criteria it seems like this change would now violate it:
> Accessible name matches visible label: The accessible name and visible label of a control match.
> Accessible name starts with visible label: The accessible name "Search for a value" begins with the text of the visible label, "Search".

Since the text is now `Edit user details` and the label is `Edit form: users`, they don't match and the accessible name doesn't start with the visible label (as opposed to `Edit form: users` just being a more descriptive match of the original text).

ZedLi (dismissed):
Looks good! Thanks for the update

hashicc (dismissed):
Thanks for helping us with our a11y!

cameronperera (dismissed):

ui/admin/app/components/form/user/index.hbs
```diff
@@ -52,7 +53,10 @@
 
   {{#if (can 'save model' @model)}}
     <form.actions
-      @enableEditText={{t 'actions.edit-form'}}
+      @ariaDescribedBy={{t
```
ZedLi, line 56: I'm not sure I understand why we linked the ID of the form itself, what text are we trying to descriptively add to this by using `aria-describedby`? It also seems weird to use a translated text as the ID when the element itself should have the translated text

ZedLi (changes requested):

ui/admin/app/components/form/user/index.hbs
venusang: I pinged you in the JIRA ticket comment but adding reference here for good measure.

(taking a screenshot so it has more context versus copy paste):

CleanShot 2025-09-10 at 09 25 10@2x

https://hashicorp.atlassian.net/browse/FE-112?focusedCommentId=828913

venusang: Going off @MelSumner thoughts in the JIRA ticket, I set the form ID value to be `user-details` versus `users` since to me it seems to make better sense that the form is specific to the `user-details`.  `Users` did not seem to make sense to me but maybe I am misunderstanding the purpose here.  

> It also seems weird to use a translated text as the ID when the element itself should have the translated text

@ZedLi from what I am getting here, it sounds like the form ID value should be semantic versus the ember generated ID.  Maybe Melanie could provide some clarity here for both of us.  I'm not all that familiar with `aria-describedby` prior to working on this.

ZedLi: Right, I think all those choices make sense but it sounds like we just need to pick one choice? It seems we could leave the button as `Edit form` if we're adding one of the aria attributes. 

With regards to the current implementation with 3, I was specifically pointing out the fact that it's currently pointing to the form ID instead of something with descriptive text (which melanie suggests as the form title):

> where the value is the ID of the form title element (in this case, the element with the text “Users”)

Though that looks like it has some extra text which would pulled in as part of the helper doc link text.

addons/core/translations/resources/en-us.yaml
```diff
@@ -570,6 +570,9 @@ user:
   actions:
     add-accounts: Add Accounts
     delete: Delete User
+    edit-details:
+      button-text: 'Edit User details'
+      aria-describedby: 'user-details'
```
nikkipurcell: change aria-describedby to 'edit-user-details' to match button text

addons/core/translations/resources/en-us.yaml
```diff
@@ -571,6 +571,9 @@ user:
   actions:
     add-accounts: Add Accounts
     delete: Delete User
+    edit-details:
+      button-text: 'Edit User details'
+      aria-describedby: 'edit-user-details'
```
ZedLi: I don't think we should have a translation for an `aria-describedby`

PR #2981, opened 2025-08-29
feat: 🎸 Refactor session recording filters

ZedLi:
# Description
A proof of concept at using sqlite to load the various filter options in session recordings page. Previously, we were loading all session recordings to then filter down the various filter options we need which takes a huge amount of time just returning all the data to filter. 

This PR has a few notable changes:
1. A simple attempt at establishing a pattern for using SELECT statements. This will be useful in the future for joins/subquery support.
2. Expanded usage of searching to be able to pass instead of just a string, an object that lets a consumer choose which columns to search on. 
3. Decouple our filters from being loaded from the route model hook which is a departure from our convention. This has the benefit of actually letting the page first load and having the filters load async so the page doesn't wait for the filters to finish. This is also much faster to query for each individual filter option from SQLite rather than returning all data to filter in application code. 
    * Having the queries be async does have some downsides in the scenario where there's a lot of filter options and a user might be waiting for the options to load if they're fast or if they goto the page with query params already selected the filter options will "pop in" after the load.

## How to Test
Use this [cluster](http://boundary-alb-dvtowfcd-362873315.us-east-1.elb.amazonaws.com:9200) and try loading session recordings compared to before this PR.

## Checklist

## PCI review checklist

### DIFF
--- a/addons/api/addon/utils/sqlite-query.js
+++ b/addons/api/addon/utils/sqlite-query.js
@@ -4,7 +4,6 @@
  */
 
 import { modelMapping } from 'api/services/sqlite';
-import { searchTables } from 'api/services/sqlite';
 import { typeOf } from '@ember/utils';
 import { underscore } from '@ember/string';
 
@@ -35,16 +34,14 @@ export function generateSQLExpressions(
   addFilterConditions({ filters, parameters, conditions });
   addSearchConditions({ search, resource, tableName, parameters, conditions });
 
+  const selectClause = constructSelectClause(select, tableName);
   const orderByClause = constructOrderByClause(resource, sort);
-
   const whereClause = conditions.length ? `WHERE ${and(conditions)}` : '';
 
   const paginationClause = page && pageSize ? `LIMIT ? OFFSET ?` : '';
   if (paginationClause) {
     parameters.push(pageSize, (page - 1) * pageSize);
   }
-
-  const selectClause = `SELECT ${select ? select.join(', ') : '*'} FROM "${tableName}"`;
 
   return {
     // Replace any empty newlines or leading whitespace on each line to be consistent with formatting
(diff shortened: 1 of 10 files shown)


addons/api/addon/utils/sqlite-query.js
```diff
@@ -142,32 +139,57 @@ function addSearchConditions({
...
-    ),
+  // Use a subquery to match against the FTS table with rowids as SQLite is
+  // much more efficient with FTS queries when using rowids or MATCH (or both).
+  // We could have also used a join here but a subquery is simpler.
+  conditions.push(
+    `rowid IN (SELECT rowid FROM ${tableName}_fts WHERE ${tableName}_fts MATCH ?)`,
   );
-  conditions.push(searchConditions);
+}
+function constructSelectClause(select = [{ field: '*' }], tableName) {
```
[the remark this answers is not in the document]
ZedLi: Hmm yeah you're right, this is a little leaky. I could make a reverse mapping of `modelMapping` (since it currently goes from column to JSON path) and use that so we don't have to be aware of the columns

ui/admin/app/controllers/scopes/scope/session-recordings/index.js
```diff
@@ -95,29 +106,95 @@ export default class ScopesScopeSessionRecordingsIndexController extends Control
...
+        query: {
+          search: { text: search, fields: config.searchFields },
         },
-      }) => {
-        if (!uniqueMap.has(id)) {
-          const projectName = name || id;
-          uniqueMap.set(id, { id, name: projectName, parent_scope_id });
-        }
+        page: 1,
+        pageSize: 250,
```
lisbet-alvarez, line 182: Should we also update the Dropdown component item limit from 500 to 250 to keep it consistent (since its used in desktop app too)?

ui/admin/app/controllers/scopes/scope/session-recordings/index.js
```diff
@@ -95,29 +106,95 @@ export default class ScopesScopeSessionRecordingsIndexController extends Control
...
-        create_time_values: {
-          target: {
-            scope: { id, name, parent_scope_id },
-          },
+  get filterConfigs() {
+    return {
+      userFilters: {
+        select: [
+          { field: 'user_id', isDistinct: true },
+          { field: 'user_name', isDistinct: true },
```
lisbet-alvarez, line 117: would it complicate the select objects too much if it were changed so that `isDistinct` is only specified once since it looks like the parser only looks to see if at least one field isDistinct? Totally get if the object is structured this way for consistency.

lisbet-alvarez (commented):
i only have two general comments but looks good and a big improvement to filters!

hashicc:
> This requires a bit of exposing query generation or adding another intermediate layer/helper though. Any thoughts on that?

I think this is where my brain went when starting to seeing sql-specific concepts like being able to do a distinct query. It would be nice to have a `db` service and have the sqlite-handler use that, and allow the `db` service to be an escape hatch in app code, too (although that might have some redundancy and then have to patch the results together). The `db` service would have the benefits of having it's arguments in terms of sql and the output in terms of the query output. I think it makes sense to offer constrained methods: `distinct`, etc, so that folks are pushed to use `this.store.query()` where possible. Maybe this could be added to the `sqlite` service but that seems like it sits at different layer.

One downside is knowing when to use the `db` service vs use the ember data store & models. The advantage of the ember data store is that it'll referentially update (you you get a list token update, push to store, now any reference to that model is magically updated). The sql results are stale-on-arrival in the sense that they're just POJOs and should be treated as readonly snapshots in time. If the lines are blurred it's possible that we might get updates on the POJO and aside from them not having a common identity, their properties are also not tracked. Maybe this is not so bad because we already have `pushToStore`. I think the downside of a `db` service that returns non-reactive, non-referential data is fine.

Maybe there's a world where sql results could be reactive, the original query stored, and refetched by some key like react-query, in combinations with the [primitives](https://canary.warp-drive.io/api/@warp-drive/core/reactive/type-aliases/ReactiveDocument#type-alias-reactivedocument-t) from warp drive... but that's just my wish for the future

ui/admin/app/controllers/scopes/scope/session-recordings/index.js
ZedLi: So the original reason was I didn't want to complicate it to take in both an array and an object for the `select` and was planning on letting users mix `isDistinct` with other options. When testing, I realized there's not a current use case to mix them so I kept it simple by just assuming you'd only be using distinct clauses if it's there but there's no reason it has to be like that in the future. I'd probably keep it like this just to make the API simpler and for future compatibility and not assume that the presence of any `isDistinct` means all of them are

cameronperera (dismissed):

ui/admin/app/routes/scopes/scope/session-recordings/index.js
```diff
@@ -157,30 +151,13 @@ export default class ScopesScopeSessionRecordingsIndexRoute extends Route {
...
         };
       }
     },
   );
 
-  /**
-   * Sets allSessionRecordings to all session recordings for filters
-   * @param {string} scope_id
-   */
-  async getAllSessionRecordings(scope_id) {
```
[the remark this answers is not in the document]
ZedLi: Good call! Totally missed that

cameronperera (approved):
Great work and tests!

PR #2996, opened 2025-09-09
fix: 🐛 Add better check for determining if we can use OPFS

ZedLi:
# Description
There's an issue with using [getDirectory](https://developer.mozilla.org/en-US/docs/Web/API/StorageManager/getDirectory) in some browsers with private mode. This API is used to access OPFS even if we're in a secure context and if it's not available, the setup will still fail. We should check to see if we have access as part of the secure check.

## How to Test
Try running in private mode and without private mode to confirm both still allow you to login and load resources.

## Checklist

## PCI review checklist


hashicc (approved):
🚀 Thanks for the fix! Tested on chrome, firefox, safari and private/non-private browsing modes

PR #2999, opened 2025-09-11
test: 💍 Refactor target functions and fix failing test

dillanb-hashi:
# Description

Changed a function to it's previous state because it broke a CE test.
Also refactored the target functions for better readability and usability. 

## How to Test

For CE
pnpm playwright test --project=chromium --config admin/playwright.config.js alias.spec.js
For ENT
pnpm playwright test --project=chromium --headed --config admin/playwright.config.js alias-ent.spec.js

## Checklist

## PCI review checklist

### DIFF
--- a/ui/desktop/app/controllers/scopes/scope/projects/targets/index.js
+++ b/ui/desktop/app/controllers/scopes/scope/projects/targets/index.js
@@ -190,6 +190,7 @@ export default class ScopesScopeProjectsTargetsIndexController extends Controlle
     const options = {
       target_id: target.id,
       token: this.session.data.authenticated.token,
+      session_max_seconds: target.session_max_seconds,
     };
 
     if (host) options.host_id = host.id;
--- a/ui/desktop/electron-app/src/helpers/spawn-promise.js
+++ b/ui/desktop/electron-app/src/helpers/spawn-promise.js
@@ -27,16 +27,19 @@ module.exports = {
    * data on either stdout or stderr.  The process is allowed to continue
    * running after the promise resolves.  This function is intended to launch
    * the local proxy.
-   * @param {string} command
+   * @param {string[]} command
+   * @param {string} token
+   * @param {number} timeout Duration in seconds
    * @return {Promise}
    */
-  spawnAsyncJSONPromise(command, token) {
+  spawnAsyncJSONPromise(command, token, timeout) {
     return new Promise((resolve, reject) => {
       const childProcess = spawn(path, command, {
         env: {
           ...process.env,
           BOUNDARY_TOKEN: token,
         },
+        timeout: timeout ? timeout * 1000 : undefined,
       });
       let outputStream = '';
       let errorStream = '';
--- a/ui/desktop/electron-app/src/ipc/handlers.js
+++ b/ui/desktop/electron-app/src/ipc/handlers.js
@@ -71,8 +71,14 @@ handle('cliExists', () => boundaryCli.exists());
 /**
  * Establishes a boundary session and returns session details.
  */
-handle('connect', ({ target_id, token, host_id }) =>
-  sessionManager.start(runtimeSettings.clusterUrl, target_id, token, host_id),
+handle('connect', ({ target_id, token, host_id, session_max_seconds }) =>
+  sessionManager.start(
+    runtimeSettings.clusterUrl,
+    target_id,
+    token,
+    host_id,
+    session_max_seconds,
+  ),
 );
 
 /**
--- a/ui/desktop/electron-app/src/models/session-manager.js
+++ b/ui/desktop/electron-app/src/models/session-manager.js
@@ -23,9 +23,16 @@ class SessionManager {
    * @param {string} target_id
    * @param {string} token
    * @param {string} host_id
+   * @param {number} session_max_seconds
    */
-  start(addr, target_id, token, host_id) {
-    const session = new Session(addr, target_id, token, host_id);
+  start(addr, target_id, token, host_id, session_max_seconds) {
+    const session = new Session(
+      addr,
+      target_id,
+      token,
+      host_id,
+      session_max_seconds,
+    );
     this.#sessions.push(session);
     return session.start();
   }
--- a/ui/desktop/electron-app/src/models/session.js
+++ b/ui/desktop/electron-app/src/models/session.js
@@ -17,6 +17,7 @@ class Session {
   #process;
   #targetId;
   #proxyDetails;
+  #sessionMaxSeconds;
 
   /**
    * Initialize a session to a controller address
@@ -25,12 +26,14 @@ class Session {
    * @param {string} targetId
    * @param {string} token
    * @param {string} hostId
+   * @param {number} sessionMaxSeconds
    */
-  constructor(addr, targetId, token, hostId) {
+  constructor(addr, targetId, token, hostId, sessionMaxSeconds) {
     this.#addr = addr;
     this.#targetId = targetId;
     this.#token = token;
     this.#hostId = hostId;
+    this.#sessionMaxSeconds = sessionMaxSeconds;
   }
 
   /**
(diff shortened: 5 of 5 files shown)


e2e-tests/admin/pages/targets.js
```diff
@@ -70,32 +70,8 @@ export class TargetsPage extends BaseResourcePage {
    * @param {string} alias alias used for the target
    * @returns Name of the target
    */
-  async createTargetWithAddressAndAlias(
-    address,
-    port,
-    alias,
-    targetType = 'tcp',
-  ) {
+  async createTargetWithAddressAndAlias(address, port, alias) {
```
moduli, line 73: thinking, would it be confusing to label this method something like `createTargetWithAddressAndAliasCe` to denote that it's intended for community edition only?

e2e-tests/admin/pages/targets.js
```diff
@@ -126,45 +101,33 @@ export class TargetsPage extends BaseResourcePage {
...
-   * Creates a new TCP target with address in boundary-enterprise
-   * Assumes you have selected the desired project.
-   * @param {string} address Address of the target
+   * Creates a new target. Assumes you have selected the desired project.
    * @param {string} port Port of the target
+   * @param {string} targetType Type of the target ("ssh", "rdp", or "tcp")
    * @returns Name of the target
    */
-  async createTcpTargetWithAddressEnt(address, port) {
+  async createTargetEnt(port, targetType) {
```
moduli, line 109: Do you think `targetType` should be the first argument? (same for the other methods)

moduli (commented):
Looks good. A couple of comments.

e2e-tests/admin/pages/targets.js
dillanb-hashi: Yeah was thinking of if I should do that or not. Probably should just to avoid confusion.

e2e-tests/admin/pages/targets.js
dillanb-hashi: Yeah I could see that. I'll make that change.

e2e-tests/admin/pages/targets.js
```diff
@@ -70,32 +70,8 @@ export class TargetsPage extends BaseResourcePage {
    * @param {string} alias alias used for the target
    * @returns Name of the target
    */
-  async createTargetWithAddressAndAlias(
```
ZedLi, line 73: It looks like this function was split into two separate ones that do nearly the same thing? I think we should start moving towards having helpers that are edition agnostic or just check the actual version we're running against if it can't be determined through the normal inputs, rather than having separate functions since most of the time they're pretty similar. I'm hoping this might also help in writing tests that can be shared between both editions and reduce the split of ENT vs CE tests. 

Could we have made this function work by instead doing a check on the targetType (and removing the default option, since I think TCP is always the default one now) and if it doesn't exist, don't try and fill out that field?

e2e-tests/admin/tests/dynamic-host-catalog-aws.spec.js
```diff
@@ -156,7 +156,7 @@ test.describe('AWS', () => {
 
         // Create a target and add DHC host set as a host source
         const targetsPage = new TargetsPage(page);
-        const targetName = await targetsPage.createTarget(targetPort);
+        const targetName = await targetsPage.createTargetCe(targetPort);
```
moduli, line 159: This is still using a `Ce` method?

e2e-tests/admin/tests/credential-store-static.spec.js
```diff
@@ -31,7 +31,7 @@ test.beforeEach(async ({ page, targetAddress, targetPort }) => {
   const projectsPage = new ProjectsPage(page);
   projectName = await projectsPage.createProject();
   const targetsPage = new TargetsPage(page);
-  targetName = await targetsPage.createTargetWithAddress(
+  targetName = await targetsPage.createTargetWithAddressCe(
```
moduli, line 34: This is still using a `Ce` method

e2e-tests/admin/tests/alias.spec.js
```diff
@@ -206,7 +206,7 @@ test.describe('Aliases', () => {
         const projectsPage = new ProjectsPage(page);
         const projectName = await projectsPage.createProject();
         const targetsPage = new TargetsPage(page);
-        const targetName = await targetsPage.createTargetWithAddress(
+        const targetName = await targetsPage.createTargetWithAddressCe(
```
moduli, line 209: This is still using a `Ce` method

e2e-tests/admin/tests/alias.spec.js
```diff
@@ -148,7 +148,7 @@ test.describe('Aliases', () => {
         await projectsPage.createProject();
         alias = 'example.alias.' + nanoid();
         const targetsPage = new TargetsPage(page);
-        const targetName = await targetsPage.createTargetWithAddressAndAlias(
+        const targetName = await targetsPage.createTargetWithAddressAndAliasCe(
```
moduli, line 151: This is still using a `Ce` method

e2e-tests/admin/tests/alias.spec.js
```diff
@@ -44,7 +44,7 @@ test.describe('Aliases', () => {
         const projectsPage = new ProjectsPage(page);
         await projectsPage.createProject();
         const targetsPage = new TargetsPage(page);
-        const targetName = await targetsPage.createTargetWithAddress(
+        const targetName = await targetsPage.createTargetWithAddressCe(
```
moduli, line 47: This is still using a `Ce` method

moduli (changes requested):

e2e-tests/admin/pages/targets.js
```diff
@@ -9,73 +9,15 @@ import { nanoid } from 'nanoid';
...
-    address,
-    port,
-    alias,
-    targetType = 'tcp',
-  ) {
+  async createTarget(targetType, port, address, alias) {
```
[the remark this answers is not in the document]
dillanb-hashi: Changed function to use optional variables and put in a default for targetType

e2e-tests/admin/pages/targets.js
```diff
@@ -9,73 +9,16 @@ import { nanoid } from 'nanoid';
...
    * @param {string} alias alias used for the target
    * @returns Name of the target
    */
-  async createTargetWithAddressAndAlias(
-    address,
-    port,
-    alias,
-    targetType = 'tcp',
-  ) {
+  async createTarget({ targetType = 'tcp', port, address, alias }) {
```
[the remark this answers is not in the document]
dillanb-hashi: It is currently working with the tests I checked, so copilot may be wrong here. Doing a full run of ce and ent just to check though.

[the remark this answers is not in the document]
moduli: 👍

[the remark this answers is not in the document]
moduli: How do we cover the CE case when there is no target type to select? Do we even want to set a default?

e2e-tests/admin/pages/targets.js
```diff
@@ -9,73 +9,16 @@ import { nanoid } from 'nanoid';
...
+    console.log(`${targetType}, ${port}, ${address}, ${alias}`);
```
ZedLi, line 21: Extra log statement here

e2e-tests/admin/pages/targets.js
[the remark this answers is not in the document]
dillanb-hashi: generic TCP is also the default for ENT so it should be good either way. 
Could also just have no default and change this to not run if targetType isn't set. Might be cleaner if we ever change the default type  
  
    if (targetTypeLabel != 'Generic TCP') {
      await this.page
        .getByRole('group', { name: 'Type' })
        .getByLabel(targetTypeLabel)
        .click();
    }

e2e-tests/admin/pages/targets.js
```diff
@@ -104,258 +47,22 @@ export class TargetsPage extends BaseResourcePage {
...
-      .click();
-    await this.page.getByLabel('Target Address').fill(address);
-    await this.page.getByLabel('Default Port').fill(port);
-    await this.page
-      .getByRole('group', { name: 'Aliases' })
-      .getByLabel('value')
-      .last()
-      .fill(alias);
-    await this.page.getByRole('button', { name: 'Add' }).click();
+    if (targetTypeLabel != 'Generic TCP') {
```
ZedLi, line 50:
```suggestion
    if (targetTypeLabel !== 'Generic TCP') {
```

e2e-tests/admin/pages/targets.js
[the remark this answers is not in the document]
ZedLi: I think not having the default is probably better, that way it's explicit and calling the function for CE implies don't need to set a type but with ent, you can still manually select it if it's ever needed (since now you can't select it and it's assumed to be defaulted to tcp)

ZedLi (approved):
This is a nice change!

e2e-tests/admin/pages/targets.js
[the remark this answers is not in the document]
moduli: Agreed with Zed.

e2e-tests/admin/pages/targets.js
```diff
@@ -104,258 +43,22 @@ export class TargetsPage extends BaseResourcePage {
...
-      .click();
-    await this.page.getByLabel('Target Address').fill(address);
-    await this.page.getByLabel('Default Port').fill(port);
-    await this.page
-      .getByRole('group', { name: 'Aliases' })
-      .getByLabel('value')
-      .last()
-      .fill(alias);
-    await this.page.getByRole('button', { name: 'Add' }).click();
+    if (targetType) {
```
moduli, line 46: Should we add a check if `targetTypeLabel` isn't defined here, fail the test? 
```
if (targetType) {
   if (targetTypeLabel isn't defined) { fail the test}
   ...
```

moduli (commented):
Looks good. One comment - let me know your thoughts

e2e-tests/admin/pages/targets.js
dillanb-hashi: Yeah I can add that.

moduli (approved):
nice changes!

PR #3002, opened 2025-09-15
fix: 🐛 Fix sessions not properly getting cleaned up

ZedLi:
✅ Closes: https://hashicorp.atlassian.net/browse/ICU-17645
✅ Closes: https://hashicorp.atlassian.net/browse/ICU-16776

# Description
This fixes two different (but related) bugs. The first is on windows when logging out or quitting, clicking OK doesn't do anything and the user can't logout or exit. This is because the `this.#process.killed` condition wasn't getting properly tracked due to us using `treeKill`. It doesn't look like there's any issue with using the standard `kill()` so I swapped back to using it. 

The second bug on windows and linux was sessions not properly getting canceled even after clicking OK on the modal. It works fine on Mac. This seems to be because just killing the underlying process we used to start the session doesn't actually cancel on windows or linux. I just manually canceled the sessions before killing the process so we can be consistent.

We us `treeKill` elsewhere but I didn't test that if we needed it. As a followup later we can confirm if it's still needed and remove it if not.

## How to Test
Test logging out and quitting the app on all three OS with active sessions. They should always cancel sessions and let you logout or close the app. Confirm on windows there are no stray processes running from electron. 

## Checklist

## PCI review checklist

### DIFF
--- a/ui/desktop/electron-app/src/models/session.js
+++ b/ui/desktop/electron-app/src/models/session.js
@@ -3,10 +3,11 @@
  * SPDX-License-Identifier: BUSL-1.1
  */
 
-const treeKill = require('tree-kill');
 const sanitizer = require('../utils/sanitizer.js');
-const { isWindows } = require('../helpers/platform.js');
-const { spawnAsyncJSONPromise } = require('../helpers/spawn-promise.js');
+const {
+  spawnAsyncJSONPromise,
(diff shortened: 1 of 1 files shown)


priya-patel04 (approved):
Tested on Mac and Windows. Works as expected 🙌 Thank you for the fix!

PR #3004, opened 2025-09-16
fix: 🐛 max depth error for active sessions filter in targets

lisbet-alvarez:
# Description

The max expression depth limit for sqlite is 1000 therefore when there are >=1k sessions then the targets page errors out when using the "Active sessions" filter. This PR introduces a temporary fix to handle this scenario by using "IN" and "NOT IN" sqlite operators instead of chaining over 1k "OR" clauses. However, this only increases the limit to around 32k targets with/without sessions since there is a maximum number of parameters that can be in a single SQL query. Note: It is very unlikely for users to hit this limit, either way a future refactor will aim to optimize our queries so this limit won't be an issue.

:tickets: [Jira ticket](https://hashicorp.atlassian.net/browse/ICU-17679)

## How to Test

Go to targets list view (using [this cluster](http://boundary-alb-feupcvur-450420035.us-east-1.elb.amazonaws.com:9200), project0) and validate "Active Sessions" filter does not error out for targets w/ less than 32k sessions.

## Checklist

## PCI review checklist

### DIFF
--- a/addons/api/addon/utils/sqlite-query.js
+++ b/addons/api/addon/utils/sqlite-query.js
@@ -8,12 +8,6 @@ import { searchTables } from 'api/services/sqlite';
 import { typeOf } from '@ember/utils';
 import { underscore } from '@ember/string';
 
-// Maximum expression tree depth is 1000 so
-// we limit the number of expressions in a query.
-// See "Maximum Depth Of An Expression Tree" in
-// https://www.sqlite.org/limits.html
-const MAX_EXPR_NUM = 999;
-
 /**
  * Takes a POJO representing a filter query and builds a SQL query.
  *
@@ -86,7 +80,7 @@ function addFilterConditions({ filters, parameters, conditions }) {
       .flatMap((item) => Object.keys(item))
       .every((op) => op === firstOperator);
     if (
-      filterValueArray.length > MAX_EXPR_NUM &&
+      filterValueArray.length > 1 &&
       (firstOperator === 'equals' || firstOperator === 'notEquals') &&
       allOperatorsEqual
     ) {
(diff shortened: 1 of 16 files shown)


addons/api/addon/utils/sqlite-query.js
```diff
@@ -75,6 +81,28 @@ function addFilterConditions({ filters, parameters, conditions }) {
       continue;
     }
 
+    const firstOperator = Object.keys(filterValueArray[0])[0];
+    const allOperatorsEqual = filterValueArray
+      .flatMap((item) => Object.keys(item))
+      .every((op) => op === firstOperator);
+    if (
+      filterValueArray.length > MAX_EXPR_NUM &&
```
lisbet-alvarez, line 89: Ended up adding the limit back in to reduce the number of filters that will need to use IN/NOT IN operators since this part doesn't make use of '?' for parameters.

ZedLi: Why can't we parametrize it? It should be possible like in our delete statements
https://github.com/hashicorp/boundary-ui/blob/4c7ea161cd6981530cd320f1b19f3b559a47e299/addons/api/addon/workers/utils/schema.js#L474

addons/api/addon/utils/sqlite-query.js
```diff
@@ -75,6 +75,31 @@ function addFilterConditions({ filters, parameters, conditions }) {
       continue;
     }
 
+    const firstOperator = Object.keys(filterValueArray[0])[0];
```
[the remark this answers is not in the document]
lisbet-alvarez: great idea!

ZedLi (approved):

priya-patel04 (approved):
Thank you for the fix!

hashicc (approved):
Nice fix! Helpful to see the change flow through to the tests

PR #3006, opened 2025-09-17
fix: 🐛 Fix schema for sqlite inserts

ZedLi:
# Description

Currently there's a bug where after deleting and inserting resources into sqlite, sometimes the search will have undefined behavior and either show the wrong rows or no rows. This is because the triggers to maintain the FTS5 tables were missing the rowid on inserts from the original table and instead using the automatic autoincremented rowids which happened to match the original as long as rows weren't deleted and re-inserted.

If they were deletions and insertions, what ends up happening is sometimes the rowids might get [re-used](https://www.sqlite.org/autoinc.html#:~:text=If%20you%20ever%20delete%20rows%20or%20if%20you%20ever%20create%20a%20row%20with%20the%20maximum%20possible%20ROWID%2C%20then%20ROWIDs%20from%20previously%20deleted%20rows%20might%20be%20reused%20when%20creating%20new%20rows%20and%20newly%20created%20ROWIDs%20might%20not%20be%20in%20strictly%20ascending%20order.) and the deletes after may also be [unpredictable](https://www.sqlite.org/fts5.html#the_delete_command:~:text=If%20the%20values%20%22inserted%22%20into%20the%20text%20columns%20as%20part%20of%20a%20%27delete%27%20command%20are%20not%20the%20same%20as%20those%20currently%20stored%20within%20the%20table%2C%20the%20results%20may%20be%20unpredictable.) and not actually delete. 

## How to Test
This is most obvious on the scopes resource (but can be reproducible with any resource) due to the fact adding or deleting a scope causes an invalid list token and we clear out the table and re-fetch.

On `main`, goto scopes and try adding a new scope. This should cause a list token invalidation and re-fetch of scopes. Try searching for different IDs and you should see that you get the wrong results or no results back.

On this branch, delete your old sqlite DB by clearing the cache (since the schema was updated and I didn't increment the version yet since we haven't released) and try the same thing as before. It should now return the right search results. 

## Checklist

## PCI review checklist

### DIFF
--- a/addons/api/addon/workers/utils/schema.js
+++ b/addons/api/addon/workers/utils/schema.js
@@ -44,9 +44,9 @@ CREATE VIRTUAL TABLE IF NOT EXISTS target_fts USING fts5(
 -- which replaces the row and doesn't execute an UPDATE trigger.
 CREATE TRIGGER IF NOT EXISTS target_ai AFTER INSERT ON target BEGIN
     INSERT INTO target_fts(
-        id, type, name, description, address, scope_id, created_time
+        rowid, id, type, name, description, address, scope_id, created_time
     ) VALUES (
-        new.id, new.type, new.name, new.description, new.address, new.scope_id, new.created_time
+        new.rowid, new.id, new.type, new.name, new.description, new.address, new.scope_id, new.created_time
     );
 END;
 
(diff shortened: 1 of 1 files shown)


lisbet-alvarez (approved):
Great catch! Tested locally and works

hashicc (approved):
I was able to test the before/after. Thanks for the fix!

hashicc (approved):
🚀 Working flawlessly!

lisbet-alvarez (approved):
Tested new change!

PR #3010, opened 2025-09-26
Use joins with filters

ZedLi:
# Description
Added support for joins with filters. Added an example using the targets table where I join with sessions to grab the active sessions instead of passing in a list of target IDs from the sessions. This removes the need to grab and return _all_ sessions which can be very slow (and even cause browser errors) with big enough data sets and is where the bulk of the time loading is spent if there's a lot of data.

## How to Test
Try using the active sessions filter on the targets page. 

## Checklist

## PCI review checklist


priya-patel04 (approved):
Nice improvement using JOIN! 🙌

PR #3021, opened 2025-10-01
feat: 🎸 Add ability to search other resources as well

ZedLi:
# Description
For pages that have tables that display data from associates resources, we currently retrieve each associated resource 1 by 1 for each shown row to get the associated data and are unable to search columns that source data that doesn't exist in the primary table.

This PR adds the ability to search across multiple tables with joins. This allows us to search across multiple tables in the case where we have columns that use data from different resources. This does however force a dependency on us to load all the associated resources first as compared to making individual `GET` requests for each associated resource. But this is likely better than making all those GET requests at once (especially if they choose 50 items per page). 

The main downside is there can be a noticeable increase in latency for the searches depending on the number of resources. 
Example with 400k aliases + 400k targets and searching between all the resources in _both_ tables. 

Before:

https://github.com/user-attachments/assets/abd00d58-97e2-492d-8cb7-317e9eaefecf

After:

https://github.com/user-attachments/assets/18321dd6-1dbd-4d19-bf28-1d58f3abd55c

## How to Test
Go to the aliases page and try searching target names or alias IDs. It should search across both resources. 

## Checklist

## PCI review checklist

### DIFF
--- a/ui/admin/app/routes/scopes/scope/aliases/index.js
+++ b/ui/admin/app/routes/scopes/scope/aliases/index.js
@@ -83,7 +83,6 @@ export default class ScopesScopeAliasesIndexRoute extends Route {
           relatedSearches: [
             {
               resource: 'target',
-              text: search,
               fields: ['name'],
               join: {
                 joinFrom: 'destination_id',
--- a/addons/api/addon/workers/utils/schema.js
+++ b/addons/api/addon/workers/utils/schema.js
@@ -25,8 +25,7 @@ const createTargetTables = `
 );
 CREATE INDEX IF NOT EXISTS idx_target_scope_id_created_time ON target(scope_id, created_time DESC);
 
--- Create a contentless FTS table as we will only use the rowids.
--- Note that this only creates the FTS index and cannot reference the target content
+-- Create a content FTS table that references the original table
 CREATE VIRTUAL TABLE IF NOT EXISTS target_fts USING fts5(
     id,
     type,


DhariniJeeva (approved):
looks good!

addons/api/tests/unit/utils/sqlite-query-test.js
```diff
@@ -419,6 +419,64 @@ module('Unit | Utility | sqlite-query', function (hooks) {
...
+                join: {
+                  joinOn: 'target_id',
+                },
+              },
+            ],
+          },
+        },
+        expectedSql: `
+          SELECT * FROM "target"
+          WHERE "target".rowid IN (SELECT rowid FROM target_fts WHERE target_fts MATCH ?
```
hashicc: Does this example make sense since `rowid` doesn't exist on the `target` table, or should it be `target_fts`?

hashicc: Oh, TIL! Yes, I was looking at the schemas.
I agree. `rowid` makes perfect sense as a default

addons/api/addon/utils/sqlite-query.js
```diff
@@ -211,25 +211,69 @@ function addSearchConditions({
...
+
+    return;
+  }
+
   // Use a subquery to match against the FTS table with rowids as SQLite is
   // much more efficient with FTS queries when using rowids or MATCH (or both).
   // We could have also used a join here but a subquery is simpler.
   conditions.push(
-    `"${tableName}".rowid IN (SELECT rowid FROM ${tableName}_fts WHERE ${tableName}_fts MATCH ?)`,
+    `"${tableName}".${search?.select ?? 'rowid'} IN (${searchSql})`,
```
hashicc:
```suggestion
    `"${tableName}".${searchSelect} IN (${searchSql})`,
```

ui/admin/app/routes/scopes/scope/aliases/index.js
```diff
@@ -79,28 +78,69 @@ export default class ScopesScopeAliasesIndexRoute extends Route {
           direction: sortDirection,
         };
 
+        const searchOptions = {
+          text: search,
+          relatedSearches: [
+            {
+              resource: 'target',
+              text: search,
```
lisbet-alvarez: I think this `text` field can be removed since its not used in the query util since its also defined on line 82.

lisbet-alvarez (approved):
This is a great improvement/feature!! Just left a couple suggestions.

PR #3026, opened 2025-10-07
chore: 🤖 update tmp package to 0.2.5

cameronperera:
# Description

## Checklist

## PCI review checklist

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -8747,8 +8747,8 @@ packages:
     resolution: {integrity: sha512-J7Z2K08jbGcdA1kkQpJSqLF6T0tdQqpR2pnSUXsIchbPdTI9v3e85cLW0d6WDhwuAleOV71j2xWs8qMPfK7nKw==}
     engines: {node: '>=6'}
 
-  tmp@0.2.3:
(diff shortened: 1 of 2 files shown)


hashicc:
Looks like the test release failed on signing without the right format for the tag.

If you check with `pnpm why -r tmp` it looks like we still have plenty of `tmp` versions that are left not upgraded. I think we might need to add a manual resolution to force a later version in the root `package.json` and one in the `electron-app/package.json`. 

It looks like there weren't any breaking changes in [changelog](https://github.com/raszi/node-tmp/blob/master/CHANGELOG.md) and the [readme has a mention of breaking changes](https://github.com/raszi/node-tmp/tree/master?tab=readme-ov-file#an-important-note-on-previously-undocumented-breaking-changes). The breaking changes seem reasonable and likely to not affect us but with the number it's hard to be sure. I think what you did with running the builds and one for the desktop app is the way to go. Given that `tmp` is a node package it can only really be running within our builds for admin, desktop, or possibly as part of the electron-app but this doesn't look like it's an issue:
```
devDependencies:
@electron-forge/maker-deb 7.8.1
└─┬ electron-installer-debian 3.2.0
  └─┬ electron-installer-common 0.10.4
    └─┬ tmp-promise 3.0.3
      └── tmp 0.2.5
```

hashicc (approved):

PR #3031, opened 2025-10-17
chore: 🤖 add ipc handlers for rdp

DhariniJeeva:
# Description

This PR adds handlers to support rdp client launch and select preferred clients

jira- https://hashicorp.atlassian.net/browse/ICU-17876

## Checklist

## PCI review checklist

### DIFF
--- a/ui/desktop/electron-app/src/services/rdp-client-manager.js
+++ b/ui/desktop/electron-app/src/services/rdp-client-manager.js
@@ -138,7 +138,7 @@ class RdpClientManager {
    */
   async launchRdpClient(sessionId, sessionManager) {
     // Get session object from session manager
-    const session = sessionManager.getSessionById?.(sessionId);
+    const session = sessionManager.getSessionById(sessionId);
 
     if (!session) {
       return;


ui/desktop/electron-app/src/services/rdp-client-manager.js
```diff
@@ -0,0 +1,205 @@
...
+const fs = require('fs');
+const which = require('which');
+const { isMac, isWindows } = require('../helpers/platform.js');
+const store = require('./electron-store-manager');
+const log = require('electron-log/main');
+
+// RDP Client Configuration
+const RDP_CLIENTS = [
+  {
+    name: 'Microsoft Remote Desktop',
```
priya-patel04: Do we need the `name` property here ? May be we can move the name to the translations and consumers can use the `value` property as key and get the corresponding translation.

ui/desktop/electron-app/src/services/rdp-client-manager.js
```diff
@@ -0,0 +1,205 @@
...
+   * @returns {Promise<string>} Preferred RDP client value
+   */
+  async getPreferredRdpClient() {
+    try {
+      let preferredClient = store.get('preferredRdpClient');
+
+      if (!preferredClient) {
+        // Auto-detect and set the best available client
+        preferredClient = await this.getBestDefaultRdpClient();
+        if (preferredClient !== 'none') {
```
priya-patel04: Wondering if we need this check i.e preferredClient !== 'none' ? Asking because if a user selects "none" option as their preferred client, then in the below `setPreferredRdpClient` method the store will end up setting preferredRdpClient value as "none".

ui/desktop/electron-app/src/services/rdp-client-manager.js
DhariniJeeva: good point, we don't need the `name` property here. removed it!

ui/desktop/electron-app/src/services/rdp-client-manager.js
```diff
@@ -0,0 +1,200 @@
...
+
+  /**
+   * Sets the user's preferred RDP client
+   * @param {string} preferredClient - The RDP client value to set as preferred
+   * @returns {boolean} True if successfully set
+   */
+  setPreferredRdpClient(preferredClient) {
+    try {
+      if (!preferredClient) {
+        store.delete('preferredRdpClient');
```
priya-patel04: Do we need to delete the `preferredRdpClient` from the store if it is not passed ? I think we can just set to `none` instead ?

ui/desktop/electron-app/src/services/rdp-client-manager.js
```diff
@@ -0,0 +1,177 @@
...
+   * Gets all available RDP clients on the current system
+   * @returns {Promise<Array>} Array of available RDP clients with value
+   */
+  async getAvailableRdpClients() {
+    const available = [];
+    for (const client of RDP_CLIENTS) {
+      if (await client.isAvailable()) {
+        available.push({
+          value: client.value,
+        });
```
ZedLi: How come we return an object that just has value instead of an array of strings?

ui/desktop/electron-app/src/services/rdp-client-manager.js
```diff
@@ -0,0 +1,177 @@
...
+   */
+  async launchRdpConnection(address, port) {
+    if (isWindows()) {
+      // Launch Windows mstsc
+      const mstscArgs = [`/v:${address}:${port}`];
+      const child = spawn('mstsc', mstscArgs, {
+        detached: true,
+        stdio: 'ignore',
+      });
+      child.unref();
```
ZedLi: Why do we need to `unref` this?

ui/desktop/electron-app/src/services/rdp-client-manager.js
```diff
@@ -0,0 +1,177 @@
...
+  /**
+   * Launches RDP connection with the specified address and port
+   * @param {string} address - Target address
+   * @param {number} port - Target port
+   */
+  async launchRdpConnection(address, port) {
+    if (isWindows()) {
+      // Launch Windows mstsc
+      const mstscArgs = [`/v:${address}:${port}`];
+      const child = spawn('mstsc', mstscArgs, {
```
ZedLi: Any reason we don't use one of our own `spawn` helpers?

What's managing this spawned process? As it is, this is a resource leak.

ui/desktop/electron-app/src/services/rdp-client-manager.js
```diff
@@ -0,0 +1,177 @@
...
+        detached: true,
+        stdio: 'ignore',
```
ZedLi: Why do we want these processes detached?

ui/desktop/electron-app/src/services/rdp-client-manager.js
```diff
@@ -0,0 +1,177 @@
...
+
+      if (!proxyDetails) {
+        return;
+      }
+
+      const { address, port } = proxyDetails;
+
+      // Launch RDP connection
+      await this.launchRdpConnection(address, port);
+    } catch (error) {
```
ZedLi: Do we want to catch this or actually throw it back up to the user?

DhariniJeeva: I think throwing it back to the user might be ideal here, will update this

ui/desktop/electron-app/src/services/rdp-client-manager.js
```diff
@@ -0,0 +1,168 @@
...
+
+  /**
+   * Launches RDP client using session ID
+   * Retrieves session object from session manager and launches appropriate RDP client
+   * @param {string} sessionId - The session ID to get session for
+   * @param {Object} sessionManager - Session manager instance to get session from
+   */
+  async launchRdpClient(sessionId, sessionManager) {
+    // Get session object from session manager
+    const session = sessionManager.getSessionById?.(sessionId);
```
ZedLi:
```suggestion
    const session = sessionManager.getSessionById(sessionId);
```

ui/desktop/electron-app/src/services/rdp-client-manager.js
```diff
@@ -0,0 +1,168 @@
...
+
+    if (!session) {
```
hashicc: This might not happen in practice but might be a good safe guard to double-check the `isRunning` status of the session before continuing: either returning with this guard or throwing an error?

PR #3032, opened 2025-10-17
Refactor hds side nav to hds app side nav and header

lisbet-alvarez:
# Description

Refactor admin & desktop UIs to use [hds app side nav](https://helios.hashicorp.design/components/app-side-nav) and [app header](https://helios.hashicorp.design/components/app-header) components. This PR also removes components + styles that are no longer needed as a result of the refactor. 

## How to Test

1. Please run a11y tests for both light and dark mode locally and verify everything passes. The only tests that currently fail are existing ones with missing ember-a11y-ignores (handled in a [separate PR](https://github.com/hashicorp/boundary-ui/pull/3159)).
2. Run e2e tests locally and verify nothing is failing. In CI desktop e2e tests are failing due to changes to admin ui, but should pass once this PR is merged in. Only one flaky admin ui (CE) is failing (unrelated to this PR).

## Checklist

## PCI review checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -53,7 +53,8 @@
       "ember-cli-babel>@babel/runtime": "7.27.6",
       "ember-try-config>package-json": "^10.0.1",
       "ember-loading>ember-concurrency": "^4.0.0",
-      "tmp": "^0.2.5"
+      "tmp": "^0.2.5",
+      "tar": "^7.5.8"
     },
     "ignoredBuiltDependencies": [
       "@parcel/watcher",
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -15,6 +15,7 @@ overrides:
   ember-try-config>package-json: ^10.0.1
   ember-loading>ember-concurrency: ^4.0.0
   tmp: ^0.2.5
+  tar: ^7.5.8
 
 patchedDependencies:
   ember-loading:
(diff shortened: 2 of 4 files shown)


ui/desktop/app/templates/application.hbs
```diff
@@ -74,15 +63,49 @@
...
+  <Frame.Sidebar>
+    <Hds::AppSideNav>
+      <Hds::AppSideNav::Portal::Target />
+    </Hds::AppSideNav>
+  </Frame.Sidebar>
 
-</Rose::Layout::Global>
+  <Frame.Main>
+    {{outlet}}
+  </Frame.Main>
```
lisbet-alvarez: since a `<main>` element is added using hds app-frame then we no longer need instances of `main` in our templates. That is why those are removed in this PR.

addons/rose/addon/styles/hds/themes/dark-mode/index.scss
```diff
@@ -132,16 +132,94 @@
...
+    /* stylelint-disable selector-class-pattern */
+    .hds-app-side-nav__overlay {
+      background-color: var(--token-color-palette-neutral-50);
+    }
+
+    .hds-app-side-nav__list-title {
+      color: var(--token-color-palette-neutral-300);
+    }
+
+    .hds-app-side-nav__list-item-link.active {
```
lisbet-alvarez: when a side nav list item is active in light mode its a dark blue color but there are color contrast a11y violations if same color is used in dark mode. So used a blue color token that has good contrast.

addons/core/addon/components/scope-picker/index.hbs
ZedLi: It seems like if there's more than 5 scopes, it'll error out here as we don't have a `scopes.scope.scopes` route (as that's admin ui only)

addons/rose/addon/styles/hds/overrides.scss
ZedLi: Can we remove this override now to confirm migrating to the new side nav "fixes" the issue

addons/core/addon/components/scope-picker/index.hbs
```diff
@@ -4,18 +4,18 @@
...
+  data-test-header-scope-dropdown
+  ...attributes
   as |D|
 >
   {{#let this.currentScope as |scope|}}
     <D.ToggleButton
       @text={{scope.name}}
       @icon={{scope.icon}}
-      class='scope-picker__toggle-button'
+      class='dropdown_toggle_color_override'
```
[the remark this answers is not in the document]
lisbet-alvarez: ah yes, u are correct!

addons/core/addon/components/scope-picker/index.hbs
lisbet-alvarez: Great catch!

ui/desktop/app/templates/application.hbs
```diff
@@ -25,10 +25,45 @@
 
       <:globalActions>
         {{#if this.session.isAuthenticated}}
-          <ScopePicker
```
ZedLi: Could we have still used the `ScopePicker` but just also pass in a flag on whether to use the truncated org list? If they diverge too much this is also fine.

The only other thing is we may want to enforce a maximum height/scroll bar since it can still overflow and we'd have to still scroll the entire window down to read them.

Image

lisbet-alvarez: yeah it ultimately felt like the dropdowns were more different than similar due to one displaying truncated lists, orgs + indented projects, and "View all" links. So thought it would probably be easier to read as its own dropdown. 
Good catch on the overflow issue!
In regards to the failing e2e test, it is failing on the `admin/pages/login.js:22`  part of global setup to save autheticated state for reuse in desktop e2e tests. My understanding is that its failing due to desktop e2e tests using boundary binary that doesn't include these changes. Which is also why global setup does pass for admin e2e tests.

PR #3033, opened 2025-10-20
Unload expired sessions from store

lisbet-alvarez:
# Description

Expired sessions get cleaned up after a while and the ember data store might still have them cached (unless there was a hard refresh) which will lead to inaccurate display of "Active sessions" on a target list page because we use `peekAll` to display active sessions. This is only an issue with the ember data store, the sqlite db correctly removes terminated sessions using `removed_ids` field from the API when those sessions get cleaned up. Therefore, this PR unloads sessions  in the targets route model hook before making API queries so that any removed/terminated sessions get cleaned up from our ember data cache.

Issue seen in Admin UI:

https://github.com/user-attachments/assets/63e88056-04bc-4835-aee1-45d0f65d6e72

Issue seen in Desktop UI:

https://github.com/user-attachments/assets/41772f48-3622-4cc2-b82a-0b42ffbd89ee

## How to Test

1. Start a short sessions (quick expiration) by connecting to a target.
2. Navigate to targets list page, validate that "Active Sessions" correctly shows on the target.
3. Navigate to sessions list page, and wait until "No Sessions Available" because then we know the session got removed & terminated through expiration.
4. Navigate to targets list page, validate that "Active Sessions" column shows nothing.

## Checklist

## PCI review checklist

### DIFF
--- a/addons/core/addon/components/filter-tags/index.js
+++ b/addons/core/addon/components/filter-tags/index.js
@@ -17,16 +17,27 @@ export default class FilterTagsIndexComponent extends Component {
 
   get filters() {
     const { allFilters, selectedFilters } = this.args.filters;
-    return Object.entries(allFilters).flatMap(([key, value]) => {
-      assert(`Tags must be an array for key ${key}`, Array.isArray(value));
-      const paramsSet = new Set(selectedFilters[key]);
-      const filters = value.filter((item) => paramsSet.has(item.id));
+    return Object.entries(allFilters).flatMap(([key, values]) => {
+      assert(`Tags must be an array for key ${key}`, Array.isArray(values));
+      const uniqueSelectedFilters = [...new Set(selectedFilters[key])];
+      if (!uniqueSelectedFilters.length) {
+        return [];
+      }
 
-      return filters.map((item) => ({
-        id: item.id,
-        name: item.name,
-        type: key,
-      }));
+      const valueMap = new Map(values.map((value) => [value.id, value.name]));
+
+      return uniqueSelectedFilters.filter(Boolean).map((item) => {
+        const filter = {
+          id: item,
+          type: key,
+        };
+
+        if (valueMap.has(item)) {
+          filter.name = valueMap.get(item);
+        }
+
+        return filter;
+      });
     });
   }
 
--- a/addons/core/tests/integration/components/filter-tags/index-test.js
+++ b/addons/core/tests/integration/components/filter-tags/index-test.js
@@ -22,4 +22,14 @@ module('Integration | Component | filter-tags/index', function (hooks) {
 
     assert.dom('.hds-tag__text').hasText('Project 1');
   });
+
+  test('it renders id if allFilter does not have corresponding value', async function (assert) {
+    this.set('filter', {
+      allFilters: { scopes: [{ id: '2', name: 'Project 2' }] },
+      selectedFilters: { scopes: ['1'] },
+    });
+    await render(hbs`<FilterTags @filters={{this.filter}} />`);
+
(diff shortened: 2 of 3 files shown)


ui/admin/app/routes/scopes/scope/targets/index.js
```diff
@@ -147,7 +151,7 @@ export default class ScopesScopeTargetsIndexRoute extends Route {
               },
             },
           },
-          { pushToStore: true, peekDb: true },
+          { peekDb: true },
```
ZedLi: Wouldn't we still need `pushToStore` since the original issue was that we weren't updating the sessions properly in the store? If we're unloading now, this `store.query` is unneeded now right?

lisbet-alvarez: Removal of `pushToStore` was mainly because its  set to `true` by [default](https://github.com/hashicorp/boundary-ui/blob/25f015b179bfa9f017e3b7e730d1adfe03a91616/addons/api/addon/handlers/sqlite-handler.js#L35). Ya ur right that extra query is uneeded now.

ui/admin/app/routes/scopes/scope/targets/index.js
```diff
@@ -92,8 +92,12 @@ export default class ScopesScopeTargetsIndexRoute extends Route {
       });
 
       if (this.can.can('list model', scope, { collection: 'sessions' })) {
+        // Before querying sessions, unload sessions currently stored in ember data store
+        // so that we remove any expired sessions that still might be cached.
+        this.store.unloadAll('session');
```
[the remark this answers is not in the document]
lisbet-alvarez: Yup that is correct, we don't properly clean up the data store. I imagine ur right in this being an issue for all resources  where it doesn't get directly removed using `destroyRecord`. Ill update to perform the cleanup in the handler instead.

addons/api/addon/handlers/sqlite-handler.js
```diff
@@ -233,6 +228,19 @@ export default class SqliteHandler {
 
     const { data: payloadData } = normalizedPayload;
 
+    // Remove any records from the DB if the API indicates they've been deleted
+    // Additionally remove deleted records from ember data store for symmetry
+    if (payload.removed_ids?.length > 0) {
+      await this.sqlite.deleteResource(type, payload.removed_ids);
+
+      payload.removed_ids.forEach((id) => {
```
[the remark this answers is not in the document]
lisbet-alvarez: ah yup good point

ZedLi (dismissed):
Works as expected, thanks for this fix!

cameronperera (approved):

priya-patel04 (approved):
Thanks for the fix!

ui/desktop/app/routes/scopes/scope/projects/targets/index.js
```diff
@@ -183,6 +185,27 @@ export default class ScopesScopeProjectsTargetsIndexRoute extends Route {
...
+        );
+        const removedSessions = storedSessionIds.difference(
+          allAssociatedSessionIds,
+        );
+        removedSessions.forEach((id) => {
+          const record = this.store.peekRecord('session', id);
+          this.store.unloadRecord(record);
+        });
+      }
+
```
lisbet-alvarez: I removed the `unloadAll('session')` and opted to removing only particular records for the loaded page (max 50 targets per page) instead since unloading all sessions would also unload attached proxy details for each session even if its still active/pending.

hashicc (approved):
Good catch, thanks for the changes!

ZedLi (approved):

PR #3034, opened 2025-10-24
feat: 🎸 preferred clients section on settings page

priya-patel04:
# Description
Adds a new Preferred Clients section on Settings Page.

 :tickets: [Jira ticket](https://hashicorp.atlassian.net/browse/ICU-17885)

### Mac:

https://github.com/user-attachments/assets/a2ed7940-90e5-4f15-acf8-d658cc276f2d

Screenshot 2025-10-27 at 1 21 26 PM

#### None detected:

https://github.com/user-attachments/assets/7e6ebe38-8f88-40d7-ae17-e64ed58946fe

### Windows:

https://github.com/user-attachments/assets/47c92850-c2c1-4362-9d1e-6944f7bb77ff

Screenshot 2025-10-27 at 1 25 33 PM

#### None detected:

https://github.com/user-attachments/assets/9750183b-9630-45fc-ab2d-b82d41bc20ce

## How to Test

1.  Navigate to Settings page and verify a new Preferred Clients section is rendered
      - Mac: Windows App should be selected by default as your preferred client
      - Windows: Mstc should be selected by default as your preferred client
2. Test updating your preferred client to None. Reopen the app and verify None selection persists.
3. Verfiy electron stores/update preferred client selection in the config:
      - Mac: /Library/Application Support/Boundary/config.json
      - Windows: \AppData\Roaming\Boundary\config.json

## Checklist

## PCI review checklist

### DIFF
--- a/ui/desktop/tests/acceptance/projects/targets/target-test.js
+++ b/ui/desktop/tests/acceptance/projects/targets/target-test.js
@@ -162,8 +162,8 @@ module('Acceptance | projects | targets | target', function (hooks) {
     this.stubCacheDaemonSearch('sessions', 'targets', 'aliases', 'sessions');
 
     // mock RDP service calls
-    let rdpService = this.owner.lookup('service:rdp');
-    sinon.stub(rdpService, 'initialize').resolves();
+    this.rdpService = this.owner.lookup('service:rdp');
+    sinon.stub(this.rdpService, 'initialize').resolves();
   });
 
   test('user can connect to a target with an address', async function (assert) {


hashicc (changes requested):
This looks great! Would this work already with a test desktop client build? I can give that a try tomorrow if so.

My feedback is:
* another test (or test assertion in the existing test) to check that the data is responding from the `change` event
* try and assign these known set of values to constants that are exported from the best source module (maybe `ui/desktop/app/services/rdp.js`)

ui/desktop/app/components/settings-card/preferred-clients/index.hbs
```diff
@@ -0,0 +1,60 @@
...
+        (hash
+          label=(t 'settings.preferred-clients.table.header.protocol')
+          width='50%'
+        )
+        (hash
+          label=(t 'settings.preferred-clients.table.header.client') width='50%'
+        )
+      }}
+    >
+      <:body as |B|>
```
DhariniJeeva: I noticed that `none detected. we recommend windows app` option from figma is missing. should we add support for that option here?

addons/api/mirage/scenarios/ipc.js
```diff
@@ -248,6 +248,14 @@ export default function initializeMockIPC(server, config) {
     resumeClientAgent() {}
     hasRunningSessions() {}
     stopAll() {}
+    getRdpClients() {
+      return ['windows-app', 'none', 'mstsc'];
```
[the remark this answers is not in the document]
hashicc: It would be a heavier lift to try and make it sharable between ember and node right now as you've discovered. We also generally have kept the electron code as a separate pnpm workspace so logically it's probably best to keep them separated. Any defined constants we'll have to keep duplicated: once in the desktop (for this PR) and then again within `electron-app`

ui/desktop/app/components/settings-card/preferred-clients/index.hbs
```diff
@@ -0,0 +1,80 @@
...
+              <Hds::Text::Body>
+                {{t
+                  'settings.preferred-clients.table.data.clients.none-detected'
+                }}
+                <Hds::Link::Inline
+                  @href={{this.rdp.recommendedRdpClient.link}}
+                  @isHrefExternal={{true}}
+                >
+                  {{t
+                    (concat
```
hashicc: This probably isn't as big of a deal since we haven't actually translated our app to anything except english, but the interpolated variable for a translation should be part of the translation: `None detected. We recommend {rdpClientAppName}`, this allows the translation to move the `{rdpClientAppName}` as needed depending on the language

addons/core/translations/en-us.yaml
```diff
@@ -141,6 +141,21 @@ settings:
...
+      header:
+        protocol: Protocol
+        client: Client
+      data:
+        protocols:
+          windows-rdp: Windows RDP
+        clients:
+          none: None
+          mstsc: Remote Desktop Connection (mstsc)
+          windows-app: Windows App
```
hashicc: Maybe you've already done this, but it might be good to double-check with product and/or our documentation site that we're consistent with the naming of how we refer to these rdp clients

ui/desktop/tests/integration/components/settings-card/preferred-clients-test.js
```diff
@@ -0,0 +1,89 @@
...
+      await render(hbs`<SettingsCard::PreferredClients />`);
+
+      assert
+        .dom('.hds-table tbody tr td:first-child')
+        .hasText('Windows RDP', 'Protocol is rendered');
+      assert
+        .dom('select option:checked')
+        .hasText('Windows App', 'Preferred client is selected');
+
+      let select = find('select');
```
hashicc: We should use the test helpers to drive the UI to manage the select instead of accessing the DOM directly and setting its value. Try out the [`select` helper](https://github.com/emberjs/ember-test-helpers/blob/master/API.md#select)

ui/desktop/app/services/rdp.js
```diff
@@ -52,19 +69,54 @@ export default class RdpService extends Service {
...
+      }
       return this.rdpClients;
     }
     try {
       this.rdpClients = await this.ipc.invoke('getRdpClients');
+      if (
+        this.rdpClients.length === 1 &&
+        this.rdpClients[0] === RDP_CLIENT_NONE
+      ) {
+        await this.setRecommendedRdpClient();
```
hashicc: Would this get called? since line 71 is already checking for `(this.rdpClients.length > 0` and then returning on line 80?

hashicc (changes requested):
I think the logic around getting the clients could be less coupled in how it's set up. Also if the methods have several conditions in a service and they're not exercised by the rendering or acceptance tests it would be good to have unit tests around them. 

I think there are some cases here that might need a few acceptance since they rely on certain state being set in the `beforeModel` hook and it would be good to ensure that the data coming from ipc is set up on the rdp service and then seen on the settings. That could be done up in a follow-up PR though.

ui/desktop/app/components/settings-card/preferred-clients/index.hbs
priya-patel04: I was debating to keep the rdp client name as variable, as we also have a link associated with it. But we can use `htmlSafe` and pass link variable as well. Will update it. Thanks!

ui/desktop/app/services/rdp.js
priya-patel04: `setRecommendedRdpClient` would get called once in the application route when we call `getRdpClients` as initially the this.rdpClients is empty. And once the value is cached, we simply return this.rdpClients on line 80.

But I do agree it make sense to set recommended rdp client explicity and ideally initialize all the required data in one go.

DhariniJeeva (approved):

PR #3050, opened 2025-11-04
fix: 🐛 Fix filters sometimes not being displayed

ZedLi:
# Description
One of the changes that was made in https://github.com/hashicorp/boundary-ui/pull/2981 was not retrieving all resources for the dropdown filters. This had an unintended consequence when displaying the selected filters as it's possible for the selected filter to not be in the list of `allFilters`. 

There's two main scenarios this happens (the prerequisite is having more filter options than we display which is 250):
1. When filtering down the dropdown filters which means it might not be in the allFilters options. This was fixed by keeping track of all the searched options to be able to display.
2. A user uses a query string or bookmarked a URL and goes to it directly. If the option isn't in the initial load, we won't have the actual option. In this case, I changed it to at least just display the ID instead of not showing up at all. Otherwise I'd have to load these selected options separately from the initial load which doesn't seem worth for this edge case.

Before:

https://github.com/user-attachments/assets/0bcd4542-8a06-4462-be4d-fd5dd080be96

After:

https://github.com/user-attachments/assets/a598487f-56ec-45fc-845d-b5b5e06ced1d

## How to Test
Use this [cluster](http://boundary-alb-voqbvgzg-1131911519.us-east-1.elb.amazonaws.com:9200). Try applying different searches for the target dropdown and confirm that the filter tags always appear.

## Checklist

## PCI review checklist

### DIFF
--- a/.github/workflows/a11y-tests.yaml
+++ b/.github/workflows/a11y-tests.yaml
@@ -17,10 +17,10 @@ jobs:
         with:
           install-dependencies: true
       - uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5.0.0
-      - uses: pnpm/action-setup@a7487c7e89a18df4991f7f222e4898a00d66ddda #v4.1.0
+      - uses: pnpm/action-setup@41ff72655975bd51cab0327fa583b6e92b6d3061 #v4.2.0
         with:
           run_install: false
-      - uses: actions/setup-node@a0853c24544627f65ddf259abe73b1d18a591444 # v5.0.0
+      - uses: actions/setup-node@2028fbc5c25fe9cf00d9f06a71cc4710d4507903 # v6.0.0
         with:
           node-version: 20
           cache: 'pnpm'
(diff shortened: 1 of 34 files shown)


ui/admin/app/controllers/scopes/scope/session-recordings/index.js
```diff
@@ -12,7 +12,23 @@ import { restartableTask } from 'ember-concurrency';
 
 class FilterOptions {
   @tracked search;
-  @tracked options = [];
+  @tracked _options = [];
+  #allOptions = new Map();
```
hashicc: I'm wondering if the interface should be: `selectedOptions` and `options`? I'm struggling with the idea that "all" is never really "all" in large filter sets, as the set could be constantly changing as 250 results get loaded based on different search terms and then `options` ends up just being a cache of any previously seen option... At least that's my understanding from reading the code. Or maybe `allOptions` needs a nice comment of what "all" means

hashicc (approved):
It looks good, I have one partial suggestion and one comment for discussion, but nothing blocking. Thanks for looking back over the recent changes for this edge case.

ui/admin/app/controllers/scopes/scope/session-recordings/index.js
ZedLi: So it's actually kind of tough, these are also not the `selectedOptions`, since that's being tracked by the query params separately which I would find even more confusing if it was called `selectedOptions`. I think a comment might be better here because you're right, it's not "all" but it's just keeping track of any previously loaded/seen options from the dropdown.

hashicc: That makes sense, I think some sort of hint would help

addons/core/addon/components/filter-tags/index.js
```diff
@@ -17,16 +17,23 @@ export default class FilterTagsIndexComponent extends Component {
...
+      assert(`Tags must be an array for key ${key}`, Array.isArray(values));
+      const uniqueSelectedFilters = [...new Set(selectedFilters[key])];
+      const valueMap = new Map(values.map((value) => [value.id, value.name]));
 
-      return filters.map((item) => ({
-        id: item.id,
-        name: item.name,
-        type: key,
-      }));
+      return uniqueSelectedFilters.filter(Boolean).map((item) => {
```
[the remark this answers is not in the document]
ZedLi: Yeah good point, it doesn't need need the map if it's empty. I added the early return to handle that

PR #3066, opened 2025-11-13
refactor: 💡 update logic to get preferred rdp client

priya-patel04:
# Description
Minor improvements for  rdp client launch:

1. No client installed initially
**Issue**: After installing the client, the Settings page still shows “None” as the auto detected preferred client.
**Expected**: The newly installed client should be auto detected and set as the preferred client in the UI
**Fix**: Don’t store the preferred client in the Electron store by default. Only save the value after the user has explicitly changed it in the Settings UI.

2. Client installed initially, then uninstalled
**Issue**: After uninstalling the client, the Settings page updates correctly with recommended client info, but the Targets list and Target details still show an “Open” button instead of “Connect.”
**Fix**: When the preferred client is not “None,” add additional check to verify that the client is actually installed

#### No client installed

https://github.com/user-attachments/assets/bceba153-5ae8-497f-8c23-a03fbd21ab7b

#### Client installed initially and then uninstalled:

https://github.com/user-attachments/assets/5f7000b8-5b3e-4534-93d9-71cc2130c1b6

#### No client installed

https://github.com/user-attachments/assets/468c2b00-e656-480c-8f61-d3fcc72ad191

#### Client installed initially and then uninstalled:

https://github.com/user-attachments/assets/909a676a-044b-4a4b-ba29-12cd9516e165

## How to Test
DM for cluster URL
Note: Reset the `~/Library/Application Support/Boundary/config.json` file  prior to testing.

1.  Install Windows App on Mac. Navigate to Settings page and verify Windows App is selected as preferred rdp client.
2.  Navigate to Targets list view and "Open" button should be visible

Uninstall and Install RDP client flow:

1. Uninstall Windows App on Mac. Navigate to Settings page and verify "None detected. We recommend "Windows App" is rendered in preferred client section
2.  Navigate to Targets list view and "Connect" button should be visible.
3. Install Windows App and refresh the window.
4. Navigate to Settings page and verify Windows App is selected as preferred rdp client.
5. Navigate to Targets list view and "Open" button should be visible

## Checklist

## PCI review checklist

### DIFF
--- a/ui/desktop/electron-app/src/services/rdp-client-manager.js
+++ b/ui/desktop/electron-app/src/services/rdp-client-manager.js
@@ -78,40 +78,23 @@ class RdpClientManager {
   }
 
   /**
-   * Gets the best default RDP client (first available that's not 'none')
-   * @returns {Promise<string>} Best RDP client value or 'none'
-   */
-  async getBestDefaultRdpClient() {
-    const availableClients = await this.getAvailableRdpClients();
-    const bestClient = availableClients.find(
-      (client) => client !== RDP_CLIENT_NONE,
-    );
-    return bestClient ?? RDP_CLIENT_NONE;
-  }
-
-  /**
    * Gets the user's preferred RDP client, auto-detecting if not set
    * @returns {Promise<string>} Preferred RDP client value
    */
   async getPreferredRdpClient() {
-    // check user stored preference
-    let preferredClient = store.get('preferredRdpClient');
+    const availableClients = await this.getAvailableRdpClients();
+    const installedClients = availableClients.filter(
+      (client) => client !== RDP_CLIENT_NONE,
+    );
 
-    if (!preferredClient) {
-      // Auto-detect and return the best available client
-      preferredClient = await this.getBestDefaultRdpClient();
-      return preferredClient;
-    }
+    if (installedClients.length === 0) return RDP_CLIENT_NONE;
 
-    // Validate preferred client is still available
-    if (preferredClient !== RDP_CLIENT_NONE) {
-      const availableClient = await this.getBestDefaultRdpClient();
-      if (availableClient === RDP_CLIENT_NONE) {
-        preferredClient = RDP_CLIENT_NONE;
-      }
-    }
-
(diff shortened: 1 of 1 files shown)


ui/desktop/electron-app/src/services/rdp-client-manager.js
```diff
@@ -94,13 +94,23 @@ class RdpClientManager {
...
+    // check user stored preference
     let preferredClient = store.get('preferredRdpClient');
 
     if (!preferredClient) {
-      // Auto-detect and set the best available client
+      // Auto-detect and return the best available client
       preferredClient = await this.getBestDefaultRdpClient();
-      store.set('preferredRdpClient', preferredClient);
+      return preferredClient;
     }
```
hashicc: I think this can be simplified to:
```js
const availableClients = getAvailableRdpClients();
const preferredClient = store.get('preferredRdpClient') ?? await this.getBestDefaultRdpClient();
const isPreferredAvailable = Boolean(availableClients.find(client => client.value === preferredClient));
return isPreferredAvailable ? preferredClient : RDP_CLIENT_NONE;
```

I think this achieves the same thing, let me know if I'm missing something. I think the goal here is to not have to special case `RDP_CLIENT_NONE` so much but just have one preferred client and then check if it's available.

I think `getBestDefaultRdpClient` could also be removed since it has a lot of overlap with finding out the preferred and `getPreferredRdpClient` is the only usage. A version without `getBestDefaultRdpClient` looks like:

```js
const availableClients = await getAvailableRdpClients();
const installedClients = availableClients.filter(client => client !== RDP_CLIENT_NONE);
const preferredClient = store.get('preferredRdpClient') ?? installedClients[0];
const isPreferredAvailable = preferredClient && Boolean(installedClients.find(client => client.value === preferredClient));
return isPreferredAvailable ? preferredClient : RDP_CLIENT_NONE;
```

hashicc (approved):
🚀 Awesome, works great, and overall more lines deleted than added

DhariniJeeva (approved):

PR #3169, opened 2026-02-16
Spike electron web contents view

priya-patel04:
# Description

## Checklist

## PCI review checklist

### DIFF
--- a/ui/desktop/app/components/session/tabs/index.js
+++ b/ui/desktop/app/components/session/tabs/index.js
@@ -5,16 +5,8 @@
 
 import Component from '@glimmer/component';
 import { action } from '@ember/object';
-import { service } from '@ember/service';
-import { Terminal } from '@xterm/xterm';
-import { FitAddon } from '@xterm/addon-fit';
-import { CanvasAddon } from '@xterm/addon-canvas';
-import { v4 as uuidv4 } from 'uuid';
 import debounce from 'lodash/debounce';
-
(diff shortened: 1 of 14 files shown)


ui/desktop/app/components/session/tabs/index.js
```diff
@@ -37,104 +35,59 @@ export default class SessionTerminalTabsComponent extends Component {
...
     window.onresize = () => {
       debouncedFit();
     };
-    xterm.onResize((size) => {
-      window.terminal.resize(size, this.id);
-    });
+  }
+
+  @action
+  openDetails() {
```
ZedLi: Why do we want the terminal view destroyed when we navigate to the details? How would a user copy over brokered credentials?

ui/desktop/app/controllers/application.js
ZedLi: Why are we doing any terminal management in the application controller?

ui/desktop/electron-app/config/web-contents-view.config.js
ZedLi: I think it's OK if it works but I found `esbuild` to be much simpler to use compared to webpack especially if it's a straightforward one off

ZedLi (commented):
Overall I think this is a great poc and I think is a very reasonable choice moving forward for better isolation of the terminal. I think we might have a few edge cases we need to properly test but I think this looks good!

PR #3194, opened 2026-03-10
Remove renderer url

ZedLi:
# Description
Refactored how we store the cluster URL for desktop client. Previously, we used the local storage of the browser in electron to keep track of the cluster URL. This meant the main electron process had to load up the renderer to be able to find out what the cluster URL we're connecting to is and we also kept track of a separate `rendererClusterUrl`.

It was a bit confusing since it wasn't obvious which one to use and there's no particular reason we should use local storage for this (likely because we weren't storing any configs yet in electron main when it was written). 

## How to Test
Make sure connecting to cluster still works and confirm that the config is correct. (on mac it should be `~/Library/Application\ Support/Boundary/config.json`.

You'll also need to force an error (easiest way is to just throw an error in code somewhere) to see the disconnect logic.

## Checklist

## PCI review checklist


cameronperera (dismissed):
Tested locally and works great. LGTM

priya-patel04 (dismissed):
Tested and looks great!

ui/desktop/tests/unit/services/cluster-url-test.js
```diff
@@ -17,37 +17,59 @@ module('Unit | Service | clusterUrl', function (hooks) {
...
     window.desktop.cluster.setClusterUrl.rejects();
-    service.setClusterUrl('invalid-origin').catch(() => {
-      assert.notOk(service.rendererClusterUrl);
+
+    try {
+      await service.setClusterUrl('http://other-origin');
+    } catch {
       assert.strictEqual(service.adapter.host, window.location.origin);
-    });
+      assert.ok(
```
lisbet-alvarez: minor suggestion: looks like `calledWithExactly` returns boolean val so should this be `assert.true`

lisbet-alvarez (dismissed):
Tested and works as expected!! Great work!

ui/desktop/tests/unit/services/cluster-url-test.js
ZedLi: Good point, updated

lisbet-alvarez (dismissed):

cameronperera (dismissed):
LGTM

DhariniJeeva (dismissed):
looks good to me!

e2e-tests/desktop/fixtures/electronTest.js
```diff
@@ -133,6 +133,8 @@ export const electronTest = test.extend({
   electronPage: async ({ electronApp }, use) => {
     const page = await electronApp.firstWindow();
     await page.evaluate(() => localStorage.clear());
+    // Reset the cluster URL stored by electron
+    await page.evaluate(() => window.desktop.cluster.resetClusterUrl());
```
ZedLi, line 137: I forgot that desktop e2e tests weren't properly resetting now that we don't store the cluster url in local storage. I ended up calling the global method directly which is a bit strange but seemed the easiest/best way to do it. Let me know if you have other suggestions

priya-patel04 (approved):

PR #3207, opened 2026-03-24
feat: 🎸 Add autocompletion to code editor

ZedLi:
# Description
Added initial code editor with autocomplete.

image

## How to Test
Goto edit grants page and test the code editor. The tables below is the logic that should be working.

### Specific id/type relationship

Assume:
- `ids` contains a **specific resource id** like `s_123` and are of the same type
- `actions` is present and valid
- the id prefix resolves to a known resource type

| Grant shape | Example | Allowed? | Rule |
|---|---|---:|---|
| specific id, **no type** | `ids=s_123;actions=read` | Yes | only id actions allowed  |
| specific id, **no type**, `create`/`list` | `ids=s_123;actions=list` | No | collection actions are disallowed without a type |
| specific id + **same type** | `ids=s_123;type=session;actions=read` | No, but the backend accepts it | `type` should be a **child** of the id’s type, not the same type |
| specific id + **child type** | `ids=hcst_123;type=host-set;actions=*` | Yes | Allowed only if the id’s `type` is a **child type** |
| specific id + **unrelated type** | `ids=s_123;type=user;actions=read` | No |`type` must be a **child** of the id’s type |
| specific id + `type=*` | `ids=hcst_123;type=*;actions=*` | Yes, sometimes | Allowed only if the id’s `type` **has child types** |
| specific id + `type=*` on leaf type | `ids=s_123;type=*;actions=*` | No | Rejected if the id’s type does **not** support child types |

### Wildcard and collection cases

| Grant shape | Example | Allowed? | Rule |
|---|---|---:|---|
| no id, no type | `actions=read` | No | Must have either id or type |
| no id + concrete type | `type=session;actions=create` | Yes, sometimes | Collection-level grants only for top level resources |
| no id + concrete type + non-collection action | `type=session;actions=read` | No | Only `create` and/or `list` allowed |
| no id + concrete type + `actions=*` | `type=scope;actions=*` | No | Wild card not allowed for type only |
| no id + `type=*` | `type=*;actions=*` | No | Wildcard type requires an id anchor |
| wildcard id + no type | `ids=*;actions=*` | No | Wildcard id requires a type |
| wildcard id + concrete type | `ids=*;type=session;actions=*` | Yes | Supported wildcard grant form, and all `type` actions |
| wildcard id + wildcard type | `ids=*;type=*;actions=*` | Yes | Supported wildcard grant form and all actions |

## Checklist

## PCI review checklist

### DIFF
--- a/ui/admin/app/components/form/role/edit-grants/index.js
+++ b/ui/admin/app/components/form/role/edit-grants/index.js
@@ -35,7 +35,6 @@ export default class FormRoleEditGrantsComponent extends Component {
   customExtensions = [
     autocompletion({
       override: [this.completionSource],
-      closeOnBlur: true,
     }),
     keymap.of(completionKeymap),
   ];
--- a/ui/admin/app/components/form/role/edit-grants/index.js
+++ b/ui/admin/app/components/form/role/edit-grants/index.js
@@ -22,9 +22,28 @@ export default class FormRoleEditGrantsComponent extends Component {
 
   exportOptionsMap = { terraform: 'terraform', nativeHCL: 'native-hcl' };
   exportOptions = Object.values(this.exportOptionsMap);
+  completionTranslatedStrings = {
+    noSuggestions: this.intl.t('resources.role.edit-grants.no-suggestions'),
+    wildcardTypes: this.intl.t(
+      'resources.role.edit-grants.completion-info.wildcard-types',
+    ),
+    wildcardIds: this.intl.t(
+      'resources.role.edit-grants.completion-info.wildcard-ids',
+    ),
+    templateValue: this.intl.t(
+      'resources.role.edit-grants.completion-info.template-value',
+    ),
+    wildcardActions: this.intl.t(
+      'resources.role.edit-grants.completion-info.wildcard-actions',
+    ),
+    allFields: this.intl.t(
+      'resources.role.edit-grants.completion-info.all-fields',
+    ),
+  };
+
   completionSource = createGrantCompletionSource(
     this.args.grantsSchema,
-    this.intl.t('resources.role.edit-grants.no-suggestions'),
+    this.completionTranslatedStrings,
   );
 
   @tracked grantStringsText = (this.args.model?.grant_strings ?? []).join('\n');
@@ -35,6 +54,8 @@ export default class FormRoleEditGrantsComponent extends Component {
   customExtensions = [
     autocompletion({
       override: [this.completionSource],
+      // Trigger autocompletion when the user completes a grant field (which we labeled as keywords)
+      activateOnCompletion: (completion) => completion.type === 'keyword',
     }),
     keymap.of(completionKeymap),
   ];
--- a/ui/admin/app/components/form/role/edit-grants/index.js
+++ b/ui/admin/app/components/form/role/edit-grants/index.js
@@ -4,21 +4,48 @@
  */
 
 import Component from '@glimmer/component';
-import { tracked } from '@glimmer/tracking';
 import { action } from '@ember/object';
+import { service } from '@ember/service';
+import { tracked } from '@glimmer/tracking';
+import {
+  autocompletion,
+  completionKeymap,
+  keymap,
+} from '@hashicorp/design-system-components/codemirror';
 
-export default class FormRoleEditGrants extends Component {
+import { createGrantCompletionSource } from 'admin/utils/grant-completions';
+
+export default class FormRoleEditGrantsComponent extends Component {
+  @service intl;
+
   // =attributes
 
   exportOptionsMap = { terraform: 'terraform', nativeHCL: 'native-hcl' };
   exportOptions = Object.values(this.exportOptionsMap);
+  completionSource = createGrantCompletionSource(
+    this.args.grantsSchema,
+    this.intl.t('resources.role.edit-grants.no-suggestions'),
+  );
 
-  // TODO: Replace with actual grant lines from code editor once implemented.
-  grantStringLines =
-    'ids=hc_123;type=host-catalog;actions=read,create,list\nids=ttcp_123;type=target;actions=list\ntype=credential;actions=create';
-
+  @tracked grantStringsText = (this.args.model?.grant_strings ?? []).join('\n');
+  @tracked currentLineText = this.args.model?.grant_strings?.[0] ?? '';
   @tracked showExportOptionsFlyout = false;
   @tracked selectedExportOption = this.exportOptions[0];
+
+  customExtensions = [
+    autocompletion({
+      override: [this.completionSource],
+      closeOnBlur: true,
+    }),
+    keymap.of(completionKeymap),
+  ];
+
+  get grantStrings() {
+    return this.grantStringsText
+      .split('\n')
+      .map((grantString) => grantString.trim())
+      .filter(Boolean);
+  }
 
   /**
    * Returns the formatted export based on the selected export option.
@@ -37,7 +64,7 @@ export default class FormRoleEditGrants extends Component {
    */
   get terraformFormattedExport() {
     let formatted = `grant_strings = [ \n`;
-    this.grantStringLines.split('\n').forEach((line) => {
+    this.grantStringsText.split('\n').forEach((line) => {
       formatted += `  "${line}",\n`;
     });
     formatted += `]\n`;
@@ -50,7 +77,7 @@ export default class FormRoleEditGrants extends Component {
    */
   get nativeHclFormattedExport() {
     let formatted = `[ \n`;
-    this.grantStringLines.split('\n').forEach((line) => {
+    this.grantStringsText.split('\n').forEach((line) => {
       formatted += `  "${line}",\n`;
     });
     formatted += `]\n`;
@@ -58,6 +85,14 @@ export default class FormRoleEditGrants extends Component {
   }
 
   // =actions
+
+  @action
+  onInput(value, view) {
+    this.grantStringsText = value;
+
+    const line = view.state.doc.lineAt(view.state.selection.main.head);
+    this.currentLineText = line.text;
+  }
 
   /**
    * Toggles the export options flyout open and closed.
--- a/ui/admin/app/utils/grant-completions.js
+++ b/ui/admin/app/utils/grant-completions.js
@@ -1,0 +1,362 @@
+/**
+ * Copyright IBM Corp. 2021, 2026
+ * SPDX-License-Identifier: BUSL-1.1
+ */
+
+const GRANT_FIELDS = ['ids', 'type', 'actions', 'output_fields'];
+const ID_TEMPLATES = ['{{.User.Id}}', '{{.Account.Id}}'];
+const filterByPrefix = (values, partial) =>
+  values.filter((value) => value.startsWith(partial));
+const withWildCard = (values = []) => ['*', ...new Set(values)];
+const parseIds = (idsValue = '') => idsValue.split(',').filter(Boolean);
+
+const getNoSuggestionsOption = (noSuggestionsLabel) => ({
+  label: noSuggestionsLabel,
+  type: 'text',
+  apply: () => {},
+  boost: -99,
+});
+
+const parseGrantFields = (lineText) =>
+  lineText
+    .split(';')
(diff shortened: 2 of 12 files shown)


ui/admin/app/utils/grant-completions.js
```diff
@@ -0,0 +1,362 @@
+/**
+ * Copyright IBM Corp. 2021, 2026
+ * SPDX-License-Identifier: BUSL-1.1
+ */
+
+const GRANT_FIELDS = ['ids', 'type', 'actions', 'output_fields'];
+const ID_TEMPLATES = ['{{.User.Id}}', '{{.Account.Id}}'];
```
lisbet-alvarez, line 7: looks like there's two more [valid formats](https://github.com/hashicorp/boundary/blob/main/internal/perms/grants.go#L533) `{{user.id}} and {{account.id}}

ui/admin/app/utils/grant-completions.js
```diff
@@ -0,0 +1,362 @@
...
+  ];
+};
+
+const getTypeOptions = (schema, idsValue) => {
+  // If there is no ids value, we can only suggest top-level resource types that have collection actions
+  if (!idsValue) {
+    return schema.resourceTypes
+      .filter(
+        (resource) =>
+          !resource.parent_type && resource.collection_actions?.length > 0,
```
lisbet-alvarez: we might want to be more specific with the collection actions that a resource should have in order to be considered a top-level resource. They can have create and/or list. The current logic would classify `billing` resource as a top level type but it is not ([ref](https://github.com/hashicorp/boundary/blob/main/internal/types/resource/resource.go#L170))

ui/admin/app/utils/grant-completions.js
```diff
@@ -0,0 +1,362 @@
...
+    resourcesByType,
+    childResourceTypesByParentType,
+    resourceTypesByIdPrefix,
+  };
+};
+
+const getResourceTypesForId = (schema, id) =>
+  id.includes('_')
+    ? [
+        schema.resourceTypesByIdPrefix.find(
```
lisbet-alvarez, line 90: im wondering if it would be simplier to have `resourceTypesByIdPrefix` map a prefix (key) to the type (value) so that we don't have to use `find` method. Also i don't think different cases are needed for an id with `_` since id.split('_')[0] the entire id if there is no `_`

ui/admin/app/utils/grant-completions.js
ZedLi: Oh that's good to know, I wasn't aware of that. Would I do anything with that for autocomplete though? I think for linting we'd have to be aware but wouldn't I just recommend one format?

ui/admin/app/utils/grant-completions.js
```diff
@@ -0,0 +1,362 @@
...
+const filterByPrefix = (values, partial) =>
+  values.filter((value) => value.startsWith(partial));
+const withWildCard = (values = []) => ['*', ...new Set(values)];
+const parseIds = (idsValue = '') => idsValue.split(',').filter(Boolean);
+
+const getNoSuggestionsOption = (noSuggestionsLabel) => ({
+  label: noSuggestionsLabel,
+  type: 'text',
+  apply: () => {},
+  boost: -99,
```
lisbet-alvarez: Looks like this will add the no suggestions to bottom of list. Is it necessary to specify `boost` here if this is shown when there are no options to suggest?

ui/admin/app/components/form/role/edit-grants/index.js
```diff
@@ -4,22 +4,49 @@
...
-
+  @tracked grantStringsText = (this.args.model?.grant_strings ?? []).join('\n');
+  @tracked currentLineText = this.args.model?.grant_strings?.[0] ?? '';
   @tracked showExportOptionsFlyout = false;
   @tracked selectedExportOption = this.exportOptions[0];
 
+  customExtensions = [
+    autocompletion({
+      override: [this.completionSource],
+      closeOnBlur: true,
```
lisbet-alvarez: Looks like this option [defaults to true](https://codemirror.net/docs/ref/#autocomplete.autocompletion^config.closeOnBlur).

ui/admin/tests/unit/utils/grant-completions-test.js
```diff
@@ -0,0 +1,312 @@
...
+  'monthly-active-users',
+  'destroy-key-version',
+  'list-keys',
+  'rotate-keys',
+  'list-key-version-destruction-jobs',
+  'attach-storage-policy',
+  'detach-storage-policy',
+  'download',
+  'reapply-storage-policy',
+];
```
lisbet-alvarez: for this main list with all the actions, should we derive it from the grantsSchemaData instead so that data is already available in beforeEach hook

ui/admin/app/utils/grant-completions.js
```diff
@@ -0,0 +1,362 @@
...
+
+const getResourceTypesForId = (schema, id) =>
+  id.includes('_')
+    ? [
+        schema.resourceTypesByIdPrefix.find(
+          ({ idPrefix }) => id.split('_')[0] === idPrefix,
+        )?.type,
+      ].filter(Boolean)
+    : schema.resourceTypesByIdPrefix
+        .filter(({ idPrefix }) => id.startsWith(idPrefix))
```
cameronperera: Should `id.startsWith(idPrefix)` be `idPrefix.startsWith(id)`?
Thinking about a TCP target a partial id of `tt` would be false currently but I imagine the intent of this fallback is for partial matches of an id?

ui/admin/app/utils/grant-completions.js
ZedLi: Good point, I removed it

ui/admin/app/utils/grant-completions.js
ZedLi: Good catch, I wasn't aware of the `billing` exception and that they consider create/list as the collection actions, I updated to check for the presence of either possible grant

ui/admin/tests/unit/utils/grant-completions-test.js
ZedLi: Yup good call, I was just being lazy 😅

ui/admin/app/utils/grant-completions.js
ZedLi: Yes great catch, I have the logic backwards and it should be the other way.

ui/admin/app/utils/grant-completions.js
ZedLi: Do you mean for example, the object is just `{ prefix: 'type' }`? We'd still have to iterate over it right? This read clearer to me using find since the implication was clear (it returns one result, the one it finds first). 

> Also i don't think different cases are needed for an id with _ since id.split('_')[0] the entire id if there is no _

Can you explain what you meant by this, I don't think I understood it.

Also, realistically this should be using a prefix tree but I was too lazy to build one or import a package, I could do it if we think it reads clearer though, thoughts?

ui/admin/app/utils/grant-completions.js
```diff
@@ -208,14 +210,22 @@ const getActionOptions = (schema, typeValue, idsValue) => {
...
-    if (
-      !selectedResource ||
-      !(getCompatibleResourceTypeForIds(schema, idsValue) === typeValue)
-    ) {
+    const matchedType = getCompatibleResourceTypeForIds(schema, idsValue);
+    const isSelectedChildType = (
+      schema.childResourceTypesByParentType[matchedType] ?? []
+    ).includes(typeValue);
+
+    // A specific ID paired with its own type is not a valid combination.
```
cameronperera, line 218: I want to make sure I understand this correct so here is an example:
`ids=ttcp_123;type=target;actions=...` is invalid because we have pinned it to a specific target so the `type` is irrelevant. However, something like this would be ok:
`ids=*;type=target;actions=...` because we want the listed actions to apply to all target types.
Am I understanding that correctly?

ZedLi: Yup exactly

ui/admin/app/utils/grant-completions.js
```diff
@@ -0,0 +1,372 @@
...
+}
+
+export const createGrantCompletionSource = (
+  grantsSchema,
+  noSuggestionsLabel,
+) => {
+  const schema = normalizeGrantsSchema(grantsSchema);
+
+  return (context) => grantCompletions(context, schema, noSuggestionsLabel);
+};
```
cameronperera, line 372: Just want to say I really like this approach. Honestly, I don't think I have seen this before but I like that it remembers the schema (and label) so it doesn't need to be computed for every completion.

cameronperera (dismissed):
Looks great and works well!
I have one non-blocking comment:
When I type `act` for example it will suggest `actions=`. I select that text and it is populated for me. Then I am sitting at `ids=*;actions=`. I personally expected that a list of suggested actions would have popped up but they don't. I have to either start typing which would limit my suggestions to actions that start with `c` if that is what I type. I found myself selecting a suggested fieldName (ids, type, or actions) then removing the `=` and typing it again to have ALL possible actions suggested. Not sure if we can make the `completionSource` run again right after inserting some text or potentially we remove the `=` from the fieldNames and let the user type it.

ui/admin/app/utils/grant-completions.js
lisbet-alvarez: I was thinking more of just an object where `resourceTypesByIdPrefix = { idPrefix: type, ...}` vs what we have now which is `resourceTypesByIdPrefix = [{ idPrefix, type }, ...]`. That way we could always just do `schema.resourceTypesByIdPrefix[id.split('_')[0]]`. Would this work for this use case. I also noticed `getResourceTypesForId` returns an array but i think it should always just be one resource type per Id perfix?

ZedLi:
> When I type `act` for example it will suggest `actions=`. I select that text and it is populated for me. Then I am sitting at `ids=*;actions=`. I personally expected that a list of suggested actions would have popped up but they don't. I have to either start typing which would limit my suggestions to actions that start with `c` if that is what I type. I found myself selecting a suggested fieldName (ids, type, or actions) then removing the `=` and typing it again to have ALL possible actions suggested. Not sure if we can make the `completionSource` run again right after inserting some text or potentially we remove the `=` from the fieldNames and let the user type it.

Hmm isn't this the behavior usually of autocomplete? Since normally it doesn't open autocomplete right after a completion, I guess I instinctively just use `Ctrl + Space` to re-open autocomplete like in an IDE. You're right about automatically including the `=`, that's open to discussion on whether we should include it but I assumed because we know that it's always required I just included it, but it is also just a delimiter so maybe not?

cameronperera:
>  I guess I instinctively just use `Ctrl + Space` to re-open autocomplete like in an IDE

Wow that does work. Sorry, I actually have never used `Ctrl + Space` in my IDE so I learned something new today.

ui/admin/app/utils/grant-completions.js
ZedLi: I get it now, well I guess it depends on if we need the else case here:
https://github.com/hashicorp/boundary-ui/blob/3c106e084dee2e0c450f6180675dd372f0c9f414/ui/admin/app/utils/grant-completions.js#L94-L96

For the same reason that it's an array instead of just a single resource type, I built in a partial check to see if there's multiple resources that can match the id prefix if a user types in something that's just the beginning before a `_`, (e.g. `h` which could match different resource types).

You're right though, if we only check autocomplete when a user actually adds a `_` or for full ids, we can avoid this extra check and use case and greatly simplify it, I also aren't sure how useful it is.

ui/admin/app/utils/grant-completions.js
lisbet-alvarez: Yeah ur right its best to just recommend one format.

ui/admin/app/utils/grant-completions.js
lisbet-alvarez: ahh ok that totally makes sense

lisbet-alvarez (dismissed):
Looks great!!

cameronperera (approved):
LGTM!

lisbet-alvarez (approved):
Looks great!

PR #3211, opened 2026-03-30
feat: 🎸 Add linting to code editor

lisbet-alvarez:
✅ Closes: https://hashicorp.atlassian.net/browse/ICU-18462

# Description

Builds off of original [POC](https://github.com/hashicorp/boundary-ui/compare/main...hds-autocomplete) created by @ZedLi to enable linting using code mirror extension to HDS code editor for grants.

image

## How to Test

1. Go to a role.
2. Go to Manage -> Edit Grants
3. Validate linter catches all violations outlined in this [table](https://github.com/hashicorp/boundary-ui/pull/3207).

## Checklist

## PCI review checklist


ui/admin/app/utils/grant-linter.js
```diff
@@ -0,0 +1,623 @@
...
+    // Check for whitespace
+    if (line.includes(' ')) {
+      const whitespaceMatch = line.match(/\s+/);
+      const firstSpace = lineStart + whitespaceMatch.index;
+      const lastSpace = firstSpace + whitespaceMatch[0].length;
+      diagnostics.push({
+        from: firstSpace,
+        to: lastSpace,
+        severity: 'error',
+        message: 'Whitespace is not allowed',
```
ZedLi, line 122: I know this is a big pain but we technically need to internationalize these strings still with ember intl

ui/admin/app/utils/grant-linter.js
```diff
@@ -0,0 +1,623 @@
...
+        actions: [createDeleteTextAction('Remove whitespace')],
```
ZedLi, line 123: I didn't realize `actions` actually lets you click it and auto correct it, that's neat.

I'm curious though, do we need the name to be unique to each one, it seems like it's usually deleting/removing, could the action just say "Remove", for example instead of something like "Remove trailing semicolon" since that feels like extra noise?

ui/admin/app/utils/grant-linter.js
```diff
@@ -0,0 +1,623 @@
...
+
+  let currentPos = 0;
+
+  lines.forEach((line) => {
+    const trimmedLine = line.trim();
+    const lineStart = currentPos;
+    const lineEnd = currentPos + line.length;
+
+    // Check for whitespace
+    if (line.includes(' ')) {
```
ZedLi, line 114: I notice the backend actually sometimes allow spaces, do we know how valid it is? (e.g. a leading space before ids and types seems ok, but not actions?) Is there any issues with these type of grants? I'm assuming this might just be a validation issue on the backend and we should just disallow it regardless?

One other thing I wanted to point out is `line.includes(' ')` only checks for actual spaces, but you do a match on _all_ whitespace with `/\s+/`, including things like tabs for examples. We should likely makes sure the if check matches the actual regex in this case.

ui/admin/app/utils/grant-linter.js
```diff
@@ -0,0 +1,623 @@
...
+      const whitespaceMatch = line.match(/\s+/);
```
ZedLi, line 115: This only matches the first instance of whitespace found, if there's more than two, it'll only pick up the first one. We should probably matchAll and return a linting error for each one

ui/admin/app/utils/grant-linter.js
```diff
@@ -0,0 +1,623 @@
...
+        from: semicolonPos,
+        to: semicolonPos + 1,
+        severity: 'error',
+        message: 'Trailing semicolon is not allowed',
+        actions: [createDeleteTextAction('Remove trailing semicolon')],
+      });
+    }
+
+    // Check for invalid characters (only allow alphanumeric characters, equal signs, commas, semicolons, underscores, hyphens, curly braces, and periods)
+    const invalidCharMatch = line.match(/[^a-zA-Z0-9=,;:_*\-.{}\s]/g) || [];
```
ZedLi, line 141: Do we allow capitals? I notice i can do `READ` and backend takes it and spits back out in lowercase, should we allow it for linting or not?

Also do we allow _any_ whitespace or just a normal space but also what type of grants use whitespace at all?

ui/admin/app/utils/grant-linter.js
```diff
@@ -0,0 +1,623 @@
...
+        to: lastSpace,
+        severity: 'error',
+        message: 'Whitespace is not allowed',
+        actions: [createDeleteTextAction('Remove whitespace')],
+      });
+    }
+
+    const trimStart = line.indexOf(trimmedLine);
+    // Check for trailing semicolon
+    if (trimmedLine.endsWith(';')) {
```
ZedLi, line 129: Should we also check for a leading semicolon too? I notice we don't flag something like `;type=host-catalog;actions=create`

ui/admin/app/utils/grant-linter.js
```diff
@@ -0,0 +1,623 @@
...
+
+      // Warn about duplicate fields - only the last instance of the field will be used for parsing
+      if (trimmedKey in parsedFields) {
+        const pos = fieldPositions[trimmedKey];
+        diagnostics.push({
+          from: pos.keyStart,
+          to: pos.valueEnd,
+          severity: 'error',
+          message: `Duplicate field "${trimmedKey}" found`,
+          actions: [createDeleteTextAction('Remove duplicate field')], // Action will remove the first instance of the duplicate field
```
ZedLi, line 189: Do we want an action for this since we don't know which one is the actual duplicate field? I wonder if this one is better left to a user to rectify. Also the action removes the first duplicate field but leaves the semicolon afterwards which is invalid.

With multiple duplicate fields, I notice it doesn't underline the last duplicate field:

Image

ui/admin/app/utils/grant-linter.js
```diff
@@ -0,0 +1,623 @@
...
+  '{{.User.Id}}',
+  '{{user.id}}',
+  '{{account.id}}',
+];
+
+const createDeleteTextAction = (name) => ({
+  name,
+  apply(view, from, to) {
+    view.dispatch({
+      changes: { from, to, insert: '' },
```
ZedLi, line 18: Is the `insert: ''` needed? I notice the docs don't include it in their examples

ui/admin/app/utils/grant-linter.js
```diff
@@ -0,0 +1,623 @@
...
+        });
+      } else {
+        validatedFields.type.value = typeValue;
+        if (!parsedFields.ids) {
+          if (!schema.topLevelTypes.includes(typeValue)) {
+            diagnostics.push({
+              from: pos.valueStart,
+              to: pos.valueEnd,
+              severity: 'error',
+              message: `Type "${typeValue}" must be a top-level type`,
```
ZedLi, line 250: Is the message here actually that `typeValue` _isn't_ a top level type and should be? I find the current usage of the non-top level variable type a bit confusing, should it just say `Type must be a top-level type` or `Type "${typeValue}" is not a top-level type`

ui/admin/app/utils/grant-linter.js
```diff
@@ -0,0 +1,623 @@
...
+
+    // Validate ids field
+    //const parsedIdsField = { value: [], exists: false, knownType: '', wildcard: false };
+    if (parsedFields.ids) {
+      validatedFields.ids = { value: [], knownType: '', wildcard: false };
+      const idsValue = parsedFields.ids.trim();
+      const pos = fieldPositions.ids;
+      const idsList = idsValue
+        .split(',')
+        .map((id) => id.trim())
```
ZedLi, line 266: One thought I have is: if we're validating whitespaces too, should we be trimming inputs? When does it make sense vs not to?

ui/admin/app/utils/grant-linter.js
```diff
@@ -0,0 +1,623 @@
...
+        .filter((id) => id);
```
ZedLi, line 267: Should we be filtering these? I notice things like `ids=,,*;` are valid now because the empty strings get caught and it doesn't recognize the wildcard being combined with others

ui/admin/app/utils/grant-linter.js
```diff
@@ -0,0 +1,623 @@
...
+      return;
+    }
+    if (validatedFieldsType.wildcard) {
+      // If type is wildcard, then pinned id(s) must support child types
+      if (!schema.parentTypes.includes(currentIdType)) {
+        diagnostics.push({
+          from: pos.valueStart,
+          to: pos.valueEnd,
+          severity: 'error',
+          message: `Ids must support child types. Invalid id "${id}"`,
```
ZedLi, line 498: Should we make this message clearer about being for pinned IDs?

ui/admin/app/utils/grant-linter.js
```diff
@@ -0,0 +1,623 @@
...
+    } else if (validatedFieldsType.value) {
+      // If type is known and not wildcard, then pinned id(s) must match the type
+      const parentType =
+        schema.resourcesByType[validatedFieldsType.value]?.parentType;
+      if (!parentType) {
+        diagnostics.push({
+          from: pos.valueStart,
+          to: pos.valueEnd,
+          severity: 'error',
+          message: `Type must support child types. Invalid type "${validatedFieldsType.value}"`,
```
ZedLi, line 511: This seems strange to have on the IDs, isn't it saying it's an invalid type on the id field? We could also say the ID must be a parent instead

ui/admin/app/utils/grant-linter.js
```diff
@@ -0,0 +1,623 @@
...
+        });
+        return;
+      }
+      if (currentIdType !== parentType) {
+        diagnostics.push({
+          from: pos.valueStart,
+          to: pos.valueEnd,
+          severity: 'error',
+          message: `Ids must match the type "${parentType}". Invalid id "${id}"`,
```
ZedLi, line 520: I think this is a confusing message, it's trying to say the ids themselves should be the `parentType` but it can be easily interpreted as the `type` _field_ should be `parentType` (which is how I read it initially)

Should we instead say something like `Ids must be "${parentType}" ids`? Or something similar that mentions it being a pinned ID?

ui/admin/app/utils/grant-linter.js
```diff
@@ -0,0 +1,623 @@
...
+      }
+    }
+
+    // Validate actions field
+    if (parsedFields.actions) {
+      const actionsValue = parsedFields.actions.trim();
+      const actionList = actionsValue
+        .split(',')
+        .map((a) => a.trim())
+        .filter((a) => a);
```
ZedLi, line 353: Same issue as with ids, it looks like `actions=,*,` ends up being a valid grant

ui/admin/app/utils/grant-linter.js
```diff
@@ -0,0 +1,623 @@
...
+  let errorMessage = (action) => `Invalid action "${action}"`;
+  if (!validatedFields.type) {
+    if (validatedFields.ids?.knownType) {
+      // Only id actions belonging to known id type (including wildcard action)
+      validActions = getIdActionsForResourceType(
+        schema,
+        validatedFields.ids.knownType,
+      );
+      errorMessage = (action) =>
+        `Invalid action "${action}". Valid id actions: ${validActions.join(', ')}`;
```
ZedLi, line 575: Should we show all the valid id actions? It can sometimes be large and hopefully autocomplete should be the one actually letting the user know what's valid. Same question for showing the other valid actions below too

Image

ui/admin/app/components/form/role/edit-grants/index.js
```diff
@@ -57,7 +61,9 @@ export default class FormRoleEditGrantsComponent extends Component {
       // Trigger autocompletion when the user completes a grant field (which we labeled as keywords)
       activateOnCompletion: (completion) => completion.type === 'keyword',
     }),
-    keymap.of(completionKeymap),
+    linter(this.linterSource),
+    lintGutter(),
+    keymap.of([...completionKeymap, ...lintKeymap]),
```
ZedLi, line 66: I had no idea there was a key mapping for linting using `Ctrl (or ⌘) + Shift + M`. Will users know about this, is it useful?

ZedLi (commented):
Wow this is incredibly cool to see in action, looks awesome so far!

ui/admin/app/utils/grant-linter.js
```diff
@@ -0,0 +1,623 @@
...
+    case 'account-template':
+      return ['*', ...schema.resourcesByType['account'].idActions];
+    case 'user-template':
+      return ['*', ...schema.resourcesByType['user'].idActions];
+    default:
+      return ['*', ...schema.resourcesByType[resourceType].idActions];
+  }
+};
+
+const validateActionsField = (
```
ZedLi, line 558: We also probably need to include a linting rule for the [`no-op`](https://developer.hashicorp.com/boundary/docs/rbac/assignable-permissions#the-no-op-action-and-listing-visibility) rule with listing resources too.

ui/admin/app/utils/grant-linter.js
lisbet-alvarez: yup good point, it is repetitive

ui/admin/app/utils/grant-linter.js
lisbet-alvarez: yeah I agree and that would also be a better ux.

ui/admin/app/utils/grant-linter.js
lisbet-alvarez: hmm ill double check If the spaces are actually valid. Im also thinking its validation issues and should not be allowed.

Ah good catch!

ui/admin/app/utils/grant-linter.js
lisbet-alvarez: I had also noticed that capitals were allowed but parsed as lowercase. However that is a good point, we can just not allow it since it gets parsed as lowercase either way.

For whitespace, when I was testing it, it broke parsing each time with "improperly formatted grant" error. So Im not aware of what whitespace is allowed. For the linter I wasn't allowing any whitespace at all since I haven't seen any grant formats that does allow it. Ill look into this further.

ui/admin/app/utils/grant-linter.js
lisbet-alvarez: oh yup good catch

ui/admin/app/utils/grant-linter.js
lisbet-alvarez: Yeah good point, user should choose which one to keep and all instances should probably be underlined

ui/admin/app/utils/grant-linter.js
lisbet-alvarez: Ah yes i see now how it's confusing, and yup the error is that the specified type is not a top-level type

ui/admin/app/utils/grant-linter.js
lisbet-alvarez: oh yup i forgot to go back and handle this case! good catch

ui/admin/app/utils/grant-linter.js
lisbet-alvarez: Good catch, i think i messed up the wording when I was refactoring it.

ui/admin/app/utils/grant-linter.js
lisbet-alvarez: Yeah, great point, its not necessary and takes up space.

ui/admin/app/components/form/role/edit-grants/index.js
lisbet-alvarez: Its to open all alerts as mentioned in the[ linting guidelines](https://helios.hashicorp.design/components/code-editor#linting:~:text=To%20view%20all%20alerts%20in%20the%20editor%2C) for hds code editor. It is useful because you can scroll through all errors and if you click on one error it highlights the line it is referring too. I had thought it was a standard code editor cmd for lint errors since there was no further mention to make the cmd known when linting is enabled. (seems eslint extension also has same cmd to view all errors panel in vscode)

ui/admin/app/utils/grant-linter.js
lisbet-alvarez: Great point, u are right they should be!

ui/admin/app/utils/grant-linter.js
lisbet-alvarez: I ended up not allowing whitespace since I was testing with varying amounts of whitespace and all were getting errors from api

ui/admin/app/utils/grant-linter.js
lisbet-alvarez: Good catch, it is not needed.

ui/admin/app/utils/grant-linter.js
```diff
@@ -11,11 +11,11 @@ const ID_TEMPLATES = [
   '{{account.id}}',
 ];
 
-const createDeleteTextAction = (name) => ({
+const createDeleteTextAction = (name = 'Remove') => ({
```
ZedLi: I think this needs to be translated as well

ui/admin/app/utils/grant-linter.js
```diff
@@ -448,54 +509,54 @@ const validateIdsField = (
...
-    const currentIdType = idType(schema, id);
     if (idsList.slice(i + 1).includes(id)) {
       diagnostics.push({
-        from: pos.valueStart,
-        to: pos.valueEnd,
+        from: idsPos.valueStart,
+        to: idsPos.valueEnd,
         severity: 'error',
-        message: `Duplicate id "${id}"`,
+        message: translate('ids.duplicate-id', { id }),
```
ZedLi, line 530: Is this one able to have an action to remove the id as well?

addons/core/translations/resources/en-us.yaml
```diff
@@ -670,6 +670,46 @@ role:
...
+        duplicate-field: Duplicate field "{field}"
+        unknown-field: Unknown field "{field}"
+        empty-value: '"{field}" value cannot be empty'
+        missing-ids-or-type: 'Missing "ids" or "type" fields'
+        missing-actions: 'Missing "actions" field'
+        missing-actions-or-output-fields: 'Missing "actions" or "output_fields" fields'
+      type:
+        invalid-type: Invalid resource type "{type}"
+        not-top-level: Type "{type}" is not a top-level type
+        not-child-type: Type "{type}" is not a child type
```
ZedLi: Should this say `Type "{type}" is not a child type of the pinned ID` to be more explicit?

ui/admin/tests/unit/utils/grant-linter-test.js
ZedLi, line 1: Can we add test cases for `no-op`?

ui/admin/app/utils/grant-linter.js
```diff
@@ -615,9 +694,32 @@ const validateActionsField = (
       return;
     }
   }
+
+  // no-op should be used with list action or else it has no effect.
+  // Also, no-op is not necessary if there is already any other action specified.
+  if (actionList.includes('no-op')) {
```
ZedLi: I think we have cases where we return before this where these checks might never get hit when it should be checked. For example `type=auth-method;actions=list,no-op` is being checked as invalid and `type=auth-method;actions=no-op` doesn't check for these no-op cases

addons/core/translations/resources/en-us.yaml
```diff
@@ -670,6 +670,46 @@ role:
...
+        must-support-child-types: Pinned-ids must support child types. Invalid id "{id}"
+        must-match-parent-type: Pinned-ids must be "{parentType}" ids. Invalid id "{id}"
+      actions:
+        empty-list: '"actions" field does not contain at least one action'
+        wildcard-cannot-combine: 'Wildcard action "*" cannot be combined with other actions'
+        duplicate-action: Duplicate action "{action}"
+        invalid-action: Invalid action "{action}"
+        id-only: Only id actions for "{type}" type are allowed. Invalid action "{action}"
+        collection-only: Only collection actions are allowed. Invalid action "{action}"
+        noop-requires-list: '"no-op" action should be used with "list" action'
```
ZedLi: Is `'"no-op" action should only be used with "list" action'` a bit stronger/clearer?

ui/admin/app/utils/grant-linter.js
```diff
@@ -615,9 +694,32 @@ const validateActionsField = (
...
+    if (!actionList.includes('list')) {
+      diagnostics.push({
+        from: pos.valueStart,
+        to: pos.valueEnd,
+        severity: 'error',
+        message: translate('actions.noop-requires-list'),
+      });
+      return;
+    }
+    if (actionList.some((action) => action !== 'no-op' && action !== 'list')) {
```
ZedLi: Do we have a case where only `list` is present and we need to tell the user they have to add `no-op` otherwise it won't work?

addons/core/translations/resources/en-us.yaml
lisbet-alvarez: yup that is clearer, thank u!

ui/admin/app/utils/grant-linter.js
lisbet-alvarez: hmm good point!

ui/admin/app/utils/grant-linter.js
lisbet-alvarez: Removed trim since like u said whitespace is already being validated along with comma separated list syntax.

ui/admin/app/utils/grant-linter.js
lisbet-alvarez: hmm looks like `type=auth-method;actions=list,no-op` is invalid since the api explicitly only allows `create, create:*, or list`

ZedLi: How does one even set something like the billing grants then, like `type=billing;actions=monthly-active-users`? Do these always require wild card for ids?

lisbet-alvarez: yup that looks to be the case `ids=*;type=billing;actions=monthly-active-users` is valid but `type=billing;actions=monthly-active-users` is not

ui/admin/app/utils/grant-linter.js
```diff
@@ -717,6 +735,17 @@ const validateActionsField = (
       return;
     }
   }
+  if (hasList) {
+    if (!hasNoOp && !hasOtherActions) {
+      diagnostics.push({
```
ZedLi: Should we add an action to automatically add the `no-op`?

ZedLi (dismissed):
Looks great!

ui/admin/app/utils/grant-linter.js
lisbet-alvarez: ooo yeah thats a good idea

addons/api/app/mirage/factories/role.js
```diff
@@ -26,7 +26,7 @@ export default factory.extend({
       'set-grant-scopes',
     ],
   grant_strings: () => [
-    'ids=*;action=*',
+    'ids=*;actions=*',
```
ZedLi, line 29: It's funny how we never noticed this 😆

ZedLi (dismissed):

ZedLi (approved):

ui/admin/tests/unit/utils/grant-linter-test.js
cameronperera, line 1: nice work on all theses test cases!

cameronperera (approved):

PR #3234, opened 2026-04-09
chore: Always run end-to-end tests

moduli:
# Description
I've noticed in PRs that when a "merge from main" update was done, end-to-end tests may not trigger if the updated files did not fall into one of these paths
```
  pull_request:
    types: [labeled, opened, synchronize, reopened]
    paths:
      - 'ui/admin/**'
      - 'e2e-tests/*'
      - 'e2e-tests/api-client/**'
      - 'e2e-tests/helpers/**'
      - 'e2e-tests/admin/**'
      - '.github/workflows/test-e2e-admin-ui.yaml'
```

To ensure that changes are fully tested alongside any new incoming changes, we should run the end-to-end tests on those updates as well. As a result, the PR is removing the `paths` filter.

This PR also splits up the e2e test job to run a new job per browser type, reducing the overall time for completion.

Before: 
Screenshot 2026-04-10 at 11 12 38 AM

After:
Screenshot 2026-04-10 at 11 17 09 AM

Also tested against `boundary-ui-enterprise` to validate that logic: https://github.com/hashicorp/boundary-ui-enterprise/actions/runs/24248679245?pr=1527

Screenshot 2026-04-10 at 11 14 46 AM

This PR also includes a fix to a slack message to account for different values based on what triggered the workflow (workflow_dispatch vs pull_request).

## Checklist

## PCI review checklist

### DIFF
--- a/.github/workflows/test-e2e-admin-ui.yaml
+++ b/.github/workflows/test-e2e-admin-ui.yaml
@@ -26,6 +26,8 @@ jobs:
     runs-on: ${{ fromJSON(vars.RUNNER_UBUNTU_22) }}
     strategy:
       matrix:
+        boundary_edition: [community, enterprise]
+        browser: [chromium, firefox, webkit]
         include:
           - boundary_edition: community
             github_repo: hashicorp/boundary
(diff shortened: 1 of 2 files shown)


.github/workflows/test-e2e-admin-ui.yaml
```diff
@@ -3,13 +3,6 @@ name: Test E2E Admin UI
 on:
   pull_request:
     types: [labeled, opened, synchronize, reopened]
-    paths:
```
ZedLi, line 6: > I've noticed in PRs that when a "merge from main" update was done, end-to-end tests may not trigger if the updated files did not fall into one of these paths

What kind of file changes weren't running e2e tests that should have been? I do see that we missed things like our addons folders. Do we want admin e2e tests running for things that are purely desktop or vice versa, e.g. electron changes 

I wonder if we should have used `paths-ignore` instead and excluded runs instead of allow listing?

moduli: I lost the example, but I had noticed some "merge updates from main" commits that then wouldn't show e2e test results in the PR status checks at the bottom. Sure, the commits themselves would likely not change the test results, but you wouldn't be able to review the test results.

Thoughts?

ZedLi: Had a quick conversation offline and don't think it's a big issue to just have the tests always run for now, especially since they use docker resources currently.

ZedLi (approved):
Looks good, I do like having the browsers broken out into their own runs!

PR #3235, opened 2026-04-09
chore(e2e): Log in via CLI during global setup to update cache

moduli:
# Description
This PR makes a slight adjustment to the e2e test setup. In the global setup, a step was added to log in to Boundary via the CLI. This will ensure that the Boundary cache is updated to the current version before starting desktop e2e tests (there is a cache check on authentication). 

This avoids a problem when running tests locally where someone may update their CLI before running e2e tests, but the desktop client behavior depends on updated cache functionality. 

## Checklist

## PCI review checklist

### DIFF
--- a/e2e-tests/global-setup.js
+++ b/e2e-tests/global-setup.js
@@ -4,6 +4,8 @@
  */
 
 import { chromium, test as baseTest, mergeTests } from '@playwright/test';
+
+import * as boundaryCli from './helpers/boundary-cli';
 import { checkEnv } from './helpers/general.js';
 import { boundaryApiClientTest } from './helpers/boundary-api-client.js';
 import { LoginPage } from './admin/pages/login.js';
@@ -23,6 +25,17 @@ async function globalSetup() {
     'E2E_PASSWORD_AUTH_METHOD_ID',
   ]);
   await authenticateToBoundary();
+
+  // log in via CLI to ensure cache is running the correct version.
+  // when running locally, there's a chance that the cache could have an older
+  // version running. logging in via the CLI will update the cache to the latest version
+  await boundaryCli.checkBoundaryCli();
+  await boundaryCli.authenticateBoundary(
+    process.env.BOUNDARY_ADDR,
+    process.env.E2E_PASSWORD_AUTH_METHOD_ID,
+    process.env.E2E_PASSWORD_ADMIN_LOGIN_NAME,
+    process.env.E2E_PASSWORD_ADMIN_PASSWORD,
+  );
 }
 
 const authenticateToBoundary = async () => {
--- a/e2e-tests/global-setup.js
+++ b/e2e-tests/global-setup.js
@@ -5,7 +5,6 @@
 
 import { chromium, test as baseTest, mergeTests } from '@playwright/test';
 
-import * as boundaryCli from './helpers/boundary-cli';
 import { checkEnv } from './helpers/general.js';
 import { boundaryApiClientTest } from './helpers/boundary-api-client.js';
 import { LoginPage } from './admin/pages/login.js';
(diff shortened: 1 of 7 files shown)


e2e-tests/global-setup.js
```diff
@@ -23,6 +25,17 @@ async function globalSetup() {
     'E2E_PASSWORD_AUTH_METHOD_ID',
   ]);
   await authenticateToBoundary();
+
+  // log in via CLI to ensure cache is running the correct version.
+  // when running locally, there's a chance that the cache could have an older
+  // version running. logging in via the CLI will update the cache to the latest version
```
ZedLi: I didn't know they'd restart the cache, interesting. But if this is the intent, why not just stop the cache directly? Should this also only be scoped for desktop tests?

moduli: > why not just stop the cache directly? 

What do you mean by this? Call `boundary cache stop`? The desktop client wants the cache running, though, right?

> Should this also only be scoped for desktop tests?

Is there a good way to do this?

ZedLi: > What do you mean by this? Call boundary cache stop? The desktop client wants the cache running, though, right?

Yes, we automatically start the cache if it's not running and if it is, we just use the existing cache.

> Is there a good way to do this?

I think there's a few ways you could do it: maybe just a beforeAll hook for the tests? Or you could just do it in a fixture in our setup. 

If you truly wanted it ran only once globally before all tests, you could just make another setup in the desktop folder as a local wrapper and then call `globalSetup` from that one and just add the desktop specific stuff. Either that or do [project dependencies](https://playwright.dev/docs/test-global-setup-teardown#option-1-project-dependencies)

moduli: - During testing, it seems like the desktop client wasn't starting the cache. I assume it has something to do with us not including the CLI in the desktop builds? I added a stop cache command to ensure it's stopped and kept the authenticatidon
- I did split up the globalSetup by making another setup in the desktop folder.

ZedLi: > During testing, it seems like the desktop client wasn't starting the cache. I assume it has something to do with us not including the CLI in the desktop builds?

Is this in CI or local? IIRC how it's working right now (in CI) is it'll use the CLI on the path since we're treating this more like an "installer build" where we don't package the CLI directly in DC and just use the CLI we setup.

If this is local, you should package it in the DC using `SETUP_CLI=true` when building.  Having to restart the cache manually with the CLI seems like it's hiding the actual issue if it's not starting the cache 🤔.

e2e-tests/package.json
```diff
@@ -24,7 +24,8 @@
     "admin:dev:chrome": "pnpm run admin --project=chromium --headed",
     "admin:dev:firefox": "pnpm run admin --project=firefox --headed",
     "admin:dev:safari": "pnpm run admin --project=webkit --headed",
-    "desktop": "pnpm run --filter desktop build && pnpm run desktop:dev",
+    "desktop:local": "SETUP_CLI=true pnpm run --filter desktop build && pnpm run desktop:dev",
+    "desktop:ci": "pnpm run --filter desktop build && pnpm run desktop:dev",
```
ZedLi: I think this could've just stayed `desktop` and if you wanted the local one it could've just been `SETUP_CLI=true pnpm run desktop`

ZedLi (approved):

e2e-tests/global-setup.js
moduli: Summary of findings
- the desktop client will use the version of boundary in `boundary-ui/ui/desktop/electron-app/cli` if it exists in there. that file gets updated to the version specified in the desktop client if you build with `SETUP_CLI=true`
- for local, we'll rely on using `SETUP_CLI=true` and use that version of Boundary for most testing. if there's a new branch you want to test out, you can manually start that cache before running the tests.

PR #3239, opened 2026-04-14
Icu 18892 add scope suffix to model

sohail-hashicorp:
# Description
Added the model changes and included the mirage data for the project level alias.

:tickets: [Model Changes](https://hashicorp.atlassian.net/browse/ICU-18892)
:tickets: [Mirage Changes](https://hashicorp.atlassian.net/browse/ICU-18893)

## How to Test
NA

## Checklist

## PCI review checklist


addons/api/addon/models/scope.js
```diff
@@ -17,6 +18,11 @@ export const TYPES_SCOPE = Object.freeze([
 export default class ScopeModel extends GeneratedScopeModel {
   // =attributes
 
+  @attr('string', {
+    description: 'The alias target suffix for project scopes.',
+  })
+  alias_target_suffix;
+
```
cameronperera, line 25: This attribute should be defined in the `GeneratedScopeModel`.
Also is `alias_target_suffix` the correct name? In the API section of the RFC it shows `alias_suffix`.

addons/api/addon/models/alias.js
```diff
@@ -4,7 +4,16 @@
...
 export const TYPES_ALIAS = Object.freeze([TYPE_ALIAS_TARGET]);
-export default class AliasModel extends GeneratedAliasModel {}
+
+export default class AliasModel extends GeneratedAliasModel {
+  // =attributes
+
+  @attr('string', {
+    description: 'The base value provided by the user before suffix appending.',
+  })
+  base_value;
```
cameronperera, line 18: Same for this attribute. Should be defined in `GeneratedAliasModel`.

addons/api/app/mirage/factories/alias.js
```diff
@@ -19,4 +19,7 @@ export default factory.extend({
       'update',
       'delete',
     ],
+  scope() {
+    return { id: 'global', type: 'global' };
+  },
```
cameronperera, line 24:
```suggestion
  scope: () => ({ id: 'global', type: 'global' }),
```
Small suggestion to better align with how we define attributes with static data. I would probably add it to `GeneratedAliasModelFactory` as well.

addons/api/app/mirage/models/alias.js
```diff
@@ -8,4 +8,8 @@ import Model from './base';
 
 export default Model.extend({
   scope: belongsTo(),
+
+  get scopeId() {
+    return this.scope_id || this.scope?.id;
+  },
```
cameronperera: I am not sure this belongs here. Maybe it needs to be on the actual Alias model, not the mirage one?

addons/api/app/mirage/config.js
```diff
@@ -269,14 +269,54 @@ function routes() {
...
+
+      if (method === 'remove-alias-target-suffix') {
+        return scope.update({
+          alias_target_suffix: null,
+          version: attrs.version,
+        });
+      }
+
+      // Fallback for unknown methods
+      return scope;
```
cameronperera, line 291: Could you explain this change from the existing pattern? We normally update the `updatedAttrs` specific to the API action and return at the end after calling `update(updatedAttrs)` on the resource.

cameronperera (commented):
I have a small suggestion. If you look at the target mirage factory, you will see where we add aliases to the target. That might be a good place to create project level aliases and attach to a target.

addons/api/addon/models/scope.js
sohail-hashicorp: Yeah, you are right. Initially we had decided on the alias_target_suffix, but now its changed to alias_suffix. Have updated. Thanks for the catch. And I have updated all the attributes to the GeneratedScopeModels.

addons/api/app/mirage/models/alias.js
sohail-hashicorp: You are right, this should be added to the alias model.

addons/api/app/mirage/config.js
sohail-hashicorp: Agreed, I updated the alias to follow the `updatedAttrs` pattern

addons/api/app/mirage/factories/target.js
```diff
@@ -179,15 +179,27 @@ export default factory.extend({
...
-        aliases.length === 0 ? undefined : faker.helpers.arrayElement(aliases);
-      if (randomlySelectedAliases) {
-        target.update({
-          aliases: randomlySelectedAliases.aliases,
+      // Create aliases scoped to this target's scope and attach them to target.
+      const aliasCount = faker.number.int({ min: 1, max: 2 });
+      const aliases = Array.from({ length: aliasCount }, () => {
+        const baseValue = faker.internet.domainWord();
+        const fullValue =
+          scope.type === 'project' && scope.alias_suffix
```
cameronperera: To follow the pattern of using TYPE_CONST can we import the `TYPE_SCOPE_PROJECT` const from `api/models/scope.js`?

addons/api/app/mirage/config.js
cameronperera: The `return scope.update(updatedAttrs);` inside each `method` check should also be removed and simply return on line 291 with `return scope.update(updatedAttrs);`. I think we want to always call update on the scope before we return.

cameronperera (approved):
LGTM

addons/api/addon/serializers/alias.js
```diff
@@ -0,0 +1,27 @@
...
+   * @param {Snapshot} snapshot
+   * @return {object}
+   */
+  serialize(snapshot) {
+    const serialized = super.serialize(...arguments);
+
+    // Ensure base_value is included in the payload
+    if (snapshot.record.base_value !== undefined) {
+      serialized.base_value = snapshot.record.base_value;
+    }
```
ZedLi: Is this serializer needed? Doesn't the `base_value` just get serialized by default with the application serializer (`super.serialize()`)?

cameronperera: Thanks for pointing this out. That is correct, we don't need this and I just looked because I wasn't exactly sure but `base_value` is a READONLY field. So we need to add the `readOnly` flag to the generated model.
After that is changed, this serializer can also be deleted as I don't think we will need any custom serialization for it now.

addons/api/addon/generated/models/alias.js
```diff
@@ -50,6 +50,11 @@ export default class GeneratedAliasModel extends BaseModel {
   })
   value;
 
+  @attr('string', {
+    description: 'The base value provided by the user before suffix appending.',
```
cameronperera, line 54: add `readOnly: true` here since the API does not expect this value with creating or editing an alias.

cameronperera (changes requested):

addons/api/addon/serializers/alias.js
sohail-hashicorp: Good catch, thanks! 

Added `readOnly: true` to the `base_value` attribute

addons/api/addon/generated/models/alias.js
sohail-hashicorp: Done

cameronperera (approved):
Thanks for the quick change!

ZedLi (approved):

PR #3246, opened 2026-04-27
Grant actions card

cameronperera:
# Description

:tickets: [Jira ticket](https://hashicorp.atlassian.net/browse/ICU-18459)

https://github.com/user-attachments/assets/b1059ba6-3f7f-42c9-b5f4-a3c8cf76b858

## How to Test

Visit the edit grants route. You should now see suggested actions on the right of the code editor. The suggestions should work as before and the suggested actions should match the actions in the completion window. Only difference is the actions card will not remove actions as they are used like the completion list.

## Checklist

## PCI review checklist


ui/admin/app/components/form/role/edit-grants/index.js
```diff
@@ -58,6 +59,12 @@ export default class FormRoleEditGrantsComponent extends Component {
       activateOnCompletion: (completion) => completion.type === 'keyword',
     }),
     keymap.of(completionKeymap),
+    EditorView.updateListener.of((update) => {
+      if (update.selectionSet) {
+        const line = update.state.doc.lineAt(update.state.selection.main.head);
+        this.currentLineText = line.text;
```
ZedLi: Should we just have all updating happen here and remove the `onInput`?

ui/admin/app/components/form/role/edit-grants/index.js
```diff
@@ -47,7 +48,7 @@ export default class FormRoleEditGrantsComponent extends Component {
   );
 
   @tracked grantStringsText = (this.args.model?.grant_strings ?? []).join('\n');
-  @tracked currentLineText = this.args.model?.grant_strings?.[0] ?? '';
```
ZedLi, line 38: Shouldn't we just default it to the first line when loading before there's any focus? Any reason we want to show nothing on first load?

ui/admin/app/components/grant-actions/index.hbs
ZedLi, line 1: Should we be showing all actions on wildcard? We might want diverge from autocomplete to limit it to a different set in this case (maybe just default CRUDL like in figma?).

Also we should not make the card expand forever down and add scrolling if needed.

ui/admin/app/utils/grant-completions.js
```diff
@@ -234,6 +262,34 @@ const getActionOptions = (schema, typeValue, idsValue) => {
...
+  grantsSchema,
+  lineText = '',
+) => {
+  const schema = getNormalizedGrantsSchema(grantsSchema);
+  const { idsValue, typeValue } = parseGrantLine(lineText);
+
+  return getActionOptions(schema, typeValue, idsValue);
+};
+
+export const getDetectedResourceTypeForGrantLine = (
```
ZedLi: Where is this used?

ui/admin/app/utils/grant-completions.js
```diff
@@ -322,7 +375,7 @@ function grantCompletions(context, schema, translatedStrings) {
     const partial = enteredActions.pop() ?? '';
     const hasEnteredActions = enteredActions.length > 0;
     const options = filterByPrefix(
-      getActionOptions(schema, typeValue, idsValue),
+      getSuggestedActionsForGrantLine(schema, line.text),
```
ZedLi: Why was this change necessary? We already parsed the grant line earlier and now it looks like we just re-parse the line again and then call `getActionOptions` again?

ui/admin/app/utils/grant-completions.js
```diff
@@ -373,3 +426,28 @@ export const createGrantCompletionSource = (
 
   return (context) => grantCompletions(context, schema, translatedStrings);
 };
+
+export const createGrantLineHelpers = (grantsSchema) => {
```
ZedLi: Is this needed here? It seems like it's a thin wrapper around the other exported functions. It also looks like you have a normalized schema in your component but still normalize the schema here.

ui/admin/app/utils/grant-completions.js
```diff
@@ -84,6 +100,18 @@ const normalizeGrantsSchema = (grantsSchema) => {
   };
 };
 
+const isNormalizedGrantsSchema = (grantsSchema) =>
```
ZedLi: Why would we need this check?

ui/admin/app/components/grant-actions/index.js
```diff
@@ -0,0 +1,172 @@
...
+import { service } from '@ember/service';
+
+import {
+  createGrantLineHelpers,
+  getCompatibleResourceTypeForIds,
+  normalizeGrantsSchema,
+  parseGrantLine,
+} from 'admin/utils/grant-completions';
+
+const ACTIONS_WITH_RESOURCE_TYPE = new Set([
```
ZedLi: What is the purpose of this? Why do we just reflect back the resource type?

ui/admin/app/components/grant-actions/index.js
```diff
@@ -0,0 +1,172 @@
...
+
+export default class GrantActionsIndex extends Component {
+  @service intl;
+
+  #normalizedSchema = normalizeGrantsSchema(this.args.grantsSchema ?? {});
+
+  #grantLineHelpers = createGrantLineHelpers(this.args.grantsSchema ?? {});
+
+  get parsedGrantLine() {
+    return parseGrantLine(this.args.grantString);
```
ZedLi: This will parse the grant line _every_ time this getter is called which seems inefficient. It looks like this can  get called a lot of times the way we've structured these getters (over 10 times by my count). I don't think it's a realistic performance issue since parsing should be relatively quick but if it ever changes we'll get caught by this. It might be worth adding the `cached` decorator here.

Though I'm curious if this indicates a more structural issue. We have a lot of getters (which I find a bit hard to read). I wonder if most of this logic should have actually been exported in an object out of a util function (whether it's autocomplete or linting) which at least lets a reader to see the different variables and logic at once instead of jumping all over.

ZedLi (commented):
Looks cool! Did an initial pass of some of the code

ui/admin/app/components/grant-actions/index.js
```diff
@@ -0,0 +1,172 @@
...
+    }
+
+    return this.parsedGrantLine.idsValue
+      .split(',')
+      .filter(Boolean)
+      .some((id) => id.startsWith('{{') && id.endsWith('}}'));
+  }
+
+  get compatibleIdsResourceType() {
+    if (!this.hasSpecificIds || this.hasTemplateIds) {
```
lisbet-alvarez: if `ids` has a valid template id we should still be able to list action suggestions

ui/admin/app/components/grant-actions/index.js
```diff
@@ -0,0 +1,172 @@
...
+  get hasInvalidType() {
+    return (
+      this.hasExplicitType &&
+      !this.#normalizedSchema.resourcesByType[this.parsedGrantLine.typeValue]
+    );
+  }
+
+  get hasInvalidIds() {
+    return (
+      this.hasSpecificIds &&
```
lisbet-alvarez: perhaps im misunderstanding here but should this be `!this.hasSpecificIds`

ui/admin/app/components/grant-actions/index.js
```diff
@@ -0,0 +1,172 @@
...
+  }
+
+  get descriptionResourceType() {
+    if (this.parsedGrantLine.typeValue === '*') {
+      return null;
+    }
+
+    return (
+      this.parsedGrantLine.typeValue ||
+      this.#grantLineHelpers.getDetectedResourceType(this.args.grantString)
```
lisbet-alvarez: is this function essentially doing the same thing as `getCompatibleResourceTypeForIds`

ui/admin/app/components/form/role/edit-grants/index.js
cameronperera: We should default. I changed this as I didn't like it listing all actions on load but that was simply due to the first grant being all wildcard values. Reverting back.

ui/admin/app/components/form/role/edit-grants/index.js
cameronperera: That is a good suggestion. I updated this to handle updating the `currentLineText` and the `grantStringsText` like `onInput`.

ui/admin/app/components/grant-actions/index.hbs
cameronperera: Totally missed the wildcard example in the mocks. I updated that and added scrolling. Much nicer for resources like targets.

ui/admin/app/components/grant-actions/index.js
cameronperera: After our discussion today, I made a larger refactor and utilized a cached decorator like you mentioned. It will be pushed up with other changes shortly.

cameronperera:
Some of the feedback is no longer relevant after a large refactor so I will resolve them now. I am not ignoring them. 😃

ui/admin/app/components/form/role/edit-grants/index.js
```diff
@@ -150,10 +151,20 @@ export default class FormRoleEditGrantsComponent extends Component {
...
+      options,
+    );
+  };
+  translateCompletionString = (key, options = {}) => {
+    if (this.isDestroying || this.isDestroyed) return '';
+    return this.intl.t(
+      `resources.role.edit-grants.completion-info.${key}`,
+      options,
+    );
+  };
```
cameronperera: This change was to help with a failing test where the `intl` was being called after test teardown. I had trouble finding a fix. This is a suggestion from Claude and was consistent with test runs. Happy to discus a different approach.

cameronperera: Fixed it. I messed up during the conflict resolutions and removed the `onSetup` that was previously added to resolve this issue.

ui/admin/app/utils/grant-completions.js
```diff
@@ -335,17 +407,18 @@ async function grantCompletions(
     };
   }
 
-  const idsValueMatch = beforeCursor.match(/ids=([\w*.{},\s-]*)$/);
+  const idsValueMatch = beforeCursor.match(/ids=([\w*.{}, -]*)$/);
```
lisbet-alvarez: shouldn't we still be checking for  `\s` any whitespace

addons/core/translations/resources/en-us.yaml
```diff
@@ -663,6 +663,7 @@ role:
   edit-grants:
     title: Edit Grants
     description: Modify existing grant strings or add new grant strings to this role.
+    no-suggestions: No suggestions
```
ZedLi: Isn't this the same as the one two lines below?

ui/admin/app/components/form/role/edit-grants/index.js
```diff
@@ -150,10 +151,18 @@ export default class FormRoleEditGrantsComponent extends Component {
       .map(({ id, name }) => ({ id, name }));
   };
 
-  translateLintingError = (key, options = {}) =>
-    this.intl.t(`resources.role.edit-grants.linting-errors.${key}`, options);
-  translateCompletionString = (key, options = {}) =>
-    this.intl.t(`resources.role.edit-grants.completion-info.${key}`, options);
+  translateLintingError = (key, options = {}) => {
```
ZedLi: Can we remove the braces again for these functions?

ui/admin/app/components/form/role/edit-grants/index.js
```diff
@@ -172,6 +181,15 @@ export default class FormRoleEditGrantsComponent extends Component {
       // Trigger autocompletion when the user completes a grant field (which we labeled as keywords)
       activateOnCompletion: (completion) => completion.type === 'keyword',
     }),
+    EditorView.updateListener.of((update) => {
+      if (update.docChanged || update.selectionSet) {
+        const line = update.state.doc.lineAt(update.state.selection.main.head);
+        this.currentLineText = line.text;
+      }
+      if (update.docChanged) {
```
ZedLi: Does this make more sense to be nested inside the first?

ui/admin/app/utils/grant-completions.js
```diff
@@ -22,7 +23,7 @@ const getNoSuggestionsOption = (noSuggestionsLabel) => ({
   apply: () => {},
 });
 
-const parseGrantFields = (lineText) =>
+export const parseGrantFields = (lineText = '') =>
```
lisbet-alvarez: looks like this function is only being used within the grants completions util, what is the purpose of exporting it

ZedLi: Are these being exported just for unit tests? I normally would suggest we just test the public API and consider these impelementation details.

ui/admin/app/utils/grant-completions.js
```diff
@@ -246,6 +263,74 @@ const getActionOptions = (schema, typeValue, idsValue) => {
...
+  const hasLiteralIds = hasSpecificIds && !hasTemplateIds;
+
+  const compatibleIdsResourceType = hasLiteralIds
+    ? getCompatibleResourceTypeForIds(schema, idsValue)
+    : null;
+
+  const hasInvalidType = hasExplicitType && !schema.resourcesByType[typeValue];
+  const hasInvalidIds = hasLiteralIds && !compatibleIdsResourceType;
+
+  const hasInvalidPinnedIdTypeCombination = (() => {
```
ZedLi: Why the usage of IIFEs in this function?

ui/admin/app/utils/grant-completions.js
```diff
@@ -246,6 +263,74 @@ const getActionOptions = (schema, typeValue, idsValue) => {
...
+    if (
+      !hasSpecificIds ||
+      !hasExplicitType ||
+      hasInvalidIds ||
+      hasInvalidType
+    ) {
```
ZedLi: Could we have just de morgan's law'ed this and chained it with the child type check?

ui/admin/app/utils/grant-completions.js
```diff
@@ -246,6 +263,74 @@ const getActionOptions = (schema, typeValue, idsValue) => {
...
+    if (!idsValue) return null;
+    return getCompatibleResourceTypeForIds(schema, idsValue);
+  })();
+
+  const actions =
+    idsValue || typeValue
+      ? getActionOptions(schema, typeValue, idsValue)
+          .filter(
+            (action) =>
+              action !== '*' && (!isWildcard || CRUDL_ACTIONS.has(action)),
```
ZedLi: This will also filter out valid pinned actions instead of showing all possible pinned actions. e.g. `ids=hcst_1234567890;type=*;actions=`

Same for wildcard IDs e.g. `ids=*;type=host-set;actions=`

ui/admin/app/utils/grant-completions.js
```diff
@@ -246,6 +263,74 @@ const getActionOptions = (schema, typeValue, idsValue) => {
...
+          )
+          .sort((left, right) => left.localeCompare(right))
+      : [];
+
+  return {
```
ZedLi, line 312: Why such a large object, did we need all of these? Can we have just returned the actual ones we need for grant actions card?

ui/admin/app/utils/grant-completions.js
```diff
@@ -362,36 +435,40 @@ async function grantCompletions(
...
-    if (getIsLoading()) {
-      allOptions.push({
-        label: translate('loading-ids'),
-        type: 'text',
-        apply: () => {},
-      });
-    }
+    const from = context.pos - partial.length;
+
+    return (async () => {
```
ZedLi: Why was this also changed to an IIFE?

lisbet-alvarez (dismissed):
Great work! this came together rly nicely!

ui/admin/app/utils/grant-completions.js
```diff
@@ -246,6 +263,74 @@ const getActionOptions = (schema, typeValue, idsValue) => {
     : actionOptions;
```
ZedLi: I think this should actually be `[]`, I don't think we should be showing all actions if we don't have a valid selected resource

addons/core/translations/resources/en-us.yaml
cameronperera: oh good catch, I removed mine and referenced yours in the template

ui/admin/app/components/form/role/edit-grants/index.js
cameronperera: yeah I can move it. Looking over it now it reads more like an if/else but it isn't really. nesting now

ui/admin/app/utils/grant-completions.js
cameronperera: yeah that is a good point. I think I was exporting before the larger refactor but I will remove it and modify the tests to make sure it is properly covered.

cameronperera: See reply to Zed's but I had the export for tests and I will be fixing that now.

ui/admin/app/utils/grant-completions.js
cameronperera: Good catch. I see where we are doing exactly that in other places as well. Updated.

ui/admin/app/utils/grant-completions.js
cameronperera: Mainly to return early but I like the readability after using the de morgan's law pattern.

ui/admin/app/utils/grant-completions.js
cameronperera: Gotcha, I see it now. Updated.

ui/admin/app/utils/grant-completions.js
cameronperera: sorry, after refactoring the edit-grants component I did not come back and refactor this out which originally had all of these used. As I simplified, less and less were needed. Updating the return now.

ui/admin/app/utils/grant-completions.js
cameronperera: Great question that I don't have a solid answer for. I was using copilot to help me refactor after pulling your changes and I think is a style choice it did and I didn't catch. I have it updated now though.

ui/admin/app/utils/grant-completions.js
cameronperera: sorry, I think this was a conflict mistake. Fixed it.

lisbet-alvarez (approved):
Looks great!

ui/admin/app/components/grant-actions/index.js
```diff
@@ -0,0 +1,69 @@
...
+      (hasInvalidType || hasInvalidIds || hasInvalidPinnedIdTypeCombination)
+    );
+  }
+
+  getActionDescription(action) {
+    const { detectedResourceType } = this.grantAnalysis;
+    const specificKey = `resources.role.edit-grants.actions.${action}`;
+    const genericKey = `resources.role.edit-grants.actions.generic.${action}`;
+
+    if (detectedResourceType && this.intl.exists(specificKey)) {
```
ZedLi, line 55: I just wanted to point out one slightly misleading description happens when doing pinned IDs:

Image

The CRUDL actions are actually on the children resources (hosts and host-sets)

addons/core/translations/resources/en-us.yaml
```diff
@@ -753,6 +753,76 @@ role:
       options:
         terraform: Terraform
         native-hcl: Native HCL
+    actions:
+      collection-actions: Collection Actions
+      id-actions: ID Actions
+      create: Create a new {resourceType}
+      list: List all {resourceType} resources
+      read: Read a {resourceType}
```
ZedLi, line 761: Nitpick but if the resource type starts with a vowel, it reads a little strange when it's `a` instead of `an`.

ZedLi (approved):
Looks good, just had a few minor comments!

PR #3247, opened 2026-04-27
Icu 18506 load resource IDs for autocomplete

ZedLi:
# Description

Added logic to handle autocompletion of IDs loaded dynamically from the DB. You can search by both name and ID which are available filtered by the allowed scope IDs on the role. 

https://github.com/user-attachments/assets/ff57bad7-fe3f-4bf6-aca8-e61515a92297

## How to Test
Go to the role you're testing and try different scopes on the role. Try editing grants and see what ids are available. 

## Checklist

## PCI review checklist

### DIFF
--- a/addons/api/addon/handlers/sqlite-handler.js
+++ b/addons/api/addon/handlers/sqlite-handler.js
@@ -62,14 +62,35 @@ export default class SqliteHandler {
           totalInsert = 0;
 
         if (!peekDb) {
-          const tokenKey = `${type}-${hashCode(remainingQuery)}`;
+          let tokenKey = `${type}-${hashCode(remainingQuery)}`;
+          // The global token key represents a recursive query from global scope,
+          // which pulls back every resource of this type.
+          const globalTokenKey = `${type}-${hashCode({ scope_id: 'global', recursive: true })}`;
           if (storeToken) {
-            const [tokenObj] = await this.sqlite.fetchResource({
-              ...generateSQLExpressions('token', {
-                filters: { id: [{ equals: tokenKey }] },
-              }),
-            });
-            listToken = tokenObj?.token;
+            // Always prefer the global token when it exists for
+            // any more fine-grained query. The global query already fetched every
+            // item for this resource type, so we should just re-use it.
+            if (tokenKey !== globalTokenKey) {
+              const [globalTokenObj] = await this.sqlite.fetchResource({
+                ...generateSQLExpressions('token', {
+                  filters: { id: [{ equals: globalTokenKey }] },
+                }),
+              });
+              if (globalTokenObj?.token) {
+                tokenKey = globalTokenKey;
+                listToken = globalTokenObj.token;
+              }
+            }
+
+            // Only look up the specific token if no global token was found.
+            if (!listToken) {
+              const [tokenObj] = await this.sqlite.fetchResource({
+                ...generateSQLExpressions('token', {
+                  filters: { id: [{ equals: tokenKey }] },
+                }),
+              });
+              listToken = tokenObj?.token;
+            }
           } else {
             // This is a temporary fix of clearing the DB (specifically for auth-methods)
             // since we are not storing the token we do not get back a list of removed_ids
--- a/addons/rose/addon/styles/hds/themes/dark-mode/index.scss
+++ b/addons/rose/addon/styles/hds/themes/dark-mode/index.scss
@@ -191,6 +191,47 @@
   }
 }
 
+// Overrides for CodeMirror 6 autocomplete and lint tooltips/panels.
+@mixin dark-mode-codemirror-override {
+  /* stylelint-disable selector-class-pattern */
+  .cm-editor {
+    // Autocomplete dropdown and lint hover tooltips
+    .cm-tooltip {
+      background-color: var(--token-color-palette-neutral-50);
+      border: 1px solid var(--token-color-palette-neutral-100);
+      color: var(--token-color-palette-neutral-400);
+    }
+
+    // Arrow pointer for tooltips shown above the cursor
+    .cm-tooltip-above::before {
+      border-top-color: var(--token-color-palette-neutral-50);
+    }
+
+    // Arrow pointer for tooltips shown below the cursor
+    .cm-tooltip-below::before {
+      border-bottom-color: var(--token-color-palette-neutral-50);
+    }
+
+    // Autocomplete list > selected item highlight
+    .cm-tooltip-autocomplete > ul > li[aria-selected] {
+      background-color: var(--token-color-palette-blue-300);
+      color: var(--token-color-palette-neutral-700);
+    }
+
+    // Lint diagnostics panel (shown at the bottom of the editor)
+    .cm-panel.cm-panel-lint ul {
+      background-color: var(--token-color-palette-neutral-0);
+    }
+
+    // Lint panel selected diagnostic item
+    .cm-panel.cm-panel-lint ul:focus li.cm-diagnostic[aria-selected] {
+      background-color: var(--token-color-palette-neutral-50);
+      color: var(--token-color-palette-neutral-400);
+    }
+  }
+  /* stylelint-enable selector-class-pattern */
+}
+
 @media (prefers-color-scheme: dark) {
   .ember-application:not(.rose-theme-light) {
     @include dark-mode;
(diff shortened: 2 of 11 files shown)


addons/api/addon/handlers/sqlite-handler.js
```diff
@@ -62,14 +62,35 @@ export default class SqliteHandler {
...
+          // which pulls back every resource of this type.
+          const globalTokenKey = `${type}-${hashCode({ scope_id: 'global', recursive: true })}`;
           if (storeToken) {
-            const [tokenObj] = await this.sqlite.fetchResource({
-              ...generateSQLExpressions('token', {
-                filters: { id: [{ equals: tokenKey }] },
-              }),
-            });
-            listToken = tokenObj?.token;
+            // Always prefer the global token when it exists for
```
ZedLi, line 70: This is something I was thinking of doing if we ever introduced global searching. With the loading of all resources for autocomplete, it's basically at that point. My main worry was the churn of loading the same resources but at different scopes. 

For example, if i've loaded all targets from the global scope recursively already (i.e. all targets), do I need to re-load and hit the API for all scopes in a specific project? Ideally I should just re-use the global scope list token and check if there were any new ones and just the the sqlite DB properly filter down the right targets.

Let me know if you have any better ideas or thoughts/concerns about this pattern!

ui/admin/app/routes/scopes/scope/roles/role/edit-grants.js
```diff
@@ -13,15 +23,32 @@ export default class ScopesScopeRolesRoleEditGrantsRoute extends Route {
...
+   */
+  loadResourcesTask = dropTask(async (scopeId) => {
+    await Promise.allSettled(
+      NAMED_RESOURCE_TYPES.map((type) =>
+        this.store.query(type, { scope_id: scopeId, recursive: true }),
+      ),
+    );
+  });
+
+  @notifyError(({ message }) => message, { catch: true })
```
lisbet-alvarez, line 51: nice improvement, using notify service is more user friendly than doing a try/catch

ui/admin/app/components/form/role/edit-grants/index.js
```diff
@@ -118,6 +228,11 @@ export default class FormRoleEditGrantsComponent extends Component {
 
   // =actions
 
+  @action
+  onSetup(view) {
+    this.#editorView = view;
+  }
```
lisbet-alvarez, line 234: Nice catch!!
For my understanding, what is the context behind the need to destroy editorView for this code editor?

ui/admin/app/routes/scopes/scope/roles/role/edit-grants.js
```diff
@@ -5,6 +5,16 @@
...
 import Route from '@ember/routing/route';
 import { service } from '@ember/service';
+import { dropTask } from 'ember-concurrency';
+import { modelMapping } from 'api/services/sqlite';
+import { notifyError } from 'core/decorators/notify';
+
+// Only resource types that have a name column are useful for ID suggestions
+// Currently this excludes session-recording and sessions as they aren't likely to be used
+// and could be expensive to load.
+export const NAMED_RESOURCE_TYPES = Object.keys(modelMapping).filter(
```
lisbet-alvarez: not sure if i'm missing the use case, but why is this`NAMED_RESOURCE_TYPES` being exported? I only see it being used in this route file

addons/api/addon/handlers/sqlite-handler.js
cameronperera: I want to make sure I am understanding this. With this change, if I fetch all scopes using the global scope recursively, we would re-use the list token which would only need to return new scopes correct?

addons/rose/addon/styles/hds/themes/dark-mode/index.scss
```diff
@@ -191,13 +191,55 @@
...
+
+    // Autocomplete list > selected item highlight
+    .cm-tooltip-autocomplete > ul > li[aria-selected] {
+      background-color: var(--token-color-palette-blue-300);
+      color: var(--token-color-palette-neutral-700);
+    }
+
+    // Lint diagnostics panel (shown at the bottom of the editor)
+    .cm-panel.cm-panel-lint ul {
+      background-color: var(--token-color-palette-neutral-0);
```
lisbet-alvarez: looks like `border-top` for the panel also needs to be overriden

addons/api/addon/handlers/sqlite-handler.js
ZedLi: Yes, but that's already how it worked. This is for the scenario where we already loaded all scopes using the global scope and I now fetch scopes on an org, it should just re-use the previously loaded scopes since it's a super set.

cameronperera: oh gotcha. Thanks for the small clarification. I see where I missed that detail. "do I need to re-load and hit the API for all scopes in a specific project"

ui/admin/app/routes/scopes/scope/roles/role/edit-grants.js
lisbet-alvarez, line 58: i think this should be a translated string

lisbet-alvarez (commented):
This is looking really good! It's impressive how auto-complete has come together!

ui/admin/app/components/form/role/edit-grants/index.js
ZedLi: I think it's a proper cleanup step regardless but I was encountering flaky unit tests because it looks like linting/autocomplete was running after the component got tear downed which ended up in something like `Uncaught Error: Cannot call '.lookup('service:intl')' after the owner has been destroyed`.

I suspect it's bc the linter has a [default delay](https://codemirror.net/docs/ref/#lint.linter^config.delay) of 750ms and some tests could run faster than that.

ui/admin/app/routes/scopes/scope/roles/role/edit-grants.js
ZedLi: Good catch, I originally was exporting it and using it in two places rather than passing it down, I'll update it

ui/admin/app/routes/scopes/scope/roles/role/edit-grants.js
ZedLi: Hmm that's an interesting point, we haven't really translated our error messages previously 🤔

cameronperera: I would leave it as is and we can maybe discuss as a team if we want to address this.

addons/rose/addon/styles/hds/themes/dark-mode/index.scss
ZedLi: I also changed the border between the gutter and content, I didn't notice that was also off

cameronperera (dismissed):
Nice work!

ui/admin/app/routes/scopes/scope/roles/role/edit-grants.js
lisbet-alvarez: yup lets just leave it as is. We do just pass through the error messages. I was looking to the instances of notifyError where we pass through a translated string instead of the error message being thrown.

lisbet-alvarez (dismissed):
Awesome work!!

ui/admin/app/components/form/role/edit-grants/index.js
```diff
@@ -17,50 +17,155 @@ import {
...
+  @tracked grantStringsText = (this.args.model?.grant_strings ?? []).join('\n');
+  @tracked currentLineText = this.args.model?.grant_strings?.[0] ?? '';
+  @tracked showExportOptionsFlyout = false;
+  @tracked selectedExportOption = this.exportOptions[0];
+
+  /**
+   * Returns the scope IDs that should be used to filter ID suggestions.
+   * @returns {Promise<string[]|null>}
+   */
+  async getAllowedScopeIds() {
```
ZedLi, line 46: Are there any gotchas limiting resources to the role scopes? One thing I was thinking of was possibly users? If a user exists in global or org level I'd imagine you'd be able to choose it in a more granular scope? But we also have templates so I'm not sure if it's an issue.

Are there other resources?

lisbet-alvarez: Just to clarify, since users only live on global/org level would attempting to set a grant for a specific user id at a project level have any effect because it would be outside of the project's grant scope?

ZedLi: Maybe not, I might have just been thinking of actually attaching principals to the role instead of permissioning for the user.

cameronperera (approved):

lisbet-alvarez (approved):

PR #3248, opened 2026-04-30
Adds alias suffix on the Project-Settings page.

sohail-hashicorp:
# Description
Adds support for project-scope alias suffixes via the new `set-alias-target-suffix`, `get-alias-target-suffix`, and `remove-alias-target-suffix` APIs.

The settings page for a project now has an `Alias suffix` sidebar similar to how we have `Storage Policy` in org-settings, where you can create a suffix, edit it, or remove it.

Also added a 127-char on the input.

:tickets: [Add Suffix Sidebar](https://hashicorp.atlassian.net/browse/ICU-18894), [Add Suffix Read flow](https://hashicorp.atlassian.net/browse/ICU-18895), [Add Suffix Create flow](https://hashicorp.atlassian.net/browse/ICU-18896), [Add Suffix Edit flow](https://hashicorp.atlassian.net/browse/ICU-18897), [Add Suffix Delete flow](https://hashicorp.atlassian.net/browse/ICU-18898)

https://github.com/user-attachments/assets/dfcf4bbe-bf11-4bdf-9053-f4fa98480ccf

## How to Test
1. `pnpm start:mirage` in `ui/admin`.
2. Navigate to any project scope's settings page, you should see the "Alias suffix" panel on the right.
3. Click **Create Alias suffix**, enter a value, save. You should see the "Alias suffix saved" toast.
4. From the sidebar dropdown, click **Edit suffix**, change the value, and save.
5. From the sidebar dropdown, click **Delete suffix**. 

## Checklist

## PCI review checklist


addons/api/tests/unit/serializers/scope-test.js
cameronperera, line 1: thanks for adding this missing serializer unit test file for scope

addons/core/translations/resources/en-us.yaml
```diff
@@ -1071,6 +1071,25 @@ scope:
     global: Global
     org: Org
     project: Project
+  alias-suffix:
+    title: Alias suffix
+    description: Append this suffix to aliases in this project scope. The alias will appear "alias.customsuffix"
```
cameronperera: non-blocking: Does `The alias will appear "alias.customsuffix"` reads better as `The alias will appear as "alias.customsuffix"`

addons/core/translations/resources/en-us.yaml
```diff
@@ -1071,6 +1071,25 @@ scope:
...
+    form:
+      helper-text: Append this suffix to aliases in this project scope. The alias will appear "alias.customsuffix" Leave empty if no suffix is needed. To delete the suffix, clear the text field.
```
cameronperera: This helper text should be updated since we changed the delete workflow.

ui/admin/app/abilities/scope.js
cameronperera, line 1: These aliases look great but I think they can be added to the scope abilities in `/addons/api` here `/boundary-ui/addons/api/addon/abilities/scope.js`. We normally use the abilities in `ui/admin` for abilities that need to verify if specific features are enabled as well.

ui/admin/app/controllers/scopes/scope/add-alias-suffix.js
```diff
@@ -0,0 +1,66 @@
...
+  @service router;
+  @service confirm;
+  @service intl;
+
+  // =properties
+
+  /**
+   * @type {boolean}
+   */
+  isEdit = false;
```
cameronperera, line 24: non-blocker: Couldn't we rely on `scope.hasSuffix` instead of using `isEdit`?

ui/admin/app/styles/app.scss
```diff
@@ -855,6 +855,28 @@
   }
 }
 
+// alias suffix sidebar
```
cameronperera: Seems like some of the styles for the sidebar are duplicated. Maybe we can create a single `scope-sidebar` class?

ui/admin/tests/acceptance/scopes/alias-suffix-test.js
```diff
@@ -0,0 +1,171 @@
...
+    await click(commonSelectors.MODAL_WARNING_CANCEL_BTN);
+
+    assert.strictEqual(
+      this.server.schema.scopes.find(instances.scopes.project.id).alias_suffix,
+      SUFFIX_VALUE,
+    );
+    assert.dom(selectors.ALIAS_SUFFIX_VALUE).hasText(SUFFIX_VALUE);
+  });
+
+  test('renders the alias suffix on load', async function (assert) {
```
cameronperera, line 191: Could we add a test for an error state when fetching the alias suffix fails?

cameronperera (changes requested):
Overall it looks great. I left a few comments and some non-blocking ones.

ui/admin/app/controllers/scopes/scope/add-alias-suffix.js
sohail-hashicorp: `scope.hasSuffix` is reactive to the form input, so it would change during create and incorrectly show the edit title and confirm modal. Hence took this approach.

addons/core/translations/resources/en-us.yaml
sohail-hashicorp: Yeah, have updated it.

ui/admin/app/abilities/scope.js
sohail-hashicorp: Gotcha. Have updated the changes.

ui/admin/tests/acceptance/scopes/alias-suffix-test.js
sohail-hashicorp: Sure, have added it.

ui/admin/tests/acceptance/scopes/alias-suffix-test.js
```diff
@@ -168,4 +169,24 @@ module('Acceptance | scopes | alias suffix', function (hooks) {
 
     assert.dom(selectors.ALIAS_SUFFIX_VALUE).hasText(SUFFIX_VALUE);
   });
+
+  test('renders empty sidebar gracefully when fetching the alias suffix fails', async function (assert) {
```
cameronperera, line 199: non-blocking: Should we show the create button if we fail to fetch the suffix value? I think this is something we can discuss with design/PM and doesn't need to be addressed in this PR.

cameronperera (dismissed):
LGTM

ui/admin/app/routes/scopes/scope/add-alias-suffix.js
```diff
@@ -0,0 +1,51 @@
...
+   */
+  async afterModel(scope) {
+    if (this.abilities.can('getAliasSuffix scope', scope)) {
+      try {
+        await this.store.findRecord('scope', scope.id, {
+          adapterOptions: { method: 'get-alias-target-suffix' },
+          reload: true,
+        });
+      } catch {
+        // do nothing
```
ZedLi: Do we actually want to do nothing in this case? Shouldn't this be a handled error since we need to display the suffix on this page?

ui/admin/app/controllers/scopes/scope/add-alias-suffix.js
```diff
@@ -0,0 +1,66 @@
...
+      try {
+        await this.confirm.confirm(
+          this.intl.t('resources.scope.alias-suffix.questions.edit.message'),
+          {
+            title: 'resources.scope.alias-suffix.questions.edit.title',
+            confirm: 'actions.confirm',
+          },
+        );
+      } catch {
+        // do nothing
```
ZedLi: If we catch this canceled confirmation, won't it still show a successful saved confirmation?

cameronperera: Ope, it does. Missed that during my testing.

ui/admin/app/routes/scopes/scope/add-alias-suffix.js
cameronperera: I left a comment on the test file but I have a similar question of what should happen if we get an error from the API.

ZedLi: It looked like that test was for showing the suffix on the project page which _might_ be OK to catch the error since we might not want it to error the page but this particular one is for the actual `add-alias-suffix` page which if it fails, there's not much for us to do (and might be misleading if it just shows as empty instead of the actual value).

cameronperera: Gotcha. Following now. Maybe something like a flash message for the error along with a short message in the sidebar suggesting a refresh or "oops error" (not that of course)?

ui/admin/tests/acceptance/scopes/alias-suffix-test.js
sohail-hashicorp: Yeah, good point. Will check with the design.

ui/admin/app/routes/scopes/scope/add-alias-suffix.js
sohail-hashicorp: Thats a good catch! Have added the changes.

ui/admin/app/controllers/scopes/scope/add-alias-suffix.js
sohail-hashicorp: hmmm, I had missed this too. I have the fix for this.

cameronperera (dismissed):

ui/admin/app/routes/scopes/scope/add-alias-suffix.js
ZedLi: I don't think we typically add a toast for these type of errors and let the API errors propagate through for an error application state to handle it (we should be handling API errors like 500 I believe?)

sohail-hashicorp: You're right. I tested it locally with mirage returning 500 and 404, and the error substate already handles it nicely. Removed the toast and updated the test. Thanks!

ZedLi (approved):
Looks good!

cameronperera (approved):

priya-patel04 (approved):
Looks great!

PR #3257, opened 2026-05-06
feat: 🎸 Handle rate limiting

ZedLi:
✅ Closes: https://hashicorp.atlassian.net/browse/ICU-12035

# Description
Added handling of 429 responses from the API which means we got rate limited. There are two scenarios to be aware of:

1. We got rate limited while trying to pull back all the data for a resource. In this case, we'd like to at least show what was currently loaded with a toast indicating that the user got rate limited and when they can refresh. In the figma, it was suggested that we use an alert banner but I ended up using a toast because the former felt like it only makes sense in the context of a list page view with a table. For other pages like going to an individual resource or on the new edit grants workflow, the banner makes less sense. The other aspect is that a 429 could happen on any request and it's easier to use a toast than try and add an alert to every page. 
2. For whatever reason, we got rate limited on the first request before we even got any data and there's nothing to load from the cache (it was fresh and we didn't have a `list_token` yet). There's not much we can do to show any data yet so in this we just propagate the error and let our normal substate error page handle it. 

Of course, in scenarios where we can't even load scopes, our app might just error out ungracefully since we require scopes do anything. 

Screenshot 2026-05-05 at 6 31 55 PM

https://github.com/user-attachments/assets/825cdfd3-6a22-4e83-a90a-62b43cbb71f3

https://github.com/user-attachments/assets/2dd8e4bc-5c5a-4c54-a649-4f4acc4a8bc6

## How to Test
Start a boundary instance with custom rate limiting (or add enough resources). This can be accomplished with [configs](https://developer.hashicorp.com/boundary/docs/api/rate-limiting) or by updating the [default values](https://github.com/hashicorp/boundary/blob/main/internal/ratelimit/config.go#L44-L53) in boundary. Test the results you get back from getting rate limited on different pages. 

## Checklist

## PCI review checklist


DhariniJeeva:
do we want to show the rate limit alert in desktop too? or is this admin only for now?

addons/core/addon/components/error/message/index.js
```diff
@@ -12,14 +12,15 @@ import Component from '@glimmer/component';
...
+ *   - 429 - `error.isRateLimited`:  the client has exceeded the rate limit
  *   - 500 - `error.isServer`:  an internal server error occurred
  *
  * For an unknown error state, i.e error state not matching to the above defined list:
  *   - unknown - `error.isUnknown`:  an error occurred, but we don't know which or
  *            we don't distinguish it yet
  */
 
-const statuses = ['401', '403', '404', '500'];
+const statuses = ['401', '403', '404', '429', '500'];
```
DhariniJeeva, line 23: could we add a test in `tests/integration/components/error/message-test.js` for 429?

addons/api/addon/adapters/application.js
```diff
@@ -131,11 +131,34 @@ export default class ApplicationAdapter extends RESTAdapter.extend(
...
+        // data has been accumulated so far rather than throwing and leaving
+        // the user stuck with nothing to show.
+        // Without a list_token this is the very first request with nothing to
+        // show, so rethrow and let normal error handling apply.
+        if (e?.errors?.[0]?.isRateLimited && queryObject.list_token) {
+          result ??= {};
+          result.items = data;
+          result.response_type = 'complete';
+          result.meta = {
+            ...result.meta,
```
lisbet-alvarez, line 147: this should be updated since `...result.meta` could error out if result = {}

addons/api/addon/adapters/application.js
```diff
@@ -223,7 +246,26 @@ export default class ApplicationAdapter extends RESTAdapter.extend(
...
+    const response = super.handleResponse(
+      status,
+      headers,
+      payload,
+      requestData,
+    );
+
+    // Attach rate-limit metadata from response headers so downstream
+    // error handlers can inform the user when they may retry.
+    if (status === 429 && response?.errors) {
```
lisbet-alvarez: in what cases would we not have the errors object if there is 429

lisbet-alvarez (commented):
Great work!! this is a fantastic improvement!

addons/api/addon/adapters/application.js
ZedLi: I believe it should be fine, you can spread an undefined variable in an object vs it erroring out in an array, see this [explanation](https://stackoverflow.com/questions/47155141/spreading-undefined-in-array-vs-object)

addons/core/addon/components/error/message/index.js
ZedLi: Good point, added a test

ZedLi:
> do we want to show the rate limit alert in desktop too? or is this admin only for now?

We could but I didn't add anything for desktop for now since the assumption was we'd be only going through cache daemon still which handles it for us.

addons/api/addon/adapters/application.js
ZedLi: Yup good point, we'd always have an errors object by this point but I also realized it made more sense to move it up in the `normalizeErrorResponse` call anyways

addons/api/addon/adapters/application.js
lisbet-alvarez: ah ok that makes sense thank u!

lisbet-alvarez (dismissed):
Looks great! This is a nice improvement!

ui/admin/app/services/store.js
ZedLi, line 1: I decided to make a new store just for admin UI for the notifications as it didn't feel right to have it in the addon. This means the store service in API doesn't really get used anymore but if we want to keep it de-coupled from the UIs, it still makes sense to have a basic store to be able to use the addon, lmk if you have thoughts on this

lisbet-alvarez (approved):

DhariniJeeva (approved):
works as expected locally. Thanks for adding this!

priya-patel04 (approved):
Tested and looks great!! 🎉

cameronperera (approved):

PR #3263, opened 2026-05-07
docs: ✏️  Add documentation for pnpm overrides

ZedLi:
# Description
Added markdown files to document our package.json overrides and updated a few. I looked into using [json5](https://json5.org/) to possibly be able to directly comment in our `package.json` but unfortunately ember tooling doesn't play nicely with it.

## How to Test
Everything should build normally.

## Checklist

## PCI review checklist


package.json
```diff
@@ -46,15 +46,14 @@
     "overrides": {
       "micromatch": "^4.0.8",
       "css-select>nth-check": "^2.0.1",
-      "node-gyp": "^10.0.0",
```
ZedLi, line 49: This was orginally needed for an issue to allow us to remove a hack during building DC:

https://github.com/hashicorp/boundary-ui/commit/8087430dd2e2847b364e6bc0633af743786e5591#diff-fa9cf0ac521faf16769cdb54ea1d67c84197a32fa095099094070c86b7939faaL140-L155

Interestingly enough, electron doesn't depend on `node-gyp` directly anymore, they actually [forked it](https://www.npmjs.com/package/@electron/node-gyp) to deal with the `distutils` issue with python 3.12. Theoretically that means we also don't need this [section](https://github.com/hashicorp/boundary-ui/blob/main/ui/desktop/README.md?plain=1#L218-L222) in the readme anymore but I haven't tested this.

package.json
```diff
@@ -46,15 +46,14 @@
...
       "ansi-html": "^0.0.9",
       "fireworm>async": "^2.6.4",
       "ember-cli-babel>@babel/runtime": "7.27.6",
       "ember-try-config>package-json": "^10.0.1",
       "ember-loading>ember-concurrency": "^4.0.0",
       "tmp": "^0.2.5",
       "tar": "^7.5.10",
-      "underscore": "^1.13.8"
+      "nomnom>underscore": "^1.13.8"
```
ZedLi, line 56: Only `nomnom` uses an invalid version underscore so I scoped it.

pnpm-lock.yaml
```diff
@@ -4214,7 +4213,7 @@ packages:
       toffee: ^0.3.6
       twig: ^1.15.2
       twing: ^5.0.2
-      underscore: ^1.13.8
+      underscore: ^1.11.0
```
lisbet-alvarez, line 4215: looks like this version of underscore used by ember-cli -> testem -> consolidate still needs to be pinned to a non vulnerable version

lisbet-alvarez (commented):
The tables are detailed and easy to read! thank u for the new docs!

pnpm-lock.yaml
ZedLi: Is it not? That current line just changed because when the override was scoped, it stopped forcing it to `^1.13.8` and back to `^1.11.0` but it looks like they all should resolve correctly to `1.13.8` unless I missed something.

lisbet-alvarez: ah u are correct, i see it now

lisbet-alvarez (approved):
Looks good! thanks for the new docs

DhariniJeeva (approved):
looks great!

priya-patel04 (approved):
Thanks for adding this!

PR #3264, opened 2026-05-08
Gating logic for aliases

sohail-hashicorp:
# Description
Implemented the gating logic for project-level aliases. When a project scope does not yet have an `alias_suffix` configured, created a unified gated empty state on both the project Aliases list page and the Target details alias sidebar, guiding users with the ability to set an Alias suffix first.

:tickets: [ICU-18900](https://hashicorp.atlassian.net/browse/ICU-18900)
:tickets: [ICU-18901](https://hashicorp.atlassian.net/browse/ICU-18901)

https://github.com/user-attachments/assets/7a25b6de-6b8d-4bee-85eb-109edde05404

## How to Test

**1. Project Aliases list (gated)**

- Navigate to a project scope that has no alias_suffix
- Open the Aliases tab in the project nav
- Verify the UI renders `Create an Alias suffix` link button
- Click the link -> should land on /scopes/{project}/add-alias-suffix

**2. Project Aliases list (ungated)**

- Set an alias suffix on the same project scope (Project Settings → Alias Suffix sidebar → Create)
- Return to the Aliases tab -> gated state is gone and the table / "no aliases yet" empty state renders.

**3. Target details sidebar (gated)**

- Remove the suffix from the project scope (or pick another no-suffix project)
- Open any target inside that project that has no aliases
- Verify the sidebar renders the Create an Alias suffix → link button
- Click the link -> should land on /scopes/{project}/add-alias-suffix

**4. Target details sidebar (ungated)**

- Set the suffix again
- Reload the target -> verify the "Add an alias" empty state renders (not the gated state)

## Checklist

## PCI review checklist

[ICU-18900]: https://hashicorp.atlassian.net/browse/ICU-18900?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[ICU-18901]: https://hashicorp.atlassian.net/browse/ICU-18901?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


ui/admin/app/templates/scopes/scope/aliases/index.hbs
```diff
@@ -31,7 +31,23 @@
...
+              data-test-aliases-gated-link
+              @icon='arrow-right'
+              @iconPosition='trailing'
+              @route='scopes.scope.add-alias-suffix'
+              @text={{t 'resources.alias.messages.gated.link'}}
+            />
+          </A.Footer>
+        {{/if}}
+      </Hds::ApplicationState>
+    {{else if @model.doAliasesExist}}
```
cameronperera, line 48: It looks like `doAliasesExist` will fetch all aliases even at the project scope level. When looking at aliases at the project scope level, should we be showing global scopes? I would imagine not and the user may not have permission to see them.

ui/admin/app/templates/scopes/scope/targets/target/index.hbs
```diff
@@ -93,21 +93,44 @@
...
-              <Hds::Text::Display @tag='h2' class='remove-margin'>
-                {{t 'resources.alias.title_plural'}}
-              </Hds::Text::Display>
-              <Hds::Text::Body @tag='p' @color='faint' class='p'>
-                {{t 'resources.alias.form.sidebar.help'}}
-              </Hds::Text::Body>
+            {{#if this.isGated}}
+              <div
+                class='target-sidebar-aliases'
+                data-test-target-aliases-gated
```
cameronperera: since this `div` has class, we don't need the `data-test-*` attribute since the class can act as a selector in tests. Normally the `data-test-*` attributes are added when we don't have a concrete way to select something we need like a class.

ui/admin/app/templates/scopes/scope/targets/target/index.hbs
```diff
@@ -93,21 +93,44 @@
...
+              >
+                <Hds::Text::Display @tag='h2'>
+                  {{t 'resources.alias.title_plural'}}
+                </Hds::Text::Display>
+                <Hds::Text::Body @tag='p' @color='faint'>
+                  {{t 'resources.alias.messages.gated.description'}}
+                </Hds::Text::Body>
+                {{#if (can 'setAliasSuffix scope' @model.project)}}
+                  <Hds::Link::Standalone
+                    data-test-target-aliases-gated-link
```
cameronperera: informative comment:
This `data-test-*` selector is in the gray area. I don't have a strong opinion if it stays or not. On one hand, it helps to be very specific, while on the other hand you could chain the class name on line 98 with the `a` tag.

ui/admin/tests/acceptance/aliases/project-gated-test.js
```diff
@@ -0,0 +1,106 @@
...
+  test('aliases nav link is visible on a project scope', async function (assert) {
+    await visit(urls.projectScope);
+
+    assert.dom(commonSelectors.SIDEBAR_NAV_LINK(urls.aliases)).isVisible();
+  });
+
+  test('shows the gated empty state when project scope has no alias suffix', async function (assert) {
+    await visit(urls.aliases);
+
+    assert.dom(selectors.GATED_STATE).isVisible();
```
cameronperera: non-blocking: This assert feels redundant with the assert below verifying the link is present which is what we truly care about in this test case.

ui/admin/tests/acceptance/aliases/project-gated-test.js
```diff
@@ -0,0 +1,106 @@
...
+  test('gated CTA is hidden when user cannot set an alias suffix', async function (assert) {
+    instances.scopes.project.update({
+      authorized_actions: instances.scopes.project.authorized_actions.filter(
+        (action) => action !== 'set-alias-target-suffix',
+      ),
+    });
+
+    await visit(urls.aliases);
+
+    assert.dom(selectors.GATED_STATE).isVisible();
```
cameronperera: If they do not have the permission to set an alias target suffix, should we be showing
"No aliases yet
To start using aliases, you need to set an Alias suffix first."
I would imagine there is a state where the user can see (list) project level aliases but not create or set the project suffix.

ui/admin/tests/acceptance/targets/aliases-gated-test.js
cameronperera, line 1: We should have a tests when the user does not have the `set-alias-target-suffix`. 
If they do not have the permission to set an alias target suffix, should we be showing the alias side bar at all?
However, if they can add an alias to a target still, I would imagine we can still now the "add alias" button? We can still add global level aliases so I think the lack of this permission (set-alias-target-suffix) shouldn't block adding aliases entirely.

ui/admin/app/templates/scopes/scope/aliases/index.hbs
sohail-hashicorp: This PR only adds the gating logic, scope filtering of the list will be taken up in the next [ticket](https://hashicorp.atlassian.net/browse/ICU-18903)

ui/admin/app/templates/scopes/scope/targets/target/index.hbs
sohail-hashicorp: Good point, resolved it.

ui/admin/tests/acceptance/aliases/project-gated-test.js
sohail-hashicorp: Good call, I have reverted the gating changes on the target sidebar entirely.

cameronperera (approved):
LGTM

PR #3269, opened 2026-05-12
refactor: 💡 update pinned IDs logic for actions table

cameronperera:
# Description

update pinned IDs logic for actions table and use the correct article for types that require "an" instead of "a".

Pinned IDs for host-catalog showing host and host-set actions:
Screenshot 2026-05-12 at 11 05 55 AM
Correct article shown:
Screenshot 2026-05-12 at 12 06 30 PM

## Checklist

## PCI review checklist


ui/admin/app/utils/grant-completions.js
```diff
@@ -293,18 +293,24 @@ export const analyzeGrantString = (grantsSchema, grantString = '') => {
...
-
-  const isBothWildcard = idsValue === '*' && typeValue === '*';
+      : compatibleIdsResourceType && typeValue === '*'
+        ? (
+            schema.childResourceTypesByParentType[compatibleIdsResourceType] ??
+            []
+          ).join(' or ') || null
+        : typeValue !== '*' && idsValue
+          ? getCompatibleResourceTypeForIds(schema, idsValue)
+          : null;
```
ZedLi: This is getting to be a pretty gnarly nested ternary, would it be easier to read with if else statements?

ZedLi (dismissed):
Looks good, just had one minor comment

ui/admin/app/utils/grant-completions.js
cameronperera: that's a good point. let me improve this

cameronperera: added comments as well

ui/admin/app/utils/grant-completions.js
```diff
@@ -290,21 +290,29 @@ export const analyzeGrantString = (grantsSchema, grantString = '') => {
...
-        : null;
+  let detectedResourceType = null;
+  if (typeValue && typeValue !== '*') {
+    // Explicit type specified, use it directly
+    detectedResourceType = typeValue;
+  } else if (compatibleIdsResourceType && typeValue === '*') {
+    // Pinned IDs with wildcard type, show the child resource types
+    const childTypes =
+      schema.childResourceTypesByParentType[compatibleIdsResourceType] ?? [];
+    detectedResourceType = childTypes.join(' or ') || null;
```
ZedLi: I think this is fine since we only have resources with two child types currently but in the future if we ever have more it might look weird chaining `or` for all of them

ZedLi (dismissed):
Looks good!

ui/admin/app/utils/grant-completions.js
cameronperera: Yeah I thought of that but adding multiple `delete` keys with slightly different descriptions seemed like a bad UX, so I landed on chaining them. Happy to brainstorm a bit and come up with a different solution.

ZedLi: Realistically the proper way to handle this wasn't manually through code since how would we internationalize this in the future? Ideally we would be using something like [formatList](https://ember-intl.github.io/ember-intl/docs/services/intl-part-1#methods-format-list) which uses `Intl.ListFormat` under the hood.

cameronperera: Thanks for the tip. I have updated it to use formatList now. I am no longer chaining them in grant-completions but returning an array of the childTypes if there are more than one then using `intl.formatList` to achieve the proper formatting.

ZedLi (approved):
Looks good!

ui/admin/app/components/grant-actions/index.js
```diff
@@ -53,7 +53,14 @@ export default class GrantActionsIndex extends Component {
...
 
     if (detectedResourceType && this.intl.exists(specificKey)) {
-      return this.intl.t(specificKey, { resourceType: detectedResourceType });
+      const resourceType = Array.isArray(detectedResourceType)
+        ? this.intl.formatList(detectedResourceType, { type: 'disjunction' })
+        : detectedResourceType;
+      // Determine the appropriate article ("a" or "an") to use based on the
+      // detected resource type. "user" is a special case that uses "a" instead
+      // of "an" because it's pronounced "you" so I don't include it below.
+      const article = /^[aeio]/i.test(resourceType) ? 'an' : 'a';
```
lisbet-alvarez, line 62: im wondering would 'an' and 'a' also have to be translated first?

lisbet-alvarez (approved):
Looks great! 
Just left one question

PR #3271, opened 2026-05-13
refactor: 💡 Update transaction style and add busy timeout PRAGMA

ZedLi:
# Description
Add another hardening for sqlite concurrency issues from webkit e2e tests. 

## How to Test
e2e tests should pass.

## Checklist

## PCI review checklist


addons/api/addon/workers/sqlite-worker.js
```diff
@@ -191,9 +202,13 @@ import {
       });
```
moduli: Did this need a `return`?

addons/api/addon/workers/sqlite-worker.js
```diff
@@ -165,11 +168,15 @@ import {
...
-      // and execute these separate insert statements in a transaction
+      // and execute these separate insert statements in a transaction.
       if (numberOfParameters > MAX_HOST_PARAMETERS) {
         const chunkSize = Math.floor(MAX_HOST_PARAMETERS / items[0].length);
 
-        return db.transaction(() => {
+        // Note: We use explicit BEGIN/COMMIT instead of db.transaction() because
+        // db.transaction() uses SAVEPOINT internally, which fails with SQLITE_BUSY
+        // when any statement is active on the connection.
+        db.exec('BEGIN');
```
moduli, line 190: (Disclaimer: I don't have expertise in this area) AI suggested the use of `BEGIN IMMEDIATE`. Does that make sense?

ZedLi: Nope, it's a bit confusing but we don't actually have multiple connections to contend with since we proxy all requests/connections through one connection so `BEGIN IMMEDIATE` doesn't have an effect here. Any `SQLITE_BUSY` issues are contentions with the same connection

addons/api/addon/workers/sqlite-worker.js
ZedLi: Ah we don't actually use the return but I'll add it regardless

moduli (dismissed):

moduli (dismissed):

cameronperera (approved):
Thanks for adding these checks to the sqlite-worker. Hopefully no more flaky tests pop up

PR #3274, opened 2026-05-15
Implement project alias CRU and scope-aware target alias rows

sohail-hashicorp:
# Description
This PR includes the project alias CRUD operation feature on top of the suffix.

# New Changes
Org Suffix
There is no `base_value` on the alias model anymore. API changes have mostly gone back to what they were before.
Project Level Aliases cannot be created until both the org and proj have a suffix defined.
When creating or editing a project level alias, the suffix segments will be shown as readonly text to the right of the text input.

:tickets: [ICU-18902](https://hashicorp.atlassian.net/browse/ICU-18902), [ICU-18903](https://hashicorp.atlassian.net/browse/ICU-18903), [ICU-18904](https://hashicorp.atlassian.net/browse/ICU-18904), [ICU-18905](https://hashicorp.atlassian.net/browse/ICU-18905), [ICU-18906](https://hashicorp.atlassian.net/browse/ICU-18906)

Global Aliases:

https://github.com/user-attachments/assets/321e6067-8bf5-48bf-9083-fa3e1cd0a70d

No Org suffix:

https://github.com/user-attachments/assets/2b565f58-bdf9-42a0-9e59-eee1cfbb0c6e

No Proj suffix:

https://github.com/user-attachments/assets/22f90bf9-d44b-4ae7-9b56-fa196abb3d1a

No suffixes:

https://github.com/user-attachments/assets/690103d8-496b-40e7-b376-d523e051973c

## How to Test
1. Start admin against the default Mirage scenario: `cd ui/admin && pnpm start`.
2. **Suffix-set project - list + CRUD**
   - Open a project with `alias_suffix = ".example"`.
   - Visit **Aliases** in the project nav - confirm only project aliases appear and the Scope column is hidden.
   - Click **New** → enter `myhost` in the value field - confirm the inline `.example` suffix decoration appears next to the input. Save and verify the row appears with composed value `myhost.example`.
   - Open the row dropdown → **Edit**. The value input is pre-populated with the base value only (`myhost`). Save changes.
   - Open the row dropdown → **Delete**. Confirm the modal and verify the row disappears.
3. **No-suffix project - gated states exists**
   - On a project without a suffix, the Aliases page should show the gated empty state with a "Create an Alias suffix" link.
   - The target details sidebar should still show the gated CTA.
4. **New Target form - scope-aware alias rows**
   - In a project with a suffix, click **New Target**. The first alias row is pre-seeded to the project scope, enter `myhost`.
   - Click **Add Alias** → the second row defaults to the project scope; flip its scope picker to **Global** and enter `globalhost`.
   - Save the target.

## Checklist

## PCI review checklist

[ICU-18902]: https://hashicorp.atlassian.net/browse/ICU-18902?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[ICU-18903]: https://hashicorp.atlassian.net/browse/ICU-18903?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[ICU-18904]: https://hashicorp.atlassian.net/browse/ICU-18904?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[ICU-18905]: https://hashicorp.atlassian.net/browse/ICU-18905?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[ICU-18906]: https://hashicorp.atlassian.net/browse/ICU-18906?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


ui/admin/app/components/form/alias/index.js
```diff
@@ -7,7 +7,38 @@ import Component from '@glimmer/component';
...
+  // =attributes
+
+  /**
+   * Returns the alias suffix prefixed with `.`
+   * @type {string|null}
+   */
+  get normalizedSuffix() {
+    const suffix = this.args.suffix;
+    if (!suffix) return null;
+    return suffix.startsWith('.') ? suffix : `.${suffix}`;
```
cameronperera: I am fairly certain the suffix has to start with a `.`. Should we still include a safety check for that in the UI?

ui/admin/app/components/form/target/details/index.hbs
```diff
@@ -244,37 +244,75 @@
...
+                value={{@model.scopeModel.id}}
+                selected={{eq R.rowData.scope_id @model.scopeModel.id}}
+              >
+                {{this.projectOptionLabel}}
+              </option>
+            {{/if}}
+          </F.Select>
+        </R.Field>
+
+        <R.Field @width='1fr' as |F|>
```
cameronperera: I believe `1fr` is the default width for the Field. We might be able to remove this.

ui/admin/app/components/form/alias/index.js
cameronperera: Confirmed with the backend, the suffix will not require the `.` to be added by the user. It will be added when the `base_value` and `suffix` are combined in the API.

ui/admin/app/routes/scopes/scope/aliases/new.js
```diff
@@ -42,4 +42,10 @@ export default class ScopesScopeAliasesNewRoute extends Route {
     record.scopeModel = scopeModel;
     return record;
   }
+
+  setupController(controller) {
```
cameronperera: Reviewing the `setupController`, I don't think we need this directly especially since the `scopeModel` is added to the record before it is returned in the model hook. I will review the other `setupController` instances added and see if they can be removed in favor of returning the `scopeModel` in the model hook.

cameronperera: Removing the unneeded setupController

ui/admin/app/components/form/alias/index.js
```diff
@@ -7,7 +7,38 @@ import Component from '@glimmer/component';
...
+  get displayBaseValue() {
+    const { model, suffix } = this.args;
+    if (!model) return '';
+    if (model.base_value) return model.base_value;
+    const value = model.value ?? '';
+    // For legacy aliases without `base_value`, recover it by removing the
+    // suffix from the composed `value`.
+    if (suffix && value.endsWith(suffix)) {
+      return value.slice(0, -suffix.length);
+    }
```
ZedLi: Why would we need a fallback here? This is only relevant if we have `suffix` right? And if we have suffix are not guaranteed that the API is returning `base_value`?

ui/admin/app/components/form/target/details/index.js
```diff
@@ -75,4 +78,81 @@ export default class FormTargetComponent extends Component {
   get showTargetTypeRadioGroup() {
     return (this.isRDPEnabled || this.isSSHEnabled) && this.args.model.isNew;
   }
+
+  /**
+   * True when the current scope is a project with a configured suffix and
+   * the user can create aliases on it.
+   * @type {boolean}
+   */
+  get canCreateAliasOnProject() {
```
ZedLi: Why isn't this part of an ability?

ui/admin/app/components/form/target/details/index.js
```diff
@@ -75,4 +78,81 @@ export default class FormTargetComponent extends Component {
...
+  }
+
+  /**
+   * Default scope id to seed onto a freshly-added alias row.
+   * @type {string}
+   */
+  get defaultAliasScopeId() {
+    return this.canCreateAliasOnProject
+      ? this.args.model.scopeModel.id
+      : (this.args.globalScope?.id ?? 'global');
```
ZedLi: Isn't `globalScope.id` always `global`?

ui/admin/app/components/form/target/details/index.js
```diff
@@ -75,4 +78,81 @@ export default class FormTargetComponent extends Component {
...
+  }
+
+  /**
+   * Dropdown label for the Project scope option, e.g.
+   * "Project name (suffix '.example')".
+   * @type {string}
+   */
+  get projectOptionLabel() {
+    const scope = this.args.model?.scopeModel;
+    const name = scope?.displayName || scope?.name || scope?.id;
```
ZedLi: Isn't `displayName` already checking for `name` and `id`?

ui/admin/app/components/form/target/details/index.js
```diff
@@ -75,4 +78,81 @@ export default class FormTargetComponent extends Component {
...
+      suffix: this.normalizedProjectSuffix,
+    });
+  }
+
+  // =actions
+
+  @action
+  addAliasRow() {
+    const existing = this.args.model.with_aliases ?? [];
+    set(this.args.model, 'with_aliases', [
```
ZedLi: Why do we need ember `set` here and on the other actions?

ui/admin/app/components/form/target/details/index.hbs
```diff
@@ -244,37 +244,75 @@
...
     (and @model.isNew (can 'create model' @globalScope collection='aliases'))
   }}
-    <Form::Field::ListWrapper
-      @layout='horizontal'
+    <Hds::Form::KeyValueInputs
       @isOptional={{true}}
-      @disabled={{form.disabled}}
+      @data={{@model.with_aliases}}
+      name='with_aliases'
+      class='target-aliases-kvi'
```
ZedLi: I think this could have just used `margin-bottom-l`

ui/admin/app/components/form/target/details/index.hbs
```diff
@@ -244,37 +244,75 @@
...
+              selected={{eq R.rowData.scope_id @globalScope.id}}
+            >
+              {{t 'resources.alias.form.alias.scope.options.global'}}
+            </option>
+            {{#if this.canCreateAliasOnProject}}
+              <option
+                value={{@model.scopeModel.id}}
+                selected={{eq R.rowData.scope_id @model.scopeModel.id}}
+              >
+                {{this.projectOptionLabel}}
```
ZedLi: Why do we display the suffix in the scope dropdown instead of showing it as another column after the alias value like we do when creating aliases?

ZedLi (commented):
Left some comments on a quick first pass

ui/admin/app/components/form/target/details/index.js
priya-patel04: +1 to this. We have similar logic in new target route [here](https://github.com/hashicorp/boundary-ui/pull/3274/changes#diff-d0bfac5ba0a5c1d43687408d0cf171252b1184778e491cff7c5b7d62dc9e4352R59-R64) and can benefit with being part of scope ability

ui/admin/app/components/form/alias/index.js
cameronperera: yeah taking a long look at this, it is unneeded. Also `base_value` is one of the things changing. I simplified this for now.

ui/admin/app/components/form/target/details/index.hbs
cameronperera: Where is `margin-bottom-l` coming from again? I remember doing this in app tokens but it didn't work for me. Am I missing something?

ui/admin/app/components/form/target/details/index.hbs
cameronperera: That is a good question and I can reach out to design. I know they wanted to make sure the user understood this was changing scope for the alias but it is not clear that the alias will have the suffix as well. 
Also, with that said, how we create aliases with a suffix might be changing in the API so we may require some design changes regardless.

ui/admin/app/components/form/target/details/index.js
cameronperera: Great feedback. Moved to an ability.

ui/admin/app/components/form/target/details/index.js
cameronperera: It sure is. Simplified this logic.

ui/admin/app/components/form/target/details/index.js
cameronperera: Yep. Fixed.

ui/admin/app/components/form/target/details/index.js
cameronperera: Great question I don't have an answer for but I did update this to our normal pattern.

ui/admin/app/components/form/target/details/index.hbs
ZedLi: It's from [core](https://github.com/hashicorp/boundary-ui/blob/main/addons/core/addon/styles/addon.scss#L94-L123), i'd double check that there's no specificity issue

cameronperera: Specificity got me again. It is and I don't remember what we have done to fix that in the past. I will keep looking to get it worked out.

ui/admin/app/components/form/target/details/index.hbs
cameronperera: Going to chat with design about this still as designs may change a bit with the intro of a org suffix.

addons/api/addon/abilities/scope.js
```diff
@@ -27,40 +27,42 @@ export default class ScopeAbility extends ModelAbility {
...
+   * is currently set.
    * @type {boolean}
    */
-  get canGetAliasSuffix() {
+  get canRemoveAliasSuffix() {
     return (
-      this.model?.isProject &&
-      this.hasAuthorizedAction('get-alias-target-suffix')
+      (this.model?.isProject || this.model?.isOrg) &&
+      Boolean(this.model?.alias_suffix) &&
```
ZedLi: Isn't this covered by the convenience getter `hasSuffix`?

addons/api/addon/abilities/scope.js
```diff
@@ -27,40 +27,42 @@ export default class ScopeAbility extends ModelAbility {
...
-   * Removing the alias target suffix is allowed on project scopes when the
-   * `remove-alias-target-suffix` authorized action is present and a suffix
-   * is currently set.
+   * Creating a project-scoped alias is allowed when the scope is a project
+   * with a configured suffix and the user has the `create` collection action
+   * on aliases.
    * @type {boolean}
    */
-  get canRemoveAliasSuffix() {
+  get canCreateProjectAlias() {
```
ZedLi: Are we just gatekeeping it to project right now and in the future allowing orgs?

addons/api/app/mirage/config.js
```diff
@@ -188,6 +188,43 @@ export default function (mirageConfig) {
   return createServer(finalConfig);
 }
 
+function normalizeSuffixSegments(suffix) {
+  if (!suffix) return null;
+  const normalized = String(suffix).split('.').filter(Boolean).join('.');
```
ZedLi: Do you expect suffix to not be a string? Also why do we strip out the leading `.`, don't we end up just adding it back?

addons/api/app/mirage/config.js
```diff
@@ -188,6 +188,43 @@ export default function (mirageConfig) {
...
+  return normalized || null;
+}
+
+function getScopeAliasSuffix(scope, scopes) {
+  if (!scope) return null;
+
+  const scopeSuffix = normalizeSuffixSegments(
+    scope.alias_suffix ?? scope.aliasSuffix,
```
ZedLi: Why do we check for both, shouldn't we just use the one field?

addons/api/app/mirage/config.js
```diff
@@ -188,6 +188,43 @@ export default function (mirageConfig) {
...
+  );
+  if (!scopeSuffix) return null;
+
+  if (scope.type !== 'project') {
+    return `.${scopeSuffix}`;
+  }
+
+  const parentScopeId = scope.scope?.id ?? scope.scope_id ?? scope.scopeId;
```
ZedLi: Why are we checking all these fields?

addons/api/app/mirage/config.js
```diff
@@ -188,6 +188,43 @@ export default function (mirageConfig) {
...
+  if (!scope) return null;
+
+  const scopeSuffix = normalizeSuffixSegments(
+    scope.alias_suffix ?? scope.aliasSuffix,
+  );
+  if (!scopeSuffix) return null;
+
+  if (scope.type !== 'project') {
+    return `.${scopeSuffix}`;
+  }
```
ZedLi: Is this also valid for global scope?

addons/api/app/mirage/config.js
```diff
@@ -188,6 +188,43 @@ export default function (mirageConfig) {
...
+
+  const parentScopeId = scope.scope?.id ?? scope.scope_id ?? scope.scopeId;
+  const parentScope = parentScopeId ? scopes.find(parentScopeId) : null;
+  const orgSuffix = normalizeSuffixSegments(
+    parentScope?.alias_suffix ?? parentScope?.aliasSuffix,
+  );
+
+  if (!orgSuffix || scopeSuffix.endsWith(`.${orgSuffix}`)) {
```
ZedLi: Do org and proj suffix need to be unique? Did we need to check the scopeSuffix with the orgSuffix?

addons/api/app/mirage/config.js
```diff
@@ -689,9 +726,40 @@ function routes() {
...
+          aliasAttrs.value = appendSuffixIfMissing(
+            aliasData.value,
+            aliasSuffix,
+          );
+        }
+        return aliases.create(aliasAttrs);
+      });
+
+      target.update({
+        aliases: createdAliases.map((a) => ({ id: a.id, value: a.value })),
```
ZedLi:
```suggestion
        aliases: createdAliases.map((alias) => ({ id: alias.id, value: alias.value })),
```

addons/core/translations/resources/en-us.yaml
```diff
@@ -1297,11 +1302,22 @@ alias:
     value:
       label: Alias Value
       help: The alias you would like to use for this resource. If using Windows, using a "." in the hostname is recommended.
+      help_project: The alias you would like to use for this resource. The Alias suffix has been set by your project administrator.
     authorize_session_arguments:
       label: Host ID
       help: Always use this host when connecting using this alias
     sidebar:
       help: Use a domain name as a shortcut for anyone with access to this target.
+      no-suffix-message: 'To start using aliases, you need to set an Alias suffix first.'
```
ZedLi: Why's this one quoted?

ui/admin/app/styles/app.scss
```diff
@@ -468,11 +468,32 @@
...
+  gap: 0.5rem;
+  width: 100%;
+
+  .hds-form-text-input {
+    flex: 0 0 50%;
+    width: 50%;
+    min-width: 0;
+  }
+
+  .alias-value-with-suffix--text {
```
ZedLi: Could have simplified with parent selector:

```suggestion
  .&--text {
```

ui/admin/app/styles/app.scss
```diff
@@ -468,11 +468,32 @@
...
+.alias-value-with-suffix {
+  display: flex;
+  align-items: center;
+  gap: 0.5rem;
+  width: 100%;
+
+  .hds-form-text-input {
+    flex: 0 0 50%;
+    width: 50%;
+    min-width: 0;
```
ZedLi: Did you need a min-width? What's the purpose of setting both flex-basis and with to 50%?

ui/admin/app/components/form/alias/index.js
ZedLi, line 1: I didn't want to leave comments everywhere but can we prefer `??` instead of `||` where it makes sense throughout this PR?

ui/admin/app/components/form/alias/index.js
```diff
@@ -5,9 +5,101 @@
...
+   * is present, strip the suffix from the model value for display.
+   * @type {string}
+   */
+  get displayBaseValue() {
+    const { model } = this.args;
+    if (!model) return '';
+
+    const fullValue = model.value || '';
+    const suffix = this.normalizedSuffix || this.trailingTwoSegmentSuffix;
+    if (!suffix || !fullValue.endsWith(suffix)) return fullValue;
```
ZedLi: Why do we want to return the full value if they don't match? If a project or org alias was changed, won't this now display the full value including the old suffixes in the user text box?

ui/admin/app/components/form/alias/index.js
```diff
@@ -5,9 +5,101 @@
...
+
+    return fullValue.slice(0, -suffix.length);
```
ZedLi: This seems a little more complex than needed. Why not just return the first value of the split string based on `.`? If there's none, great, if there is, it's always the display base value.

Using `-suffix.length` seems more error prone because now we have to verify that there's no off by 1 errors from calculating the suffix.

ui/admin/app/components/form/alias/index.js
```diff
@@ -5,9 +5,101 @@
...
+    if (!suffix) return null;
+    return `.${suffix}`;
+  }
+
+  /**
+   * Fallback for project aliases when a suffix cannot be resolved from scope:
+   * derive `.<segmentN-1>.<segmentN>` from the current model value.
+   * @type {string|null}
+   */
+  get trailingTwoSegmentSuffix() {
```
ZedLi: Should we be falling back in the first place, what necessitates this?

ui/admin/app/components/form/alias/index.js
```diff
@@ -17,4 +109,15 @@ export default class FormAliasComponent extends Component {
...
+
+  /**
+   * Updates the alias value as the user types.
+   */
+  @action
+  handleBaseValueChange({ target: { value } }) {
+    const { model } = this.args;
+    const suffix = this.normalizedSuffix || this.trailingTwoSegmentSuffix;
+
+    model.value = suffix && value ? `${value}${suffix}` : value;
```
ZedLi: Could we have just had suffix default to empty string and just return `${value}${suffix}` always?

ui/admin/app/components/form/target/details/index.js
```diff
@@ -75,4 +79,106 @@ export default class FormTargetComponent extends Component {
   get showTargetTypeRadioGroup() {
     return (this.isRDPEnabled || this.isSSHEnabled) && this.args.model.isNew;
   }
+
+  /**
+   * True when the current scope is a project with a configured suffix and
+   * the user can create aliases on it.
+   * @type {boolean}
+   */
+  get canCreateAliasOnProject() {
```
ZedLi: Why did this need to remain a getter, shouldn't we have just used the ability directly?

ui/admin/app/components/form/target/details/index.hbs
```diff
@@ -242,39 +242,84 @@
...
-    >
-      <:fieldset as |F|>
-        <F.Legend>
-          {{t 'resources.alias.form.alias.label'}}
-        </F.Legend>
-        <F.HelperText>
-          {{t 'resources.alias.form.alias.help'}}
-        </F.HelperText>
+    <div class='margin-bottom-l'>
+      <Hds::Form::KeyValueInputs
```
ZedLi, line 254: I think @DhariniJeeva is trying to get in our component for this too FYI

ui/admin/app/controllers/scopes/scope/aliases/index.js
```diff
@@ -60,18 +61,43 @@ export default class ScopesScopeAliasesIndexController extends Controller {
...
+      !this.scope?.alias_suffix &&
       this.abilities.can('setAliasSuffix scope', this.scope)
     );
   }
 
+  /**
+   * True when the parent org scope is missing an alias suffix.
+   * @type {boolean}
+   */
+  get isOrgSuffixMissing() {
```
ZedLi: Should `isProjectSuffixMissing` and `isProjectSuffixMissing` be encapsulated as part of the ability instead of getters here? It looks like we use it in more than one place and it seems like their usage is just more akin to gatekeeping

addons/api/app/mirage/factories/alias.js
```diff
@@ -9,6 +9,29 @@ import generateId from '../helpers/id';
...
+  const scopeSuffix = normalizeSuffixSegments(scope?.alias_suffix);
+  if (!scopeSuffix) return null;
+
+  if (scope.type !== 'project') return `.${scopeSuffix}`;
+
+  const orgScopeId = scope.scope?.id;
+  const orgScope = orgScopeId ? server.schema.scopes.find(orgScopeId) : null;
+  const orgSuffix = normalizeSuffixSegments(orgScope?.alias_suffix);
+
+  if (!orgSuffix || scopeSuffix.endsWith(`.${orgSuffix}`)) {
```
lisbet-alvarez: in which case would the project suffix already end with the org suffix?

ui/admin/app/templates/scopes/scope/aliases/index.hbs
```diff
@@ -128,9 +154,17 @@
...
                   {{/if}}
                 </B.Td>
-                <B.Td>
-                  <Hds::Badge @text={{t 'titles.global'}} @icon='globe' />
-                </B.Td>
+                {{#unless this.scope.isProject}}
+                  <B.Td>
+                    <Hds::Badge
+                      @text={{or
+                        data.alias.scopeModel.displayName
```
ZedLi: Does display name not already check for ID?

ZedLi (commented):
Initial pass of just code, will come back and test the actual branch

addons/api/app/mirage/factories/alias.js
```diff
@@ -20,4 +43,18 @@ export default factory.extend({
       'delete',
     ],
   scope: () => ({ id: 'global', type: 'global' }),
+
+  afterCreate(alias, server) {
+    const scopeId = alias.scope_id || alias.scope?.id;
```
lisbet-alvarez: shouldn't `scope?.id` work why are we also checking for `scope_id`

addons/api/app/mirage/config.js
```diff
@@ -949,7 +1017,24 @@ function routes() {
...
+    function ({ aliases, scopes }, { params: { id } }) {
+      const alias = aliases.find(id);
+      const attrs = this.normalizedRequestAttrs();
+
+      // Mirror the create-alias suffix logic on update.
+      const scopeId = alias.scope_id;
+      const scope = scopes.find(scopeId);
+      const aliasSuffix = getScopeAliasSuffix(scope, scopes);
+      const updatedAttrs = { ...attrs };
+      if (attrs.value !== undefined && aliasSuffix) {
```
lisbet-alvarez: why is it that here we check the `value` is not undefined but in line 748 we just check for existence of `value` field?

addons/core/translations/resources/en-us.yaml
```diff
@@ -1073,23 +1073,25 @@ scope:
     project: Project
   alias-suffix:
     title: Alias suffix
-    description: Append this suffix to aliases in this project scope. The alias will appear as "alias.customsuffix"
-    form:
-      helper-text: Append this suffix to aliases in this project scope. The alias will appear as "alias.customsuffix".
+    description: Append this suffix to aliases in this project scope. The alias will appear as “alias.projectsuffix.orgsuffix”
```
lisbet-alvarez: the quotes around `alias.projectsuffix.orgsuffix` look a bit off i think they are supposed to be `"` instead of `“`?

addons/core/translations/resources/en-us.yaml
```diff
@@ -1297,11 +1302,22 @@ alias:
     value:
       label: Alias Value
       help: The alias you would like to use for this resource. If using Windows, using a "." in the hostname is recommended.
+      help_project: The alias you would like to use for this resource. The Alias suffix has been set by your project administrator.
```
lisbet-alvarez, line 1305: should `alias` in the first sentence also be capitalized because alias in the second sentence and other references is capitalized?

addons/api/addon/abilities/scope.js
lisbet-alvarez: im wondering, since this getter isn't for a specific authorized action returned from the api and more so derived from other properties on other models, should this be moved admin level abilities?

ui/admin/app/routes/scopes/scope/aliases/index.js
lisbet-alvarez, line 163: i think we need to add a scope filter to this query because right now i am getting "No results found" message instead of a "No aliases yet" message

ui/admin/app/components/form/target/details/index.hbs
DhariniJeeva: Thanks for the heads up! I can reconcile my changes if this pr goes in first.

addons/api/app/mirage/factories/alias.js
cameronperera: There isn't one. This logic has been removed.

addons/api/addon/abilities/scope.js
cameronperera: yes, updated

addons/api/addon/abilities/scope.js
cameronperera: As of right now, there is no plan to create org level aliases. So only global and project scopes can create aliases. The global scope only needs to check for the `authorized_collection_actions` as we currently do.

addons/api/app/mirage/config.js
cameronperera: It will be a string and that is removed. For the `.`, originally we thought the API would include the `.` but it doesn't. Updating this to reflect that.

addons/api/app/mirage/config.js
cameronperera: Fixed.

addons/api/app/mirage/config.js
cameronperera: I want to blame API slop but it's my fault for not catching. Fixed.

addons/api/addon/abilities/scope.js
```diff
@@ -27,40 +27,71 @@ export default class ScopeAbility extends ModelAbility {
...
    * @type {boolean}
    */
-  get canRemoveAliasSuffix() {
+  get canSeeOrgSuffixPrompt() {
+    const org = this.model?.scopeModel;
     return (
       this.model?.isProject &&
-      Boolean(this.model?.alias_suffix) &&
-      this.hasAuthorizedAction('remove-alias-target-suffix')
+      !!org &&
```
lisbet-alvarez: is this just supposed to be `org &` ?

addons/api/app/mirage/factories/target.js
```diff
@@ -181,26 +180,38 @@ export default factory.extend({
...
-            ? `${baseValue}${scope.alias_suffix}`
-            : baseValue;
+      // Only create aliases if the scope has a suffix (and for projects, the parent org also has one).
+      const scopeRecord = server.schema.scopes.find(scope.id);
+      const hasScopeSuffix = Boolean(scopeRecord?.alias_suffix);
+      const parentOrgId = scopeRecord?.scope?.id;
+      const parentOrg = parentOrgId
+        ? server.schema.scopes.find(parentOrgId)
+        : null;
+      const hasOrgSuffix =
```
lisbet-alvarez: could this check be simplified to just be `Boolean(parentOrg?.alias_suffix);`?

ui/admin/app/components/form/alias/index.js
```diff
@@ -7,7 +7,90 @@ import Component from '@glimmer/component';
...
+      return null;
+    }
+    const normalized = String(suffix).split('.').filter(Boolean).join('.');
+    return normalized || null;
+  }
+
+  /**
+   * Returns the combined suffix segments for the current alias scope.
+   * For project aliases: `projectSuffix.orgSuffix`.
+   * For other scopes: `scopeSuffix`.
```
lisbet-alvarez: Perhaps im misunderstanding but should this just be "For org aliases: `scopeSuffix`"

ui/admin/app/components/form/alias/index.js
```diff
@@ -7,7 +7,90 @@ import Component from '@glimmer/component';
...
+   * @type {string|null}
+   */
+  get combinedSuffixSegments() {
+    const projectSuffix = this.normalizeSuffixSegments(this.args.suffix);
+    const scope = this.args.model?.scopeModel;
+
+    if (!scope?.isProject) {
+      return projectSuffix;
```
lisbet-alvarez: should `projectSuffix` be renamed to just `scopeSuffix` since it is the suffix of the current scope which could be org or project?

ui/admin/app/templates/scopes/scope/aliases/index.hbs
```diff
@@ -19,7 +19,17 @@
       </PH.Description>
       <PH.Actions>
         {{#if (can 'create model' this.scope collection='aliases')}}
-          {{#if (and @model.doAliasesExist (not this.isGated))}}
+          {{#if
+            (and
+              @model.doAliasesExist
+              (not
```
lisbet-alvarez: perhaps this could be more readable if we use `and` & `cannot` helpers instead. So itd be (and (cannot {ability}) (cannot {ability}))?

lisbet-alvarez (commented):
Looking really good!!
I left a couple suggestions/questions

addons/api/app/mirage/config.js
cameronperera: It isn't and I see why this check is confusing. I updated this logic.

addons/api/app/mirage/config.js
cameronperera: I have reached out to product about this. I will update this comment.

addons/core/translations/resources/en-us.yaml
cameronperera: Not sure. Fixed.

ui/admin/app/styles/app.scss
cameronperera: Good catch. I am not sure why both were added but I removed the redundant parts.

ui/admin/app/components/form/alias/index.js
cameronperera: This file got a big overhaul but this is fixed now.

ui/admin/app/components/form/alias/index.js
cameronperera: Agreed. I have updated this to simplify this a good bit. One thing I do want to point out. The "base value" of an alias, the part the user can edit or create, can be multi-segmented. So it could be `alias.proj.org` or it could be `alias.seg.men.ted.proj.org`. Both are valid aliases and we only ever want to remove the last 2 segments.

ui/admin/app/components/form/alias/index.js
cameronperera: Also updated and fixed.

ui/admin/app/components/form/alias/index.js
cameronperera: also heavy updated.

ui/admin/app/components/form/target/details/index.js
cameronperera: sorry, missed when I created the ability, updated now

ui/admin/app/controllers/scopes/scope/aliases/index.js
cameronperera: Good call, moved to an ability

ui/admin/app/templates/scopes/scope/aliases/index.hbs
cameronperera: Sorry, this was left over from when we were going to show all aliases at the global level but we have decided against that for now. This has been updated to just use `data.alias.scopeModel.displayName` now.

addons/api/app/mirage/factories/alias.js
cameronperera: poor initial review on my part, fixed now

addons/api/app/mirage/config.js
cameronperera: a lot of the mirage config has been updated and this was removed

addons/core/translations/resources/en-us.yaml
cameronperera: oh good catch, I copied the text out of figma so that is odd, fixed now

addons/core/translations/resources/en-us.yaml
cameronperera: I talked with design about this and agree that it is weird but keeping it as is. The reason is alias in the first sentence is referring to the "alias value" while the alias in the second sentence is talking about an Alias resource. Happy to chat more about this.

addons/api/addon/abilities/scope.js
cameronperera: That is a good callout. I will move this.

addons/api/addon/abilities/scope.js
cameronperera: Yes, updated

addons/api/app/mirage/factories/target.js
cameronperera: It sure can. Updated.

ui/admin/app/components/form/alias/index.js
cameronperera: Yes that is confusing. Updating now.

ui/admin/app/templates/scopes/scope/aliases/index.hbs
cameronperera: ooh that reads nicer, thanks!

lisbet-alvarez (dismissed):
Looks great!

ui/admin/app/components/form/alias/index.js
cameronperera: Sorry, I missed this comment as I think a push marked it outdated. This was updated. Ultimately, if the scope isn't a project scope, I don't want any suffix returned so this returns `null` now. I still use projectSuffix but that is to better differentiate between the orgSuffix that is also used if the scope is a project.

ui/admin/app/components/form/alias/index.js
```diff
@@ -116,8 +85,8 @@ export default class FormAliasComponent extends Component {
   @action
   handleBaseValueChange({ target: { value } }) {
     const { model } = this.args;
-    const suffix = this.normalizedSuffix || this.trailingTwoSegmentSuffix;
+    const suffix = this.normalizedSuffix ?? '';
 
-    model.value = suffix && value ? `${value}${suffix}` : value;
+    model.value = value ? `${value}${suffix}` : value;
```
ZedLi: Can this just be simplified to `model.value = ${value}${suffix}` since if there's no suffix it would just be an empty string anyway

ui/admin/app/components/form/alias/index.js
```diff
@@ -7,7 +7,68 @@ import Component from '@glimmer/component';
...
+  get combinedSuffixSegments() {
+    const scope = this.args.model?.scopeModel;
+    if (!scope?.isProject) {
+      return null;
+    }
+
+    const projectSuffix = this.args.suffix;
+    const orgSuffix = this.args.orgSuffix;
+
+    return `${projectSuffix}.${orgSuffix}`;
```
ZedLi, line 27: So we're forcing users to always use both proj and org suffixes if they want to use this, it can never be one or the other?

ui/admin/app/components/form/alias/index.js
```diff
@@ -7,7 +7,68 @@ import Component from '@glimmer/component';
...
+  }
+
+  // =actions
+
+  @action
+  handleSubmit() {
+    const { model } = this.args;
+    const suffix = this.normalizedSuffix;
+    if (model && suffix) {
+      model.value = `${this.displayBaseValue}${suffix}`;
```
ZedLi, line 67: Does backend validate this and force the suffix? If we ever submit a value that doesn't use the right suffix do we get an error back?

ZedLi (dismissed):
Looks good, just had a few questions

ui/admin/app/components/form/alias/index.js
cameronperera: That is correct. They have to have both set before using project level aliases.

ui/admin/app/components/form/alias/index.js
cameronperera: Yep, the backend will validate this and give a proper error back if they don't match.

ui/admin/app/components/form/alias/index.js
cameronperera: Great point. Updated.

ZedLi (approved):

lisbet-alvarez (approved):

PR #3276, opened 2026-05-15
chore: 🤖 migrate host catalog and host set list wrapper fields to key value

DhariniJeeva:
# Description

This PR migrates the listwrapper fields in host catalog and host set to keyValue component

Screenshot 2026-05-14 at 5 22 02 PM

Screenshot 2026-05-14 at 5 21 50 PM

Screenshot 2026-05-14 at 5 21 39 PM

## How to Test

1. go to host catalog, toggle credential type and try adding/deleting role_tags 
2. edit existing host catalog and try adding/deleting role_tags
3. Go to host set, try adding/deleting preferred endpoints and filters (azure and aws)
4. edit existing host set, try adding/deleting preferred endpoints and filters (azure and aws)

## Checklist

## PCI review checklist

### DIFF
--- a/ui/admin/app/components/form/field/key-value/index.hbs
+++ b/ui/admin/app/components/form/field/key-value/index.hbs
@@ -21,8 +21,8 @@
         Field=R.Field
         DeleteRowButton=R.DeleteRowButton
         rowData=R.rowData
-        removeRow=(fn @onRemove R.rowData)
-        updateRow=(fn @onUpdate R.rowData)
+        removeRow=(fn (or @onRemove this.handleRemove) R.rowData)
+        updateRow=(fn (or @onUpdate this.handleUpdate) R.rowData)
         canDelete=this.canDelete
       )
       to='row'
@@ -32,7 +32,7 @@
     <F.AddRowButton
       @text={{t 'actions.add'}}
       data-test-add-button
-      @onClick={{@onAdd}}
+      @onClick={{or @onAdd this.handleAdd}}
     />
     {{#if @errors}}
       <F.Error as |E|>
--- a/ui/admin/app/components/form/field/key-value/index.hbs
+++ b/ui/admin/app/components/form/field/key-value/index.hbs
@@ -21,8 +21,8 @@
         Field=R.Field
         DeleteRowButton=R.DeleteRowButton
         rowData=R.rowData
-        removeRow=(fn (or @onRemove this.handleRemove) R.rowData)
-        updateRow=(fn (or @onUpdate this.handleUpdate) R.rowData)
+        removeRow=(fn this.handleRemove R.rowData)
+        updateRow=(fn this.handleUpdate R.rowData)
         canDelete=this.canDelete
       )
       to='row'
@@ -32,7 +32,7 @@
     <F.AddRowButton
       @text={{t 'actions.add'}}
       data-test-add-button
-      @onClick={{or @onAdd this.handleAdd}}
+      @onClick={{this.handleAdd}}
     />
     {{#if @errors}}
       <F.Error as |E|>
--- a/ui/admin/app/components/form/field/key-value/index.js
+++ b/ui/admin/app/components/form/field/key-value/index.js
@@ -4,6 +4,7 @@
  */
 
 import Component from '@glimmer/component';
+import { action } from '@ember/object';
 
 export default class FormFieldKeyValueComponent extends Component {
   /**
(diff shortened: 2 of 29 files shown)


ui/admin/app/components/form/host-catalog/index.js
```diff
@@ -62,4 +62,48 @@ export default class FormHostCatalogIndexComponent extends Component {
...
+   * @param {object} rowData - The row data to remove
+   */
+  @action
+  removeRow(field, rowData) {
+    let rows = (this.args.model[field] ?? []).filter(
+      (item) => item !== rowData,
+    );
+
+    if (rows.length === 0) {
+      const properties = Object.keys(rowData).length
```
ZedLi: will `rowData` ever not be passed in for this?

ui/admin/app/components/form/host-catalog/index.hbs
```diff
@@ -5,10 +5,14 @@
...
 <this.hostCatalogFormComponent
   @model={{@model}}
+  @edit={{@edit}}
   @submit={{@submit}}
   @cancel={{@cancel}}
   @changeType={{@changeType}}
   @changeCredentialType={{@changeCredentialType}}
+  @addRow={{this.addRow}}
+  @removeRow={{this.removeRow}}
+  @updateRow={{this.updateRow}}
```
ZedLi: I noticed in some of your functions you made it so it could handle a more general case. I forget if we already tried this and there was a reason this wouldn't make sense but I'm noticing these three functions to handle rows can get quite repetitive and we end up with nearly the same logic for all of them.

Was there a reason we couldn't have done something like this directly on the component for each of the row functions as a default? 

```
  handleAdd() {
    const { model, field, properties = ['value'] } = this.args;
    if (model && field) {
      const emptyRow = Object.fromEntries(properties.map((prop) => [prop, '']));
      model[field] = [...(model[field] ?? []), emptyRow];
    } else {
      this.args.onAdd?.();
    }
  }
```

ZedLi (dismissed):
Looks good, I just had a question regarding how we use the component

ui/admin/app/components/form/host-catalog/index.hbs
DhariniJeeva: hmm I originally had the row handler logic inside the KeyValue comp, but based on feedback, I moved it to the consumers.  Now, the consumer handles all data changes and just passes handler functions down to KeyValue comp. I agree it is getting repetitive- what do you think about creating a shared util for these row handlers? That way we can keep the logic in one place and still let each consumer customize if required. Open to suggestions!

ui/admin/app/components/form/host-catalog/index.js
DhariniJeeva: in the current flow, rowData is always passed, but I could add a guard to check for it.

ZedLi: Sorry my question was about the current guard checking for `rowData` length, since it always was passed in by us, I was wondering if it was needed

ui/admin/app/components/form/host-catalog/index.hbs
ZedLi: IIRC the main issue was the hard coding of either `key` or `value` fields and being able to handle different fields for each row. We'd still be manually yielding the individual row elements but it seems like we can have default options for the most widely used cases and still be able to use `rowData` or something like `@properties={{array 'key' 'value'}}` to properly handle the different fields. 

It just seems like having a `@model` and `@field` passed in makes it much easier as a default but there's still an escape hatch of `@onAdd`, `@onUpdate`, etc if we ever did need it (for cases where it's not being saved to a model as an example)

DhariniJeeva: makes sense! looking into this approach

ui/admin/app/components/form/host-catalog/index.js
DhariniJeeva: sorry yeah it wasn't needed!

ui/admin/app/components/form/host-catalog/index.hbs
DhariniJeeva: updated!

ui/admin/app/components/form/field/key-value/index.hbs
```diff
@@ -21,8 +21,8 @@
         Field=R.Field
         DeleteRowButton=R.DeleteRowButton
         rowData=R.rowData
-        removeRow=(fn @onRemove R.rowData)
-        updateRow=(fn @onUpdate R.rowData)
+        removeRow=(fn (or @onRemove this.handleRemove) R.rowData)
+        updateRow=(fn (or @onUpdate this.handleUpdate) R.rowData)
```
ZedLi: How come we need the `or`? Isn't this already being handled by the `handle{x}` actions? If we're worried about the `on{x}` actions being called we can check for those first

ui/admin/app/components/form/field/key-value/index.js
```diff
@@ -25,4 +26,58 @@ export default class FormFieldKeyValueComponent extends Component {
...
+   * Uses @model, @name, and @properties if provided, otherwise calls @onAdd.
+   */
+  @action
+  handleAdd() {
+    const { model, name, properties } = this.args;
+    if (model && name && properties) {
+      const emptyRow = Object.fromEntries(properties.map((prop) => [prop, '']));
+      model[name] = [...(model[name] ?? []), emptyRow];
+    } else {
+      this.args.onAdd?.();
```
ZedLi: We can invert these if statements to check if `this.args.onAdd` exists and just call it immediately if it does that way we don't need the or statements in the templates. We also can optionally add asserts for the other fields if we don't supply a custom handler

ZedLi (dismissed):
Looks great! Thanks for the migration

ui/admin/app/components/form/field/key-value/index.hbs
DhariniJeeva: good point! it can be def be simplified. Updated!

ZedLi (approved):

PR #3284, opened 2026-05-27
chore: migrate listwrapper Target and worker fields to keyvalue component

DhariniJeeva:
# Description

This pr migrates listwrapper fields in target and worker filter. 

Screenshot 2026-05-27 at 11 01 33 AM

Screenshot 2026-05-27 at 11 01 13 AM

## Checklist

## PCI review checklist

### DIFF
--- a/ui/admin/app/components/form/field/key-value/index.js
+++ b/ui/admin/app/components/form/field/key-value/index.js
@@ -34,6 +34,11 @@ export default class FormFieldKeyValueComponent extends Component {
   }
 
   get errors() {
+    // If errors are passed directly as an argument, use those
+    if (this.args.errors) {
+      return this.args.errors;
+    }
+
     const { model, name } = this.args;
 
     if (model && name && model.errors) {
--- a/ui/admin/app/components/form/field/key-value/index.js
+++ b/ui/admin/app/components/form/field/key-value/index.js
@@ -22,14 +22,11 @@ export default class FormFieldKeyValueComponent extends Component {
     // For a single row, only show if it has any non-empty values
     if (rows.length === 1) {
       const row = rows[0];
-
-      // Pass @properties for rows with tracked fields so canDelete reactively checks the tracked properties.
-      const properties = this.args.properties || [];
-      if (properties.length) {
-        return properties.some((property) => row?.[property]);
+      for (const key in row) {
+        if (row[key]) {
+          return true;
+        }
       }
-
-      return Object.values(row).some((value) => value);
     }
 
     // Return false if there are no rows
--- a/ui/admin/app/components/form/field/key-value/index.js
+++ b/ui/admin/app/components/form/field/key-value/index.js
@@ -29,7 +29,7 @@ export default class FormFieldKeyValueComponent extends Component {
         return properties.some((property) => row?.[property]);
       }
 
-      return Object.values(rows[0]).some((value) => value);
+      return Object.values(row).some((value) => value);
     }
 
     // Return false if there are no rows
--- a/ui/admin/app/components/form/field/key-value/index.js
+++ b/ui/admin/app/components/form/field/key-value/index.js
@@ -33,6 +33,16 @@ export default class FormFieldKeyValueComponent extends Component {
     return false;
   }
 
+  get errors() {
+    const { model, name } = this.args;
+
+    if (model && name && model.errors) {
+      return model.errors[name];
+    }
+
+    return null;
+  }
+
   /**
    * Default handler for adding a row.
    * Uses @model, @name, and @properties if provided, otherwise calls @onAdd.
--- a/ui/admin/app/components/form/field/key-value/index.js
+++ b/ui/admin/app/components/form/field/key-value/index.js
@@ -21,6 +21,14 @@ export default class FormFieldKeyValueComponent extends Component {
 
     // For a single row, only show if it has any non-empty values
     if (rows.length === 1) {
+      const row = rows[0];
+
+      // Pass @properties for rows with tracked fields so canDelete reactively checks the tracked properties.
+      const properties = this.args.properties || [];
+      if (properties.length) {
+        return properties.some((property) => row?.[property]);
+      }
+
       return Object.values(rows[0]).some((value) => value);
     }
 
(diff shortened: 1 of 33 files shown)


ui/admin/app/components/form/worker/create-tags/index.hbs
```diff
@@ -10,24 +10,39 @@
...
+          data-test-key-input
+          @value={{R.rowData.key}}
+          {{on 'input' (fn R.updateRow 'key')}}
+        />
+      </R.Field>
+
+      <R.Field as |F|>
+        <F.Label>{{t 'form.value.label'}}</F.Label>
+        <F.TextInput
+          data-test-value-input
```
priya-patel04, line 35: Is the `value` arg missing here?

ui/admin/app/components/form/worker/create-worker-led/index.hbs
```diff
@@ -78,35 +78,40 @@
...
-          <:value as |V|>
-            <V.text />
-          </:value>
-        </F.KeyValue>
-      </:field>
-    </Form::Field::ListWrapper>
+        <R.Field as |F|>
+          <F.TextInput
+            data-test-value-input
+            {{on 'input' (fn R.updateRow 'value')}}
```
priya-patel04, line 106: Same here as well for the missing `value` arg

DhariniJeeva: oops yes it should be there. ty!

priya-patel04 (dismissed):
Looks good!

ui/admin/app/components/form/field/key-value/index.js
```diff
@@ -21,7 +21,15 @@ export default class FormFieldKeyValueComponent extends Component {
 
     // For a single row, only show if it has any non-empty values
     if (rows.length === 1) {
-      return Object.values(rows[0]).some((value) => value);
+      const row = rows[0];
+
+      // Pass @properties for rows with tracked fields so canDelete reactively checks the tracked properties.
+      const properties = this.args.properties || [];
```
ZedLi:
```suggestion
      const properties = this.args.properties ?? [];
```

ui/admin/app/components/form/worker/create-tags/index.hbs
```diff
@@ -10,24 +10,40 @@
...
-          <V.text />
-        </:value>
-      </F.KeyValue>
-    </:field>
-  </Form::Field::ListWrapper>
+  <Form::Field::KeyValue
+    @name='api_tags'
+    @data={{@apiTags}}
+    @model={{this}}
+    @properties={{array 'key' 'value'}}
```
ZedLi: Do you need `model` and `properties` if you're using the custom `on{Action}` handlers?

ui/admin/app/components/form/worker/create-worker-led/index.hbs
```diff
@@ -78,35 +78,41 @@
...
     {{/if}}
-    <Form::Field::ListWrapper
-      @layout='horizontal'
-      @disabled={{form.disabled}}
+
+    <Form::Field::KeyValue
+      @name='worker_tags'
+      @data={{this.workerTags}}
+      @model={{this}}
+      @properties={{array 'key' 'value'}}
```
ZedLi: Same here, are these needed?

ui/admin/app/components/form/field/key-value/index.js
```diff
@@ -21,7 +21,15 @@ export default class FormFieldKeyValueComponent extends Component {
 
     // For a single row, only show if it has any non-empty values
     if (rows.length === 1) {
-      return Object.values(rows[0]).some((value) => value);
+      const row = rows[0];
+
+      // Pass @properties for rows with tracked fields so canDelete reactively checks the tracked properties.
```
ZedLi: I'm trying to understand what the underlying issue is here, is it because we're not getting reactivity from ember tracking from the passed in `this.args.data`? (This is how I read the comment)

Or is the issue because we have an object (our Tags class?) that has been decorated with tracked properties that isn't working with `Object.values`? If it's the latter, could we just use a [`for..in`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for...in) loop? `Object.values` only enumerates its own plain object properties while `for..in` will enumerate over the prototype chain and the inherited properties which is how ember implements its tracking.

DhariniJeeva: it's the latter with Tag class and tracked properties and yes we can use a for..in loop. I was considering that approach but wasn't sure which was the best one. With for..in, we also wouldn't need to pass properties when using custom handlers.

ui/admin/app/components/form/worker/create-tags/index.hbs
DhariniJeeva: it's not needed, but I made a change yesterday where we derive `errors` from the model instead of passing in the `error` arg but this wouldn't work in case of custom handlers. I will revert that change and update this pr.

ui/admin/app/components/form/worker/create-worker-led/index.hbs
DhariniJeeva: removd!

ui/admin/app/components/form/worker/create-tags/index.hbs
ZedLi: Couldn't we have checked if we pass in the `errors` arg but still default to using the model like we did for the custom handlers? It doesn't look like the tag ones can have errors either right, is there scenarios where we'd have errors when we don't use a `model`?

DhariniJeeva: hmm, we don't currently have any scenarios where we'd have errors without a model.

DhariniJeeva: updated this!

ui/admin/app/components/form/field/key-value/index.js
```diff
@@ -21,13 +21,28 @@ export default class FormFieldKeyValueComponent extends Component {
...
+          return true;
+        }
+      }
     }
 
     // Return false if there are no rows
     return false;
   }
 
+  get errors() {
```
ZedLi, line 36: We might as well still handle using `@errors` if it's passed in if we have it, though I agree it's unlikely we'd need it.

ZedLi (dismissed):

ZedLi (approved):
Looks good!

PR #3293, opened 2026-06-01
Llb/key value migration

DhariniJeeva:
# Description

This LLB contains PRs that migrate `listwrapper` fields to hds keyValueInputs component. 

jira - https://hashicorp.atlassian.net/browse/ICU-17397

## Checklist

## PCI review checklist


ui/admin/app/components/form/auth-method/oidc/index.hbs
```diff
@@ -98,7 +98,6 @@
     name='client_id'
     @value={{@model.client_id}}
     @isRequired={{true}}
-    @isInvalid={{@model.errors.client_id}}
```
priya-patel04, line 101: I dont have the context, but was this intentional to remove `isInvalid` arg here ?

ui/admin/app/routes/scopes/scope/storage-buckets/new.js
```diff
@@ -78,6 +80,7 @@ export default class ScopesScopeStorageBucketsNewRoute extends Route {
       bucket_prefix,
       region,
       worker_filter,
+      role_tags: role_tags || [{ key: '', value: '' }],
```
priya-patel04: Similar to host catalogs logic [here](https://github.com/hashicorp/boundary-ui/pull/3293/changes#diff-1bb40a68ad00e9db56a2a7ee6f2371b5a50b6094d9ad862e3557a512d8f7c719R64), should we add `role_tags` only for dynamic credentials ?

priya-patel04 (commented):
Overall looks great! Left couple minor comments

ui/admin/app/components/form/auth-method/oidc/index.hbs
DhariniJeeva: yes this was intentional. In a previous review we discussed this- there's no key/value specific error message from the api. If we get an error, it would be for the field (`client_id`) as a whole. Passing `@isInvalid` will cause all key/value to appear invalid, so we rely on the error message to display the error.

ui/admin/app/routes/scopes/scope/storage-buckets/new.js
DhariniJeeva: we can remove that change from the route for both since we inject the empty rows from the changeCredentialType in the controllers. added it [here](https://github.com/hashicorp/boundary-ui/pull/3302) ty for catching!

ui/admin/app/components/form/auth-method/oidc/index.hbs
priya-patel04: Ahh got it! Thanks for the explanation

ZedLi (approved):

PR #3302, opened 2026-06-01
test: 💍 a11y

DhariniJeeva:
# Description

Screenshot 2026-06-02 at 11 52 49 AM

## Checklist

## PCI review checklist

### DIFF
--- a/e2e-tests/admin/pages/targets.js
+++ b/e2e-tests/admin/pages/targets.js
@@ -58,7 +58,7 @@ export class TargetsPage extends BaseResourcePage {
     if (alias) {
       await this.page
         .getByRole('group', { name: 'Aliases' })
-        .getByLabel('Alias')
+        .getByLabel('Aliases')
         .last()
         .fill(alias);
       await this.page.getByRole('button', { name: 'Add' }).click();
--- a/e2e-tests/admin/pages/targets.js
+++ b/e2e-tests/admin/pages/targets.js
@@ -58,7 +58,7 @@ export class TargetsPage extends BaseResourcePage {
     if (alias) {
       await this.page
         .getByRole('group', { name: 'Aliases' })
-        .getByLabel('value')
+        .getByLabel('Alias')
         .last()
         .fill(alias);
       await this.page.getByRole('button', { name: 'Add' }).click();
--- a/ui/admin/app/routes/scopes/scope/host-catalogs/new.js
+++ b/ui/admin/app/routes/scopes/scope/host-catalogs/new.js
@@ -5,7 +5,6 @@
 
 import Route from '@ember/routing/route';
 import { service } from '@ember/service';
-import { TYPE_HOST_CATALOG_PLUGIN_AWS } from 'api/models/host-catalog';
 
 export default class ScopesScopeHostCatalogsNewRoute extends Route {
   // =services
@@ -60,11 +59,6 @@ export default class ScopesScopeHostCatalogsNewRoute extends Route {
       description,
       disable_credential_rotation,
     });
-
-    if (compositeType === TYPE_HOST_CATALOG_PLUGIN_AWS) {
-      // Ensure one editable role_tags row exists for dynamic credentials.
-      record.role_tags = [{ key: '', value: '' }];
-    }
 
     record.scopeModel = scopeModel;
     return record;
--- a/ui/admin/tests/acceptance/auth-methods/oidc-test.js
+++ b/ui/admin/tests/acceptance/auth-methods/oidc-test.js
@@ -80,6 +80,9 @@ module('Acceptance | auth-methods | oidc', function (hooks) {
       id: instances.authMethod.id,
     });
 
+    // Re-open dropdown to check the selected state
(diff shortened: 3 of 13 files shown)


ui/admin/tests/acceptance/accounts/list-test.js
```diff
@@ -49,6 +49,15 @@ module('Acceptance | accounts | list', function (hooks) {
   });
 
   test('Users can navigate to accounts with proper authorization', async function (assert) {
+    setRunOptions({
```
ZedLi: Are these not valid a11y violations we need to fix?

DhariniJeeva: oh I thought we usually ignored the `label` ones. Is that not right?

DhariniJeeva: maybe for this component, we could add aria-label for textInput and other inputs.   I may be misunderstood looking at the pattern.

ZedLi: Do we know what it fails on exactly? I just assumed the label one is legit and should be something we need to rectify

DhariniJeeva: yeah you are right. There were a couple of labels missing and I added aria-labels to places where we didn't need to show a label in the UI. 

All the tests pass locally now except `oidc: can update oidc state to active-public` which is timing out and doesn't seem relevant to the changes I made.

ZedLi: Do you know if the e2e test is also related to some of the changes?

DhariniJeeva: I think it is related to the keyvalue field. looking into this

DhariniJeeva: fixed the failing e2e test!

e2e-tests/admin/pages/targets.js
```diff
@@ -58,7 +58,7 @@ export class TargetsPage extends BaseResourcePage {
     if (alias) {
       await this.page
         .getByRole('group', { name: 'Aliases' })
-        .getByLabel('value')
+        .getByLabel('Aliases')
```
DhariniJeeva, line 61: this is due to the aria-label change [here](https://github.com/hashicorp/boundary-ui/pull/3302/changes#diff-af75bbf4eb601c24977bfcf4cb91cb95cb40e95ce1358d0092bd8cf0fc1c1bb9R259-R260). Previously this was `Value` when we used the listWrapper component

ui/admin/app/routes/scopes/scope/host-catalogs/new.js
```diff
@@ -61,11 +60,6 @@ export default class ScopesScopeHostCatalogsNewRoute extends Route {
       disable_credential_rotation,
     });
 
-    if (compositeType === TYPE_HOST_CATALOG_PLUGIN_AWS) {
```
DhariniJeeva, line 64: we don't need this as we handle this logic in the changeCredentialType. same for storage buckets.

ui/admin/tests/acceptance/accounts/list-test.js
DhariniJeeva: `oidc: can update oidc state to active-public` seems to be passing in main. looking into this.

ui/admin/tests/acceptance/auth-methods/oidc-test.js
```diff
@@ -80,6 +80,9 @@ module('Acceptance | auth-methods | oidc', function (hooks) {
       id: instances.authMethod.id,
     });
 
+    // Re-open dropdown to check the selected state
+    await click(selectors.CHANGE_STATE_DROPDOWN);
```
DhariniJeeva, line 84: Reopening the dropdown here bc the migration to keyvalue introduced a re-render(due to direct data binding) after the state change. This re-render closes the dropdown before the assertion runs, causing the selected option check to fail in a11y tests. Open to alternative approaches if there's a cleaner way to handle this.

ZedLi: Can you help explain what happens a bit more, I don't see this test touch anything with the key value changes so I'm not sure why it affects the dropdown?

DhariniJeeva: when a state is selected from the dropdown, the model seems to be re-rendering due to the migration. If I remove the migrated fields from the template, the dropdown doesn't close after making selecting an option.  I'm not sure what is causing this, I think it could be the `emptyArrayIfMissing` option we have.. So all these fields are getting reset when the state changes causing the dropdown to close.

ZedLi (approved):
Thanks for the fix!

cameronperera (approved):

ui/admin/tests/acceptance/auth-methods/oidc-test.js
ZedLi: Huh that seems strange and likely something we should investigate but I don't think is something that's blocking

PR #3306, opened 2026-06-03
chore: make sub claim mapping readonly after creation

DhariniJeeva:
# Description

API doesn't allow `sub` on update. If there is a `sub` claim mapping, disable the field.

[jira](https://hashicorp.atlassian.net/browse/ICU-18579)

https://github.com/user-attachments/assets/2e77ae50-d6df-4176-a703-b2e0e85ec9f0

Screenshot 2026-06-03 at 4 28 33 PM

Screenshot 2026-06-03 at 4 28 16 PM

Screenshot 2026-06-03 at 4 28 33 PM

## How to Test

1. run boundary-ent controller
2. you will need an oidc auth provider, I used [auth0](https://developer.hashicorp.com/boundary/tutorials/identity-management/oidc-auth0) which is pretty simple to setup
3. add any sub value and save it, field should be disabled after creation 
4. create another oidc form, this time add a name or email claim mapping, after save, the field should be still editable and `sub` should not be shown in the dropdown list. 

## Checklist

- [x] I have added `a11y-tests` label to run a11y audit tests if needed

## PCI review checklist


lisbet-alvarez (approved):
Looks great!

addons/api/addon/serializers/auth-method.js
```diff
@@ -48,6 +48,18 @@ export default class AuthMethodSerializer extends ApplicationSerializer {
...
+
+    // Don't send account_claim_maps on update if there's a saved 'sub' mapping
+    // The API doesn't allow updating when 'sub' claim mappings exist
+    if (!snapshot.record.isNew && snapshot.record.account_claim_maps) {
+      const hasSubClaimMap = snapshot.record.account_claim_maps.some(
+        (item) => item?.value === 'sub',
+      );
+
+      if (hasSubClaimMap) {
+        delete serialized.attributes.account_claim_maps;
```
ZedLi, line 60: So we're preventing users from updating all account claim maps if there's a `sub`? Can we just not omit the `sub` value and disable only that and still update other fields?

ui/admin/app/components/form/auth-method/oidc/index.js
```diff
@@ -39,6 +39,35 @@ export default class FormAuthMethodOidcComponent extends Component {
...
+   * @returns {string[]}
+   */
+  toClaimOptions = (rowData) => {
+    // On creation, show all options
+    if (this.args.model?.isNew) {
+      return this.toClaims;
+    }
+
+    // On update, filter "sub" unless it's the current value
+    return this.toClaims.filter(
```
ZedLi, line 66: What's the purpose of filtering out sub if the field is disabled anyways and a user can't modify it?

ZedLi (approved):
Looks good!

addons/api/addon/serializers/auth-method.js
DhariniJeeva: we discussed this offline- but for context, yes, we are. The api update call fails if there is already a `sub` value so it won't other fields to be updated either.

ui/admin/app/components/form/auth-method/oidc/index.js
DhariniJeeva: This filtering is for the case where account_claim_maps is empty and since sub is only allowed during creation, we still want to make sure it doesn't get included in the dropdown for update.

PR #3323, opened 2026-06-17
Use codemirror tooltip extension to set parent to fix positioning bug

lisbet-alvarez:
# Description

In firefox only (works normally in Safari and Chrome), the autocomplete suggestions tooltip position gets set to `absolute` as a fallback. Usually this happens due to an ancestor of the tooltip element having a transform applied. After inspecting elements that does not seem to be the case so it seems to be an issue with firefox browser itself. To avoid positioning issue a [workaround](https://discuss.codemirror.net/t/tooltip-position-as-fixed-as-absolute/8776) is to set tooltip parent to `document.body`

Before:
image

After:
image

## How to Test

1. Check in all major browsers (Firefox, Chrome, Safari) that autocomplete tooltip is not hidden and overlaps code editor instead.
2. Make sure tooltip is still visible in dark mode
3. Make sure tooltip is still visible when editor is in full screen mode

## Checklist

## PCI review checklist

- [X] I have documented a clear reason for, and description of, the change I am making.

### DIFF
--- a/ui/admin/app/components/form/role/edit-grants/index.js
+++ b/ui/admin/app/components/form/role/edit-grants/index.js
@@ -170,8 +170,8 @@ export default class FormRoleEditGrantsComponent extends Component {
   );
 
   customExtensions = [
-    // Configure tooltips to render in the body to avoid overflow issues within the editor container.
-    tooltips({ parent: document.body }),
+    // Configure tooltips to render in page layout to avoid overflow issues within the editor container.
+    tooltips({ parent: document.querySelector('.rose-layout-page') }),
     autocompletion({
       override: [this.completionSource],
       // Trigger autocompletion when the user completes a grant field (which we labeled as keywords)
--- a/ui/admin/app/components/form/role/edit-grants/index.js
+++ b/ui/admin/app/components/form/role/edit-grants/index.js
@@ -170,6 +170,7 @@ export default class FormRoleEditGrantsComponent extends Component {
   );
 
   customExtensions = [
+    // Configure tooltips to render in the body to avoid overflow issues within the editor container.
     tooltips({ parent: document.body }),
     autocompletion({
       override: [this.completionSource],
--- a/ui/admin/app/components/form/role/edit-grants/index.js
+++ b/ui/admin/app/components/form/role/edit-grants/index.js
@@ -8,6 +8,7 @@ import { action } from '@ember/object';
 import { service } from '@ember/service';
 import { tracked } from '@glimmer/tracking';
 import {
+  tooltips,
   autocompletion,
   completionKeymap,
   startCompletion,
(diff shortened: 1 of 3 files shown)


ui/admin/app/components/form/role/edit-grants/index.js
```diff
@@ -169,6 +170,8 @@ export default class FormRoleEditGrantsComponent extends Component {
   );
 
   customExtensions = [
+    // Configure tooltips to render in the body to avoid overflow issues within the editor container.
+    tooltips({ parent: document.body }),
```
ZedLi: I think this will end up breaking our dark mode overrides too since the selectors will now be different than expected

lisbet-alvarez: ah yes that is right!
updated

ui/admin/app/styles/app.scss
```diff
@@ -1042,3 +1042,8 @@
   max-height: 50vh;
   overflow-y: auto;
 }
+
+// Grants editor autocomplete tooltip
+.rose-layout-page div .cm-tooltip-autocomplete {
+  z-index: 1000;
```
ZedLi, line 1049: Why do we need this z-index? It's usually not great when something like this is needed

ZedLi (dismissed):
Works as expected, great catch and thanks for the fix!

ui/admin/app/styles/app.scss
lisbet-alvarez: when code editor is in full screen mode the tooltip gets hidden behind the editor. I tried making the parent of the tooltip the hds code editor itself but then we run into the same issue with the tooltip getting cut off (when not in fullscreen mode).

ZedLi (approved):

priya-patel04 (approved):
Tested and looks great!

ui/admin/app/components/form/role/edit-grants/index.js
```diff
@@ -169,6 +170,8 @@ export default class FormRoleEditGrantsComponent extends Component {
   );
 
   customExtensions = [
+    // Configure tooltips to render in page layout to avoid overflow issues within the editor container.
+    tooltips({ parent: document.querySelector('.rose-layout-page') }),
```
ZedLi, line 174: In chrome and safari, what does codemirror use for the parent when it correctly uses the `fixed` layout before this change?

lisbet-alvarez: Screenshot 2026-06-18 at 8 56 42 AM

parent is code mirror editor in all browsers, above image is chrome
this image below is in firefox
Screenshot 2026-06-18 at 9 00 37 AM

ZedLi: Gotcha, I wasn't sure if they were different, what an interesting behavior

PR #3339, opened 2026-07-31
fix: targets pagination shows wrong counts and empty pages

### LINKED ISSUE #3338: Incorrect pagination in Targets view causes missing items and empty pages
### **Description**

The pagination in the **Targets** view appears to be incorrect. The number of displayed targets does not match the selected **Items per page** value, resulting in missing targets and empty pages.

For example, I have **5 targets**, but with the default **Items per page = 10**, only **1 target** is displayed. Pagination incorrectly shows **8 pages**, and pages 2 through 8 are empty.

Changing **Items per page** to **50** partially improves the situation, but only **4 targets** are shown on the first page and the fifth target appears on a second page. Since there are only five targets in total, they should all fit on a single page regardless of whether the page size is 10 or 50.

Multiple developers on our team have observed the same behavior.

### **Steps to Reproduce**

1. Open the Boundary Desktop application.
2. Navigate to the **Targets** view.
3. Ensure there are only a few targets (for example, 5 targets).
4. Leave **Items per page** set to the default value of **10**.
5. Observe that only one target is displayed and that pagination shows multiple pages.
6. Navigate through the additional pages.
7. Change **Items per page** to **50**.

## **Expected behavior**

With only five targets available:

* **Items per page = 10** should display all five targets on a single page.
* **Items per page = 50** should also display all five targets on a single page.
* Pagination should not display additional empty pages.

## **Actual behavior**

* With **Items per page = 10**, only one target is displayed.
* Pagination incorrectly displays **8 pages**, although only five targets exist.
* Pages **2 through 8** are empty.
* With **Items per page = 50**, four targets are displayed on the first page and the fifth target appears on a second page.

## **Screenshots (if applicable)**

Attached screenshots demonstrate:

1. **Items per page = 10**, showing only one target and pagination from 1 to 8 with empty pages.
2. **Items per page = 50**, showing four targets on the first page and the remaining target on page 2.

## **Environment**

* **OS and Version:** macOS (lake tahoe)
* **Architecture:** arm64 (Apple Silicon)
* **Browser/Version:** N/A
* **Desktop Client Version:** 2.6.2
* **Controller/worker versions:** 0.3.0

## **Logs & Console Output (if applicable)**

No relevant errors were observed. This appears to be a UI/pagination issue.

## **Additional Context**

This issue negatively impacts usability because it gives users the impression that targets are missing. Several developers on our team have experienced the same behavior, suggesting this is not an isolated issue.

Image
Image

cschindlbeck:
## Summary

Fixes #3338. Two bugs combined to break pagination in the desktop client's Targets view:

- The "connect" ability filter (`this.abilities.can('connect target', target)`) was applied **after** the target list was already paginated. With permissions scattered unevenly across a larger raw target set, a given page could end up showing far fewer connectable targets than fit (or none at all), while the pagination total was computed from the raw, permission-unaware count. The filter is now applied before pagination, and totals are computed from the filtered set.
- The local cache daemon can return multiple raw rows for the same underlying target (e.g. stale/duplicate entries in its cache). These are now deduplicated by id before totals are computed and pages are sliced.
- `paginateResults` returned a duplicate of the last item instead of an empty array when asked for a page beyond the available results; fixed to return `[]`.

## Test plan

- [x] Added a unit test for `paginateResults` covering the out-of-range page case
- [x] Added an acceptance test reproducing the duplicate-cache-daemon-row scenario
- [x] Added an acceptance test reproducing the ability-filter-before-pagination scenario (many targets, only a few connectable)
- [x] `pnpm test` passes in `addons/api` and `ui/desktop`
- [x] Verified manually against a real cluster and running cache daemon

### DIFF
--- a/addons/api/addon/handlers/cache-daemon-handler.js
+++ b/addons/api/addon/handlers/cache-daemon-handler.js
@@ -147,7 +147,20 @@ export default class CacheDaemonHandler {
           }
         }
 
-        const results = cacheDaemonResults[underscore(resourceName)];
+        // The cache daemon may return multiple raw rows for the same
+        // underlying resource (e.g. stale or duplicate entries in its local
+        // cache). Pagination must be computed from the unique set of
+        // resources, not the raw row count, otherwise a single resource can
+        // be repeated or fragmented across many pages.
+        const rawResults = cacheDaemonResults[underscore(resourceName)];
+        const seenIds = new Set();
+        const results = rawResults?.filter((result) => {
+          if (seenIds.has(result.id)) {
+            return false;
+          }
+          seenIds.add(result.id);
+          return true;
+        });
         const payload = { items: paginateResults(results, page, pageSize) };
 
         const schema = store.modelFor(type);
--- a/addons/api/addon/utils/paginate-results.js
+++ b/addons/api/addon/utils/paginate-results.js
@@ -22,7 +22,7 @@ export const paginateResults = (array, page, pageSize) => {
   }
 
   const offset = (page - 1) * pageSize;
-  const start = Math.min(length - 1, offset);
+  const start = Math.min(length, offset);
   const end = Math.min(length, offset + pageSize);
 
   return array.slice(start, end);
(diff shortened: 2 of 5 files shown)


cschindlbeck:
@ZedLi can you take a look at this, it is a real bug many of our devs complain about

